PR #202, opened 2016-01-20
shallow clone for install

edouard-lopez:
Most people won't need to pull the whole history

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ Laptop and Desktop browser for OS X and Windows.
 
 ## Installation
 
-1. Clone the git repository from GitHub:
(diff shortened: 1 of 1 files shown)


bridiver:
These instructions are primarily targeted at developers who want to contribute so I'm not sure if we'd want to change that. @bbondy ?

bbondy:
Thanks for the PR! 
I prefer getting the whole history locally, I suspect a lot of devs do too.  Could you amend the commit to provide an alternate installation step?

bbondy:
Thanks!

bridiver:
great addition, thanks!

PR #207, opened 2016-01-20
Repository points to incorrect location, updating

eliperelman:


package.json
```diff
@@ -24,10 +24,7 @@
     "postinstall": "node ./tools/rebuildNativeModules.js&&webpack",
     "test": "NODE_ENV=test mocha --compilers js:babel-register --recursive $(find test -name '*Test.js')"
   },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/bbondy/electron-sample.git"
-  },
+  "repository": "brave/browser-laptop",
```
bbondy: https://docs.npmjs.com/files/package.json#repository
Is this maybe the preferred format? If so mind doing an amend of this pull request?
Thanks for the PR.

eliperelman: I believe it is preferred for GH et al repos as it creates a nice shorthand that is consistent between package.json and `npm install`. If you look further down in that same section, it shows the shorthands.

So if you could clarify, which would you prefer? I can change this to be the longer URL format, or update all the git URLs to use the short format.

bbondy: OK that sounds good, thanks for clarifying and sorry for missing the scroll down ;)  Merging.

eliperelman: No worries, and thanks!

PR #211, opened 2016-01-20
Integrate NSP security checks into build scripts

aekeus:
@bbondy @yan this adds the NSP security checks into both the OSx and win64 build scripts. Tested build on both platforms.

### DIFF
--- a/build-darwin.js
+++ b/build-darwin.js
@@ -15,6 +15,7 @@ var cmds = [
   'rm -rf Brave-darwin-x64',
   'NODE_ENV=production ./node_modules/.bin/webpack',
   'rm -f dist/Brave.dmg',
+  'npm run checks',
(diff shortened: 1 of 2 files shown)


bbondy:
Great idea!

bbondy:
Does the script give output if it fails or does it fail the script?
If it is supposed to fail the script, please verify that a security error returns non zero return code and that non zero triggers a fail of the script. 

If it is output only that's ok as is without further checks.

bbondy:
Feel free to merge after, looks like some conflict.

aekeus:
From the docs:

  --warn-only: optional

    report errors, but always quit with an exit code of 0

Without the parameter (which we are not sending) it will fail with a non-zero return code / stopping the build process.

bbondy:
great, thanks for checking.

therealklanni:
Was planning on looking into this later, thanks @aekeus :+1:

PR #216, opened 2016-01-21
fix git url dependency

craigmichaelmartin:
cf https://docs.npmjs.com/files/package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   "dependencies": {
     "abp-filter-parser-cpp": "^1.0.22",
     "electron-localshortcut": "^0.6.0",
-    "electron-prebuilt": "https://github.com/brave/electron-prebuilt",
+    "electron-prebuilt": "brave/electron-prebuilt",
     "electron-squirrel-startup": "^0.1.4",
     "file-loader": "^0.8.5",
     "font-awesome": "^4.5.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
   "dependencies": {
     "abp-filter-parser-cpp": "^1.0.22",
     "electron-localshortcut": "^0.6.0",
-    "electron-prebuilt": "https://github.com/brave/electron-prebuilt",
+    "electron-prebuilt": "git+https://github.com/brave/electron-prebuilt",
```
eliperelman: This could be shortened to:

```json
"electron-prebuilt": "brave/electron-prebuilt",
```

package.json
```diff
@@ -65,7 +65,7 @@
     "tracking-protection": "^1.0.7",
     "url-loader": "^0.5.7",
     "urlutil.js": "^0.1.1",
-    "vibrant": "https://github.com/jariz/vibrant.js#1.0"
+    "vibrant": "git+https://github.com/jariz/vibrant.js#1.0"
```
eliperelman: This could be shortened to:

```json
"vibrant": "jariz/vibrant.js#1.0",
```

eliperelman:
I made some comments, which would also make the git dependencies consistent with the `repository` field in package.json.

craigmichaelmartin:
Great, thanks @eliperelman!

bbondy:
Looks great, thanks!

PR #222, opened 2016-01-21
fix undefined active frame for back/forward button after tab close by cl…

jiangyang:
…icking

### DIFF
--- a/js/components/tab.js
+++ b/js/components/tab.js
@@ -101,7 +101,8 @@ class Tab extends ImmutableComponent {
     WindowActions.setActiveFrame(this.props.frameProps)
   }
 
-  onCloseFrame () {
+  onCloseFrame (event) {
+    event.stopPropagation()
     WindowActions.closeFrame(this.props.frames, this.props.frameProps)
   }
 


bridiver:
Thanks for the PR! Is there a test we can add for this to prevent regressions?

bridiver:
It looks to me like the props are not updating in main.js when the activeFrameKey changes. I think the issue may be in shouldComponentUpdate.

bridiver:
The root cause of this is a call to set the active frame after the frame is closed. Should have a fix for it very soon. Thanks for your help! The steps to replicate the issue you provided were critical to fixing it.

bbondy:
@bridiver should this be closed now?

bridiver:
No, still working on the fix for this issue. The other PR was just a side issue that I initially thought was related

bridiver:
actually I guess we can close it because I'll have a new PR for it

bridiver:
Closing this and I'll reference it from my PR. Thanks again @jiangyang !

bridiver:
I was initially skeptical, but after more investigation those does look like the right solution. It doesn't appear to need ```event.nativeEvent.stopImmediatePropagation()``` though. I'm putting together a test case for it now

jiangyang:
late to reply...thanks for looking into this. I didn't go for the tests as I wasn't sure how to assert on the active frame but oh well. I've removed the nativeEvent stop as requested. Thanks again.

bridiver:
closed by #232

bridiver:
ha, sorry, beat you to it but your commit is in there

jiangyang:
lol, no worries.

PR #223, opened 2016-01-21
Fixed #200: add "engines" field to package.json to indicate node version

### LINKED ISSUE #200: Figure out missing dependencies not documented
Because why not!

yurenju:
Noted, this field can't avoid user to install it with wrong version of node, so it just indicates what version of node we should use.

reference: http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,5 +120,9 @@
   },
   "pre-commit": [
     "lint"
-  ]
+  ],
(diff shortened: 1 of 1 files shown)


bbondy:
Awesome, thanks so much!

PR #225, opened 2016-01-21
Immutable.is only works on Immutable objects

bridiver:
Immutable.fromJS can be called on Immutables as well as plain JS so this works for both cases

### DIFF
--- a/js/components/immutableComponent.js
+++ b/js/components/immutableComponent.js
@@ -7,7 +7,7 @@ const Immutable = require('immutable')
 
 class ImmutableComponent extends React.Component {
   shouldComponentUpdate (nextProps, nextState) {
-    return Object.keys(nextProps).some(prop => !Immutable.is(nextState, this.props[prop]))
+    return !Immutable.is(Immutable.fromJS(this.props), Immutable.fromJS(nextProps))
   }
 }
 


bridiver:
@bbondy

js/components/immutableComponent.js
```diff
@@ -7,7 +7,7 @@ const Immutable = require('immutable')
 
 class ImmutableComponent extends React.Component {
   shouldComponentUpdate (nextProps, nextState) {
-    return Object.keys(nextProps).some(prop => !Immutable.is(nextState, this.props[prop]))
+    return !Immutable.is(Immutable.fromJS(this.props), Immutable.fromJS(nextProps))
```
bbondy: Please double verify that Immutable.formJS will not cause any slowdown if you already have an immutable object. I bet it is fine, but please double check.  Just because the performance of the entire app relies on this check.

bbondy:
Double check my comment, but good to post after that. Thanks for finding this!

bridiver:
```
function fromJSDefault(json) {
  if (Array.isArray(json)) {
    return IndexedSeq(json).map(fromJSDefault).toList();
  }
  if (isPlainObj(json)) {
    return KeyedSeq(json).map(fromJSDefault).toMap();
  }
  return json;
}
```
basically just a passthrough if it's already Immutable

bbondy:
Perfect, thanks for checking

bridiver:
It's something we should probably take a closer look at later and see what percentage of components actually require update on average. If the percentage is very high then it might makes sense to drop the check entirely and return true like the default. Need to compare the extra render time to the overhead of making all the checks

bbondy:
Fairly certain that we want to efficiently compute this (as we do) and return false most of the time or else we'll suffer a performance problem.

bridiver:
Agreed in principle, but that will take some tuning of the properties we send to each component. Right now we pass around a lot of extra information that ends up triggering renders for components when they are not strictly necessary. Just think we should get some empirical data to make sure our assumptions are correct.

bbondy:
From what I've read efficiently computing this exact line of code (which is why we use ImmutableJS) and returning false a lot are key.  But yep I didn't profile this yet.

PR #233, opened 2016-01-22
use npm run alias

craigmichaelmartin:
This is indeed what is listed elsewhere, eg in the readme.
cf https://docs.npmjs.com/cli/run-script#synopsis


bridiver:
nice catch, thanks!

PR #234, opened 2016-01-22
Add build script for linux x64

chreble:
[Review on Reviewable](https://reviewable.io/reviews/brave/browser-laptop/234)


bbondy:
Thanks, this is great! (note that I'll refactor the top level directory build scripts soonish for better code re-use and move to tools/)

PR #240, opened 2016-01-22
More modern look and cleaner layout

teemuteemu:


eliperelman:
@teemuteemu would you happen to have a before and after screenshot?

teemuteemu:
Sure, here before:
brave-before

and after:
brave-after

bbondy:
I'm a fan of the changes, @bradleyrichter do you agree?

bbondy:
I think we'd want the nav-sphere (back / forward) re-added, and tab page back/forward to keep more of our identity, but need to get some more weigh in from our designer before you spend more time on it. I do prefer the other tweaks though.

bridiver:
I prefer the less round tabs as well, but also agree about keeping some distinct design elements. This design does give me an interesting idea: collapsing the entire tab navigation section leaving only the navbar. To avoid confusing new users it should probably be always open by default, but it could have "auto" mode like the OS X dock and slide in/out of view when changing tabs by keyboard shortcut or navbar mouseover. An always closed mode would also make a lot of sense along with some of the alternative tab navigation ideas we've talked about.

bradleyrichter:
@teemuteemu  - thanks for the suggestions. I want to create a light version of the chrome and this is looking good. Refining the tabs is on my short list. They do need to be simplified. As @bbondy  said, the round treatment in the URL bar has become part of the brand/style. I wonder how it will look in a light version? If you have time, I'd love to see it.

@bridiver Also on the idea list. Hide the tab bar for new windows. Often, you need a new window that will be closed before other tabs are added. That would tidy up the chrome in that instance. When we get going on prefs, we can look at adding some of these things.

diracdeltas:
just my 2 cents: i like the lightness, would prefer keeping round tabs

bradleyrichter:
I'm excited to collect feedback through the launch. I expect a lot of debate over light vs dark. We can have both by 1.0. And track the usage eventually to decide on the default. One of the main goals is to stand out from the other browsers. We can achieve that with a light theme but not as easily.

bradleyrichter:
Ok, let's pull this in and try it out. I'll adjust things if we run into conflicts with other buttons and colors. (save dark style for later in prefs). Thanks @teemuteemu.

bbondy:
Thanks for your contributions here @teemuteemu, highly appreciated! Merging.

PR #245, opened 2016-01-22
Adds nextTab and prevTab tabs elements only when needed on DOM

joselitojunior1:


bbondy:
@bradleyrichter when you get a chance, let me know if you think having those UI elements hidden for a single page would be better or not.  Let me know if you need a build to consider it.

bradleyrichter:
@bbondy I'd like to try this in a build. Sounds like a good change if it doesn't reduce the discoverability of how the tab sets work. It might even improve it? Need to try...

thanks @joselitojunior1 !

joselitojunior1:
@bradleyrichter I make that primarily because I got myself trying to back pages clicking on the `prevTab` button instead of the (bigger) correct one, near the url bar. This implementation looks like Firefox one, but uses the pagination rules of Brave.

bbondy:
If we do hide them, might make sense to also hide the tab page bar in that case.  Please also comment on that when you take a look @bradleyrichter. Thanks guys.

bradleyrichter:
I agree this is a usability issue. Happened to me too. It was worse in the previous builds. I think we can further reduce the contrast of the button in the disabled state if hiding it introducing a new problem. (one of which may be some hanging buttons and pinned tabs.)

@bbondy I do agree to also hide the tab indicator bar. We can revert if it reduces usability or put it in the future UI prefs panel.

diracdeltas:
I'm pretty in favor of this change. The only thing I would fix is that the padding on the left side without a `prevTab` element is too small. It should be the same as the padding in the second screenshot below.

screen shot 2016-01-22 at 12 13 02
screen shot 2016-01-22 at 12 13 02

bradleyrichter:
@diracdeltas we may need to adjust when pinned tabs are landed. That margin space is also needed as a drop-zone for pinning.

bbondy:
Thanks, I'll merge this in and we can tweak if needed.

PR #247, opened 2016-01-22
Adds copy address on context menus

joselitojunior1:
Adds "Copy link address" and "Copy image address" when opening a context menu for `A` and `IMG` elements

### DIFF
--- a/js/contextMenus.js
+++ b/js/contextMenus.js
@@ -4,6 +4,7 @@
 
 const remote = require('remote')
 const Menu = remote.require('menu')
+const Clipboard = require('clipboard')
 const messages = require('./constants/messages')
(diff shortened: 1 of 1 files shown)


joselitojunior1:
**Screenshots**:
*Links*
screen shot 2016-01-22 at 1 50 32 pm

*Images*
screen shot 2016-01-22 at 1 50 23 pm

bridiver:
Thanks for the PR! There is a related issue for links that I just posted #248 which will likely require some refactoring of the context menus, but this may be ok for now to support the simple case. What do you think @bbondy ?

bbondy:
Yes this is great for dogfood right now, thanks for the contribution!
Merging.

PR #254, opened 2016-01-22
Build Windows binaries in dist folder

aekeus:


bbondy:
Thanks, please also update .gitignore and the ignore list inside the build win64 script.  Please merge this when done that, thanks.

bbondy:
I'll just merge this now but please do in a followup later.

PR #256, opened 2016-01-22
Partitioned tabs

bbondy:
Reviewer: @diracdeltas

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -209,7 +209,7 @@ const doAction = (action) => {
       break
     case WindowConstants.WINDOW_NEW_FRAME:
       let nextKey = incrementNextKey()
-      let nextPartitionNumber = currentPartitionNumber
+      let nextPartitionNumber = 0
       if (action.frameOpts.isPartitioned) {
         nextPartitionNumber = incrementPartitionNumber()
       }
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -55,7 +55,9 @@ const updateTabPageIndex = (frameProps) => {
 }
 
 let currentKey = 0
+let currentPartitionNumber = 0
 const incrementNextKey = () => ++currentKey
+const incrementPartitionNumber = () => ++currentPartitionNumber
 
 class WindowStore extends EventEmitter {
 
@@ -112,6 +114,7 @@ const doAction = (action) => {
     case WindowConstants.WINDOW_SET_STATE:
       windowState = action.windowState
       currentKey = windowState.get('frames').reduce((previousVal, frame) => Math.max(previousVal, frame.get('key')), 0)
+      currentPartitionNumber = windowState.get('frames').reduce((previousVal, frame) => Math.max(previousVal, frame.get('partitionNumber')), 0)
       windowStore.emitChange()
       break
     case WindowConstants.WINDOW_SET_URL:
@@ -206,8 +209,12 @@ const doAction = (action) => {
       break
     case WindowConstants.WINDOW_NEW_FRAME:
       let nextKey = incrementNextKey()
+      let nextPartitionNumber = currentPartitionNumber
+      if (action.frameOpts.isPartitioned) {
+        nextPartitionNumber = incrementPartitionNumber()
+      }
       windowState = windowState.merge(FrameStateUtil.addFrame(windowState.get('frames'), action.frameOpts,
-        nextKey, action.openInForeground ? nextKey : windowState.get('activeFrameKey')))
+        nextKey, nextPartitionNumber, action.openInForeground ? nextKey : windowState.get('activeFrameKey')))
       if (action.openInForeground) {
         updateTabPageIndex(FrameStateUtil.getActiveFrame(windowState))
       }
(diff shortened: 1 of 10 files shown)


diracdeltas:
HTTPS Everywhere appears to not be working in the new partition

js/stores/windowStore.js
```diff
@@ -206,8 +209,12 @@ const doAction = (action) => {
       break
     case WindowConstants.WINDOW_NEW_FRAME:
       let nextKey = incrementNextKey()
+      let nextPartitionNumber = currentPartitionNumber
+      if (action.frameOpts.isPartitioned) {
+        nextPartitionNumber = incrementPartitionNumber()
+      }
```
diracdeltas: This seems buggy. if i open a new tab (via cmd+t), it seems to always uses the highest partition number. Ex: after closing all my partition-2 tabs, I open a tab from my default partition tab and it is in partition 2.

I expected new tabs to inherit the partition of the opener tab.

diracdeltas:
This is a good start. overall i think my biggest concern is that it's hard to tell what partition any tab is in. i think there needs to be UI like the "Person 1" indicator from Chrome:

screen shot 2016-01-22 at 18 30 51

Also, what are use cases for partitioned tabs that aren't already possible with private tabs? (Or with a fresh session per private tab)

js/stores/windowStore.js
bbondy: This is actually a bug. The correct functionality for new tab is to always open with the default session 0.  
Tabs and windows opened from a partitioned tab inherit that tab though.
Will fix, nice find.

test/tabTest.js
```diff
@@ -29,9 +29,19 @@ describe('tabs', function () {
...
     it('creates a private new tab when signaled', function *() {
       yield this.app.client
-        .ipcSend(messages.SHORTCUT_NEW_FRAME, 'http://www.brianbondy.com', true)
+        .ipcSend(messages.SHORTCUT_NEW_FRAME, 'http://www.brave.com', { isPrivate: true })
         .waitForExist('.tab[data-frame-key="3"]')
-        .waitForVisible('webview[partition]')
+        .waitForVisible('webview[partition="private-1"]')
+    })
+
+    it('creates a partitioned new tab when signaled', function *() {
```
diracdeltas: this fails for me

```
> npm run test -- --grep="partition"

> brave@0.7.7 test /Users/yan/repos/browser-electron
> NODE_ENV=test mocha --compilers js:babel-register --recursive $(find test -name '*Test.js') "--grep=partition"



  tabs
    new tab
      1) creates a partitioned new tab when signaled


  0 passing (10s)
  1 failing

  1) tabs new tab creates a partitioned new tab when signaled:
     Error: element (.tab[data-frame-key="4"]) still not existing after 5000ms
      at Object.<anonymous> (node_modules/webdriverio/lib/commands/waitForExist.js:50:19)
      at Object.safeExecute (node_modules/webdriverio/lib/utils/safeExecute.js:19:24)
      at Object.WebdriverIO.resolve (node_modules/webdriverio/lib/webdriverio.js:87:29)
      at node_modules/webdriverio/lib/webdriverio.js:227:32
      at _rejected (node_modules/webdriverio/node_modules/q/q.js:830:24)
      at node_modules/webdriverio/node_modules/q/q.js:856:30
      at Promise.when (node_modules/webdriverio/node_modules/q/q.js:1108:31)
      at Promise.promise.promiseDispatch (node_modules/webdriverio/node_modules/q/q.js:774:41)
      at node_modules/webdriverio/node_modules/q/q.js:590:44
      at runSingle (node_modules/webdriverio/node_modules/q/q.js:123:13)
      at flush (node_modules/webdriverio/node_modules/q/q.js:111:13)

```

diracdeltas: although it seems to pass when i run the entire test suite

bbondy:
Great questions:

>This is a good start. overall i think my biggest concern is that it's hard to tell what partition any tab is in. i think there needs to be UI like the "Person 1" indicator from Chrome:

I believe that the most important time you need to know of a partition is when you first create it.  I'm fairly sure that after you login to your second twitter account you don't really care.  This is because all tabs open in session 0 by default unless explicitly specified otherwise (modulo the bug above).  We will probably add some kind of visual indicator after but let's figure that out after.  You can see in the SiteInfo for now. 

> Also, what are use cases for partitioned tabs that aren't already possible with private tabs? (Or with a fresh session per private tab)

Persistent storage is the biggest.  For example I will have 3 pinned tabs fro twitter for Link Bubble, Brave Software and myself. They will persist and each live in their own session. I won't touch those tabs again.

Partitioned tabs are not about privacy they are a fix to people needing to use multiple browsers.  Private tabs are about privacy.  We may want persistent private tabs later.

test/tabTest.js
bbondy: Unfortunately those tests in that category depend on each other, should be refactored a bit later.

bbondy:
> HTTPS Everywhere appears to not be working in the new partition

This could be related to that private tab bug I posted before with HTTPS Everywhere not working there.

bbondy:
> This could be related to that private tab bug I posted before with HTTPS Everywhere not working there.

Seems so, I reopened that and I'll merge this.  Fixed the bug you found.

bbondy:
Posted #263 to explore UI concepts more.

Issue #250, opened 2016-01-22
Alternate ports


bkardell:
I'd really like to be able to run against something other than :8080 - can we get a command-line arg to specify port?

bbondy:
Sounds good, would accept PR for this if someone has time.
To make env vars cross platform we probably need the npm command to start a script in the tools/ directory like some other examples there.   Trying to avoid needing gulp if possible.

bkardell:
a configuration file is possibly easier/better because you need to match it on the server/browser, right?

bbondy:
I was just anticipating some problems dealing with env vars cross platform.

Let's just do something env var based like:
PORT=9091 npm run watch
on Windows that might be:
set PORT=9091&& npm run watch

Command line arg that npm passes to a tools/ script would be fine with me too.
npm run watch --port=9091

bbondy:
ditto PORT=9091 npm run start

I'd be ok with a config file but not making it required too.
git add one time and then add to .gitignore

rmoriz:
this still doesn't work, neither setting the envs 'PORT',  'BRAVE_PORT' or the npm config 'brave:port' has any effect.

luixxiul:
Me neither on Ubuntu 14.04. None of them works.

PR #276, opened 2016-01-23
Support alternate ports

bkardell:
re: Issue #250 

Not sure this is the best way to go about this, I don't like that it rewrites the file, but it seems to work for now?


bkardell:
actually, don't accept this even if you were going to -- I hate it, I'll fix that first.

PR #277, opened 2016-01-23
Add support for alternate ports

bkardell:
re: #250 Adds support for setting an env variable "BRAVE_PORT" which is respected by both the client and server if set.


diracdeltas:
thanks for this brian! please squash commits (`git rebase -i master`) so this is a single commit instead of 9

PR #280, opened 2016-01-23
add babel react plugins

### LINKED ISSUE #53: Check that we're using these babel / webpack optimizations
Check that we're using these babel / webpack optimizations:
> "optimisation.react.constantElements",
> "optimisation.react.inlineElements",

hzoo:
Fixes #53

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,7 @@
 {
-  "presets": ["react", "es2015"]
+  "presets": ["react", "es2015"],
+  "plugins": [
+    "transform-react-inline-elements",
+    "transform-react-constant-elements"
+  ]
 }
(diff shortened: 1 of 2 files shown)


hzoo:
Are the PR's running travis/tests? (I'm still `npm i` locally)

bbondy:
They were but I think when we opened the repo it got messed, will fix soon.

bbondy:
Thank you so much for this!

hzoo:
Looking at the webpack config - I think we can remove the ones there since the babelrc should cover it?

https://github.com/brave/browser-laptop/blob/master/webpack.config.js#L37 - doesn't seem to be a reason to keep both.

PR #287, opened 2016-01-24
List needed dependancies

PFCKrutonium:
For building, you know?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,13 @@ Laptop and Desktop browser for OS X, Windows and Linux.
 
 ## Installation
 
-1. Shallow clone the git repository from GitHub:
+1. Ensure you have the following installed (For Compiling and Running):   
+[Node.js x64](https://nodejs.org/en/)  
+[Python 2.7](https://www.python.org/downloads/)  
+[Visual Studio](https://www.visualstudio.com/en-us/products/visual-studio-community-vs.aspx)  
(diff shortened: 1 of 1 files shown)


bbondy:
Thanks for the pull request. 

- What does it fail on if you don't have Python? We aren't using it explicitly anywhere that I know of.
- I don't believe NSIS is a dependency anymore, there is some outdated info about an old installer we had in the README mentioning it but is being removed in a pull request I did earlier today. 
- Please add a new docs/prerequisites.md and link to it from the readme instead.

PFCKrutonium:
If you build without python, sqlite3 fails to compile.

I will remove the NSIS Link

Will do.

bbondy:
Thanks! btw you could also make changes, `git add -u`, `git commit --amend`, and then `git push -f` into your branch and it'll auto update the pull request.
Thanks again for doing this and for the follow up changes.

PFCKrutonium:
TBH I have no idea how to, I did this all online via the website. I'm going to do it all properly via normal git and make a new PR for my own sake.

bbondy:
that works too, thanks!

PFCKrutonium:
@bbondy Question, should I set up the file to easily document reqs for Linux/OS X? Also, does it compile on Linux/OS X yet?

bbondy:
Yes but we can put it all in the same doc under subsections

docs/prerequisites.md

```
# Prerequisites

## Windows

...
```

We can fill in other platforms over time.

PR #289, opened 2016-01-24
One Click Launch for Brave on Windows

PFCKrutonium:
Allows one click launching with the latest git build at all times.

### DIFF
--- a/LaunchBrave.bat
+++ b/LaunchBrave.bat
@@ -1,0 +1,9 @@
+@echo off
+
+REM Enables one-click launching of the latest and greates build at any time, though it takes a moment since it has to first check for an update
+REM and if found, download it. It will then compile it regardless, because git exit codes are useless. (0 is success, 1 is error.)
+
+git pull
+call CMD /C npm install
+call CMD /C start npm run watch
+call CMD /C start npm start


bbondy:
Sorry I don't think we should optimize in one script to run things 1 time only, would rather not clutter up the root of the directory.  Fine for you to fork and put in your fork though!

bgiesing:
Are we allowed to add this as a tip to the wiki @bbondy?

I have done a similar script for Linux that checks for updates first. If there's changes, it runs `git pull` and `npm install` first, otherwise it just runs the app. It also silences the npm output unless you add `-v`.

```bash
#!/bin/bash

LOCAL=$(git rev-parse @)
REMOTE=$(git rev-parse @{u})
BASE=$(git merge-base @ @{u})
verbose=

  case "$1" in
    -v|--verbose)
         verbose=1
         shift ;;
  esac

if [ $LOCAL = $REMOTE ]; then
    echo "No Updates"
elif [ $LOCAL = $BASE ]; then
    echo "Updating Brave..."
    git pull
    if [ "$verbose" = 1 ]; then
        npm install
    else
        npm install &> /dev/null
    fi
fi

echo "Starting Brave"
if [ "$verbose" = 1 ]; then
    npm run watch &
    npm start &
else
    {
        npm run watch &
        npm start &
    } &> /dev/null
fi
```

PR #292, opened 2016-01-24
add optional alternative port support through env variable BRAVE_PORT

bkardell:
add optional alternative port support through env variable BRAVE_PORT

### DIFF
--- a/app/index-dev.html
+++ b/app/index-dev.html
@@ -8,6 +8,7 @@
     <script>
       require("babel-polyfill")
       require("vibrant/dist/Vibrant.min.js")
+      const bravePort = require("./env-echo.js").getEnv("BRAVE_PORT");
     </script>
     <script src="./ext/l20n.min.js" defer></script>
     <link rel="localization" href="locales/{locale}/app.l20n">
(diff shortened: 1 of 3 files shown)


app/index-dev.html
```diff
@@ -22,12 +23,26 @@
...
     </div>
   </body>
-  <script src="http://localhost:8080/webpack-dev-server.js"
-          onerror="document.querySelector('#setupError').style.display ='block'"></script>
-  <script src="http://localhost:8080/built/bundle.js"></script>
+  <script>
+    const createScript = function (scriptPath) {
+      return new Promise(function (resolve, reject) {
+        var script = document.createElement("script")
+        script.type = "text/javascript"
```
bbondy: nit: for js pls use single quotes here and in this file. (html double is fine)

bbondy:
Looks good to me, pls adjust the 2 nits and I think we can merge it in.

bbondy:
Thanks for your work on this, this is great.

app/index-dev.html
```diff
@@ -8,6 +8,7 @@
     <script>
       require("babel-polyfill")
       require("vibrant/dist/Vibrant.min.js")
+      const bravePort = require("./env-echo.js").getEnv("BRAVE_PORT");
```
bbondy: Why not just process.env.BRAVE_PORT here? Less chars and 1 less file.

app/index-dev.html
bkardell: Yeah sure, I agree with this actually - generally I try to adopt the conventions in the file - the issue was that it didn't seem consistent already (requires for example used double above) so I just made a call.  I'll adjust.

app/index-dev.html
bkardell: Yeah, I was originally getting a little fancier with it, looking for a port instead of just using the env in which case I didn't want to repeat it just never backed it out all the way here and I was tired... will adjust.

bkardell:
Sorry, I messed this pull up via rebase, just closing it - will resubmit.

PR #299, opened 2016-01-24
Remove dependency on gnu rm for Windows builds

aviolentgrace:
Windows doesn't have the gnu coreutils (including rm) by default, as such this script (which powers 'npm run build-package') cannot run successfully without the user first installing GnuWin or a similar tool bundle and wiring up the PATH environment variable to point to the location of the rm command.

This change removes the dependency on rm and uses Windows' own rmdir instead.

It also checks to see if any deletions are necessary by prefacing with 'if exist', and it eliminates all of the dist/* deletions, as they seemed unnecessary. (Perhaps they're required for other workflows I haven't got to yet, but on my Windows 10 machine, I never saw a dist folder created by the build.

### DIFF
--- a/tools/buildPackage.js
+++ b/tools/buildPackage.js
@@ -19,24 +19,33 @@ var env = {
   NODE_ENV: 'production'
 }
 
-console.log('Cleaning build target')
+var cmds = ['echo cleaning up target...']
 
 if (isWindows) {
-  var cmds = [
-    'if exist ' + buildDir + ' rmdir /s /q ' + buildDir,
-  ]
-  execute(cmds, env, console.log.bind(null, 'done'))
+  cmds = cmds.concat([
+    '(if exist ' + buildDir + ' rmdir /s /q ' + buildDir + ')',
+    '(if exist dist\*.dmg del /q dist\*.dmg)',
+    '(if exist dist\*.nupkg del /q dist\*.nupkg)',
+    '(if exist dist\*.exe del /q dist\*.exe)',
+    '(if exist dist\*.msi del /q dist\*.msi)',
+    '(if exist dist\RELEASES del /q dist\RELEASES)',
+    '(if exist dist\*.zip del /q dist\*.zip)'
+  ])
 } else {
-  var cmds = [
+  cmds = cmds.concat([
     'rm -rf ' + buildDir,
-    'rm -f dist/*.dmg dist/*.nupkg dist/*.exe dist/*.msi dist/RELEASES dist/*.zip',
-  ]
-  execute(cmds, env, console.log.bind(null, 'done'))
+    'rm -f dist/*.dmg dist/*.nupkg dist/*.exe dist/*.msi dist/RELEASES dist/*.zip'
+  ])
 }
+
+cmds = cmds.concat([
+  'echo done',
+  'echo starting build...'
+])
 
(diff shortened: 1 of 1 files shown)


tools/buildPackage.js
```diff
@@ -19,11 +19,24 @@ var env = {
...
+    'if exist ' + buildDir + ' rmdir /s /q ' + buildDir,
+  ]
+  execute(cmds, env, console.log.bind(null, 'done'))
+} else {
+  var cmds = [
+    'rm -rf ' + buildDir,
+    'rm -f dist/*.dmg dist/*.nupkg dist/*.exe dist/*.msi dist/RELEASES dist/*.zip',
+  ]
+  execute(cmds, env, console.log.bind(null, 'done'))
+}
```
bbondy: Thanks.  Fairly sure this won't always work because execute runs async right now, please fix.

bbondy: Also no need to have a console message when the clean is done, it would be a bit misleading that way.

The easiest way to fix is to mutate the cmds array.

aviolentgrace: I think i'll nix the cleaning message due to the async nature of execute. (I'm fairly new at node.js, is there a way to make execute serialize for things like cleanup that should finish before continuing?)

bbondy: Please do this:

Put the `var cmds = []`

At the top

Then do `cmds.push(...)` for the remove call per platform.
Then below where you removed the commands do:
`cmds = cmds.append([...])`

aviolentgrace:
I made the recommended changes. Thanks!

tools/buildPackage.js
aviolentgrace: Ah, thanks! I just saw this after coming up with something similar (but
different). I can re-shape it to use append/push as appropriate. Let me
know!

aviolentgrace: I played around with push/append. Looks like `.append()` (for adding multiple elements to an array) doesn't exist in this version of js; i'm already using `.concat()` in my latest commit, so as long as the overall approach doesn't grate, it should work!

aviolentgrace:
As an aside, I'd love to know how to squash these two commits... I use git a lot locally for my own code so I'm no stranger to squash/fixup and rebasing my changes, but i'm new to the way GitHub organizes & names the branches. 

For example, I don't have a clue how to pull down this 'aviolentgrace:patch-2' branch GitHub is hosting for me... Is there an easy way for me to get branches (such as this patch-2 branch?) set up locally with remote tracking so that I can more easily work with these kinds of changes?

(In the future, I suppose I should just make a topic branch on my machine and then `push origin HEAD` from the get-go...)

bbondy:
I believe you can rebase and squash with `git rebase -i master`, you can also `git commit --amend` changes and `git push -f` into the branch which I believe you have access to, it will update the pull request automatically.

bbondy:
As for your question about how to test a random pull request locally, see:
https://help.github.com/articles/checking-out-pull-requests-locally/

aviolentgrace:
Thanks, although it seems the instructions there are out of date or don't work if they're  your own changes!

e.g., this step is invisible for me:

`At the bottom of the pull request, click command line. Follow the sequence of steps to bring down the proposed pull request.`

bbondy:
Looks good, thanks.

aviolentgrace:
Thanks for the help, Brian!

PR #301, opened 2016-01-24
Add optional support for alternative port

bkardell:
Add support for optional alternative port through BRAVE_PORT env setting

Apologies for all of the confusion on the pulls here, I'm somewhat embarrassed to admit that I've _mostly_ avoided the command line for github somehow thus far which is very unlike me - all of the craziness surrounding this was part of my unfortunate learning process.  I'm glad that getting myself into some trouble here provided an impetus to figure it out.

### DIFF
--- a/js/components/urlBar.js
+++ b/js/components/urlBar.js
@@ -15,6 +15,7 @@ const remote = global.require('electron').remote
 
 const UrlBarSuggestions = require('./urlBarSuggestions.js')
 const messages = require('../constants/messages')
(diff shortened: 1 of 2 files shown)


bbondy:
Sorry for the hassle but pls rebase / squash the 15 commits into one.

bkardell:
No, it's not a hassle, I was actually trying to do this and instead of shrinking, the number of commits just kept growing :-p  so... I'll figure that out later I guess.

PR #316, opened 2016-01-25
Replace string ref attributes with callbacks

razh:
The first commit removes the explicit requirement of `react-dom`'s `findDOMNode` to access the underlying DOM node of DOM components:

```jsx
<div ref="div" />

// Before
React.findDOMNode(this.refs.div)

// After
this.refs.div
```

The second commit switches to using callback attributes for refs (ignoring any unused ones). I left that one separate in case you wish to continue using string refs.


bbondy:
Great refactor, thanks for doing it!

PR #320, opened 2016-01-25
Adding prefix view-source on title in url bar.

Adejair:
view source prefix in bar


diracdeltas:
Note this still doesn't set the URL bar location to `view-source:..` but helpful anyways. thanks!

Issue #138, opened 2016-01-11
Make private tabs pass the grep test


bbondy:
visit a site you've never been to before (goodeggs.com) in a private tab and then grep for goodeggs in /Users/yan/Library/Application Support/Brave. Found matches in /Cache/data_1 and /Cache/data_3.

diracdeltas:
I think I figured out two reasons that this is happening:

1. favicons from private tabs (which are loaded outside of the webview partition) are hitting the cache. i confirmed that disabling `setFavicon` and `faviconUtil.getFavicon` made private tabs pass the grep test.
2. independently, searching via cmd+k while a private tab is active seems to be writing the search query to cache. this is probably because we are sending one XHR per keystroke to the search autocomplete URL from the urlbar, which is also outside of the webview partition.

bbondy:
Nice finds.

PR #367, opened 2016-01-27
0.7.10 release notes

bbondy:


diracdeltas:
+1

diracdeltas:
Does 0.7.10 include #138 ? if so that should go in there

diracdeltas:
nvm above comment, i see it in 0.7.9

PR #374, opened 2016-01-28
Windows About menu dialog

aekeus:


bbondy:
Can we just use the same dialog for win and osx? Please do and then merge, thanks.

bbondy:
I prefer to keep the electron base version there too if possible

aekeus:
Adding the electron version should not be a problem. We will need to create a custom form if we want the About windows to look identical between versions.

bbondy:
> Adding the electron version should not be a problem. We will need to create a custom form if we want the About windows to look identical between versions.

I just meant that I wanted the OSX and Windows ones to use the same dialog. Currently the OSX uses the built in role one which doesn't match.   Differences between windows on different Windows version is good and expected.

aekeus:
Added electron and chrome version numbers. Confirmed on both Windows and OSx.

bbondy:
Thanks @aekeus 
Brendan will be happy with this change!

bbondy:
Just noting the electron one should probably read: libchromiumcontent and not Chromium if you get a sec for a followup.

PR #385, opened 2016-01-28
Add hover effect on the "back forward" button

RacioN:
- Add hover effect on the "back forward" button.
(When I saw "back forward" button, I couldn't understanding in an instant.)

- Change url bar selection color.

### DIFF
--- a/less/variables.less
+++ b/less/variables.less
@@ -24,7 +24,7 @@
 @progressBarColor: #3498DB;
 @defaultFontSize: 13px;
 @audioColor: @highlightBlue;
-@focusUrlbarOutline: @highlightBlue;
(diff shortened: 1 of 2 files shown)


bbondy:
@bradleyrichter could you do a design review on this? Thanks!

bradleyrichter:
Thanks @RacioN - I'll review it today. Adding more prominent hover states to all buttons to the do-list.

bbondy:
Brad should this be closed? I think you mentioned it would be incorporated to your hover states changes.

bradleyrichter:
@bbondy yes, I like the look. Incorporating it into the next round of UI improvements.

PR #400, opened 2016-01-29
Fix Bravery menu to match new spec

### LINKED ISSUE #330: Cleanup Bravery menu
They should work like radio buttons:
screenshot 2016-01-26 09 50 29

- Replace ads does the safe ad insertion, adblock, and TP
- Block ads does the same as above but without the insertion
- Allow ads and tracking (sounds so filthy!:)) will just load pages without adblock and without TP.

diracdeltas:
This also refactors the Bravery menu into a common menu so it can be used
in the hamburger menu later.

Fix #330 

Auditors: @bbondy

### DIFF
--- a/js/commonMenu.js
+++ b/js/commonMenu.js
@@ -8,6 +8,11 @@ const AppConfig = require('./constants/appConfig')
 const AppActions = require('../js/actions/appActions')
 const messages = require('../js/constants/messages')
 
+const httpsEverywhere = 'httpsEverywhere'
+const adblock = 'adblock'
+const adInsertion = 'adInsertion'
+const trackingProtection = 'trackingProtection'
+
 let electron
 try {
   electron = require('electron')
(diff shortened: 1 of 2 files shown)


js/commonMenu.js
```diff
@@ -95,3 +100,66 @@ module.exports.findOnPageMenuItem = {
...
+module.exports.buildBraveryMenu = function (settings, init) {
+  const replaceAds = settings[adInsertion] || false
+  const blockAds = settings[adblock] || false
+  const blockTracking = settings[trackingProtection] || false
+  const useHttps = settings[httpsEverywhere] || false
+  return {
+    label: 'Bravery',
+    submenu: [
+      {
+        type: 'radio',
```
bbondy: I didn't know radio existed, nice.

bbondy:
++

PR #420, opened 2016-01-29
Add bravery menu to the hamburger menu

diracdeltas:
Auditors: @bbondy

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -111,7 +111,7 @@ class Main extends ImmutableComponent {
     })
     // whether the current page is bookmarked. needed to re-initialize the
     // application menu.
-    settings.bookmarked = this.refs.navBar.bookmarked
+    settings.bookmarked = this.navBar.bookmarked
     contextMenus.onHamburgerMenu(settings)
   }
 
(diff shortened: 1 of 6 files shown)


js/components/main.js
```diff
@@ -102,6 +102,19 @@ class Main extends ImmutableComponent {
...
+  onHamburgerMenu () {
+    let settings = {}
+    const resourceNames = ['httpsEverywhere', 'adblock', 'trackingProtection', 'adInsertion']
+    resourceNames.forEach((name) => {
+      let enabled = this.props.appState.getIn([name, 'enabled'])
+      settings[name] = enabled === undefined ? AppConfig[name].enabled : enabled
+    })
+    // whether the current page is bookmarked. needed to re-initialize the
+    // application menu.
+    settings.bookmarked = this.refs.navBar.bookmarked
```
diracdeltas: I guess we're no longer using refs? lmk if i should update this to be consistent with https://github.com/brave/browser-laptop/pull/316/files

bbondy: yes the approach in https://github.com/brave/browser-laptop/pull/316/files is better thanks.

js/components/main.js
```diff
@@ -102,6 +102,19 @@ class Main extends ImmutableComponent {
     // TODO
   }
 
+  onHamburgerMenu () {
+    let settings = {}
+    const resourceNames = ['httpsEverywhere', 'adblock', 'trackingProtection', 'adInsertion']
```
bbondy: would be better if we pulled these in from a constants file in case the names ever change.

bbondy:
Couple of minor things, please feel free to merge with edits.

PR #433, opened 2016-01-29
Send updater query parameters based on calendar date

aekeus:
* Compare calendar dates when sending daily flag instead of 86400 seconds
  * Send first=true if first update request
  * Use moment.js for date handling

### DIFF
--- a/app/dates.js
+++ b/app/dates.js
@@ -1,0 +1,5 @@
+const moment = require('moment')
+
+exports.todayYMD = () => {
+  return moment().format('YYYY-MM-DD')
+}
--- a/app/updater.js
+++ b/app/updater.js
@@ -14,6 +14,7 @@ const querystring = require('querystring')
 const AppStore = require('../js/stores/appStore')
 const AppActions = require('../js/actions/appActions')
 const Immutable = require('immutable')
+const dates = require('./dates')
 
 const fs = require('fs')
 const path = require('path')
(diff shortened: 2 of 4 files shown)


bbondy:
Will this mean that many instances will report first-time twice in the same day when this lands? If so can that be fixed?

app/dates.js
```diff
@@ -0,0 +1,15 @@
...
+}
+
+exports.todayWOY = () => {
+  return moment().isoWeek()
+}
+
+// Months a returned zero based from moment
+// We add 1 to make sure January does not fail a truth test
+exports.todayMonth = () => {
+  return moment().month() + 1
```
bbondy: I hate to suggest this since you already tested, but moment adds another 1MB to the package. Is it worth it for this? I think this is possible with plain JS.

app/updater.js
```diff
@@ -86,67 +88,44 @@ exports.init = (platform, ver) => {
...
 }
 
-const secondsPerDay = 24 * 60 * 60
-const secondsPerWeek = secondsPerDay * 7
-const secondsPerMonth = secondsPerDay * 30
-
 // Build a set of three params providing flags determining when the last update occurred
 // This is a privacy preserving policy. Instead of passing personally identifying
-// information, the browser will pass the three boolean values indicating when the last
+// information, the browser will pass thefour boolean values indicating when the last
```
bbondy: nit: typo spacing

bbondy:
Merging this for now and then doing release. Please do the moment removal commit for next release though.

PR #482, opened 2016-01-31
Initial window size to workspace size

### LINKED ISSUE #477: Initial window size far too small
On Mac, after initial install and launching brave 0.7.11 the browser window is tiny. Make the default window size bigger.

bbondy:
Fix #477

Reviewers: @bridiver

I'm fairly sure the should.eventually.be tests aren't doing anything because they didn't fail, but updated to what it should be for now until we look into that.  (I did rebuild webpack btw)


bbondy:
Merging but please review when you have time.

Issue #481, opened 2016-01-31
suppress (brief flash of ) tab loading indicator on new tabs 


asadotzler:
On the 0.7.11 build on Mac, new tabs briefly flash a loading indicator on the left side before hiding it. Makes creating new tabs feel janky. Probably also a good idea to suppress the pageload time displayed in the addressbar. This is a blank tab and it shouldn't feel or label itself as slower than "instant".

diracdeltas:
i tried to get rid of this jank while fixing https://github.com/brave/browser-laptop/issues/373 but haven't figured out how to make 'about:blank' load in <10ms. other than not setting the `src` attribute on the webview at all, but that caused a lot more bugs.

bbondy:
let's just not show the loading indicator for about: sourced URLs.

psimyn:
this can be closed now

PR #553, opened 2016-02-02
don't show loading spinner for 'about:' urls, prevents brief flicker …

psimyn:
…on new tab creation mentioned in https://github.com/brave/browser-laptop/issues/481

### DIFF
--- a/docs/windowActions.md
+++ b/docs/windowActions.md
@@ -110,13 +110,15 @@ Sets a frame as pinned
 
 
 
-### onWebviewLoadStart(frameProps) 
+### onWebviewLoadStart(frameProps, indicateLoading) 
 
 Dispatches a message to the store to indicate that the webview is loading.
 
 **Parameters**
 
 **frameProps**: `Object`, The frame properties for the webview in question.
+
+**indicateLoading**: `boolean`, should a loading spinner be shown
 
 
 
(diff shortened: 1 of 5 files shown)


bbondy:
Looks great, can you also document this in `docs/state.md`?  After that I'll gladly merge :+1:

psimyn:
sure, does that mean adding a comment to loading?

```
    loading: boolean, // show the loading indicator
```

Is there a way to run `npm test` in minimized mode? I tried running it locally before pushing but the constant flashing/focus-stealing forced me to kill it

bbondy:
Sorry actually please either store `indicateLoading` in state and then use that to determine whether to render. Or an alternate way would be just to check the frameProps's location protocol inside the render directly as well.  The way you have it might have slightly better perf though, but the other way would be a more simpler patch.

psimyn:
ah, putting in loading getter of tab component is much simpler. Thoughts on this:

```
  get loading () {
    return this.props.frameProps &&
    this.props.frameProps.get('loading') &&
+    !this.props.frameProps.get('location').startsWith('about:')
  }
```

Should this be different PR/squashed if I'm going to revert other commit?

bbondy:
yep that's what I was thinking, let's go with that exactly.
Whatever's easier for you.  If you want to close and open a new one or change and then git commit --amend and force push to your branch.

psimyn:
forget to revert when amending, then abandoned ship. New PR at https://github.com/brave/browser-laptop/pull/554 :)

PR #554, opened 2016-02-02
don't show loading spinner for 'about:' urls, prevents brief flicker …

psimyn:
…on new tab creation, from #481


bbondy:
++ thanks!

Issue #248, opened 2016-01-22
Context menus for links don't handle onClick links or links with that contain span, etc...


bridiver:
The context menu will only recognize a link and get the correct source for the most basic case of a simple link with only text inside it.

PR #566, opened 2016-02-02
Context menu for non-text links

psimyn:
Fixes https://github.com/brave/browser-laptop/issues/248

- Check parentNodes to see if the Link context menu items should be shown. 
- Allow showing both link and image menu items (for image-links)
- Show context menu for non-text input types

### DIFF
--- a/app/content/webviewPreload.js
+++ b/app/content/webviewPreload.js
@@ -197,9 +197,21 @@ function isPlatformOSX () {
 
 document.addEventListener('contextmenu', (e) => {
   var name = e.target.nodeName.toUpperCase()
+
+  var href
+  var maybeLink = e.target
+  while (maybeLink.parentNode) {
+    if (maybeLink.nodeName.toUpperCase() === 'A') {
+      href = maybeLink.href
+      break
+    }
+    maybeLink = maybeLink.parentNode
+  }
+
   var nodeProps = {
     name: name,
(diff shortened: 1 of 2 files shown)


app/content/webviewPreload.js
```diff
@@ -156,7 +156,8 @@ ipc.on(messages.SET_AD_DIV_CANDIDATES, function (e, adDivCandidates, placeholder
 })
 
 function hasSelection (node) {
-  if (node && node.selectionStart !== undefined &&
+  if (node && /password|search|tel|text|url/.test(node.type) &&
```
bridiver: I'm not sure if we need this extra check, won't it prevent the copy menu from working on plain text? Is there a case that is broken that this fixes?

bridiver:
Thanks for the PR! This is a much needed fix. Just had one small question on it, but otherwise looks great to me. It would be awesome to get some tests for these.

app/content/webviewPreload.js
psimyn: Was specifically for input types. Currently error on input type submit, for example

psimyn:
@bridiver addressed/fixed question - thanks for picking that up!

bridiver:
Thanks! There are currently some conflicts, if you can resolve those and bring it up-to-date I'll get it merged

psimyn:
done

bridiver:
Merged, thanks!

PR #581, opened 2016-02-02
replace webviewPreload ipc with postMessage

bridiver:
@bbondy

### DIFF
--- a/app/content/webviewPreload.js
+++ b/app/content/webviewPreload.js
@@ -2,9 +2,28 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var ipc = require('electron').ipcRenderer
 var messages = require('../../js/constants/messages')
 var KeyCodes = require('../../js/constants/keyCodes')
+
+var ipc = {
+  events: {},
+  embedder: null,
+  processMessage: function (event) {
+    if (event.origin === 'file://') {
+      this.embedder = event.source
+      var cb = this.events[event.data[0]]
+      cb && cb.apply(null, event.data)
+    }
+  },
+  on: function (name, cb) {
+    this.events[name] = cb
+  },
+  send: function () {
+    var args = Array.prototype.slice.call(arguments)
+    this.embedder && this.embedder.postMessage(args, 'file://')
+  }
+}
+addEventListener('message', ipc.processMessage.bind(ipc)) // eslint-disable-line
 
 /**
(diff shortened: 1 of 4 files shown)


bridiver:
this needs to be amended to fix ipc.send for the key events

bridiver:
now it's ok

bbondy:
Instead of all the eslint disabling pls use `window`. as a prefix

bbondy:
Looks good to me, I'll merge.
@diracdeltas please do a security audit too, if any issues we can do in follow ups.

app/content/webviewPreload.js
```diff
@@ -2,10 +2,29 @@
...
 
-var ipc = require('electron').ipcRenderer
 var messages = require('../../js/constants/messages')
 var KeyCodes = require('../../js/constants/keyCodes')
 
+var ipc = {
+  events: {},
+  embedder: null,
+  processMessage: function (event) {
+    if (event.origin === 'file://') {
```
diracdeltas: Can these origin checks be for `file:///full/path/to/index.html` instead of just `file://`? Otherwise any HTML file that the user has downloaded and opened in a tab/iframe could set itself as the embedder.

app/content/webviewPreload.js
```diff
@@ -2,10 +2,29 @@
...
+      this.embedder = event.source
```
diracdeltas: Should be `this.embedder` only be set if it's not set already? (is it ever expected to change?)

app/content/webviewPreload.js
bridiver: It only sends file:// unfortunately. The updates I'm making now to Electron will fire a "HELO" message to register the embedder as soon as the script context is created and before the page will have any opportunity to preempt it.

js/components/frame.js
```diff
@@ -54,6 +54,10 @@ class Frame extends ImmutableComponent {
 
   componentDidMount () {
     this.updateWebview()
+    // forward postMessage events from webview to webContents
+    window.addEventListener('message', function (event) {
```
bridiver: I think I should add an origin check here. The page itself will have no access to the embedder and can't send it any messages, but I think this could end up firing multiple messages for the same event because "window" is the same for all frames. Going to create a quick patch for it right now.

bridiver: added in https://github.com/brave/browser-laptop/pull/599

app/content/webviewPreload.js
bridiver: No, you're right, it should only be set once. I had that in there, but I think I took it out when things weren't working right and apparently forgot to put it back.

Issue #587, opened 2016-02-03
"open in new tab" should open new tab in background


stshank:
When right-clicking on a link and opening it in a new tab, the current web page should remain the active one. Instead, Brave switches to the new tab. Perhaps this is deliberate, but it's not what I prefer, and it's not what mainstream browsers do.

PR #592, opened 2016-02-03
'Open in new tab' shouldn't focus newly opened tab

psimyn:
Now matches behaviour of cmd/ctrl click.

Fixes https://github.com/brave/browser-laptop/issues/587

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -86,7 +86,7 @@ class Main extends ImmutableComponent {
         location: url || Config.defaultUrl,
         isPrivate: !!options.isPrivate,
         isPartitioned: !!options.isPartitioned
-      })
+      }, options.openInForeground)
 
       // Focus URL bar when adding tab via shortcut
       electron.remote.getCurrentWebContents().send(messages.SHORTCUT_FOCUS_URL)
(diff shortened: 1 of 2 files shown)


js/components/main.js
```diff
@@ -86,7 +86,7 @@ class Main extends ImmutableComponent {
         location: url || Config.defaultUrl,
         isPrivate: !!options.isPrivate,
         isPartitioned: !!options.isPartitioned
-      })
+      }, options.openInForeground)
```
bbondy: You'll have to either update all the callers of `SHORTCUT_NEW_FRAME` to preserve their old behaviour, or do:
`options.openInForeground === undefined ? true : option.openInForeground`

bbondy:
Thanks for the patch, looking good, just one change pls.

js/components/main.js
bbondy: Btw you can check with `git grep SHORTCUT_NEW_FRAME`

psimyn: is that just for added safety? WindowActions.newFrame defaults the argument to true (including if undefined)

bbondy: Oops, you're right! I'll merge as is, thanks.

PR #619, opened 2016-02-05
Block 3rd party cookies and refer(r)er(r)

diracdeltas:
Note that the cookies/referer still appear in network devtools with a "provisional headers" warning. however, running wireshark on slashdot.org shows that cookies/referer to slashdot are sent and cookies/referer to other origins are not.

TODO: add pref to turn this off/on

Auditors: @bbondy

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -32,30 +32,26 @@ function registerForSession (session) {
     }
 
     let results
-    let cbArgs = {}
     for (let i = 0; i < filteringFns.length; i++) {
       let currentResults = filteringFns[i](details)
       if (currentResults && !module.exports.isResourceEnabled(currentResults.resourceName)) {
         continue
       }
       results = currentResults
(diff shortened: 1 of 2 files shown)


app/filtering.js
```diff
@@ -31,36 +32,44 @@ function registerForSession (session) {
...
 
     let results
-    let cbArgs
     for (let i = 0; i < filteringFns.length; i++) {
       let currentResults = filteringFns[i](details)
       if (currentResults && !module.exports.isResourceEnabled(currentResults.resourceName)) {
         continue
       }
       results = currentResults
-      cbArgs = cbArgs || results.cbArgs
```
diracdeltas: I got rid of this because `results.cbArgs` seemed to always be undefined or `{}`

bbondy:
++

bbondy:
Please hookup a pref for it (store it in preferences from now on for things like this). Just in case sites break.

PR #625, opened 2016-02-05
show url location on link hover. may resolve #528

### LINKED ISSUE #528: implement link hover URL preview
When I hover over a link, I'd like to see a preview of the address I'm about to visit. May depend on status bar implementation at #328 (or not!)

psimyn:
this will likely need some work, but seems to work.

screen shot 2016-02-06 at 04 00 44

- Possibly a wrong way to go about it
- There is undoubtedly a better way to handle event delegation in webview


bbondy:
Awesome! Adding @bridiver to review since he's refactoring some of this related code. He might have some tips.

psimyn:
also related - I don't think https://github.com/atom/electron/issues/1080 is resolved but that may be the more appropriate place for events

therealglazou:
This is fine unless the pointer happens to hover over a link precisely under the generated `hrefPreview` div element. In that case, you may want to position the div not at `letf: 0px`but at `right: 0px`...

therealglazou:
So the best way to implement what I said above is probably to add an event listener on hover on hrefPreview that toggles a class on the div on and off. Then you have two CSS rules for hrefPreview, one with the class and one without, the former being 'right: 0px' and the latter 'left: 0px'. You can't directly use a single :hover css rule because that won't easily allow to switch between the two different settings and you still need to handle the case where the link is under an hrefPreview sticking to the right side of the viewport...

psimyn:
Thanks @therealglazou! I think to toggle the class on hover should cover most cases (if it defaults to the left, then a "right" modifier class sets left auto).

Haven’t been able to test yet; electron was being a bit funny with ipc messages not triggering, may need to update. I'll try hover test later today

bridiver:
is this going to be a performance issue? The mouseover event will walk up the dom to the root (potentially) for every node you mouseover. I think @psimyn is right that there is a better way to do this inside Electron, but at least in the short-term I don't see any issues (other than performance potentially) here after resolving the conflicts.

psimyn:
@therealglazou I ended up sending the clientX and clientY from the mouseEvent, then determining whether or not to show it on left or right. Any listeners on hrefPreview caused flickering, as the mouseout from the link itself was fired and made it disappear. If this way seems horrendous can try something different. This seems to handle everything I've tried so far

@bridiver unsure of the performance hit. Have only done limited scroll/hover testing so far. I considered adding a ~50ms timeout before displaying, then clearing it. This would eliminate when mousing around or scrolling. I'll see if I can notice an actual issue first.

Do you want commits to be squashed pre-merge?

bridiver:
Sorry for the delayed response. Squashing them would be great, I try to keep everything in a single commit. We're doing some performance optimizations right now and I'd like to profile this before merging to see if adding a slight delay is necessary or not.

bridiver:
seems to be ok

bbondy:
very cool, thanks @psimyn!

PR #632, opened 2016-02-06
use content scripts instead of preload

bridiver:
https://github.com/brave/electron/pull/10

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -35,7 +35,7 @@ class Frame extends ImmutableComponent {
     this.webview = this.webview || document.createElement('webview')
     this.webview.setAttribute('allowDisplayingInsecureContent', true)
     this.webview.setAttribute('data-frame-key', this.props.frame.get('key'))
-    this.webview.setAttribute('preload', 'content/webviewPreload.js')
+    this.webview.setAttribute('contentScripts', 'content/webviewPreload.js')
     if (this.props.frame.get('isPrivate')) {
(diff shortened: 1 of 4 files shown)


bbondy:
This currently breaks preferences

bridiver:
Probably because I removed the message event listener. It wasn't needed for the webview preload anymore and I didn't think about the preferences

js/components/frame.js
```diff
@@ -54,12 +54,6 @@ class Frame extends ImmutableComponent {
 
   componentDidMount () {
     this.updateWebview()
-    // forward postMessage events from webview to webContents
-    window.addEventListener('message', function (event) {
```
bridiver: This probably broke the preferences. If we are going to keep using contentWindow for them we should lock down the origin for it

js/components/frame.js
```diff
@@ -54,12 +54,6 @@ class Frame extends ImmutableComponent {
...
-      if (this.webview.getAttribute('src').match(event.origin)) {
```
bbondy: Yep per slack match origin will also match things like brianbondy.com#file:// and allow it. So this should be startsWith I think and also I'm not sure how to lock this to a full path and not just the file protocol.

bridiver: does setting document.domain work?

bbondy: document.domain will be set to blank and also I think origin for a file URL will always return file://

bbondy:
Merged this and will fix preferences in master

PR #646, opened 2016-02-07
Add line-height at .tabTitle

RacioN:
It's minor change.
Improve visibility when using japanese.


bradleyrichter:
@RacioN Great fix. Thanks!

@bbondy Let's take this as soon as its convenient.

bbondy:
@RacioN Thanks for the fix, and @bradleyrichter thanks for the design review!

PR #651, opened 2016-02-07
tabPages should not show if only pinned tabs open. fixes #647

### LINKED ISSUE #647: Pinned tab height and position of new tab icon inconsistencies
When I have only pinned tabs in my browser the height is off.
Also when there are only pinned tabs the 'new tab' icon spontaneously moves to the far right hand side of the nav bar.

screen shot 2016-02-07 at 7 16 19 pm
screen shot 2016-02-07 at 7 16 27 pm

psimyn:
shrinks the tab page next/prev buttons a bit, but more sensible display for the add new tab button

### DIFF
--- a/js/components/tabPages.js
+++ b/js/components/tabPages.js
@@ -33,10 +33,10 @@ class TabPages extends ImmutableComponent {
     return <div
       className={cx({
         tabPages: true,
-        singlePage: this.tabPageCount === 1
+        singlePage: this.tabPageCount <= 1
       })}>
     {
-      this.tabPageCount !== 1 &&
+      this.tabPageCount > 1 &&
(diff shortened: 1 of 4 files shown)


bradleyrichter:
@psimyn I'm still adjusting this tab page buttons but this sounds like a good improvement. If convenient, can you please add a screen shot?

And thanks for the pinned tabs/sets fix. Good idea!

-Brad

bbondy:
No screenshot needed here @bradleyrichter it's just a bugfix where we show extra tab bar height when there are no non-pinned tabs. I noticed this as well the other day but didn't have time to fix. 

@psimyn thanks for the fix!

bbondy:
Fixes work great, merging.

Issue #657, opened 2016-02-07
Brave shows blank screen when accessing certain sites


thykkutathil:
When accessing www.newsblur.com on the MacOS build, the Browser Console shows "Inspected target disconnected." and the screen is entirely blank. I am able to open a new Window and resume normal browsing.
![screen shot 2016-02-07 at 9 57 51 am](https://cloud.githubusercontent.com/assets/143114/12873023/4a21db64-cd81-11e5-8010-38c75a3f0f47.png)

Version: 0.7.13
Electron: 0.36.7
libchromiumcontent: 47.0.2526.110

MacOS El Capitan 10.11.3

TheFallenOneGOTH:
Partial confirmation on a Windows 10 non-touchscreen platform. Same Brave specs.

Going to NewsBlur loads for me for a few seconds before going to a black screen and becoming unresponsive.

bbondy:
Due to vibrant.js and broken favicons.
Verified this is fixed now that we don't use vibrant.js anymore.
https://github.com/brave/browser-laptop/commit/4d9c80dbdd0db582b1159e657f7b4991877eb204

Issue #600, opened 2016-02-03
Visiting the following page completely stops browser on MacOS


butlermatt:
Trying to view this article:
http://www.xaverian.ca/2014/10/30/mother-webbs-exposed-post-on-mother-webbs-facebook-page-brings-to-light-accounts-of-sexual-harassment/

Brave goes to a blank gray screen. Cannot close it, cannot quit. Need to use kill -KILL <pid> to end the session after which I can then close the window and restart.

Mac OS X 10.11.3
Brave: 0.7.12

Codes4Fun:
I reproduced this on Window 7, except it goes black (navigation bar included).

I traced it down to this line of html:
```
<link rel="Shortcut Icon" href="http://www.xaverian.ca/wp-content/themes/xaverian/images/favicon.ico" type="image/x-icon">
```

Codes4Fun:
The icon file itself seems to be bugged, it reports a width of 256x256 but the internal PNG is 16x16, if I take the icon into a hexeditor and change bytes 6 and 7 to 0x10 (16) the bug doesn't happen. Looking at other browsers, the other browsers don't use that icon, I'm trying to find which icons they do use and where it is located.

Codes4Fun:
So the icon the other browsers use come from this html:
```
<link rel="icon" href="http://www.xaverian.ca/wp-content/uploads/2015/09/cropped-Circle-logo-32x32.png" sizes="32x32">
```
Chrome uses the first icon it finds in the html and if it fails tries to use the next icon.
Firefox uses the last icon and if it fails displays no icon.
So it looks like Brave simply needs to detect the bad icon and respond accordingly.

Codes4Fun:
This issue seems to be with vibrant.js drag/dropping the icon on their site http://jariz.github.io/vibrant.js/ makes chrome display the sad face, but firefox throws an exception cmap.vboxes is undefined.

In Brave if I pass an empty string instead of getFavicon the issue doesn't happen, the icon doesn't show up on the tab.

bbondy:
Nice find! You guys made it much easier to fix, I'll take the fix.

bbondy:
So I found that it hangs as mentioned before in Vibrant, but more specifically its' not vibrant's fault, it's libchromiumcontent's context.drawImage when passed an invalid image 
`context.drawImage(b, 0, 0, 256, 256)`

[...]

PR #667, opened 2016-02-08
Computed theme color

bbondy:
Going to merge all of this, but @diracdeltas please review when you get a chance.


diracdeltas:
lgtm

diracdeltas:
this is showing different colors for my two gmail tabs
screen shot 2016-02-09 at 2 11 04 pm
screen shot 2016-02-09 at 2 10 57 pm

bbondy:
beige/yellow color present at all towards the top -- left, middle or right?
Does page have a theme color meta tag?
Reproducible ?

diracdeltas:
the yellow is from the gmail "loading" status bar at the top middle, which
appears for a second and then goes away. i'm guessing the color is getting
set during the interval where the bar is visible.

bbondy:
We could perhaps keep a list of overrides for tab color per host. Thoughts?

diracdeltas:
That could work. Curious why we got rid of Vibrant; i generally liked those colors better

bbondy:
Vibrant uses a lot of CPU, selected wrong colors sometimes and caused some pages to crash the renderer process which there is no easy fix to, see:
https://github.com/brave/browser-laptop/issues/657
https://github.com/brave/browser-laptop/issues/600

See my note about context.drawImage.

PR #670, opened 2016-02-08
use oip endpoint

bridiver:
@bbondy @diracdeltas

### DIFF
--- a/js/constants/config.js
+++ b/js/constants/config.js
@@ -7,7 +7,7 @@
 // https://vault-staging.brave.com for a dev build
 // http://localhost:3000 for production
 const vaultHost = process.env.VAULT_HOST || 'https://vault-staging.brave.com'
-const adHost = process.env.AD_HOST || '//cdn.brave.com'
+const adHost = process.env.AD_HOST || 'https://oip.brave.com'
 
 export default {
   zoom: {
--- a/app/content/webviewPreload.js
+++ b/app/content/webviewPreload.js
@@ -82,33 +82,45 @@
     var segment_expiration_time = 0 // no expiration
 
     // ref param for referrer when possible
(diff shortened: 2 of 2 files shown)


js/constants/config.js
```diff
@@ -7,7 +7,7 @@
 // https://vault-staging.brave.com for a dev build
 // http://localhost:3000 for production
 const vaultHost = process.env.VAULT_HOST || 'https://vault-staging.brave.com'
-const adHost = process.env.AD_HOST || '//cdn.brave.com'
+const adHost = process.env.AD_HOST || '//oip.brave.com'
```
bbondy: Is this env var only used at webpack time?  I.e. does not have an effect for builds after  `npm run build` package is done and you run it?

bridiver: I'm pretty sure it is only used at runtime so you should be able to set it whenever

bbondy: I think it's resolved at webpack time, which I think is preferred.

diracdeltas: Any reason not to just change this to https instead of a protocol-relative URL? Otherwise a passive MITM sees information about a user's ad preferences that otherwise wouldn't be exposed.

bridiver: Cost and overhead I suppose. There is no assumption that the data is going to a trusted 3rd-party so the only reason we use https at all is to avoid mixed content.

diracdeltas: HTTPS = a non-trusted third party sees your data
HTTP = anyone on your local network, your ISP, the NSA, etc sees your data
so there is still a security gain from HTTPS.

see arguments at https://istlsfastyet.com/ against cost/performance overhead of HTTPS.

bridiver: Well, the point is that the browser should consider anything it exposes to be "public knowledge" and make decisions accordingly. I thought there was an additional cost in dollars at Fastly for SSL and that's what I was concerned about, but looking back at their pricing page I guess there isn't so I suppose there is no downside to always having SSL always on

diracdeltas: :+1:

PR #678, opened 2016-02-09
center title within window. fixes #671

### LINKED ISSUE #671: Page title is slightly not centered
Measure the pixel count on the left and the right with Commadn+Shift+4
It is off by about 10 pixels I think because of the lock icon.

psimyn:
remove min-width from urlBarIcon when no icon is set


bbondy:
This is great, thanks!

PR #679, opened 2016-02-09
build index-dev.html with webpack allows custom port to work with CSP

### LINKED ISSUE #664: Running on an alternate port is currently broken in dev mode
The CSP which was added on HTML files currently makes the browse not start with alternate dev ports.
I can't think of an easy way to fix but  hopefully someone else can.

Otherwise maybe we should consider a less common port to use and maybe give people instructions about how to modify the CSP meta tags for local dev on a diff port.

psimyn:
Fixes #664

`npm run watch` won't output the file, so could use some docs/package.json changes


bridiver:
I'm going to pull some of this into the changes into #753

bridiver:
Thanks for this btw, it solves another problem I was dealing with. I'm not seeing the generated file when running the webpack dev server, am I missing something?

psimyn:
The webpack config changes in that PR look great! 

Are you running watch-all? dev server doesn't write anything to disk, so need either watch all or build test to output the file.

bridiver:
Yea, I realized that it doesn't write anything out with only the dev server running and that turns out to be ok with the historyFallback option. Otherwise it sometimes can't reload properly when running everything through localhost

PR #736, opened 2016-02-11
default title to the url

bridiver:
requires https://github.com/brave/electron/commit/f260d583217a7b752c29f3b67dcc8b62bd2f36b7
@bbondy

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -140,29 +140,28 @@ class Frame extends ImmutableComponent {
     this.webview.addEventListener('new-window', (e, url, frameName, disposition, options) => {
       e.preventDefault()
 
-      const guestInstanceId = e.options && e.options.webPreferences && e.options.webPreferences.guestInstanceId
-      const windowOptions = e.options && e.options.windowOptions || {}
-      windowOptions.parentWindowKey = remote.getCurrentWindow().id
-      windowOptions.disposition = e.disposition
+      let guestInstanceId = e.options && e.options.webPreferences && e.options.webPreferences.guestInstanceId
+      let windowOpts = e.options && e.options.windowOptions || {}
+      windowOpts.parentWindowKey = remote.getCurrentWindow().id
+      windowOpts.disposition = e.disposition
+      let delayedLoadUrl = e.options && e.options.delayedLoadUrl
+
+      let frameOpts = {
+        location: e.url,
+        parentFrameKey: this.props.frame.get('key'),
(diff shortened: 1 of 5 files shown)


js/actions/windowActions.js
```diff
@@ -213,6 +213,9 @@ const WindowActions = {
     if (frameOpts.location && UrlUtil.isURL(frameOpts.location)) {
       frameOpts.location = UrlUtil.getUrlFromInput(frameOpts.location)
     }
+    // default the title to the location
+    frameOpts.title = frameOpts.title || frameOpts.location
```
bbondy: I don't think this is correct, why is it needed? 
We render title based on if there is no title then use location.

bbondy:
Is there an issue on file for the motivation for these changes?

js/actions/windowActions.js
bridiver: When a new tab/window is created with an opener the location isn't sent initially. We create a new about:blank tab and when it's ready the opener sets the location. I am now sending the "delayedLoadUrl" with the new window/tab event so we can display the url until the page is loaded and the real title is available.

app/about-blank.html
```diff
@@ -0,0 +1,11 @@
+<html>
```
bbondy: about:blank should be a blank html document
screenshot 2016-02-11 16 18 02

bridiver: yea, I'm not sure if this change is needed now

bridiver:
removed some changes that didn't actually turn out to be necessary

js/components/frame.js
```diff
@@ -140,29 +140,28 @@ class Frame extends ImmutableComponent {
...
+      windowOpts.disposition = e.disposition
+      let delayedLoadUrl = e.options && e.options.delayedLoadUrl
+
+      let frameOpts = {
+        location: e.url,
+        parentFrameKey: this.props.frame.get('key'),
+        isPrivate: this.props.frame.get('isPrivate'),
+        partitionNumber: this.props.frame.get('partitionNumber'),
+        // use the delayed load url for the temporary title
+        title: delayedLoadUrl,
```
bbondy: Since this i set to the title in the state I think it will cause title mode to display still.

Usually when a page has no title the UI should not go into title mode. 
I.e. showing

somedomain.com | http://somedomain.com/test 
would be wrong and it's preferred to just show the urlbar.

bridiver:
any other issues with this one?

bbondy:
looks good to me

PR #744, opened 2016-02-12
Add rudimentary cert warning pages

diracdeltas:
Currently there is no way to click through the warning or inspect the cert. Will add later.

Test sites: https://lenovo.com and pages at https://badssl.com

Auditors: @bbondy

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -80,9 +80,9 @@ app.on('ready', function () {
     // window webcontents, not the webview webcontents
     BrowserWindow.getAllWindows().map((win) => {
       win.webContents.send(messages.CERT_ERROR, {
-        url: url,
+        url,
         certError: error,
-        cert: cert
+        cert
       })
     })
   })
--- a/app/index.js
+++ b/app/index.js
@@ -75,6 +75,17 @@ const saveIfAllCollected = () => {
 }
 
 app.on('ready', function () {
+  app.on('certificate-error', function (e, webContents, url, error, cert, cb) {
+    // Tell the page to show an unlocked icon. Note this is sent to the main
+    // window webcontents, not the webview webcontents
+    BrowserWindow.getAllWindows().map((win) => {
+      win.webContents.send(messages.CERT_ERROR, {
+        url: url,
+        certError: error,
+        cert: cert
+      })
+    })
+  })
   app.on('window-all-closed', function () {
     // On OS X it is common for applications and their menu bar
     // to stay active until the user quits explicitly with Cmd + Q
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -25,16 +25,14 @@ class Frame extends ImmutableComponent {
 
   updateWebview () {
     let src = this.props.frame.get('src')
-    const isAboutURL = isSourceAboutUrl(src)
-    if (isAboutURL) {
-      src = getTargetAboutUrl(src)
-    }
     let contentScripts = ['content/webviewPreload.js']
-    if (this.props.frame.get('location') === 'about:preferences') {
+    // NOTE: use src instead of location here because location is sometimes
+    // not updated when redirecting to about:certerror
+    if (src === 'about:preferences' || src === 'about:certerror') {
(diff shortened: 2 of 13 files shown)


js/components/urlBar.js
```diff
@@ -159,6 +159,18 @@ class UrlBar extends ImmutableComponent {
     })
     // escape key handling
     ipc.on(messages.SHORTCUT_ACTIVE_FRAME_STOP, this.onActiveFrameStop.bind(this))
+    ipc.on(messages.CERT_ERROR, (e, details) => {
+      // Sometimes the cert error fires before the active frame location
+      // has been updated, so wait 100ms.
+      window.setTimeout(() => {
```
bbondy: Seems like this might lead to a race condition that happens infrequently

js/components/frame.js
```diff
@@ -25,16 +25,14 @@ class Frame extends ImmutableComponent {
...
     let src = this.props.frame.get('src')
-    const isAboutURL = isSourceAboutUrl(src)
-    if (isAboutURL) {
-      src = getTargetAboutUrl(src)
-    }
     let contentScripts = ['content/webviewPreload.js']
-    if (this.props.frame.get('location') === 'about:preferences') {
+    // NOTE: use src instead of location here because location is sometimes
+    // not updated when redirecting to about:certerror
+    if (src === 'about:preferences' || src === 'about:certerror') {
```
bbondy: Noting that because of this change if we ever link to a page within an about: page which replaces the about page (i.e. not target=_blank) then that page will have elevated privs.  Seems a bit scary.  
Likewise if a content page links to about:preferences or about:certerror then the navigated frame won't work.

app/index.js
```diff
@@ -75,6 +75,17 @@ const saveIfAllCollected = () => {
...
 
 app.on('ready', function () {
+  app.on('certificate-error', function (e, webContents, url, error, cert, cb) {
+    // Tell the page to show an unlocked icon. Note this is sent to the main
+    // window webcontents, not the webview webcontents
+    BrowserWindow.getAllWindows().map((win) => {
+      win.webContents.send(messages.CERT_ERROR, {
+        url: url,
+        certError: error,
+        cert: cert
```
bbondy: nit: cert,

js/components/urlBar.js
diracdeltas: I couldn't figure out a way to get around it otherwise. The race condition worst-case is that the page is blank (which is what happens now when you hit a page with a cert error). In this case, the user will probably reload the page and then it will definitely show about:certerror on reload.

js/components/frame.js
diracdeltas: Any idea why `location` would still be the previous URL here? It seems like it should always be the same as `src` when a fresh `about:` page is loaded.

app/index.js
diracdeltas: ?
adding a comma throws a lint warning

bbondy: sorry just ignore the comma, I was just giving a nit that the value doesn't need to be specified when it is the same as the key.
So should have read:

nit: cert

js/components/frame.js
bbondy: `location` would never be the previous value. Always the most recent value.

`src` on the other hand is what the iframe src is  And if the iframe is navigated, src stays the same.

js/components/urlBar.js
bbondy: that doesn't seem bad, just post an issue pls though to handle this in electron later.

js/components/frame.js
diracdeltas: I was definitely seeing a bug where the URL bar was `about:certerror`, the page shown was `about:certerror`, but `this.props.frame.get('location')` was `https://lenovo.com` (the page that caused the cert error). Will try to give repro steps.

diracdeltas: Can't seem to get it to happen again. Will change it back to location instead of src.

diracdeltas:
Closing this in favor of the version about to be pushed to master which is rebased with the fixes above

PR #746, opened 2016-02-12
"open in new tab" shouldn't focus urlbar. fixes #745

### LINKED ISSUE #745: right-click + open in new tab puts focus in address bar
I've seen this behavior on Feedly, Imgur and CNET, so I imagine it's universal. (Tested on Windows, not Mac.)

1. Right click on a link
2. Select "open in new tab"
3. The new page opens successfully in new tab, but for the existing page, the focus moves to the address bar

psimyn:
damn that almost synced up amazingly

screen shot 2016-02-13 at 01 24 03

fixes #745

### DIFF
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -45,6 +45,7 @@ const WindowActions = {
   loadUrl: function (activeFrame, location) {
     location = location.trim()
     let newFrame = false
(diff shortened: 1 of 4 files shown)


bridiver:
I think there is a more general urlbar focus issue that needs to be addressed. The same thing happens when new popup windows are opened

psimyn:
awesome, thanks @bridiver

PR #753, opened 2016-02-13
webpack fixes and optimizations

bridiver:
use webpack dev server for about pages
newtab loading is slightly faster in production


bridiver:
@bbondy

app/about-preferences.html
```diff
@@ -3,19 +3,19 @@
...
-  <script src="gen/aboutPages.entry.js" defer></script>
-  <script src="./ext/l20n.min.js" defer></script>
-</head>
-<body>
-  <div id="appContainer"/>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <!-- TODO: Refactor away all unsafe-inline content -->
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; form-action 'none'; referrer no-referrer; script-src 'self'; connect-src 'self' ws://localhost:8080; style-src 'self' 'unsafe-inline'; font-src 'self'">
```
bbondy: what is ws://localhost:8080?

bridiver: some websocket that webpack uses

js/components/frame.js
```diff
@@ -224,29 +231,23 @@ class Frame extends ImmutableComponent {
...
+      if (this.props.frame.get('location') === 'about:certerror' &&
+          security && security.get('certDetails')) {
+        // Don't send certDetails.cert since it is big and crashes the page
+        this.webview.send(messages.CERT_DETAILS_UPDATED, {
+          url: security.get('certDetails').url,
+          error: security.get('certDetails').error
+        })
+      }
+    })
+    this.webview.addEventListener('did-navigate-in-page', () => {
```
bbondy: I'm fairly sure moving this in did-finish-load isn't right. If you allow ads and load slashdot I think this gets called a bunch of times before the frame is actually loaded. WHy was this changed away from did-frame-finish-load w/ isMainFrame check?

bridiver: did-finish-load is only called for the main frame
`void WebContents::DidFinishLoad(content::RenderFrameHost* render_frame_host,
                                const GURL& validated_url) {
  bool is_main_frame = !render_frame_host->GetParent();
  Emit("did-frame-finish-load", is_main_frame);

  if (is_main_frame)
    Emit("did-finish-load");
}
`

bbondy: Did it change from some fixes in electron? Because it was being called multiple times before.

bridiver: not sure, but I'll double check that it only gets called once

bridiver: I only see one call per page on slashdot and several others

bridiver: looks like some other change already removed the duplicate, but this puts it in the event for the main frame only

bridiver:
newtab load time is now < 30ms

app/about-about.html
```diff
@@ -3,17 +3,23 @@
...
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; form-action 'none'; referrer no-referrer; script-src 'self' 'unsafe-inline'; connect-src 'self' ws://localhost:8080">
+    <meta name="availableLanguages" content="en-US">
+    <meta name="defaultLanguage" content="en-US">
+    <title data-l10n-id="aboutPages"></title>
+    <script>
+      window.setTimeout(function() {
+        var po = document.createElement('script'); po.type = 'text/javascript'; po.async = true;
+        po.src = 'gen/aboutPages.entry.js';
+        var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(po, s);
+      }, 30);
```
bridiver: seems to work with a slight lower delay, but it can't be arbitrarily low because at some point it reverts back to synchronous loading so I kept it at 30 to be safe

bridiver: actually a better fix is probably to do this with a dom event. The 30ms probably puts it outside of the load

bbondy:
Looks great, merging!

PR #754, opened 2016-02-13
use sendToHost in webviewPreload

bridiver:
really fixes 'undefined method get' when opening a popup window from a link
@bbondy


bbondy:
++ thanks for cleaning this up. Merging.

PR #755, opened 2016-02-13
only default urlbar focus for new empty active tabs

### LINKED ISSUE #745: right-click + open in new tab puts focus in address bar

bridiver:
general case of #746 that fixes both new tabs and new windows
fixes #745


bbondy:
++

PR #757, opened 2016-02-13
url querystring / hash changes shouldn't clear title. fixes 752

### LINKED ISSUE #752: Title loses its value when browsing a Medium article
The title in the URL bar seems to disappear when I browse a Medium article[1]. It starts with the title populate but then it disappears after the window is ready.

[1] https://backchannel.com/our-medical-data-must-become-free-f6d533db6bed#.6akqmsnvo

psimyn:
fixes #752

There is probably a nicer way of stripping it

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -13,6 +13,8 @@ const ipc = global.require('electron').ipcRenderer
 const messages = require('../constants/messages')
 const debounce = require('../lib/debounce.js')
 
+import { currentPage } from '../lib/appUrlUtil'
+
 let windowState = Immutable.fromJS({
   activeFrameKey: null,
   frames: [],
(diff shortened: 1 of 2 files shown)


js/stores/windowStore.js
```diff
@@ -142,14 +142,15 @@ const doAction = (action) => {
       break
     case WindowConstants.WINDOW_SET_LOCATION:
       const key = action.key || windowState.get('activeFrameKey')
-      const lastLocation = windowState.getIn(frameStatePath(key).concat(['location']))
+      const nextLocation = action.location.split('?')[0].split('#')[0]
```
bridiver: won't this pull the named anchor from the wrong place? http://tools.ietf.org/html/rfc3986#section-4.1
Having it before the query string might be non-standard behavior that we also have to support, but either way it's probably better to use URL.parse/format

bridiver:
Thanks for the PR! After the fixes above I'm good to merge this as a fix for now, but I think we should implement DidStartProvisionalLoadForFrame in Electron and set the location there. Unlike DidCommitProvisionalLoadForFrame (which fires the load-commit event), DidStartProvisionalLoadForFrame doesn't fire for reference fragment changes. We should link this commit to https://github.com/brave/browser-laptop/issues/759 so we can update this when it is done

bridiver:
it would also be great if you could add a test for this

js/stores/windowStore.js
psimyn: updated to use origin + pathname

psimyn:
updated URL parsing. I tried adding a test but having issues accessing webview elements in tabTest.

PR #761, opened 2016-02-14
always show bookmark icon, to better identify bookmarked pages.

### LINKED ISSUE #724: Show bookmarked status when urlBar is on titleMode
Currently the reload button (`startButtons`) is hidden together with the star button (`endButtons`) when on titleMode. If you just unhide `endButtons`, the star is partly occluded by the title text/ellipsis.

This is bad because:
* When visiting a bookmarked page, user can't tell that it's bookmarked.
* User bookmarks a page with Ctrl+D, sees no feedback that it worked.

psimyn:
fixes #724

screen shot 2016-02-15 at 00 50 40

screen shot 2016-02-14 at 01 32 27
screen shot 2016-02-14 at 01 32 37

Possible additions:
 - reduce opacity of star when in titlemode
 - should star be appended to page title in title mode, or keep position fixed?

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -226,51 +226,48 @@ class Main extends ImmutableComponent {
     const showBookmarksToolbar = getSetting(settingsState, settings.SHOW_BOOKMARKS_TOOLBAR)
     return <div id='window' ref={node => this.mainWindow = node}>
       <div className='top'>
-        <div className='navigatorOuterWrapper'>
-          <div className='navigatorWrapper'>
-            <div className='backforward'>
-              <span data-l10n-id='backButton'
-                className='back fa fa-angle-left'
-                disabled={!activeFrame || !activeFrame.get('canGoBack')}
-                onClick={this.onBack.bind(this)} />
-              <span data-l10n-id='forwardButton'
-                className='forward fa fa-angle-right'
-                disabled={!activeFrame || !activeFrame.get('canGoForward')}
-                onClick={this.onForward.bind(this)} />
-            </div>
-            <NavigationBar
-              ref={node => this.navBar = node}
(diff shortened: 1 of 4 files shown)


bbondy:
CC @bradleyrichter what is your preference here?

bradleyrichter:
Good idea. If we are always truncating title to match the URL bar width, then keep the same position but let's try transitioning to a small filled star in light grey or orange if marked. This should visually align better with the locked icon on the left.

psimyn:
@bradleyrichter scaled, and lighter grey for disabled one

screen shot 2016-02-15 at 22 04 08
screen shot 2016-02-15 at 22 04 14

bradleyrichter:
@psimyn It seems like the chance of "floating stars" will be quite high given short titles. Can you append it to the title using the same method as the lock icon?

psimyn:
@bradleyrichter yeah it looks pretty weird just dangling there. I'll resolve conflicts and update it

psimyn:
screen shot 2016-02-18 at 21 47 13
screen shot 2016-02-18 at 21 47 23

bradleyrichter:
@bbondy Let's pull this in. I may want to hide the non-marked star state, similar to the locked icon approach but let's try it this way first.

Thanks Simon!

Brad

psimyn:
@bradleyrichter hiding non-marked star state is good idea, can be added easily enough with css. I also want to make the positioning of things on the topbar a bit less flaky - currently I'm constantly afraid that changes will throw things out of alignment

bbondy:
> hiding non-marked star state is good idea, can be added easily enough with css

Merging for now, will accept other patches afterwards if you do.

PR #762, opened 2016-02-14
add undo and redo items to context menu for text fields

### LINKED ISSUE #721: Right Click Menu -- Undo Option (and others like Redo) Missing when Inputting Text
Brave Menu: 

Chrome Menu which Includes Undo Option:

psimyn:
resolves #721

they should probably be disabled, though unsure how to determine this using role menu items


bbondy:
Thanks!  pls post a follow up for adding canUndo and canRedo to electron and doing this without roles

psimyn:
should I add that to here or to the electron fork? or both?

bbondy:
tracking as an issue in this repo is fine, thanks.

Issue #132, opened 2016-01-10
Add tests for adblock 


bbondy:
Including notifications are sent correctly to frames.

luixxiul:
Can this be closed with https://github.com/brave/browser-laptop/commit/0aaab3b56baf26ff1a57249e30990738555a79bd?

bbondy:
yes thanks!

PR #765, opened 2016-02-14
Add adblock test

rob-blasetti:
Kicking off the adblock tests as per issue #132

### DIFF
--- a/test/app/adBlockTest.js
+++ b/test/app/adBlockTest.js
@@ -1,0 +1,27 @@
+/* global describe, it, before */
+
+const Brave = require('../lib/brave')
+const selectors = require('../lib/selectors')
+const messages = require('../../js/constants/messages')
+
+describe('adBlock', function () {
+  function * setup (client) {
+    yield client
+      .waitUntilWindowLoaded()
+      .waitForVisible('#window')
+      .waitForVisible(selectors.urlInput)
+  }
+
+  describe('blocks ruled ads', function () {
+    Brave.beforeAll(this)
+    before(function *() {
+      yield setup(this.app.client)
+    })
+
+    it('blocks a known rule on a local url', function *() {
+      yield this.app.client
+        .ipcSend(messages.SHORTCUT_NEW_FRAME, 'http://www.newsbtc.com/2016/02/14/bitcoin-price-weekly-analysis-primed-for-lift-off-2/')
+        .isExisting('.adfixedLeft').should.eventually.be.false
+    })
+  })
+})


test/app/adBlockTest.js
```diff
@@ -0,0 +1,27 @@
...
+  describe('blocks ruled ads', function () {
+    Brave.beforeAll(this)
+    before(function *() {
+      yield setup(this.app.client)
+    })
+
+    it('blocks a known rule on a local url', function *() {
+      yield this.app.client
+        .ipcSend(messages.SHORTCUT_NEW_FRAME, 'http://www.newsbtc.com/2016/02/14/bitcoin-price-weekly-analysis-primed-for-lift-off-2/')
+        .isExisting('.adfixedLeft').should.eventually.be.false
```
bbondy: I don't think this is working, and also no remote URLs.

So this should be a html file in the test/fixtures directory and it should contain maybe a script, image, or iframe with a parameter that will get blocked, like:

?ad_box_=1

I'm not sure how to detect that the resource got blocked offf hand though.

bridiver: each tab is actually its own window inside the main window. See "windw.open with click" for examples of interacting with the tab vs main window. Also, checking for non-existence is tricky and this would be subject to a race condition because "should.eventually.be" does not actually do what it sounds like it does. In the test/fixture that @bbondy mentioned you could create an element after dom load and check for the non-existence of the ad after that element is created.

bbondy: You can click on the favicon in the urlbar and make sure the adblock l10n-args is correct for the number of things blocked.  Since ads are blocked you could even use a remote resource.

rob-blasetti: The test is working in the CI check, others are failing though so the build is broken.
I'll have a crack at fixing it according to the feedback, thank guys.

bbondy: We have some intermittent fails in travis especially.
Mainly we just mean that the test isn't testing the feature like it appears to.  So passing in this case isn't attached to the feature it is supposed to be testing. Thanks for taking a look!

bridiver: @rob-blasetti just to expand on what @bbondy said, the main window and the webview in the tab are completely separate (they don't even run in the same process). You cannot access anything in the tab webview if you are attached to the main window and vice-versa. A single window with one tab will actually create two windows from webdriver's perspective. A window with two tabs will have 3 and two windows with one tab each will have 4

bbondy:
Please re-submit when ready, just clearing queue for now so closing for now.

PR #776, opened 2016-02-15
Add tooltips for nav/tab sections

### LINKED ISSUE #622: Usability - tooltips
Related to issue #621 which may be confusing for new/novice users, I noticed there are no tooltips on any buttons. Whilst the iconography is fairly standard, when the "new tab" button moves, it is not immediately obvious that the "+" icon still represents "new tab" as its context has changed. Tooltips on hover would assist.

psimyn:
Resolves #622 

Probably needs localisation though

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -227,11 +227,11 @@ class Main extends ImmutableComponent {
         <div className='navigatorOuterWrapper'>
           <div className='navigatorWrapper'>
             <div className='backforward'>
-              <span
+              <span data-l10n-id='backButton'
                 className='back fa fa-angle-left'
                 disabled={!activeFrame || !activeFrame.get('canGoBack')}
                 onClick={this.onBack.bind(this)} />
-              <span
+              <span data-l10n-id='forwardButton'
                 className='forward fa fa-angle-right'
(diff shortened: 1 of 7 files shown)


js/components/main.js
```diff
@@ -224,11 +224,11 @@ class Main extends ImmutableComponent {
       <div className='top'>
         <div className='navigator-wrapper'>
           <div className='backforward'>
-            <span
+            <span title='Go back'
```
bbondy: These should be defined in the l20n file please, thanks!

bbondy: `app/locales/en-US/app.l20n`
Example:
`<backButton
  title: "Go back">`

Then set the data-l10n-id

psimyn: you are a wizard. Was wondering how to set other values. Thanks!

psimyn: @bbondy updated with data-l10n-ids 

let me know if there is some kind of system in the locale file that I've missed, I just put them near related items

bbondy:
Looks great, pls rebase and then I'll merge in.

psimyn:
rebased. Thanks, @bbondy

bbondy:
Thanks!

PR #778, opened 2016-02-15
highlight text in searchbar when opening it

### LINKED ISSUE #416: Text inside the find bar should be highlighted when pressing Ctrl+F for the 2nd time
capture

When the find bar is already opened (used before), pressing Ctrl+F doesn't highlight the text. On other browsers, pressing Ctrl+F for the 2nd time will highlight the text, so that users can resume their search and immediately type new words.

Windows 7 Home Premium 64-bit

psimyn:
fixes #416

I'll try to add some tests; have been having troubling triggering the search in webdriver

### DIFF
--- a/js/components/findbar.js
+++ b/js/components/findbar.js
@@ -12,6 +12,12 @@ const WindowActions = require('../actions/windowActions')
 export default class FindBar extends ImmutableComponent {
   constructor () {
     super()
+    this.onBlur = this.onBlur.bind(this)
+    this.onKeyDown = this.onKeyDown.bind(this)
+    this.onChange = this.onChange.bind(this)
+    this.onFindPrev = this.onFindPrev.bind(this)
+    this.onFindNext = this.onFindNext.bind(this)
+    this.onCaseSensitivityChange = this.onCaseSensitivityChange.bind(this)
   }
 
   onChange (e) {
@@ -54,6 +60,9 @@ export default class FindBar extends ImmutableComponent {
   }
 
   componentDidUpdate (prevProps) {
+    if (this.props.selected) {
+      this.focus()
+    }
     if (this.props.findDetail && !prevProps.findDetail ||
         this.props.findDetail.get('searchString') !== prevProps.findDetail.get('searchString') ||
         this.props.findDetail.get('caseSensitivity') !== prevProps.findDetail.get('caseSensitivity')) {
@@ -80,6 +89,10 @@ export default class FindBar extends ImmutableComponent {
         }
         break
     }
+  }
+
+  onBlur (e) {
+    WindowActions.setFindbarSelected(this.props.frame, false)
   }
 
   get numberOfMatches () {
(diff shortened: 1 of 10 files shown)


bbondy:
This is one of my biggest dogfood gripes, thanks for taking it. 
I think there's an issue currently though in that if you open the find bar then the urlbar won't be focusable.
I think we have to do something like how `activeShortcut` works in frame.js where you reset the value afterwards.

psimyn:
@bbondy I added an onBlur and a new Selected item to state which seems to fix urlbar issue

Trying to get tests working has been hit and miss - stuff in frameStateUtil sometimes seems to throw `get is not a function` for frameprops. This may be an issue with test setup

and I gather there is something horrifically wrong with the tests; 2 of them pass sometimes, 1 always and 1 never :8ball:

js/components/findbar.js
```diff
@@ -129,7 +137,7 @@ export default class FindBar extends ImmutableComponent {
         data-l10n-id='findResultMatches'/>
     }
 
-    return <div className='findBar'>
+    return <div className='findBar' onBlur={() => this.onBlur()}>
```
bbondy: fine as is but for future ref you can just do onBlur={this.onBlur}

js/state/frameStateUtil.js
```diff
@@ -84,7 +84,16 @@ export function findIndexForFrameKey (frames, key) {
...
 export function getFramePropsIndex (frames, frameProps) {
-  return frames.findIndex(found => found.get('key') === frameProps.get('key'))
+  return frames.findIndex(found => {
+    try {
+      return found.get('key') === frameProps.get('key')
+    } catch (e) {
+      // todo
+      // catch '.get is not a function'
+      // this seems to happen when testing
+      // possibly dispatchToIPC related
```
bbondy: Sounds like you have `found` which is not an Immutable object.  Example:

{a: 3} vs Immutable.fromJS({a: 3})

bbondy:
Looking great but just waiting to merge in case it breaks the tests.  Let me know.  Thanks for all the test work in here too.

js/components/findbar.js
psimyn: updated this and other .binds in here

js/state/frameStateUtil.js
psimyn: I think it is frameProps - added the following to the `catch` block

```
console.log(found.get && found.toJS() || found)
console.log(frameProps.get && frameProps.toJS() || frameProps)
```

and it shows this for frameProps:

```
Object
@@__IMMUTABLE_ITERABLE__@@: true
@@__IMMUTABLE_KEYED__@@: true
@@__IMMUTABLE_MAP__@@: true
__altered: false
_root: Object
  entries: Array[3]
    0: Array[2]
      0: "findbarShown"
      1: false
      length: 2
    1: Array[2]
    2: Array[2]
    length: 3
ownerID: Object
size: 3
```

Never happens outside of test mode. I'll look into it a bit more

psimyn:
@bbondy all tests are passing except the favicon ones fixed by https://github.com/brave/browser-laptop/pull/814

Changed windowActions and store in test mode to allow frameProps to work, let me know if it is sane

bbondy:
Thanks! merging.

PR #792, opened 2016-02-16
check for all windows before running any assertions

bridiver:
checking travis with PR


bbondy:
fine with me if you want to merge

PR #803, opened 2016-02-17
add reopen last tab to tab context menu

### LINKED ISSUE #789: Add "Reopen Last Closed Tab" to Tab's context menu
Currently Brave have the functionality of undo close tab, however it's only available under the History menu, which makes it difficult to find for users coming from other browsers.

## Steps to reproduce:

- with a bunch of tabs opened, close one tab
- right-click on one of the remaining tabs
- try to reopen the last closed tab

## Expected results:

An option for reopening last closed tab should be present, like Chrome ("Reopen Close Tab"), or Firefox("Undo Close Tab" ).

## Actual results:

Only those options are present:

screen shot 2016-02-16 at 4 46 56 pm

psimyn:
fixes #789

### DIFF
--- a/js/commonMenu.js
+++ b/js/commonMenu.js
@@ -84,6 +84,14 @@ module.exports.newWindowMenuItem = {
   click: () => AppActions.newWindow()
 }
 
+module.exports.reopenLastClosedTabItem = {
+  label: 'Reopen Last Closed Tab',
+  accelerator: 'Shift+CmdOrCtrl+T',
+  click: function (item, focusedWindow) {
+    module.exports.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_UNDO_CLOSED_FRAME])
+  }
+}
+
 module.exports.separatorMenuItem = {
   type: 'separator'
 }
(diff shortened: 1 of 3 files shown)


js/contextMenus.js
```diff
@@ -134,6 +135,11 @@ function tabTemplateInit (frameProps) {
     })
   }
 
+  items.push(Object.assign({},
+    CommonMenu.reopenLastClosedTabItem,
+    { enabled: WindowStore.getState().get('closedFrames').size > 0 }
```
bbondy: I'm not a fan of including windowStore here, in favor of instead passing in info to contextrMenu but this is ok for now. Thanks.

js/commonMenu.js
```diff
@@ -84,6 +84,14 @@ module.exports.newWindowMenuItem = {
   click: () => AppActions.newWindow()
 }
 
+module.exports.reopenLastClosedTabItem = {
+  label: 'Reopen Last Closed Tab',
```
bbondy: I think we might have a capitalization inconsistency.  Would you mind fixing?

bbondy:
merging

js/contextMenus.js
psimyn: do you mean passing it with the event to `onTabContextMenu` ? I wasn't big on including it here either, but figured passing it with the event would mean going through Tabs' props

bbondy: ok this is fine.

PR #814, opened 2016-02-18
Favicon test fixes

bridiver:


bbondy:
I think this is right but I'd like to learn if we can get access to favicons via url alone from user's cache with just the remote url when the user is offline.

PR #825, opened 2016-02-19
in titlemode, only show bookmark for bookmarked pages

psimyn:
And `position: absolute` the star to page title doesn't jump when bookmarking with ctrl D

screen shot 2016-02-20 at 00 46 20
screen shot 2016-02-20 at 00 46 31


bbondy:
++ thanks

PR #841, opened 2016-02-20
Get rid of sqlite for HTTPS Everywhere

### LINKED ISSUE #673: switch to JSON for HTTPS everywhere
https://github.com/EFForg/https-everywhere/pull/4110/files switches upstream over to JSON; we should too when it lands in stable

diracdeltas:
Fix #673 
Also fixes all SQLITE issues :)

I will open separate issues for iOS/Android to get rid of sqlite. @garvankeeley @SergeyZhukovsky here is a summary of the changes:
* In `https://s3.amazonaws.com/https-everywhere-data/5.1.3/`, there is now a single file `httpse.json` that replaces `rulesets.sqlite` and `httpse-targets.json`.
* `httpse.json` has two fields: `rulesetStrings` and `targets`. 
* `targets` is the same as what used to be `httpse-targets.json` (map of hostname patterns to ruleset IDs).
* `rulesetStrings` is a map of ruleset IDs to the actual ruleset, which can be applied using the same code as before.

Now instead of doing a sqlite query for `ruleset_id` to get the ruleset, you can just get the ruleset by `rulesetStrings[ruleset_id]`.

Auditors: @bbondy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,5 +41,5 @@ app/gen
 *.pfx
 
 # HTTPS Everywhere files
-httpse-targets.json
-rulesets.sqlite
+rulesets.json
+httpse.json
--- a/app/dataFile.js
+++ b/app/dataFile.js
@@ -61,45 +61,23 @@ function downloadSingleFile (resourceName, url, version, force, resolve, reject)
 }
 
 module.exports.downloadDataFile = (resourceName, url, version, force) => {
-  if (resourceName === 'httpsEverywhere') {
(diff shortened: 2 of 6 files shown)


bbondy:
Looks great, could you attach the files here that I need to upload in the 5.1.3 bucket?

diracdeltas:
already uploaded

diracdeltas:
merging now

bbondy:
++ thanks

Issue #720, opened 2016-02-11
flash of selected tab when hovering across the border between two unselected tabs


asadotzler:
With tab preview on hover enabled, 0.7.14 build on Mac.

1. open three tabs, mozilla.com, brave.com, and cnn.com
2. select the mozilla tab.
3. hover over the brave tab and see a brave preview
4. hover over the cnn.com tab and see a cnn preview
5. hover over the space between the brave and cnn tabs and see no preview, just the mozilla.com tab. 

A common behavior I find is to sweep across several tabs looking at the previews to locate a lost tab. As I slide my mouse across the tab strip, each time it moves across the gap between tabs there's a flash of the selected tab that's really jarring. I think for the purposes of tab preview on hover, there should be no gap between tabs.

weems:
This preview feature is useful, but between two tabs, I have confirmed the steps to recreate are valid inbetween: 

1. tweetdeck.twitter.com
2. vox.com
3. yahoo.com
4. github.com 

(Though the code at hand, relates to all tab preview functionality so it does not depend on specific URLs)

bbondy:
I think this is a dupe of #142. I'm thinking we can solve this without changing tab width just by debouncing the clear / show previews by something like 100ms.

bbondy:
@bradleyrichter would you prefer a timeout here or to remove the margin between tabs?

weems:
@bbondy perhaps it is @bradleyrichter expertise required but it would seem to me that the most serious issue is caused by failure of margin between tabs to function properly.

psimyn:
I think debouncing makes the most sense. even if there was no margin there is still that tabbar area above them that would have the same effect.

timeout will prevent the flicker of preview->non-preview->preview when mousing between tabs

weems:
perhaps timeout + margin will mitigate both of those issues, @psimyn

bradleyrichter:
@bbondy I think there are a few ways to solve it. Adding may will still fail when either moving slowly or accidentally drifting slowly between tabs. Adding enough of a delay to cover that slow-move case might be annoying.

We do need to maintain a visual gap between tabs and I may also add a border or even a side-border  style to non-selected tabs. Could we nest the tab in a clear DIV that allows for no gutters, and assign the function there? 

Let's try a few things in a branch and I'll play around with it until it feels right.

The problem would still exist above the tabs though as @psimyn mentioned.


Thanks for the ideas guys!

bbondy:
[...]

Issue #142, opened 2016-01-11
Tweak preview tabs timeout


bbondy:
1) It should delay a bit more when moving from 1 tab to the other so there is no active tab in between. 
2) Play with the hover time to avoid accidental previews but also don't make it too long.
3) The time it takes to reset to the "you need to hover" seems a bit too long

bbondy:
@BrendanEich suggest doing this based on pointer velocity instead of just timeout.  I think that's a great idea.

luixxiul:
+1 from support

PR #848, opened 2016-02-21
Increase delay to tab previews on hover

bogas04:
I personally find the tab previews to be very abrupt which becomes annoying sometimes.
Ideally we should allow the timings to be configurable, and probably we can choose a default that everyone is happy with, via some telemetry or voting.

### DIFF
--- a/js/components/tab.js
+++ b/js/components/tab.js
@@ -102,7 +102,7 @@ class Tab extends ImmutableComponent {
     const previewMode = new Date().getTime() - this.lastPreviewClearTime < 1500
     window.clearTimeout(this.hoverClearTimeout)
     this.hoverTimeout =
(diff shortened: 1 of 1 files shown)


psimyn:
Thanks! Definitely in favour of removing the 0 (prevents flicker when mousing up to url bar then back down). On the fence about going as high as 750ms, but open to increasing it from current.

Related issues for this:
https://github.com/brave/browser-laptop/issues/142
https://github.com/brave/browser-laptop/issues/720

bridiver:
The delay should probably be configurable as an advanced setting eventually. I'd actually prefer a slightly shorter delay with no preview mode

bbondy:
there's some concern that it looks like lag when there's a delay, I'm going to close this in favour of the above 2 issue links and possibly a refactoring of how this works completely.

PR #852, opened 2016-02-22
Revert "Rebuild sqlite3 for electron"

diracdeltas:
This reverts commit 2964be26e757feb37988749eb1e312393644a03e, which I believe is making Travis error out after #673 was fixed.

Auditors: @bbondy


bbondy:
++ we may need this again unfortunately at a later date when we get bookmark importing from firefox.

PR #853, opened 2016-02-22
wrapper for urlbar inputs

psimyn:
this probably causes more problems than it solves...

Main issue I was trying to resolve was `position: absolute` on lock icon, loadtime and bookmark icon causing issues with centering/layout of title.

Also adds some conditional rendering for title/input modes, instead of both always being there

### DIFF
--- a/js/components/navigationBar.js
+++ b/js/components/navigationBar.js
@@ -94,7 +94,7 @@ class NavigationBar extends ImmutableComponent {
           titleMode: this.titleMode
         })}>
 
-      { isSourceAboutUrl(frameProps.get('location')) ? null
+      { isSourceAboutUrl(frameProps.get('location')) || this.titleMode ? null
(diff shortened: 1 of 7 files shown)


bbondy:
> this probably causes more problems than it solves...

That line makes me worry, is this ready for review?

psimyn:
@bbondy that line was due to it not being linked to a specific issue :)

Hold off on review until I fix tests though - 2 remaining breakages due to urlBar element not existing in titleMode (and vice-versa). If that ends up being huge effort I'll go back to hiding with css, but only rendering relevant elements seems a better solution.

psimyn:
@bbondy tests should be fixed now. Main functional change from this is removing title/urlbar when not in use - made some tests a bit funny but they match up with what a user actually sees.

Let me know what you think

bbondy:
Thanks, merging

PR #855, opened 2016-02-22
Fix glitch on (repeated) logo with retina display

elia:
screen shot 2016-02-22 at 14 59 09


bbondy:
++ thanks

weems:
so the logo is a error? funny, I thought that was expected :P #retinaUser

bbondy:
tiny orange line on the right

weems:
I see it now, just thought it was intentional :smile:

PR #863, opened 2016-02-22
Make cert error pages work for non-active frames

### LINKED ISSUE #832: Right click and open in new tab on badssl.com is not showing cert pages

diracdeltas:
Fix #832

Auditors: @bbondy

### DIFF
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -62,7 +62,8 @@ const WindowActions = {
   loadUrl: function (frame, location) {
     location = location.trim()
     let newFrame = false
-    if (frame.get('pinnedLocation')) {
+    if (frame.get('pinnedLocation') && location !== 'about:certerror' &&
+        frame.get('location') !== 'about:certerror') {
       try {
         const origin1 = new window.URL(frame.get('location')).origin
         const origin2 = new window.URL(location).origin
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -92,6 +92,24 @@ const WindowActions = {
   },
 
   /**
+   * Similar to loadUrl, but loads the URL in a specified frame instead of the
+   * active one.
+   * @param {string} location - url to load
+   * @param {number} key - key of the frame to load in
+   */
+  loadUrlInFrame: function (location, key) {
+    location = location.trim()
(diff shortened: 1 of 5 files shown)


js/actions/windowActions.js
```diff
@@ -92,6 +92,24 @@ const WindowActions = {
   },
 
   /**
+   * Similar to loadUrl, but loads the URL in a specified frame instead of the
+   * active one.
+   * @param {string} location - url to load
+   * @param {number} key - key of the frame to load in
+   */
+  loadUrlInFrame: function (location, key) {
```
bbondy: Instead of this please modify loadUrl, it currently takes in a frameProps already but it doesn't pass the key to the store.  It would be better for it to pass the key to the store and for the store to use it.

js/stores/windowStore.js
```diff
@@ -139,6 +139,17 @@ const doAction = (action) => {
...
+      windowState = windowState.mergeIn(frameStatePath(action.key), {
+        src: action.location,
+        location: action.location,
+        audioPlaybackActive: false,
+        icon: undefined,
+        themeColor: undefined,
+        computedThemeColor: undefined,
+        title: ''
+      })
+      break
```
bbondy: Remove this too

bbondy:
Looks good, just the one note which would simplify complexity a lot.  Pls verify that is ok to do first, but I think it is.

diracdeltas:
yep will do

diracdeltas:
@bbondy done. please re-audit

js/components/frame.js
```diff
@@ -70,12 +71,24 @@ class Frame extends ImmutableComponent {
   }
 
   componentDidMount () {
+    ipc.on(messages.CERT_ERROR, (e, details) => {
```
bbondy: We used to have some ipc listener registration in frame.js but we had to move it out because creating lots of frames was giving there might be a leak warning. Might be worth checking for similar warnings with lots of tabs.

bbondy:
++ I've been wanting loadUrl to be not active frame specific anyway, independent of this task.  Thanks for doing it.

js/components/frame.js
diracdeltas: will keep an eye out; haven't seen any warnings yet

PR #867, opened 2016-02-23
Refactor onBeforeRequest to use Filtering.js

### LINKED ISSUE #845: Refactor https everywhere to use filtering.js
In order to do https://github.com/brave/browser-laptop/issues/748 we need some refactoring so that we can use `onBeforeRequest` for adblock & tracking protection.

Electron only allows you to have one handler for these webRequest methods.

@diracdeltas would you mind taking this? I'll then take #748.

diracdeltas:
Fix #845 

Auditors: @bbondy

### DIFF
--- a/app/adBlock.js
+++ b/app/adBlock.js
@@ -25,7 +25,7 @@ let mapFilterType = {
 }
 
 const startAdBlocking = () => {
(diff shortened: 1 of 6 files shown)


bbondy:
yay thanks!

PR #868, opened 2016-02-23
specify draggable regions

bridiver:
this only fixes the titlebar issue with https://github.com/brave/electron/pull/13, but the changes won't break anything without it


bbondy:
yay++ CC @bradleyrichter

PR #869, opened 2016-02-23
default file paths to use file:// scheme instead of http://

### LINKED ISSUE #862: when opening html files on local disk the url still contains http://
Firstly fantastic job, I'm using this on mobile and laptop. Good Job.
Glenn

Environment OSX, El Capitan
Brave Version 0.7.14

command-O opens a local file as 
http://users/developer/dev/some-path/docs/index.html rather than
file:///users/developer/dev/some-path/docs/index.html

editing the http to file in the url resolves the issue.

psimyn:
fixes #862

### DIFF
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -4,8 +4,11 @@
 
 'use strict'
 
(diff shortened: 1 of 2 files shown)


bbondy:
Great to have the tests too, thanks!

PR #870, opened 2016-02-23
Added bookmark toolbar signals show/hide test.

rob-blasetti:
I assume when the bookmarks toolbar is opened from the render window we want to kick the call back to the main window to do the work. These tests should work when we are able to communicate the messages to the main window. [Skipped for now so we don't break the build.]

### DIFF
--- a/js/constants/messages.js
+++ b/js/constants/messages.js
@@ -71,7 +71,10 @@ const messages = {
   NEW_FRAME: _,
   // HTTPS
   CERT_DETAILS_UPDATED: _, /** @arg {Object} security state of the active frame */
-  CERT_ERROR_ACCEPTED: _ /** @arg {string} url where a cert error was accepted */
+  CERT_ERROR_ACCEPTED: _, /** @arg {string} url where a cert error was accepted */
+  SHOW_BOOKMARKS_TOOLBAR: _,
+  HIDE_BOOKMARKS_TOOLBAR: _
+
 }
 
 module.exports = mapValuesByKeys(messages)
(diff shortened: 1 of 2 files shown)


js/constants/messages.js
```diff
@@ -71,7 +71,10 @@ const messages = {
   NEW_FRAME: _,
   // HTTPS
   CERT_DETAILS_UPDATED: _, /** @arg {Object} security state of the active frame */
-  CERT_ERROR_ACCEPTED: _ /** @arg {string} url where a cert error was accepted */
+  CERT_ERROR_ACCEPTED: _, /** @arg {string} url where a cert error was accepted */
+  SHOW_BOOKMARKS_TOOLBAR: _,
+  HIDE_BOOKMARKS_TOOLBAR: _
```
bbondy: These aren't hooked up to anything

test/components/bookmarksToolbarTest.js
```diff
@@ -0,0 +1,34 @@
...
+        .ipcSend(messages.SHOW_BOOKMARKS_TOOLBAR)
+        .waitForExist('.bookmarksToolbar')
+    })
+
+    it.skip('should close toolbar when signaled', function *() {
+      yield this.app.client
+        .ipcSend(messages.SHOW_BOOKMARKS_TOOLBAR)
+        .ipcSend(messages.HIDE_BOOKMARKS_TOOLBAR)
+        .isExisting('.bookmarksToolbar').should.eventually.be.false
+    })
```
bbondy: Does this work? I don't think show/hide bookmarks is hooked up to anything so I can't imagine it would.   Previous test commits also didn't work when I ran the tests. I think you need to run `npm run watch-all` in one terminal and in the other do npm run test.

You can change a describe block to describe.only to run only that test or you can run `npm run test -- --grep=should\ close\ toolbar\ when\ signaled` for example.

Also I think that an IPC message is not correct for this, instead it should be a toggle of this pref: `bookmarks.toolbar.show`.

Issue #780, opened 2016-02-15
Facebook Third Party Sign In Button Error


weems:
Support Request to Fix Facebook Sign in to Websites.

1. Go to https://quora.com
2. Click Sign in Using Facebook
3. Observe Nothing Happens
4. Try with Same Type of Button For Google when Signed in to Google and Notice you get a Window to allow Signing in to Quora using Google

Trying the Above Steps for https://fitbit.com and https://digg.com/reader#sign-in renders the same result. I think it is likely due to third party tracking being blocked. Perhaps we should discuss if we want users to block third party tracking or allow these buttons to function. It seems we are allowing the Google Button to function.

psimyn:
Possibly duplicate of https://github.com/brave/browser-laptop/issues/449 ?

And maybe https://github.com/brave/browser-laptop/issues/503

weems:
@psimyn for #449 I would agree, but Tracking is Off Yet Signing in using Google works.

weems:
@psimyn I don't think it is relevant for #503 because in the latest Brave I can sign in on FB with tracking on.

bbondy:
@diracdeltas could you take a look?

diracdeltas:
This is due to tracking protection blocking requests from `www.quora.com` to `www.facebook.com` and possibly also `staticxx.facebook.com`.  Quora, Digg, and Fitbit login with FB are also broken in the same way in Firefox in Private Browsing mode (which has tracking protection enabled).

I think we shouldn't whitelist FB in this case because it's a solvable bug in the implementation of FB login. There is no reason the site needs to make a third-party AJAX request to a facebook domain in order to popup the login window AFAICT. As the OP noted, Google login works fine. 

We could try to detect if the page is making requests to the FB login endpoint and pop up a dialog that says basically "turn off tracking protection on this site in order to login with FB".

weems:
@diracdeltas curious if (Firefox, others) consider the FB buttons more of a tracking risk than the way Google has designed them in some way more securely? Or if the reason Google is allowed through is something more mundane.

diracdeltas:
> Or if the reason Google is allowed through is something more mundane.

The reason Google is allowed through is that the button functionality doesn't require 3rd party requests.

weems:
If only FB was more responsible. :/

BrendanEich:
[...]

PR #877, opened 2016-02-23
Temporarily whitelist domains needed for fb login to work

diracdeltas:
this temporarily fixes the broken sites reported in https://github.com/brave/browser-laptop/issues/780 but i want to get others' thoughts on whether it's too tracking-permissive before merging.


bbondy:
++ for now

PR #878, opened 2016-02-23
add linux vm

bridiver:
@bbondy


bbondy:
neat setup, I haven't tried them yet but I'll merge for now and will try at some point.

PR #884, opened 2016-02-24
only update the navbar input if the location has changed

### LINKED ISSUE #847: Delay when opening new tab
There is some quite big delay when opening a new tab, you have quite to wait some bit before you are able to type in the url of the page you want to visit.

bridiver:
fixes #847


bbondy:
++

PR #888, opened 2016-02-24
urlbar should update on url fragment changes

bridiver:


bbondy:
++ and grea to have tests for this becaue I could see it breaking otherwise. Thanks.

Issue #710, opened 2016-02-10
tab title cleared on refresh


psimyn:
1. Load page
1. Refresh
1. Tab title set to location :(

PR #934, opened 2016-02-27
clearing the title in WINDOW_SET_LOCATION breaks titles for tabs crea…

### LINKED ISSUE #927: Tab titles not updating correctly with google docs
From support.
As in the picture provided.
Reproduced with Windows 10 7.16 version.

image

psimyn:
…ted with window.open

fixes #927 and doesn't regress #710 

Not sure why title was previously being cleared here

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -135,14 +135,10 @@ const doAction = (action) => {
     case WindowConstants.WINDOW_SET_LOCATION:
       const key = action.key || windowState.get('activeFrameKey')
       const lastLocation = windowState.getIn(frameStatePath(key).concat(['location']))
-      const lastTitle = windowState.getIn(frameStatePath(key).concat(['title']))
-      let locationChanged = !action.location || !lastLocation ||
-        action.location.split('#')[0] !== lastLocation.split('#')[0]
       windowState = windowState.mergeIn(frameStatePath(key), {
         audioPlaybackActive: false,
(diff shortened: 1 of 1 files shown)


js/stores/windowStore.js
```diff
@@ -135,14 +135,10 @@ const doAction = (action) => {
...
       const key = action.key || windowState.get('activeFrameKey')
       const lastLocation = windowState.getIn(frameStatePath(key).concat(['location']))
-      const lastTitle = windowState.getIn(frameStatePath(key).concat(['title']))
-      let locationChanged = !action.location || !lastLocation ||
-        action.location.split('#')[0] !== lastLocation.split('#')[0]
       windowState = windowState.mergeIn(frameStatePath(key), {
         audioPlaybackActive: false,
         adblock: {},
         trackingProtection: {},
-        title: locationChanged ? '' : lastTitle,
```
bbondy: I'm fairly sure that doing this will break when you navigate to a page with a title and then after that to a page with no title.  This area of code is desperately in need of tests.

bbondy: Actually I'm surprised if this passes the test `page without a title`, maybe that test needs a tweak if it does pass.

psimyn: Ahahaha that's the one. Knew I'd overlooked super obvious one. I'll look into fixing the tests around page titles and adding a few more

PR #935, opened 2016-02-27
remove box shadow from urlbar inner input

psimyn:
I think I derped a rebase for the urlbar wrapper changes. Removes this top line (right about "search for address"):

screen shot 2016-02-27 at 19 52 11


bbondy:
thx

PR #942, opened 2016-02-28
don't show contextmenu if a page has its own

### LINKED ISSUE #936: Context menu doesn't respect overrides (in Google Docs)
Google docs provides its own context menu, but Brave's context menu is still appearing upon right clicking:

Double context menu

This doesn't seem to be the case everywhere. S3 console's custom context menu seems to prevent the standard menu from being activated.

psimyn:
evt.defaultPrevented probably indicates there was a custom context menu event on document. setTimeout used to allow other listeners to trigger first, otherwise defaultPrevented will not have been set

fixes #936

also fix typo in test filename


bbondy:
++

bbondy:
I think we can consider the site has a bug if they don't prevent default.

PR #944, opened 2016-02-28
Support mouse navigaton

### LINKED ISSUE #938: Add mouse forward/back button support
Plenty of mice these days have side buttons for forward/back while browsing the web. I'd like to see support for this in Brave.

I did some poking around at it for a while, but unfortunately came up with no fruitful results. There doesn't seem to be a reasonable way to detect special mouse button presses (or anything beyond left/right/middle click) within normal client JS-land, thus my approach of putting a listener in the main component did not work out.

I tried to do some research into electron to see what kind of native mouse event support there was, but nothing I could find was too useful.

Long story short, no great avenue seems immediately available. If anyone has knowledge in this area to shed some light, I'd be happy to work on a pull request proposal for this.

twisterghost:
Adds support for mouse navigation. Listen to the mouse commands in `appStore.js` and send the forward/back shortcut messages through `ipc`.

This is to fix #938


bbondy:
Awesome, great work!

Issue #926, opened 2016-02-26
CTRL+ALT+B does not work


SLBondy:
Windows 10 the shortcut (CTRL+ALT+B) does not work

twisterghost:
Can you give some information on whatt this shortcut is supposed to do?

bbondy:
bookmarks manager I believe.

bbondy:
See `app/localShortcuts.js` if you want to take, it should be for all platforms not just windows, but on OS X it's Command and not Control.
`CmdOrCtrl+Alt+B`

twisterghost:
I can give it a shot so long as I understand what its supposed to do. To be clear, it just needs to open Brave's bookmarks manager? I'm running chrome and IE and neither seem to respond to ctrl+alt+b, but I'm happy to add it.

twisterghost:
Yeah I see the hotkey labelled on the bookmarks manager. I'll see if I can get it working.

bbondy:
Actually just the accelerator in js/commonMenu.js should be needed.  It does seem to work on OS X.

twisterghost:
Ah, I got it working in localShortcuts, but it looks like the accelerator already exists for it. Wonder why its borked.

twisterghost:
This issue might be of use: https://github.com/atom/atom/issues/3930

Looks like its a common issue for `ctrl+alt+b` to be hijacked on windows. Changing the accelerator to just `alt-b` on my local windows machine makes it work fine, though I don't know how into the idea of changing the hotkey you guys are.

Maybe fork the hotkey per-OS? Or perhaps an FAQ point would be better for the sake of standardization.

twisterghost:
This has been fixed by #948

bbondy:
Thanks @twisterghost !

PR #948, opened 2016-02-28
Fork the bookmarks manager hotkey

twisterghost:
This is a potential fix for #926. Use Shift+Alt+B to open the bookmarks manager on windows since Ctrl+Alt+B seems to be often hijacked.

As far as I can tell the only other way to fix this to make it work universally is to add a hotkey for `ctrl+alt+b` to `localShortcuts.js` to open a new window pointing at `about:bookmarks`. I have a working copy of that if that is preferred.

### DIFF
--- a/js/commonMenu.js
+++ b/js/commonMenu.js
@@ -18,6 +18,7 @@ const cookieblock = AppConfig.resourceNames.COOKIEBLOCK
 const settings = require('./constants/settings')
 const getSetting = require('./settings').getSetting
 const issuesUrl = 'https://github.com/brave/browser-laptop/issues'
+const isDarwin = process.platform === 'darwin'
 
 let electron
 try {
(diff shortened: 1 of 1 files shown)


js/commonMenu.js
```diff
@@ -151,7 +151,7 @@ module.exports.preferencesMenuItem = {
 
 module.exports.bookmarksMenuItem = {
   label: 'Bookmarks manager...',
-  accelerator: 'CmdOrCtrl+Alt+b',
+  accelerator: process.platform === 'win32' ? 'Shift+Alt+B' : 'CmdOrCtrl+Alt+B',
```
bbondy: Right idea to fork it but pls use the same shortcut as Chrome here for familiarity's sake.  That's  Control+Shift+O for bookmarks manager on Windows.  And Command+Alt+B for OS X.

twisterghost: Will do

bbondy: Actually Linux is always the same as Windows as well so this should be `const isDarwin = process.platform === 'darwin'` w/ an isDarwin ? X : Y.

bbondy: See `app/menu.js:` we usually put that const at the top

twisterghost: But wouldnt we want it to be windows or darwin?

bbondy: we want windows + linux to be Command+Shift+O and OS X to be 'CmdOrCtrl+Alt+B',

twisterghost: 024ebdc

bbondy:
++ great fixes, thanks!

PR #949, opened 2016-02-28
Add support for alt+arrow key navigation

twisterghost:
Even more hotkeys, woo.

Looks like most other browsers support `alt+left/right` to navigate. This PR adds support for that.

### DIFF
--- a/app/localShortcuts.js
+++ b/app/localShortcuts.js
@@ -20,9 +20,7 @@ module.exports.register = (win) => {
     ['CmdOrCtrl+Alt+Left', messages.SHORTCUT_PREV_TAB],
     ['Ctrl+PageDown', messages.SHORTCUT_NEXT_TAB],
     ['Ctrl+PageUp', messages.SHORTCUT_PREV_TAB],
-    ['CmdOrCtrl+9', messages.SHORTCUT_SET_ACTIVE_FRAME_TO_LAST],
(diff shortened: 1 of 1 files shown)


app/localShortcuts.js
```diff
@@ -20,7 +20,9 @@ module.exports.register = (win) => {
     ['CmdOrCtrl+Alt+Left', messages.SHORTCUT_PREV_TAB],
     ['Ctrl+PageDown', messages.SHORTCUT_NEXT_TAB],
     ['Ctrl+PageUp', messages.SHORTCUT_PREV_TAB],
-    ['CmdOrCtrl+9', messages.SHORTCUT_SET_ACTIVE_FRAME_TO_LAST]
+    ['CmdOrCtrl+9', messages.SHORTCUT_SET_ACTIVE_FRAME_TO_LAST],
+    ['Alt+Left', messages.SHORTCUT_ACTIVE_FRAME_BACK],
+    ['Alt+Right', messages.SHORTCUT_ACTIVE_FRAME_FORWARD]
```
bbondy: This looks right to me but we need to have these conditionally added w/ another `if (!isDarwin)` because on OS X alt + left goes to previous workd and alt + right goes to next word.

twisterghost: b2f332a

bbondy:
++ thanks!

PR #962, opened 2016-03-01
Refactor set url and reload handling

### LINKED ISSUE #799: Typing an address then going back before typing same address
Bug replication procedure : 
- Type address or website name (www.youtube.com)
- Go to website
- Use the back button
- Type same address

Will not go to the address in question

twisterghost:
This is a minor refactor of how reloading vs navigating words, as well as an addition of a `navigationTime` prop to frameProps.

This is a to fix #799, but does not solve one of the underlying problems of going back/forward not syncing the new location of the frame to the store.

I made an attempt to fix that underlying problem by adding an action to update the `src`/`location` on the listener for `did-navigate` on `frame.js`, but it always ended up causing a double navigation, which is bad, and also causes problems with the history stack.

This PR fixes the issue and maintains the history stack by changing where the logic lives for reloading vs navigating, while also adding a new way for the window to know when a navigation happens by looking at the navigation time as well as the `src` since the `src` doesn't always change (i.e. when going back or forward)


bridiver:
Thanks for the PR! Can you put together a test case for this? We've had a lot of regressions around url bar handling and I want to make sure that we are covering everything moving forward.

twisterghost:
Yeah, I'll see what I can do

bridiver:
Thanks! If you need any help let me know, but there should be enough existing examples to get you started

twisterghost:
For the sake of keeping open PRs trimmed, I'll close this. When I get some time to work on this again, I'll take another crack at it (if its not fixed in the meantime :smile:)

bbondy:
OK thanks, I'm not sure if what you did is right or wrong but we definitely need tests for this before we make significant changes to it like this because we've broken functionality around here very often.

PR #965, opened 2016-03-01
Fix rendering error

abiggerhammer:
The angle bracket in your `title` tag broke the rendered Markdown. Quoting the tag fixes it.


bbondy:
This actually needs to be fixed in that javadoc in js/actions/windowActions.  This file is generated.

bbondy:
Closing for now, will accept a fixed pull request per the suggestion above though.

Issue #967, opened 2016-03-01
Store lastDataFileSize and do comparison 


bbondy:
Android had a problem where data files weren't always the full file.  This is probably because the process was killed while it was being downloaded or written.  
We should also store the content length from the content length http header and do a comparison.  If it changed then re-download.
The read size must match the header to use it.

CC @garvankeeley would be good to have this on all platforms.

bbondy:
CC @diracdeltas I think this is where the sqlite surplus of crashes was routing from as well.

garvankeeley:
interesting! I was wondering how you were getting a corrupt file. iOS has 'atomic' flags on the operations I use (which means things happen to a temp file first) and then 'atomically' move to their final destination, which is done as a single filesystem rename call. It is considered failsafe to do this.

bbondy:
ok great, so no change needed on iOS, I think this is possibly a bug on laptop though.

PR #969, opened 2016-03-01
Only set etag and last check date if download succeeds

diracdeltas:
Related to #967, if the resource file download doesn't resolve, the browser
should at least redownload when it restarts.

Auditors: @bbondy


bbondy:
++ thanks for taking this.

PR #1033, opened 2016-03-08
small batch of style changes for URL bar

bradleyrichter:
@bbondy still needs code to separate selected lock icon size


bbondy:
Looks great, thanks. Merging.

Issue #923, opened 2016-02-26
Make permission and password prompts testable and base page interactive


bbondy:
https://github.com/brave/browser-laptop/commit/35efd0ecdcaf98deee1d29c1a319fbbfe04a0c59#commitcomment-16356058

diracdeltas:
Pretty sure this would require the page to be reloaded after the user answers the permission prompt. Not a blocker but may be annoying.

bbondy:
I don't think a reload would be required, at least it shouldn't be.

diracdeltas:
The Electron permission handler API requires the allow/deny callback to be called synchronously (see https://github.com/brave/browser-laptop/blob/master/app/filtering.js#L191). I guess we could wait a fixed amount of time for the user to make a decision and then deny the initial permission request by default if they don't.

bbondy:
> The Electron permission handler API requires the allow/deny callback to be called synchronously

Let's fix that for this bug, (no rush)

diracdeltas:
Also could be useful for https://github.com/atom/electron/blob/777ff450cf4a0722d29bce44c29fe02470952823/docs/api/app.md#event-certificate-error

diracdeltas:
@bbondy did basically this in https://github.com/brave/browser-laptop/commit/bb27ec23021680ecb314e33fa24b7a3dfd8a2d33, so i take back what i said about this issue requiring changes to Electron

diracdeltas:
same thing needs to be done for password save prompts. https://github.com/brave/browser-laptop/pull/1051

diracdeltas:
any thoughts about whether these prompts should be popup windows or a notification bar at the top of the frame?

bbondy:
mainly comes down to if you want to be able to have have a password prompt in one tab and then you can switch tab to a different tab that has a password prompt.  And then you can go back to the first password prompt tab.  My guess is per frame is desirable.

bbondy:
Moving to 0.9.1

PR #1051, opened 2016-03-10
Start password saving

diracdeltas:
This adds basic support for saving username/passwords and autofilling them later. It seems to work for most simple forms like https://trac.torproject.org/projects/tor/login.

follow-ups:
- [ ] replace electron's dialog with our custom dialog (#923)
- [x]  more crypto unit tests
- [ ] add prefs page to view and manage passwords (#1101)
- [x] better multi-account support (detect when a username has been entered and find the corresponding password)

login details are stored in `appState`; the password is encrypted with a key stored in the OS keychain before being written to disk. 

Auditors: @bbondy (aiming for 0.8.2)

### DIFF
--- a/app/content/scripts/webviewPreload.js
+++ b/app/content/scripts/webviewPreload.js
@@ -197,11 +197,25 @@
         }
         // Ask the main process for the credentials
         ipcRenderer.send('get-password', formOrigin, action)
-        console.log('got password field', formOrigin, action, usernameElem, passwordElem)
       }
+
+      // Whenever a form is submitted, offer to save it in the password manager
+      // if the credentials have changed.
+      form.addEventListener('submit', e => {
+        var fields = getFormFields(form, true)
+        // Re-get action in case it has changed
+        var action = form.action || document.location.href
+        var usernameElem = fields[0] || {}
+        var passwordElem = fields[1] || {}
+        ipcRenderer.send('save-password', usernameElem.value, passwordElem.value,
+                         formOrigin, action)
+      })
     })
 
     ipcRenderer.on('got-password', (e, username, password, origin, action) => {
+      // TODO: If there are multiple accounts, this always autofills the
+      // most recently-added one. We should detect when the username is
+      // entered and get the corresponding password.
       var elems = credentials[action]
       if (formOrigin === origin && elems) {
         elems.forEach((elem) => {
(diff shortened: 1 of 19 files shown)


docs/state.md
```diff
@@ -21,6 +21,14 @@ AppStore
     startTime: number, // datetime.getTime()
     endTime: number // datetime.getTime()
   }],
+  passwords: [{
```
diracdeltas: on second thought, maybe this data structure should be:
```
passwords: {
  encryptedData: string, // encrypted serialization of [{origin: ..., action: ..., username:..., password: ...}]
  authTag: string,
  iv: string
}
```
so that the complete list of login details is encrypted, not just the individual passwords. This would get decrypted the first time a login detail is needed and re-encrypted every time the state is saved persistently.

app/content/scripts/webviewPreload.js
```diff
@@ -165,6 +165,180 @@
...
+    if (passwords.length === 0 || passwords.length > 3) {
+      return [null, null, null]
+    }
+
+    // look for any form field that has username-ish attributes
+    var username = form.querySelector(['input[type=text][autocomplete=email i]']) ||
+        form.querySelector(['input[type=text][autocomplete=username i]']) ||
+        form.querySelector(['input[type=text][name=email i]']) ||
+        form.querySelector(['input[type=text][name=username i]']) ||
+        form.querySelector(['input[type=text][name=user i]'])
```
bbondy: Does Firefox or Chrome look more extensively for an input without any of those attributes? Possibly by looking at siblings that come before the password page if none of these are found?

bbondy: Nevermind I should have kept reading the code :)

js/components/main.js
```diff
@@ -173,6 +173,10 @@ class Main extends ImmutableComponent {
       windowActions.loadUrl(FrameStateUtil.getFrameByKey(self.props.windowState, frameKey), previousLocation)
     })
 
+    ipc.on(messages.SHOW_USERNAME_LIST, (e, usernames, origin, action, boundingRect) => {
+      contextMenus.onShowUsernameMenu(usernames, origin, action, boundingRect)
```
bbondy: Good idea to use a context menu here

js/contextMenus.js
```diff
@@ -197,6 +197,25 @@ function moreBookmarksTemplateInit (allBookmarkItems, bookmarks, activeFrame) {
...
 }
 
+function usernameTemplateInit (usernames, origin, action) {
+  let items = []
+  for (var username in usernames) {
+    let password = usernames[username]
+    items.push({
+      label: username,
+      click: (item, focusedWindow) => {
+        console.log('item clicked', username, password)
```
bbondy: WIP I assume

bbondy:
Great that you could leverage keytar and thanks for writing tests.
Looking very good so far.

js/contextMenus.js
diracdeltas: Yep, but this should be good to go by end of day. I'll open separate pull req for the other follow ups.

docs/state.md
diracdeltas: Still undecided whether I want to make this change. The benefits of having it the way it is now are:
1. There are some operations that can be done without getting the master password, like seeing which passwords have been saved and deleting them once we have a password manager settings page.
2. No need to figure out `encryptedData` serialization/deserialization.

I think (1) is reason enough to keep it as is for now.

diracdeltas:
@bbondy i think this is ready to merge. seems to work ok on most sites i've tried (gmail, twitter, facebook, zenefits, https://trac.torproject.org/projects/tor/login, etc.) Doesn't work on some fancy forms like quora. will open follow-up issues.

bbondy:
Great work on this!  Merging so it'll be included in 0.8.2.

PR #1079, opened 2016-03-15
tab fixes to realign, and change prev-next tab buttons

bradleyrichter:
@bbondy

### DIFF
--- a/less/bookmarksToolbar.less
+++ b/less/bookmarksToolbar.less
@@ -6,7 +6,7 @@
 
 .bookmarksToolbar {
   background: @toolbarBackground;
-  border-bottom: 1px solid #BBBBBB;
+  border-bottom: 1px solid #aaaaaa;
   border-top: 1px solid #FFFFFF;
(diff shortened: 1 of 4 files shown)


bradleyrichter:
The main change is in the prev/next tab style. I want to reduce the contrast further to reduce the muscle-memory confusion between this button and the back button. Hover turns white to remind that it's a different, and tab set specific button. (later we will also link the next tab set indicator in this hover state.

screen grabs of new prev/next tab style:

normal state:
![screen shot 2016-03-15 at 10 12 40 am]
(https://cloud.githubusercontent.com/assets/13509546/13786824/ada41e7c-ea96-11e5-860c-18e64b9c8309.png)


hover state:
screen shot 2016-03-15 at 10 12 51 am

bbondy:
++ will merge, but please watch for formatting for future commits.
i) color is too far indented 
ii) We shouldn't ever keep around commented code since we have source control which does that for free.
Thx

screenshot 2016-03-15 13 42 40

PR #1085, opened 2016-03-16
bn-BD localization of current version

seeam:
As the title states, created the localized version of current l20n files into Bengali language. Hope it will be merged soon and I also believe brave l20n will be implemented to any localization platform like pontoon or transifex, that will surely pave the way for more people to chime in.


bbondy:
Thank you!

bbondy:
We're going to convert these to l10n files and move to transifex.

PR #1148, opened 2016-03-22
Created the inspect element for right click including the correct locations for the console

bobvanluijt:
Right click inspect element, it will also go to the correct element in the console

PR #1150, opened 2016-03-23
Created inspect element option in contextMenu including the correct locations (x&y) for the console

bobvanluijt:
Right click inspect element, it will also go to the correct element location in the console...

PS:
This is a new request based on: https://github.com/brave/browser-laptop/pull/1148/files

### DIFF
--- a/js/contextMenus.js
+++ b/js/contextMenus.js
@@ -241,7 +241,7 @@ function tabTemplateInit (frameProps) {
           // Handle converting the current tab window into a pinned site
           windowActions.setPinned(frameProps, false)
           // Handle setting it in app storage for the other windows
-          appActions.removeSite(siteUtil.getDetailFromFrame(frameProps, siteTags.PINNED), siteTags.PINNED)
+          appActions.removeSite(siteUtil.getDetailFromFrame(frameProps), siteTags.PINNED)
         }
       })
     } else {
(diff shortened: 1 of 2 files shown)


js/contextMenus.js
```diff
@@ -501,6 +505,19 @@ function mainTemplateInit (nodeProps, frame) {
...
     })
+
+  template.push(CommonMenu.separatorMenuItem)
+  
+  template.push({
+    label: 'Inspect Element',
+    click: (item, focusedWindow) => {
+      if (focusedWindow) {
+        document.querySelector( 'webview' ).openDevTools();
+        document.querySelector( 'webview' ).inspectElement(nodeProps.offsetX, nodeProps.offsetY);
```
bbondy: This is looking a lot better now thanks for cleaning it up. 
Couple of minor suggestions:
i) create a `js/actions/windowAction.js` and put this code in there.  Call it something like inspectElement (x, y).  
ii) Only query the dom element once and store it in a variable, and then do an if (webview) {} in case the dom element doesn't exist.   Also I think you may only need the inpsectElement and not the openDevTools() but whatever is needed is fine.

bobvanluijt:
Cool, what would be the best approach for you, closing this request and create a new one?

bbondy:
No need just do a second commit and push it, it'll auto add to the pull request.  You can also ping me in a new message on this PR after you do that so I see it faster. Thanks!

bbondy:
I'll just merge this in and I'll do the minor cleanup in a folllow up commit. Thanks again for the work, great addition!

bobvanluijt:
Awesome and welcome :+1:
Sorry that I didn't have the time to do those fixes yet. Let me know if you need additional help with this issue...

Issue #1110, opened 2016-03-17
Zoom-in/Zoom-out option


Haze-Aulaian:
Hello Dev team,

i noticed that Control key+mouse scroll doesn't minimize the page like in firefox and chrome .

can that be added to brave?
i notice that Control key with +/- keys only allow for minimize or maximize the page zoom.
one more thing is that the browser resets that zoom selection once you open any site again without saving the preference of the zoom.

regards,
Haze

bbondy:
Fixed via https://github.com/brave/browser-laptop/commit/d70bb786382e512b28af962c6fd8d3c467a5cee3

PR #1182, opened 2016-03-27
Handle mousewheel zoom

nashe:
Aloha,

Should fix issue #1110. However, I did not found (because of no experience with Electron and React at all) why the current `zoomLevel` is not saved to the `WindowStore`.

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -27,6 +27,7 @@ class Frame extends ImmutableComponent {
   constructor () {
     super()
     this.previousLocation = 'about:newtab'
+    this.mouseWheelZoomCount = 0
   }
 
   updateWebview () {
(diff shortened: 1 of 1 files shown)


bbondy:
Sweet, thanks for the patch! I'll take a look tomorrow when I have a mouse.

bbondy:
I think we need a couple changes before we can merge it in.
i) Instead of mouseWheelZoomCount I think we should add another parameter for the action to zoom in / out for step size, and set it to some smaller number.   Currently it feels not responsive enough.
ii) Currently the page scrolls and it zooms when it hits this code, the scroll should be prevented. Possibly by calling event.preventDefault will stop it from scrolling.

bbondy:
For i) Works nicely with the trackpad btw just not the mouse. 
For ii) the problem is still there with the trackpad but it's not as bad as with the mouse.  
I'll take a look at the tweaks as per the Slack conversation.

bbondy:
Merged the follow ups here:
https://github.com/brave/browser-laptop/pull/1200

Thanks again for the work!

PR #1189, opened 2016-03-28
Drop use of defunct frame prop from Main.onCloseFrame

Empact:
There is no frame passed in by the caller of Main, instead use the
props passed in as an argument.


diracdeltas:
looks good to me

PR #1190, opened 2016-03-28
Don’t pass settings into NavigationBar, etc

Empact:
It’s unused there - no need to access it outside of about:preferences

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -355,7 +355,6 @@ class Main extends ImmutableComponent {
     const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
 
     this.frames = {}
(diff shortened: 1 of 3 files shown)


diracdeltas:
Looks fine to me but I'll defer to @bbondy who made the settings page

bbondy:
Looks good but please also remove the urlBar.js one which passes this.props.settings to UrlBarSuggestions since it'll just be passing an undefined value now.  It used to exist because getSettings required this as a param btw.

Empact:
@bbondy Yep that change is in the PR, actually the only thing referred to by the "etc". :)

Only uses now are passing to Frame and internal use in preferences.js, where AboutPreferences passes them down from window.initSettings.

bbondy:
oh ok thanks! merging.

Issue #1184, opened 2016-03-28
HTTPS Everywhere performance improvement


diracdeltas:
It should keep a cache of recently-applied rulesets. HTTPS Everywhere Chromium already does this I think.

Empact:
I'm looking into this.

diracdeltas:
yay

PR #1195, opened 2016-03-28
Cache HttpsEverywhere lookups

Empact:
I have a few questions for this:

* Is expiry necessary?
* Would you prefer to extract a separate function for lookup only, apart from the caching?

#1184


diracdeltas:
ben working on LRU cache right now

Empact:
Updated with an LRU cache now - but haven't done any benchmarking yet.

diracdeltas:
++, lgtm! but @bbondy should probably sign off on the Tracking Protection changes.

bbondy:
lgtm too, merging.

PR #1207, opened 2016-03-31
l20n -> l10n conversion

aekeus:
* UI Translations
  * Menu Translations
  * Preferences Settings
  * l20n -> l10n .properties conversion


app/locale.js
```diff
@@ -0,0 +1,181 @@
...
+    'bookmarksToolbar',
+    'bravery',
+    'replaceAds',
+    'blockAds',
+    'allowAdsAndTracking',
+    'block3rdPartyCookie',
+    'blockPopups',
+    'httpsEverywhere'
+  ]
+}
```
bbondy: would be good to eventually have this figured out dynamically but the scope for that is too big for now.

aekeus: Agreed. Would need to pre-read en-US files to retrieve tokens and fill translation cache.

app/locale.js
```diff
@@ -0,0 +1,181 @@
...
+  // Default to noop callback
+  cb = cb || function () {}
+
+  lang = language
+
+  // Languages to support
+  const langs = [
+    { code: 'en-US' },
+    { code: 'nl-NL' },
+    { code: 'pt-BR' }
```
bbondy: missing: bn-BD

Would be good to have a `docs/localizationDevelopment.md` file or similar which explains some of the manual things we need to do relating to localization.  For example list localizing menu items, what you need to do when adding a new language. etc.

app/locale.js
```diff
@@ -0,0 +1,181 @@
...
+  ]
+
+  // Property files to parse
+  const propertyFiles = [
+    path.join(__dirname, 'locales', lang, 'menu.properties'),
+    path.join(__dirname, 'locales', lang, 'app.properties'),
+    path.join(__dirname, 'locales', lang, 'password.properties')
```
bbondy: I'd be worth trying to automate this, perhaps on a pre-commit hook to verify that the files in en-us match the files here at a minimum.

app/locale.js
bbondy: pre-commit script would apply here too, see below.

app/menu.js
```diff
@@ -38,6 +38,9 @@ let lastSettingsState, lastArgs
  *   bookmarked
  */
 const init = (settingsState, args) => {
+  // This needs to be within the init method to handle translations
```
bbondy: Perhaps explain more about why.

js/about/preferences.js
```diff
@@ -92,6 +92,14 @@ class GeneralTab extends ImmutableComponent {
           value={getSetting(settings.HOMEPAGE, this.props.settings)}
           onChange={changeSetting.bind(null, this.props.onChangeSetting, settings.HOMEPAGE)} />
       </SettingItem>
+      <SettingItem dataL10nId='selectedLanguage'>
+        <select value={getSetting(settings.LANGUAGE, this.props.settings)}
+          onChange={changeSetting.bind(null, this.props.onChangeSetting, settings.LANGUAGE)} >
+          <option data-l10n-id='en-US' value='en-US'/>
+          <option data-l10n-id='nl-NL' value='nl-NL'/>
+          <option data-l10n-id='pt-BR' value='pt-BR'/>
```
bbondy: Pls get this from some centralized required file here and in other places.

js/l10n.js
```diff
@@ -0,0 +1,12 @@
+const ipcRenderer = require('electron').ipcRenderer
+const locale = require('../app/locale')
+
+exports.translation = (token) => {
+  if (ipcRenderer) {
+    // If in the renderer process - ask for translation from main process
+    return ipcRenderer.sendSync('translation', token)
```
bbondy: This is an important one, we should async send the translations to the renderer when the window is created and not on each request for a translated string sync.

app/locale.js
bbondy: A nice way to do this might be via convention. Like anything with `menu` prefix will be included here.

js/l10n.js
```diff
@@ -0,0 +1,24 @@
...
+
+// As for a translation for the current language
+exports.translation = (token) => {
+  // If we are in the renderer process
+  if (ipcRenderer) {
+    // If the token does not exist in the renderer translations cache
+    if (!rendererTranslationCache[token]) {
+      // Ask for all translations from the main process and cache (this will happen once
+      // per renderer process)
+      rendererTranslationCache = ipcRenderer.sendSync('translations')
```
bbondy: I'll merge this anyway, but just noting that I'd prefer no sync IPC here.

bbondy:
Thanks, merging.

luixxiul:
@bbondy, would you please explain how to test translated files locally? Information about what needs to be added or modified to build Brave locally would be helpful. I would like to check the quality and consistency of translation before sending PR.

PR #1211, opened 2016-03-31
Retrieve and store translation files from Transifex

aekeus:


tools/download-languages.js
```diff
@@ -0,0 +1,57 @@
...
+var languages = fs.readdirSync(path.join(__dirname, '..', 'app', 'locales')).filter(function (language) {
+  return language !== 'en-US'
+}).map(function (language) {
+  return language.replace('-', '_')
+})
+
+// Setup the credentials
+const username = process.env.USERNAME
+const password = process.env.PASSWORD
+if (!(username && password)) {
```
bbondy: np here but pls use De Morgan's law as it's more clear for future commits

tools/download-languages.js
```diff
@@ -0,0 +1,57 @@
...
+  throw new Error('The USERNAME and PASSWORD environment variables must be set to the Transifex credentials')
+}
+
+// URI and resource list
+const TEMPLATE = 'http://www.transifex.com/api/2/project/brave-laptop/resource/RESOURCE_SLUG/translation/LANG_CODE/?file'
+const resources = [
+  'app', 'menu', 'downloads', 'preferences', 'passwords', 'bookmarks'
```
bbondy: Can this be obtained dynamically based on the files in locales/en-us ?
Would make 1 less step of maintenance as we add new files.

aekeus: That is a good idea. Will implement.

bbondy:
++ merging

PR #1227, opened 2016-04-02
Consistently unpack the needed selectors under test

Empact:
This was the practice elsewhere, and is more explicit & DRY


bbondy:
++ thanks

PR #1228, opened 2016-04-02
Add waits to the pinnedTabs tests to address intermittent failures

Empact:
On my local environment, this gets test green on a fairly regular basis. Still some intermittent failures hiding out but nothing too egregious.


bbondy:
++ thanks.

PR #1229, opened 2016-04-02
Fix an occasional navigationBarTest failure

Empact:
By applying some setup that is present e.g. on line 102 of this file,
to reveal the urlInput.


bbondy:
++ thanks

PR #1231, opened 2016-04-02
update pt-br translation

charbelrami:


bbondy:
These should be pulled down by a script hat @aekeus is working on from Transifex now. Any changes should be done in Transifex also.  So I will close this based on that info. Thanks!

PR #1233, opened 2016-04-03
Safer 'save password' dialog and better update status

replacepreg:
It’s safer for everyone, specially newbies.
And it’s less stupidly.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -482,7 +482,7 @@ app.on('ready', () => {
         title: 'Save password?',
         message: message,
         buttons: ['Yes', 'No'],
-        defaultId: 0,
+        defaultId: 1,
         cancelId: 1
       }, (buttonId) => {
         if (buttonId !== 0) {
--- a/app/locales/en-US/app.l20n
+++ b/app/locales/en-US/app.l20n
@@ -92,7 +92,7 @@
 <updateOops "Oops!">
 <updateNotYet "Not yet!">
 <updateAvail "A Brave update was just released.">
-<updateNotAvail "Sorry, there are no updates yet. Please check back soon.">
+<updateNotAvail "Brave is up to date. Please check back soon.">
 <updateRequiresRelaunch "Requires a quick relaunch...">
 <updateViewLog "View log">
 <updateHide "Hide">


app/index.js
```diff
@@ -482,7 +482,7 @@ app.on('ready', () => {
         title: 'Save password?',
         message: message,
         buttons: ['Yes', 'No'],
-        defaultId: 0,
+        defaultId: 1,
```
diracdeltas: If the user wants to use a different password manager or none at all, they can turn it off in `about:preferences`. If they have the password manager turned on, they probably want to save passwords most of the time. So the default should be 'yes'. 

BTW, it's usually safer to use a password manager which saves passwords locally encrypted to disk (which we do) than none at all.

app/locales/en-US/app.l20n
```diff
@@ -92,7 +92,7 @@
 <updateOops "Oops!">
 <updateNotYet "Not yet!">
 <updateAvail "A Brave update was just released.">
-<updateNotAvail "Sorry, there are no updates yet. Please check back soon.">
+<updateNotAvail "Brave is up to date. Please check back soon.">
```
diracdeltas: Thanks, I cherry-picked this commit onto master.

diracdeltas:
cherry-picked second commit.

app/index.js
replacepreg: This version (which you do) is suck, because everyone is not expert!
So, for example, I can use my friends computer, I can login to facebook and I can login that way:
1. Login Facebook
2. Fill username, password
3. Type 'Enter'
    A dialog came, I did not read it, or my hand went to key with carelessness.
4. Type 'Enter'

Finally: I fucked up. At least you can remove 'Default focus' from the button. Chrome doing it.

diracdeltas: If someone has your computer for 1 minute, they can install a rootkit on it. They can steal any cookies from sites where you are logged in. They can also open Chrome and get your Chrome passwords from the Chrome built-in password manager.

Removing default focus from a button when the user has chosen "Save passwords" in preferences is just inconvenient. However, I think the issue you are addressing (user accidentally saving a password) will be fixed when the password prompt UI becomes a notification bar instead of a dialog anyway.

replacepreg: Stop expert bullshits.
I think, your job is doing your fucking job. You're sayin 'safer' browser, in fact. It doesn't.
If you using someone's computer for few minutes, going password manager, and taking a screenshot and upload it internet, easier than installing rootkits etc. I'm talking about asking windows/osx password before showing saved fucking passwords.

bbondy: @replacepreg your comment was removed. Feel free to try again without profanity if you'd like to have a productive conversation.

replacepreg: Fuck off cunt.

PR #1293, opened 2016-04-09
Fix for netflix issue

jneumann:
Code written by MegaUber

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -99,6 +99,7 @@ class Frame extends ImmutableComponent {
     const allowRunningInsecureContent = !!(hack && hack.allowRunningInsecureContent)
     if (didSrcChange || didLocationChange && location === 'about:certerror' || !this.webview ||
         allowRunningInsecureContent !== this.webview.allowRunningInsecureContent) {
+      this.webview.setAttribute('plugins', 'true')
       this.updateWebview()
     }
     if (didLocationChange && location !== 'about:certerror' &&
--- a/app/index.js
+++ b/app/index.js
@@ -54,6 +54,44 @@ let authCallbacks = {}
 // Don't show the keytar prompt more than once per 5 minutes
 let throttleKeytar = false
 
+// Widevine include
+let widevineVersion = '1.4.8.866'
+let baseWidevinePath = '../js/plugins/widevine/' + widevineVersion + '/_platform_specific/'
+var widevine_adapter_path = ''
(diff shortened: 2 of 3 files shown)


MegaUber:
I have the plugin files zipped up and ready to be used if someone wants or needs them, well you kind of need them for the plugins to work :)

I uploaded the plugins because they are over 10 megs to [Plugins](http://www.megauberepic.com/downloads/plugins.zip)

they need to be put in the js directory IE: /js/plugins

bbondy:
Which codecs and licensing are the dlls? We need to figure this out before we can merge something like this.  The plugins should also go into a subdir, unless they are GPL, in that case we'd want to download them.  Any patent restrictions on widevine?  

We could alternately not include the DLLS just yet and have it fail silently if they aren't manually copied in.

MegaUber:
I'm not 100% sure, I just put it all together :) 

Makes sense that you'd have to find out all the restrictions of the dlls and other files.

bbondy:
OK Please indicate where the DLLs are from and also put in the pull request for now.

MegaUber:
Something changed in the newest version of Brave has caused this solution to no longer work, when i do navigator.plugins it shows that the plugin in loaded, but it doesnt seem to function. Both flash and Widevine are showing as loaded, but when i go to netflix or amazon prime they still seem to show that widevine needs to be loaded. Any ideas why this would happen?

bridiver:
@MegaUber have you tried clearing ~/.electron and node_modules to make sure you have everything up-to-date? The last release had a lot of changes to add extensions, but I don't think it should have any effect on this

bbondy:
Thanks for this work, it helped us get started with the Flash work going into the next version. I'm going to close this for now since it won't be pulled directly though.

js/components/frame.js
```diff
@@ -99,6 +99,7 @@ class Frame extends ImmutableComponent {
     const allowRunningInsecureContent = !!(hack && hack.allowRunningInsecureContent)
     if (didSrcChange || didLocationChange && location === 'about:certerror' || !this.webview ||
         allowRunningInsecureContent !== this.webview.allowRunningInsecureContent) {
+      this.webview.setAttribute('plugins', 'true')
```
bridiver: is there a reason this is set outside of updateWebview and only on update? Seems like this would fail to set the attribute on a window restore and probably some other cases as well

PR #1301, opened 2016-04-09
Update README.md

Petros0:


bbondy:
Sorry, although I think the wording here can be improved, this doesn't sound right.

PR #1328, opened 2016-04-11
Too many clicks

bridiver:
This also gives activate/open/etc.. consistent behavior that matches Safari and Chrome


bbondy:
Looks good thanks!

Issue #1305, opened 2016-04-09
Password session storage sometimes causing app level exception


bbondy:
We have a couple of reports of this:
<img width="434" alt="screenshot 2016-04-09 12 45 35" src="https://cloud.githubusercontent.com/assets/831718/14404949/165d9876-fe51-11e5-9c61-19373f982246.png">

It looks related to session storage and passwords.

bbondy:
When you fix this problem, please also handle the case that if someone is already in this state it'll gracefully fix it for them. Thanks!

diracdeltas:
argh, it turns out url.parse throws this error if the input is `null`, which is probably what was happening on some pages. fixing

weems:
for historical knowledge I had this app level error when I installed the first 0.9 RC. Seems to be fixed in RC2
![2016-04-08 22_18_28-preferences](https://cloud.githubusercontent.com/assets/470705/14408107/6bc6ccc8-feac-11e5-9753-2e795464f6ab.png)

bbondy:
I think if we used `flow` this would have been prevented for free.  Adding a label so we can keep track if it happens often or not.

diracdeltas:
I think this could have been prevented with `flow` + a string type annotation on the form action. i just tried out `flow weak` on some files and it wasn't as noisy as i thought it would be.

PR #1350, opened 2016-04-13
flow type checking

diracdeltas:
This adds Flow type checking to `app/extensions/brave/brave-default.js` to see if it could have prevented #1305. Note that the Flow type checker will not throw errors when it does not have enough context/annotations to infer types. This means that annotated code will often cause more errors than unannotated code.

Conclusion: #1305 could have probably been prevented with Flow though it would have been a lot of manual work to annotate all types being sent across the IPC boundary. I caught some other potential bugs while working on making `npm run-script flow` pass, so that was good.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,10 +1,9 @@
 [ignore]
-.*/node_modules/.*
 .*/app/extensions/brave/gen/.*
+.*/test/.*/*.json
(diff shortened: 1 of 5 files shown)


bbondy:
Awesome, great start. Thanks!

PR #1351, opened 2016-04-13
Green tests

Empact:
This attempts to solve a few outstanding issues preventing the test suite from being reliable, some of them recently introduced.


bbondy:
Thanks for the fixes,  merging.

PR #1354, opened 2016-04-14
Block (most) canvas and webgl fingerprinting by default

### LINKED ISSUE #694: Fix canvas fingerprinting
PLEASE REMOVE WEBRTC AND CANVAS

http://www.browserleaks.com/webrtc
https://www.perfect-privacy.com/webrtc-leaktest/

http://www.browserleaks.com/canvas
http://venturebeat.com/2014/07/30/canvas-fingerprinting-is-tracking-you-and-you-dont-even-know-what-it-is/

diracdeltas:
This blocks the most common methods of webgl/canvas fingerprinting that rely on scripts like https://github.com/Valve/fingerprintjs2/blob/49e2c11cf08c13e12bf38185cec9ae159025e526/fingerprint2.js. The detection method is adapted from https://github.com/EFForg/privacybadgerchrome.

TODOs:
* Doesn't seem to work 100% of the time, perhaps due to a race condition where the page accesses canvas/gl properties before the content script is loaded.
* This should have a corresponding pref. Probably set to default-off until we have a prompt that lets the user whitelist canvas/gl fingerprinting on a per-site basis.

Test case: see the results on https://panopticlick.eff.org/results?#fingerprintTable for fingerprint and webgl before and after this change.

Fix #694

### DIFF
--- a/app/extensions/brave/locales/en-US/preferences.l20n
+++ b/app/extensions/brave/locales/en-US/preferences.l20n
@@ -38,6 +38,7 @@
 <bookmarks "Bookmarks">
 <openedTabs "Open tabs">
 <doNotTrack "Send a 'Do Not Track' header with browsing requests (requires browser restart)">
+<blockCanvasFingerprinting "Block HTML canvas and WebGL fingerprinting">
 <advancedPrivacySettings "Advanced privacy settings:">
 
 /* Security settings page */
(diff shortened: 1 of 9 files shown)


app/extensions/brave/locales/en-US/preferences.l20n
```diff
@@ -38,6 +38,7 @@
 <bookmarks "Bookmarks">
 <openedTabs "Open tabs">
 <doNotTrack "Send a 'Do Not Track' header with browsing requests (requires browser restart)">
+<blockCanvasFingerprinting "Block HTML canvas and WebGL fingerprinting">
```
bbondy: might be good to add a warning here that it could break some pages.

bbondy:
Sorry for the delay, looks good adn thanks.

app/extensions/brave/locales/en-US/preferences.l20n
diracdeltas: will do

PR #1384, opened 2016-04-15
Retrieve the default locale from electron.app.getLocale()

aekeus:
* Retrieve default locale code
  * Inspect the default locale and modify to conform to xx-XX format

### DIFF
--- a/app/locale.js
+++ b/app/locale.js
@@ -7,6 +7,8 @@
 const L20n = require('l20n')
 const path = require('path')
 const ipcMain = require('electron').ipcMain
+const electron = require('electron')
+const app = electron.app
 
 // Exhaustive list of identifiers used by top and context menus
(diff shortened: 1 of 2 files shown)


app/locale.js
```diff
@@ -147,14 +180,13 @@ exports.init = (language, cb) => {
...
-  lang = language
+  lang = language || defaultLocale()
 
   // Languages to support - TODO retrieve this dynamically
   const langs = [
     { code: 'en-US' },
     { code: 'nl-NL' },
-    { code: 'pt-BR' },
-    { code: 'bn-BD' }
+    { code: 'pt-BR' }
```
bbondy: Even within the same file we have 2 of the same list of languages, please calculate one from the other.

Please merge directly after that change and with the other localization merge that was backed out. Thanks!

PR #1430, opened 2016-04-21
Set proper winBaseUrl based on arch

aekeus:


app/updater.js
```diff
@@ -25,6 +25,7 @@ const updateLogPath = path.join(app.getPath('userData'), 'updateLog.log')
 // in built mode console.log output is not emitted to the terminal
 // in prod mode we pipe to a file
 var debug = function (contents) {
+    console.log(contents)
```
bbondy: nit: remove

app/updater.js
```diff
@@ -47,13 +48,13 @@ exports.updateUrl = function (updates, platform, arch) {
...
   }
   platformBaseUrl = `${updates.baseUrl}/${Channel.channel()}/${version}/${platforms[platform]}`
-  console.log(platformBaseUrl)
   debug(`platformBaseUrl = ${platformBaseUrl}`)
   if (platform === 'darwin') {
     return platformBaseUrl
   } else {
-    debug(updates.winBaseUrl.replace('CHANNEL', Channel.channel()))
-    return updates.winBaseUrl.replace('CHANNEL', Channel.channel())
+    var windowsUpdateUrlWithArch = updates.winBaseUrl.replace('CHANNEL', Channel.channel()) + platforms[platform]
```
bbondy: Variable name a bit off since linux will go in this clause too won't it? Or if not why not?

bbondy:
++ otherwise

app/updater.js
aekeus: Yes, Linux will go here. Naming is a vestige from OSX + Windows only.

app/updater.js
```diff
@@ -47,13 +48,17 @@ exports.updateUrl = function (updates, platform, arch) {
...
   }
   platformBaseUrl = `${updates.baseUrl}/${Channel.channel()}/${version}/${platforms[platform]}`
-  console.log(platformBaseUrl)
   debug(`platformBaseUrl = ${platformBaseUrl}`)
   if (platform === 'darwin') {
     return platformBaseUrl
   } else {
-    debug(updates.winBaseUrl.replace('CHANNEL', Channel.channel()))
-    return updates.winBaseUrl.replace('CHANNEL', Channel.channel())
+    if (platform.match(/^win32/)) {
```
bbondy: is there a possibility for something suffixed to win32?

bbondy: Also just a nit but else if block instead of if/else nested in else.

bbondy: merging as is though

PR #1436, opened 2016-04-22
Latest language download

aekeus:


app/extensions/brave/locales/bn-BD/app.properties
```diff
@@ -0,0 +1,111 @@
+reload=Reload
```
bbondy: It looks like bn-BD properties files is loading in english strings.  Is that expected?

aekeus: Yes, that is what Transfex does when an identifier is not present in the destination language

bbondy:
OK thanks, merging.

PR #1443, opened 2016-04-22
less/updateBar: cleaner look, change buttons to show pointer cursor

williamboman:
*Before*

*After*


bbondy:
Looks good to me but CC'ing @bradleyrichter for sign off on the design look change.

less/updateBar.less
```diff
@@ -31,11 +31,14 @@
   }
 
   .updateButton {
+    cursor: pointer;
```
bbondy: I don't think we want pointer's on buttons at all because it makes it look less native looking that way and less html looking.  Like for example we don't have a pointer on the back / forward buttons.

CC @bradleyrichter for thoughts.

bradleyrichter:
button mouse over style change looks nice. I am not sure about ALL CAPS for the text. Please leave as mixed case for now. (fine with inital cap...)

thank you @williamboman !

bbondy:
So jus the mouse overs, thanks.

williamboman:
Thanks for reviewing! I've updated the patch now.

I feel like the button hover changes should probably be made in the button component itself to keep it consistent throughout the UI, i.e. add hover and active effects to the modifier classes of `.browserButton`?

williamboman:
Also, not really relevant to this PR, but should localized strings be able to have trailing whitespace? If not, I'd suggest adding [this](https://github.com/brave/browser-laptop/commit/4ae961ec27e1481357e0c557f725bccafca88276).

bbondy:
they should be able to have trailing whitespace

bbondy:
> I feel like the button hover changes should probably be made in the button component itself to keep it consistent throughout the UI, i.e. add hover and active effects to the modifier classes of .browserButton? I think the current hover effects on buttons are too subtle.

There's js/components/button.js so it should use that, but I wouldn't worry about it for now.

app/extensions/brave/locales/pt-BR/app.properties
```diff
@@ -78,11 +78,11 @@ blockedTrackingElements={{blockedTrackingElementsSize}} Elementos de rastreament
...
 blockedAds={{blockedAdsSize}} Anúncios bloqueados
 redirectedResources={{redirectedResourcesSize}} Recursos atualizados para HTTPS
-updateNow=atualizar agora
+updateNow=Atualizar agora
 updateChecking=Estamos verificando atualizações...
 updateDownloading=Atualização encontrada, estamos baixando...
-updateLater=mais tarde
-updateDetails=detalhes
+updateLater=Mais tarde
+updateDetails=Detalhes
```
bbondy: These changes will be overwritten, you would need to change them on Transifex for them to take effect.

bbondy: You can consider all localization files except for english to be auto generated from transifex.

bbondy: You can request access here: https://www.transifex.com/brave/brave-laptop/dashboard/

bbondy:
Thanks for the patch! merging.

PR #1446, opened 2016-04-23
extension api updates for https://github.com/brave/electron/pull/18

bridiver:
requires https://github.com/brave/electron/pull/18


bbondy:
We're at 0.37.7 now with your change in as well, so merging this.

Issue #1114, opened 2016-03-17
bookmarks bar site icons


Crocodile73:
would like to see the option, to have the icon of the website on the bookmark bar next too the name. looks better to me.

bsclifton:
I'd love to see the favicons also- checking out this issue...

It would be ideal if their display could be toggleable. Let's see how hard this is :smile:

bsclifton:
I was able to get that area of the code toggling pretty easy:
https://github.com/bsclifton/browser-laptop/commit/2c78f41bdf092e156d26d1ba50bcb53a61615c86

The hard part would be storing the icons and then loading them. I'll keep looking into this

bsclifton:
I got this mostly working, can definitely use an assist cc: @bbondy 

I squashed the previous commit + another coding session into a single commit:
https://github.com/bsclifton/browser-laptop/commit/1cde3bf174233d96d5fb9b22f0924d93a632a5a6

The preferences screen now shows Bookmarks Toolbar / favicon checkboxes that do work; the icon will show (not in the folder unfortunately, need to see why). The positioning is all jacked though :/

bsclifton:
Submitted https://github.com/brave/browser-laptop/pull/1454 which adds icons for the buttons in the Bookmarks Toolbar. If accepted, I'd like to make a 2nd PR which puts the icons into the folders (context menus).

bbondy:
That works, I think Brad wanted the borders though, but I want him to reply first.  @bradleyrichter

bsclifton:
I think we can address everything with the PR:
* Bookmarks toolbar fixes that already work (DONE)
* Adding the icon to context menu, per the comments in the PR (pending)
* UX decisions (borders needed on top of bookmarks toolbar?) and tweaks

I'll try to wrap up the context menu part tonite, will stay tuned for comments :smiley:

bradleyrichter:
@bsclifton @bbondy 

Favicons certainly add to the visual noise, but I know they do help with bookmark recognition. Here is a mockup showing the borders removed:

![image](https://cloud.githubusercontent.com/assets/13509546/14773623/030e7c1a-0a61-11e6-9bd2-ed50fd14a273.png)


I think I prefer the borders but they could be softened later.

I would also prefer to retain the thin height for favicons-off if possible. I don't think there is a happy medium unless it becomes too problematic.

bbondy:
I think we're ready to merge after Brad's comments. Thanks again for the awesome work @bsclifton

[...]

PR #1454, opened 2016-04-23
Bookmarks Toolbar - shows in prefs and icons also show

bsclifton:
Part 1 of a fix for https://github.com/brave/browser-laptop/issues/1114

Added configuration for toggling Bookmarks Toolbar / favicon. Favicon now shows in Bookmarks Toolbar (not in folders though- those use the ContextMenu object).

Here's a pic which shows the preferences along with how the icons look in the menu. Ones that receive a 404 unfortunately don't have a default icon.

bookmarks_toolbar

### DIFF
--- a/js/components/bookmarksToolbar.js
+++ b/js/components/bookmarksToolbar.js
@@ -126,6 +126,8 @@ class BookmarkToolbarButton extends ImmutableComponent {
           backgroundSize: iconSize,
           height: iconSize
         })
+      } else {
+        showFavicon = false
       }
     }
 
@@ -255,12 +257,15 @@ class BookmarksToolbar extends ImmutableComponent {
     contextMenus.onTabsToolbarContextMenu(this.props.activeFrame, closest && closest.props.bookmark || undefined, closest.isDroppedOn, e)
   }
   render () {
+    let showFavicon = getSetting(settings.SHOW_BOOKMARKS_TOOLBAR_FAVICON) === true
     this.bookmarkRefs = []
     return <div
-      className={cx({
-        bookmarksToolbar: true,
-        allowDragging: this.props.shouldAllowWindowDrag
-      })}
+      className={
+        cx({
+          bookmarksToolbar: true,
+          allowDragging: this.props.shouldAllowWindowDrag
+        }) + (showFavicon ? ' showFavicons' : '')
+      }
       onDrop={this.onDrop.bind(this)}
       onDragEnter={this.onDragEnter.bind(this)}
       onDragOver={this.onDragOver.bind(this)}
--- a/js/components/bookmarksToolbar.js
+++ b/js/components/bookmarksToolbar.js
@@ -16,6 +16,8 @@ const Button = require('../components/button')
 const cx = require('../lib/classSet.js')
 const dnd = require('../dnd')
 const dndData = require('../dndData')
+const settings = require('../constants/settings')
+const getSetting = require('../settings').getSetting
 
 // TODO: Obtain from the less file
 const bookmarkMaxWidth = 100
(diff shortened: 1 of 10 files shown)


bbondy:
Looking great so far, thanks!

If you add a bookmark folder, then the items inside those menus should have an icon as well.
You can specify the icon in `js/contextMenus.js`.  Just add a new property on `bookmarkItemsInit`'s `templateItem` object.

Then add handling to display that icon in `js/components/contextMenu.js`'s `ContextMenuItem`'s `render`.

I also noticed that for long bookmark names I think it used to have ... for them and now they abruptly get cut off.  

More comments to come inline.

bbondy:
The overflow menu when you have a lot of bookmarks in the bookmark toolbar also needs support.  I think it will be handled in the same `bookmarkItemsInit` though.

less/bookmarksToolbar.less
```diff
@@ -12,7 +12,7 @@
   display: flex;
   flex: 1;
   padding: 3px 10px;
-  height: 23px;
+  height: 30px;
```
bbondy: I know we want to be competitive on the amount of height our browser chrome takes up (it matters a lot to some people).  Increasing by 7px always wouldn't fly.  

I see 2 options:
i) Add a className in React code that's handled here that specifies whether or not to use 23px for no favicon or 30px for favicons. (call the class something like `showFavicons`).
or ii) Always use 23px and just resize the favicons smaller.
CC @bradleyrichter

bbondy:
everything else looks great to me!

less/bookmarksToolbar.less
bsclifton: Conditionally making it 30px would be easy :smiley: For reference, Chrome is 27px (between the url bar and the bottom of bookmarks toolbar). In Brave though, we have a bordered style above the bookmarks toolbar, which makes it harder to fix this in reasonably (padding, etc).

There are also some borders:
* class style "navigatorWrapper" has 1px border-bottom w/ #aaa
* class style "bookmarksToolbar" has 1px border-top w/ #fff

In addition to conditionally setting the size to 30px, I'd like to propose conditionally eliminating these two borders *IF* the bookmarks tb is shown (it's needed when not using the bookmarks tb, since there is no other border below)

bsclifton:
Per your comment above @bbondy (collapsed by GitHub), I changed bookmarks tb to be conditionally larger if icons show. I also conditionally hid the border on the top, to make things look cleaner (like Chrome). Since there's not longer a border on top and bottom, I could reduce the size of the bookmarks tb (see screenshot below, which shows an item with mouse over)

bookmark-toolbar-v2

bsclifton:
Sorry for all the rebasing, I had to switch between two PCs when file modes get jacked up on Windows :/ Got that fixed in my git config now though

bsclifton:
Should be GTG! 

The only known issue is that the ellipsis no longer shows for items. This happens because the "bookmarkToolbarButton" class was modified to add
* align-items: center;
* display: flex;

This was needed to align the icon/text properly (styles copied from the tab).

To fix the ellipsis issue, I'll definitely need some help (my flex game is weak!)
bookmarks_toolbar_context_menu

js/components/bookmarksToolbar.js
```diff
@@ -231,12 +257,15 @@ class BookmarksToolbar extends ImmutableComponent {
...
     return <div
-      className={cx({
-        bookmarksToolbar: true,
-        allowDragging: this.props.shouldAllowWindowDrag
-      })}
+      className={
+        cx({
+          bookmarksToolbar: true,
+          allowDragging: this.props.shouldAllowWindowDrag
+        }) + (showFavicon ? ' showFavicons' : '')
```
bbondy: I'll merge this and fix but just an fyi that you can do this:
```
 cx({
  bookmarksToolbar: true,
  allowDragging: this.props.shouldAllowWindowDrag,
  showFavicons: showFavicon
})
```

It's basically just a way to have object properties with true/false values and if they are true the key name is converted to a string.  It's a nicer format than a bunch of ternary checks because it's so common to include / exclude for classes.

bbondy:
looks good to me thanks!

bradleyrichter:
perfect. thanks!

PR #1456, opened 2016-04-23
fix text selection detection

### LINKED ISSUE #1455: context menu doesn't open in email fields
email doesn't have a selectionStart or selectionEnd so it throws an exception in brave-default.js hasSelection

bridiver:
fixes #1455


bbondy:
nice

PR #1484, opened 2016-04-27
Replace modal dialogs with notification bars

### LINKED ISSUE #923: Make permission and password prompts testable and base page interactive

### LINKED ISSUE #1049: permission prompts should be able to save decisions between restarts
Currently it can only save the decision until the browser restarts; would be good to have an option to save the decision on the site persistently. This blocks somewhat on #923 because otherwise the UI to present the 6 different permission options (`[Allow, Deny] * [This time only, Until browser restart, Persistently]`) will be unpleasant.

There also needs to be some sort of UI in `about:preferences` to manage all the site-specific decisions that the user has made. This is also needed for when we have site-specific Bravery settings. Chrome has something like this in chrome://settings -> content settings.
screen shot 2016-03-09 at 11 36 38 am

### LINKED ISSUE #1201: the password manager prompt should have an option to remember the setting
when prompted to save a password or for notification/etc.. permissions, there should be an option to save the response (past the current session). This probably also requires a UI to manage the saved responses or at the very least a method to reset all of them.

### LINKED ISSUE #1206: password save prompt stops feedly login window from closing
Steps to reproduce:
1. Log out of feedly and any service you use as a login provider (google, etc..)
2. Click the login link and login using the popup window
3. When prompted to save your password click "no"
4. Feedly usually opens in the main window, but the popup doesn't close. Sometimes feedly also fails to reload

diracdeltas:
* replaces the modal dialogs for password saving / permission requests with a notification bar at the top of the browser. I decided that notification bars should *not* be per-frame, since they refer to top-level app state.
* add site-specific settings for permissions
* adds "never save password on this site" option

Fix #1049, fix #1206, fix #923, fix #1201 

Auditors: @bbondy

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -45,9 +45,9 @@ const downloadMap = {}
 const registeredSessions = {}
 
 /**
- * Maps notification messages to their callback
+ * Maps permission notification bar messages to their callback
  */
-const notificationCallbacks = {}
+const permissionCallbacks = {}
 
 module.exports.registerBeforeSendHeadersFilteringCB = (filteringFn) => {
   beforeSendHeadersFilteringFns.push(filteringFn)
@@ -254,17 +254,20 @@ function registerPermissionHandler (session) {
       return
     }
     const message = `Allow ${host} to ${permissions[permission].action}?`
-    if (message in notificationCallbacks) {
+    if (message in permissionCallbacks) {
       // This notification is already shown
       return
     }
 
     appActions.showMessageBox({
       buttons: ['Deny', 'Allow'],
+      options: {
+        persist: true
+      },
       message
     })
-    notificationCallbacks[message] = (buttonIndex, persist) => {
-      let result = !!(buttonIndex)
+    permissionCallbacks[message] = (buttonIndex, persist) => {
+      const result = !!(buttonIndex)
       cb(result)
       if (persist) {
         // remember site setting for this host over http(s)
@@ -403,9 +406,9 @@ module.exports.init = () => {
     }
   })
   ipcMain.on(messages.NOTIFICATION_RESPONSE, (e, message, buttonIndex, persist) => {
-    if (notificationCallbacks[message]) {
-      notificationCallbacks[message](buttonIndex, persist)
-      delete notificationCallbacks[message]
+    if (permissionCallbacks[message]) {
+      permissionCallbacks[message](buttonIndex, persist)
+      delete permissionCallbacks[message]
     }
     appActions.hideMessageBox(message)
   })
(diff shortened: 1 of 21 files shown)


app/filtering.js
```diff
@@ -198,32 +204,25 @@ function registerPermissionHandler (session) {
   // TODO: Localize strings
   let permissions = {
     media: {
-      action: 'use your camera and/or microphone',
-      hosts: {}
+      action: 'use your camera and/or microphone'
```
bbondy: You don't have to now but this will require a bit of localization work. The strings should go into locale.js and also the string action shouldn't be a variable.  Only host would be a variable.  The action might move around in the sentence in different languages.

app/index.js
```diff
@@ -494,21 +504,33 @@ app.on('ready', () => {
...
-        cancelId: 1
-      }, (buttonId) => {
-        if (buttonId !== 0) {
+        ? `Would you like Brave to remember the password for ${username} on ${origin}?`
+        : `Would you like Brave to remember your password on ${origin}?`
+
+      if (!(message in passwordCallbacks)) {
+        // Notification not shown already
+        appActions.showMessageBox({
+          buttons: ['Yes', 'No', 'Never for this site'],
```
bbondy: always for this site?

bbondy:
Fantastic, thanks! Merging.

app/index.js
bbondy: Nevermind, that is implied by Yes.

docs/state.md
```diff
@@ -32,7 +32,15 @@ AppStore
...
-      zoomLevel: number
+      zoomLevel: number,
+      mediaPermission: boolean,
+      geolocationPermission: boolean,
+      notificationsPermission: boolean,
+      midiSysexPermission: boolean,
+      pointerLockPermission: boolean,
+      fullscreenPermission: boolean,
+      openExternalPermission: boolean,
+      savePasswords: boolean // Only false for now
```
bbondy: Only false or undefined for now right?

app/index.js
diracdeltas: Sort of. "Always for this site" would mean that any password you enter on foo.com is always saved/updated automatically. I don't think this is a useful option because if someone enters a wrong password, this will permanently delete their previously-saved password.

docs/state.md
diracdeltas: yep it can be undefined. will update

PR #1491, opened 2016-04-27
Added missing entries from non-english locales

bsclifton:
Something that was missed from https://github.com/brave/browser-laptop/pull/1454


bbondy:
These are auto generated files only so if you wanted to add them you'd have to request access on 
https://www.transifex.com/brave/brave-laptop/dashboard/

But there's no benefit to do it unless you actually have translated strings because it falls back to english.

PR #1505, opened 2016-04-29
.editorconfig: override trim_trailing_whitespace for *.properties as well

williamboman:
ref https://github.com/brave/browser-laptop/pull/1443#issuecomment-214965704

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,5 +12,5 @@ end_of_line = lf
 charset = utf-8
 insert_final_newline = true
 
-[*.md]
+[*.{md,properties}]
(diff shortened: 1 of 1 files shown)


bbondy:
There are some translated strings that have trailing whitespace that do not appear correctly in the UI currently.  I'd like to have this but that needs to be fixed first and changes to the components that use those strings should be updated first.  There's at least one I think.

bbondy:
Fixed that last string with a required trailing whitespace so  this can be merged now.

bbondy:
Thanks!

PR #1507, opened 2016-04-29
show urlbar when hovering tabs & loadtime hidden for about pages

### LINKED ISSUE #1486: trigger title bar/URL bar change on Tab-Bar mouse enter
This will give users more time to aim for the hidden buttons like reload, bookmark, and the future home button.

### LINKED ISSUE #1503: hide "page load speed" for new tabs since they are empty and should load in zero time
we may also want to convert milliseconds to seconds if we want more users to understand it.

psimyn:
show urlbar when mouse is over tabbar, for easier navigation of nav
resolves #1486

hide loadtime for about: pages
always show loadtime in seconds for consistency
fixes #1503

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -259,7 +259,7 @@ class Main extends ImmutableComponent {
   }
 
   checkForTitleMode (pageY) {
-    const navigator = document.querySelector('#navigator')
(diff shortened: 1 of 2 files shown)


bbondy:
great, thank you!

PR #1523, opened 2016-05-01
Fixed issue w/ overflow on hamburger menu; it was causing scrollbars …

bsclifton:
Possibly only an issue on Windows? Caused by this commit:
https://github.com/brave/browser-laptop/commit/b3759626137c673be083b0bf26692dd258b59186#diff-368c5017488bea1c7d085fe1054d0684R21

Here's what the issue looks like before this patch:
hamburger

And here's a screencap w/ the dev tools open showing how the submenus have overflow:

### DIFF
--- a/less/contextMenu.less
+++ b/less/contextMenu.less
@@ -18,7 +18,6 @@
   // and bookmarks overflow menu reaching down too low.
   @usedUpChrome: @navbarHeight + @bookmarksToolbarWithFaviconsHeight;
   max-height: calc(~'100% - @{usedUpChrome}');
-  overflow: auto;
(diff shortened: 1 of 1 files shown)


bbondy:
Could you try leaving `auto` for `overflow-y` and not for `overflow-x`?  This `overflow: auto` was just recently added to make bookmark toolbar folders scrollable and the hamburger menu scrollable.   I think that will fix both problems.

bsclifton:
Unfortunately, it only seems to work with:
`overflow-y: visible`

:cry:

bbondy:
Thanks!

PR #1551, opened 2016-05-03
Move inline JS into external files

diracdeltas:
so that @bridiver doesn't have to update the manifest CSP every release

auditors: @bbondy


diracdeltas:
my build is taking forever so i haven't checked for CSP violation errors

bbondy:
sorry for the delay, got side tracked with an urgent extra release today. Thanks for doing this.

PR #1564, opened 2016-05-04
Add NoScript feature

### LINKED ISSUE #231: add NoScript-like feature
some power users have asked for the ability to block js from running except on whitelisted sites. Chrome has a feature to whitelist/blacklist scripts by hostname pattern, which may be sufficient.

diracdeltas:
This adds a noscript feature which can be turned off globally, for the current site temporarily, or for the current site permanently.

I chose to put the noscript site options in a popup on the right side of the urlbar, which is what chrome does. Ideally this would be merged with the UI for per-site bravery settings when that exists. CC @bradleyrichter for design thoughts.

Fix #231 

screen shot 2016-05-04 at 3 12 14 pm
^ noscript popup appears when clicking on the red ban icon. ideally it would be something similar to the noscript logo, which is what noscript users are used to clicking on.
noscript_logo


diracdeltas:
Will fix merge conflicts once errors on master are fixed

bbondy:
Noting a couple of follow ups we'll need:
- Ability to clear once permanently accepted? Perhaps while on the page or an about page whatever is easier / better.
- Ability to turn this feature on / off in preferences, kind of like canvas fingerprinting.

bbondy:
minor bug: text looks selectable by cursor in the popup, just set no text selection in that popup.

js/state/siteUtil.js
```diff
@@ -334,3 +335,22 @@ module.exports.clearSitesWithoutTags = function (sites) {
...
+ * @return {string?}
+ */
+module.exports.getOrigin = function (location) {
+  // Returns scheme + hostname + port
+  if (typeof location !== 'string') {
+    return null
+  }
+  let parsed = urlParse(location)
+  if (parsed.host) {
+    return location.split(parsed.host)[0] + parsed.host
```
bbondy: minor bug here where if a hostname is the same as a protocol name then this won't work correctly.   Maybe I could have a computer on my network called http for example.  Far fetched example but could happen.

bbondy:
Another minor one, notification window should not drag the whole window.

app/noScript.js
```diff
@@ -0,0 +1,72 @@
...
+  if (!Filtering.isResourceEnabled(module.exports.resourceName)) {
+    return result
+  }
+  // Ignore whitelisted URL schemes and non-applicable resource types
+  let parsed = urlParse(details.firstPartyUrl)
+  if (['about:', 'chrome:', 'chrome-extension:'].includes(parsed.protocol) || ['stylesheet', 'script', 'image'].includes(details.resourceType) || parsed.hostname === 'localhost') {
+    return result
+  }
+
+  let origin = siteUtil.getOrigin(details.firstPartyUrl)
```
bbondy: I think there's a minor problem here from what I'd expect.  For example on my website (brianbondy.com) I have a video in the first blog post.  If I accept once the video doesn't load, I think because the iframe has a different first party url.   I'm ok with landing this as is though but just post an issue for that to handle it after in a different way (whitelisting the whole frame once we can reliably determine frame from filtering functions)

docs/state.md
```diff
@@ -86,6 +87,9 @@ AppStore
   cookieblock: {
     enabled: boolean // Enable 3p cookie/referer blocking
   },
+  cookieblock: {
```
bbondy: I think this should be `noScript`

app/noScript.js
bbondy: Note I could see a slight argument for not allowing the iframe origins but allowing a permanent exception seems to allow the iframes afterwards.

js/components/frame.js
```diff
@@ -433,6 +435,16 @@ class Frame extends ImmutableComponent {
     this.webview.addEventListener('media-paused', ({title}) => {
       windowActions.setAudioPlaybackActive(this.props.frame, false)
     })
+    this.webview.addEventListener('console-message', (e) => {
+      if (this.props.enableNoScript && e.level === 2 &&
+          e.message && e.message.includes(consoleStrings.SCRIPT_BLOCKED)) {
+        // Note that the site was blocked
+        // TODO: Parse out the location of the script that was blocked and send
+        // it too
+        windowActions.setBlockedBy(this.props.frame,
```
bbondy: Would prefer to store a list of URLs instead of the full message in case we need t use the info for something else later on.

js/constants/console.js
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const strings = {
+  // Noscript
+  SCRIPT_BLOCKED: 'violates the following Content Security Policy directive: "script-src \'none\'"'
```
bbondy: Ideally this would be localized but not overly concerned about it

bbondy:
Overall this is looking great. Feel free to merge and work on the follow ups as new tasks.

bradleyrichter:
@diracdeltas I hope we don't end up here from a power-user-only perspective:

image

diracdeltas:
> Noting a couple of follow ups we'll need:
> 
> Ability to clear once permanently accepted? Perhaps while on the page or an about page whatever is easier / better.

Yeah, this will have to exist for all per-site bravery settings. Maybe in Preferences > Privacy where the per-site permissions are? Chrome puts them all in content settings.

> Ability to turn this feature on / off in preferences, kind of like canvas fingerprinting.

Should this go in Security like canvas fp, or in the future Bravery tab in prefs?

diracdeltas:
> @diracdeltas I'm concerned that we will up here from a power-user-only perspective:

@bradleyrichter noscript's UI is overkill for most users, probably even most NoScript users. My sense is that most people don't need to whitelist on a per-script basis. Compare to Chrome's native script-blocking UI:
screen shot 2016-05-04 at 4 34 58 pm

I would lean towards looking more like Chrome than NoScript in the long-run.

js/components/frame.js
diracdeltas: The URL is included in the full message, it just needs to be parsed out at some point

bbondy: ok, it was just strange to see the full message in session store.

app/noScript.js
diracdeltas: I think I see the issue. foo.com embeds frame from bar.com. User whitelists foo.com once, and the page is reloaded. Whichever frame loads first (probably the main frame) gets allowed through in the code block below, and then `foo.com` gets removed from `temporarilyAllowed`, so the second frame load is blocked. Will fix.

docs/state.md
diracdeltas: ++

js/constants/console.js
diracdeltas: it's a console string from blink (not us) and most users will never see it; not sure if localization is needed

js/state/siteUtil.js
diracdeltas: nice catch. I just realized `url.parse` has a `slashes` property, which means it can reconstruct the origin unambiguously. will fix.

js/components/frame.js
diracdeltas: note to self: need to clean noscript data from frames when the session ends

acvanzant:
I want to vote for more granular whitelisting and temporary whitelisting. I don't want to have to actually, you know, view the code to know all of the scripts that are going to be run when I allow all javascript for the whole domain. 

I don't use the full noscript UI but I also rarely allow all scripts for any site so I'd personally like something more than what is currently here.

diracdeltas:
fixed except for https://github.com/brave/browser-laptop/issues/1586

PR #1587, opened 2016-05-06
use extension urls in dev

bridiver:
use extension urls in dev and dynamically generate manifest file
auditors: @bbondy


bbondy:
thanks for doing this!

PR #1629, opened 2016-05-08
use extension urls in dev

bridiver:
fix for earlier PR - should always match on extension url for about content scripts
@bbondy @bsclifton


bbondy:
Thanks!

Issue #1524, opened 2016-05-01
Show menu by default on Windows and allow to turn it off


bbondy:
A lot of times Windows users think we're missing functionality because there's no menu bar visible in Windows.  Of course like in most Windows apps you can press Alt to show it, but a lot of users don't see that.

This task is to:
i) Show the menu bar by default on Windows
ii) Add another context menu item next to the Bookmarks Toolbar one for Menu Bar. 

CC @bradleyrichter if he has objections.

Originally it was hidden by default because on Windows we have extra window chrome from the titlebar. Ideally we could get rid of that first bu this task would be a lot easier to do first.

bbondy:
btw I'd still like us to have the on / off context menu even if we don't change the default. 
Also you can change this `autoHideMenuBar: true,` in appStore.js.
CC @bsclifton  in case he's interested in taking and typing this to a pref.

bsclifton:
Should be a slam dunk- will check out later today

bradleyrichter:
We need this. Too many support requests related to not finding the menu.

bsclifton:
Started to work on this...
https://github.com/bsclifton/browser-laptop/commit/97f789898b080933ecd88c4c0763b8b082a6805b

Requires a browser restart at the moment, trying to find a clean way to hook into electron::BrowserWindow#setMenuBarVisibility(visible) when the property changes

Should this change only be visible to Windows users? Or visible to all but only defaulted to show the menu on Windows?

SJ-Diamond:
Would make sense to make the hamburger menu disappear if the menu bar is shown. Would be nice if you would implement that.

bradleyrichter:
I am a little concerned about things moving around when hiding the burger. 

Before we add this hiding ability, let me look at how "those other browsers" deal with this issue. 

Tx - Brad

bbondy:
I think the hamburger menu is still useful for ease of access for common things like zoom for example.  If we're going to hide it I'd rather just have a second checkbox like the Show Home Button option.

bsclifton:
Should be completely fixed by https://github.com/brave/browser-laptop/pull/1630

Ready for review :smile:

luixxiul:
Closing as this was already fixed.

PR #1630, opened 2016-05-08
about:preferences now has "autohide menu" option

bsclifton:
Updates to resolve https://github.com/brave/browser-laptop/issues/1524

- Home button and this new setting pulled into an "Appearance settings" area of the general tab.
- Change takes effect immediately

~~What is missing (will address in future PR):~~
- ~~context menu item in the bookmarks toolbar~~
- ~~default this new option to "false", but only for Windows users~~

### DIFF
--- a/app/extensions/brave/locales/en-US/menu.properties
+++ b/app/extensions/brave/locales/en-US/menu.properties
@@ -120,4 +120,4 @@ zoom=Zoom
 new=New
 learnSpelling=Learn Spelling
 ignoreSpelling=Ignore Spelling
-autoHideMenuBar=Hide the menu bar
+autoHideMenuBar=Hide Menu Bar
--- a/app/extensions/brave/locales/en-US/menu.properties
+++ b/app/extensions/brave/locales/en-US/menu.properties
@@ -120,3 +120,4 @@ zoom=Zoom
 new=New
 learnSpelling=Learn Spelling
 ignoreSpelling=Ignore Spelling
+autoHideMenuBar=Hide the menu bar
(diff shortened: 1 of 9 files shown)


app/extensions/brave/locales/en-US/menu.properties
```diff
@@ -120,3 +120,4 @@ zoom=Zoom
 new=New
 learnSpelling=Learn Spelling
 ignoreSpelling=Ignore Spelling
+autoHideMenuBar=Hide the menu bar
```
bbondy: nit: Hide Menu Bar

bbondy:
This is looking good but the main thing I see before this can be merged in is that there should be no option for this at all for OS X (process.platform === 'darwin')

bsclifton:
Good catch, thanks :smiley: I'll tweak this PR to hide the preference (and the context menu) on OS X

bsclifton:
Screenshots (NOTE: changes only show for Windows / Linux)

New option in preferences


New context menu item when right clicking on bookmarks toolbar

bbondy:
Manually merged because of a conflict, thanks!

luixxiul:
On Ubuntu Hide Menu Bar seems to hide nothing. Is it only for Windows?

bbondy:
I actually went back and forth about linux, does it not hide anything for all distros? I noticed the electron code only ignores OS X.

luixxiul:
Turning on that option does not hide the bookmark toolbar. Rather than that I'm not sure what it is supposed to do because it is unclear to me where the menu bar is.

bsclifton:
Interesting... is the menu in BrowserWindow a Windows only feature? 

It's supposed to hide this menu (see picture) which is toggleable by hitting alt (at least on Windows) and hidden by default (before this PR- which sets it to shown for Windows by default):
menu

luixxiul:
@bsclifton On Unity (Ubuntu) like this:



Not sure about other desktop environments.

bsclifton:
ah- it seems at least with Ubuntu/Unity, the menu ALWAYS shows

bbondy:
I'll post separate t fix for Ubuntu/Unity but probably not the highest priority.  The option is respected and works properly in Dedbian with Gnome.

Issue #1628, opened 2016-05-08
Unnatural space next to back button (Windows 10)


luixxiul:
On Windows 10 there is unnatural space next to the back button, under "New Tab" (title bar) 

![](http://i.imgur.com/zL4xQZN.jpg)

Intended?

bsclifton:
I also don't like this- it would be easy to fix
cc: @bradleyrichter

bbondy:
This is just because of the OS X controls on the left, on Windows the gap shouldn't be there.

I'd imagine adding some css class top level that's called isWindows via check for `process.platform === 'win32'` would do the trick.

bradleyrichter:
I'd love to have this fixed.

bsclifton:
https://github.com/brave/browser-laptop/pull/1638 ready for review :smile:

duff2481:
@luixxiul using 0.9.5 version on Windows machine and I have the start bar on top of my screen, not the bottom.  I constantly need to readjust each time brave is loaded.

bbondy:
What is the start bar? Do you mean the Windows task bar?

duff2481:
yes, task bar.

luixxiul:
Would you please take a screenshot?
2016/05/13 22:36 "duff2481" <notifications@github.com>:

duff2481:
![brave_default_opening](https://cloud.githubusercontent.com/assets/19331241/15253922/57b62416-1902-11e6-957b-475e2a97cb4d.png)

luixxiul:
By the way on Windows 10,

![](http://i.imgur.com/VCUjf0n.jpg)

Attention to the top right corner.

Also I opened the new issue #1727 for that issue, thanks.

PR #1638, opened 2016-05-09
Fix for https://github.com/brave/browser-laptop/issues/1628

bsclifton:
An elegant fix was available, thanks to:
https://github.com/brave/browser-laptop/blob/master/js/components/window.js#L55

This PR:
- removes the hardcoded 70px margin from navigatorWrapper
- properly calculates center offset if margin is 0
- will add 70px margin & account for margin when calculating center offset **if platform is darwin**

I tested the manually and verified this works.  Mac users should be good to go- I tested their style by using ".platform--win32" and confirmed it looks the same.

Screencap on Windows after the fix:
spacing-issue-fix


bbondy:
Looks great, thanks!

PR #1653, opened 2016-05-10
Update install instruction for Debian

zgrimshell:
dpkg -i shouldn't be used as there is a good chance of missing dependencies and it can't handle that. So instead gdebi should be used. Also there is no need for "./" in front of name of the package as the command will be issued in same directory where wget downloaded it (assuming user doesn't close and reopen terminal between commands in which case ./ still wouldn't be sane as it is used to detect executable and not a path). This issues is probably the same for other Debian based distributions (such as Ubuntu and Mint) so feel free to edit them too (I don't use them so I can't comment 100% on it).


bbondy:
sounds good! Worked for me in Debian.

PR #1654, opened 2016-05-10
send brave extension manifest as an object

bridiver:


bbondy:
lgtm

bbondy:
this was reverted because it was causing a startup crash

PR #1655, opened 2016-05-10
Fix to get urlBar context menu working again

bsclifton:
Was broke w/ this:
https://github.com/brave/browser-laptop/commit/d5c102e9079fb792ef3e100cf4eb09e912198c1f#diff-bf31d40b9ccd66ab63bd8cf24b30eab0R255

Just needed to propagate event object


bbondy:
nice find, thanks!

PR #1662, opened 2016-05-10
use data urls for brave extension because chrome-extension urls don't…

### LINKED ISSUE #1656: About page favicon
They were originally added here:
https://github.com/brave/browser-laptop/issues/898#issuecomment-218091747

Maybe broken with extensions @bridiver

bridiver:
… load correctly

fixes #1656

### DIFF
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -54,7 +54,8 @@ let generateBraveManfiest = () => {
       'default-src': '\'self\'',
       'form-action': '\'none\'',
       'referrer': 'no-referrer',
-      'style-src': '\'self\' \'unsafe-inline\''
+      'style-src': '\'self\' \'unsafe-inline\'',
+      'img-src': '\'self\' data:'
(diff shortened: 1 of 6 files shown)


app/extensions.js
```diff
@@ -54,7 +54,8 @@ let generateBraveManfiest = () => {
       'default-src': '\'self\'',
       'form-action': '\'none\'',
       'referrer': 'no-referrer',
-      'style-src': '\'self\' \'unsafe-inline\''
+      'style-src': '\'self\' \'unsafe-inline\'',
+      'img-src': '\'self\' data:'
```
bbondy: wouldn't it be possible just to specify the file path of the icon instead?

bridiver: the loading issue actually has nothing to do with CSP, this was just required to allow the data urls

bbondy:
lgtm

PR #1665, opened 2016-05-10
update url for all protocols

### LINKED ISSUE #1618: Address bar doesn't change during "tabnapping" phishing attempt
Tested on Brave 0.9.4.

Brave, like many other browsers, is vulnerable to "[tabnapping](https://en.wikipedia.org/wiki/Tabnabbing)", where an attacker-controlled application opens a foreground tab to a trusted website, and after the user judges the website is safe, the attacker uses `window.location` to replace the foreground tab with a malicious document.

A proof of concept is available here: http://lcamtuf.coredump.cx/switch/

Clicking "Do it" opens a new tab with a bank login page. After 7.5 seconds, the page will be transparently replaced with a `data:` URI of a fake bank login page.

In all other browsers, the address bar will change to the `data:` URI, thus alerting a keen-eyed user.
In Brave, the address bar does **not** change. (To make matters worse, even if this is fixed, in brave-laptop you also have to hover over the title bar to even see the address.
The only tell-tale sign that something is happening is a new page is added to the history list, so the "back" button becomes active.

To test the above PoC on Brave, you might want to save the page and edit the "bank_html" to some other text, as a proof that the page is being replaced. Otherwise, there's no indication of it happening.

Even if Brave doesn't wish to block this kind of attack altogether (and possibly break behaving websites that use it for legitimate reasons), at least change the address in the location bar, and perhaps make it visible for a moment, so the user would have a chance to see something strange is going on.

brave-android behaves differently:
Once the `window.location` is modified with a `data:` URI, it launches Chrome.

The browser might also forget to change the address bar during a "[reverse tabnapping](https://sites.google.com/site/bughunteruniversity/nonvuln/phishing-with-window-opener)", where the opened tab can affect the opener tab using `window.opener.location.assign`, but I did not test that.

bridiver:
fixes #1618

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -337,8 +337,9 @@ class Frame extends ImmutableComponent {
           this.props.frame)
         const key = this.props.frame.get('key')
         const parsedUrl = urlParse(event.url)
+        // don't change url for non-display protocols like mailto
         if (['http:', 'https:', 'about:', 'chrome:', 'chrome-extension:', 'file:',
-             'view-source:'].includes(parsedUrl.protocol)) {
+             'view-source:', 'ftp:', 'data:'].includes(parsedUrl.protocol)) {
           windowActions.setLocation(event.url, key)
         }
         const hack = siteHacks[parsedUrl.hostname]
--- a/test/components/navigationBarTest.js
+++ b/test/components/navigationBarTest.js
@@ -45,6 +45,28 @@ describe('urlbar', function () {
   }
 
   describe('navigation', function () {
+    describe.only('tabnapping', function () {
+      Brave.beforeAll(this)
+
+      before(function * () {
+        var page1 = Brave.server.url('tabnapping.html')
+        yield setup(this.app.client)
+        yield this.app.client
+          .tabByUrl(Brave.newTabUrl)
+          .url(page1)
+          .waitForUrl(page1)
+          .leftClick('#open_target')
+      })
+
+      it('updates the location in the navbar when changed by the opener', function * () {
+        yield this.app.client
+          .windowByUrl(Brave.browserWindowUrl)
+          .waitUntil(function () {
+            return this.getValue(urlInput).then((val) => val === 'data:text/html;,%3Ctitle%3ETabnapping%20Target%3C/title%3E')
+          })
+      })
+    })
+
     describe('page with a title', function () {
       Brave.beforeAll(this)
 
--- a/test/fixtures/tabnapping.html
+++ b/test/fixtures/tabnapping.html
@@ -1,0 +1,21 @@
+<html>
+<body>
+<script>
+
+var target_html =
+  '<title>Tabnapping Target</title>'
+
+var w = null
+function open_target() {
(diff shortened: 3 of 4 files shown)


js/components/frame.js
```diff
@@ -337,8 +337,7 @@ class Frame extends ImmutableComponent {
           this.props.frame)
         const key = this.props.frame.get('key')
         const parsedUrl = urlParse(event.url)
-        if (['http:', 'https:', 'about:', 'chrome:', 'chrome-extension:', 'file:',
-             'view-source:'].includes(parsedUrl.protocol)) {
+        if (parsedUrl.protocol) {
```
bridiver: I'm not sure if this check is really necessary. Is there a circumstance under which we would not want to update the url?

bridiver:
working on a test for this, but wanted to get the fix out asap

js/components/frame.js
bbondy: See here:
https://github.com/brave/browser-laptop/commit/27e61f2f48ddb805549f4c48ced302478bc95343

CC'ing @diracdeltas

bbondy:
looks good, thanks.

js/components/frame.js
```diff
@@ -337,8 +337,9 @@ class Frame extends ImmutableComponent {
           this.props.frame)
         const key = this.props.frame.get('key')
         const parsedUrl = urlParse(event.url)
+        // don't change url for non-display protocols like mailto
         if (['http:', 'https:', 'about:', 'chrome:', 'chrome-extension:', 'file:',
-             'view-source:'].includes(parsedUrl.protocol)) {
+             'view-source:', 'ftp:', 'data:'].includes(parsedUrl.protocol)) {
```
diracdeltas: i think this should also include `sftp:` and probably `blob:`. 

would be nice if we could get a list of all the protocols that Electron handles by default and pull this out into a constants file for reuse later

bridiver: I just threw ftp in there because I noticed it was missing, but do we want to make this a whitelist or a blacklist? At the moment the only protocol we are aware of that causes a problem is mailto, but there could potentially be other security issues related to not changing urls

diracdeltas: Isn't the `mailto:` problem applicable to every protocol which Brave doesn't handle? `skype:`, `magnet:`, etc.

bridiver: I guess it's a question of the severity of the problem. Changing the url for skype, etc... is a bad user experience, but not changing the protocol for data was a real security issue. Given that, I'd rather blacklist than whitelist or come up with a different solution that doesn't even fire the event for protocols that shouldn't be handled by it

diracdeltas: agree that blacklisting is better; `mailto:` and `javascript:` are probably the only ones that come up frequently enough to be an issue.

bridiver: wouldn't there be a javascript hack similar to the data hack?

diracdeltas: good point - let's change this to a blacklist with only `mailto:` for now

Issue #1407, opened 2016-04-19
Add a Paste and Go context menu option to the navbar.


bbondy:

bbondy:
CC @bsclifton in case this would interest you.

bsclifton:
Submitted this to get things started:
https://github.com/brave/browser-laptop/pull/1655

bsclifton:
Should be fixed with https://github.com/brave/browser-laptop/pull/1670

bbondy:
yep!

PR #1670, opened 2016-05-11
Fix for https://github.com/brave/browser-laptop/issues/1407

bsclifton:
- added "Paste and Go" option for urls
- added "Paste and Search" option for non-urls
- empty clipboard shows disabled "Paste and Search", similar to other browsers
- urlBar now has different context menu than text input
- similar to Chrome, cut/copy/paste always visible in context menu... but disabled if no action can be taken


bsclifton:
Screen caps:

bbondy:
looks great!

Issue #1253, opened 2016-04-05
contextual menus upgrade


bradleyrichter:
Brave users will expect familiar contextual menus that include the commonly used items found in most browsers, plus convienent access to new Brave features.

Please implement as shown in the spec images below: 

![image](https://cloud.githubusercontent.com/assets/13509546/14269062/e450c646-fa95-11e5-9939-bcc2cea16576.png)

![image](https://cloud.githubusercontent.com/assets/13509546/14269066/ea5951e8-fa95-11e5-822e-3a13516088c8.png)

![image](https://cloud.githubusercontent.com/assets/13509546/14269071/fc2a979c-fa95-11e5-97e4-0bc8d0cbb4bd.png)

![image](https://cloud.githubusercontent.com/assets/13509546/14269075/04967df6-fa96-11e5-9ecc-e5bab5386698.png)

TheDigitalOrchard:
You've done quite a bit of work preparing these mockups. However, I personally disagree with some of the proposed menu labels. "Open Link in New Tab", for example, You're not opening a "link", you're opening the page that the link points to... a "page". Therefore, "Open in New Tab" is better wording, in my opinion.

I do agree with many of the other changes, particularly small stuff like "Proper Title Case".

bradleyrichter:
@TheDigitalOrchard thanks for the feedback. I'm all for improving things, and correcting lingo. But in this case, is it better to follow than lead? Safari, FX and Chrome all use "Open Link in..."

![image](https://cloud.githubusercontent.com/assets/13509546/14328797/1ee67960-fbed-11e5-8031-18a51983189b.png)

Maybe "Open Linked Page in..." would be more correct but takes more space.

weems:
I think New Tab is general enough, people will know that pages are links. Short and sweet.

bsclifton:
Checking this one out... :smiley: might have some questions

diracdeltas:
@bsclifton how's it going on this? i can take it but don't want to duplicate work.

bsclifton:
@diracdeltas things are going great :smile: I hope to wrap this up tonite.  I'll be on Slack

bsclifton:
@bradleyrichter in the first spec image (NEW MENU: on link), what does "Copy" do (versus copy link)?

bradleyrichter:
I think it copies the selected rich text to clipboard as if it were not linked. The existing menu has it as well.

bsclifton:
[...]

PR #1672, opened 2016-05-11
See https://github.com/brave/browser-laptop/issues/1253

bsclifton:
Added search to context menu (for urls and for selection). Text could use updating, per the spec ("Search **engine** for '**term**'").


bsclifton:
cc: @bradleyrichter

bbondy:
@bradleyrichter I think Copy is used a lot more than search, I know the spec says differently but could we put those commands first, and then the search related ones in its own section after?

For reference:
https://github.com/brave/browser-laptop/issues/1253

Also I think we need to update the text as it looks to me currently like it's something related to OpenSearch which is an xml format.  And the ... is reserved for when there's a secondary action coming up.

bradleyrichter:
@bbondy I'm guessing that the order used by chrome and safari places the unique actions on top for access priority and "mode recognition" over "Copy" because copy the most common keyboard shortcut? 

Firefox has the order as you are suggesting but I prefer the chrome/safari approach in this case.

Regarding Open Search... - this is a place holder until we can connect the variables to the menu?

bsclifton:
@bradleyrichter yes, that would be the next step :smile: I'd have to dig in more to read the search engine preference. That change would need to include a change to the menu.properties file with a token replaceable entry like "Search {{engine}} with '{{term}}'" so that other languages would display properly (since they could have a different ordering for the words)

bradleyrichter:
@bsclifton Can you start with "Search the web for [selected variable]"?

(if it's not too long)

bsclifton:
Will do :+1:

bsclifton:
@bradleyrichter fixed, per your comment (went with Search for [selected variable])

Here's a pic

bbondy:
Great addition, thanks!

Issue #1699, opened 2016-05-12
Change Hide Menu context option to Show Menu Bar


bbondy:
Just flipping this would be a bit better since the check mark indicates on, keep the pref the same though.  
CC @bsclifton

bsclifton:
Fixed with https://github.com/brave/browser-laptop/pull/1700

PR #1700, opened 2016-05-12
Update for hide/show menu bar in context menu

bsclifton:
Fix for https://github.com/brave/browser-laptop/issues/1699

Fixes for the context menu which appears when right clicking bookmarks toolbar
- Removes the check mark
- If menu is visible, it will show "Hide Menu Bar"
- If menu is hidden, it will show "Show Menu Bar"

The preferences screen and underlying setting for this were not changed

### DIFF
--- a/app/extensions/brave/locales/en-US/menu.properties
+++ b/app/extensions/brave/locales/en-US/menu.properties
@@ -122,4 +122,4 @@ zoom=Zoom
 new=New
 learnSpelling=Learn Spelling
 ignoreSpelling=Ignore Spelling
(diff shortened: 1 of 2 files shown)


bbondy:
Could we do only "Show Menu Bar" with a checkmark when it's on? I'm looking for consistency with 
"Bookmarks Toolbar".  I guess we should call it just "Menu Bar".

bsclifton:
Updated! Should be good to go :smile:

bbondy:
Perfect, thank you!

Issue #1774, opened 2016-05-16
Add context menu option to empty items from a bookmark folder


aekeus:

NejcZdovc:
Duplication of #4455

PR #1775, opened 2016-05-16
Add context menu entry to remove all items from a bookmark folder

aekeus:
Implements #1774

### DIFF
--- a/js/state/siteUtil.js
+++ b/js/state/siteUtil.js
@@ -136,6 +136,23 @@ module.exports.removeSite = function (sites, siteDetail, tag) {
   return sites.setIn([index, 'tags'], tags.toSet().remove(tag).toList())
 }
 
+/**
+ * Removes all sites from a siteDetail
+ *
+ * @param sites The application state's Immutable sites list
+ * @param siteDetail The siteDetail to remove
+ * @return The new sites Immutable object
+ */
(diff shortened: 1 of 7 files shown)


js/state/siteUtil.js
```diff
@@ -136,6 +136,23 @@ module.exports.removeSite = function (sites, siteDetail, tag) {
...
+ * @param sites The application state's Immutable sites list
+ * @param siteDetail The siteDetail to remove
+ * @return The new sites Immutable object
+ */
+module.exports.emptySite = function (sites, siteDetail, tag) {
+  const index = module.exports.getSiteIndex(sites, siteDetail, tag)
+  if (index === -1) {
+    return sites
+  }
+  return sites.filter((site) => {
```
bbondy: We should leave the site item here and just remove the bookmark tag from it.  That way it still stays in history.

bbondy:
Looks great otherwise, just one small change. Thanks!

js/state/siteUtil.js
aekeus: Ok, right. I didn't think about that. Will do.

bbondy:
this needs rebasing too when fixed

bbondy:
since this is more than a month old I'll just close for now. Please re-issue a pull request if/when you get back to it. Thanks!

aekeus:
Thanks will do.

PR #1793, opened 2016-05-18
add error pages

### LINKED ISSUE #1611: Implement error pages
When going to page which can't be reached, I see a blank page.

Example URL: http://localhost:1234/ or http://foobar/

Chrome shows the error:

> This site can’t be reached. localhost refused to connect. ERR_CONNECTION_REFUSED

Brave shows empty page.

I'd be expecting some error pages with the details why the page doesn't load.

bridiver:

fixes #1611

@bbondy


app/index.js
```diff
@@ -183,16 +183,17 @@ app.on('ready', () => {
...
+
     // Tell the page to show an unlocked icon. Note this is sent to the main
     // window webcontents, not the webview webcontents
-    BrowserWindow.getAllWindows().map((win) => {
-      win.webContents.send(messages.CERT_ERROR, {
-        url,
-        error,
-        cert
-      })
+    webContents.hostWebContents.send(messages.CERT_ERROR, {
```
diracdeltas: this throws `cannot read property 'send' of null` on startup when the restored page has an invalid cert - need to check that hostWebContents exists

js/about/certerror.js
```diff
@@ -3,39 +3,36 @@
...
+    })
   }
 
   onSafety () {
-    aboutActions.rejectCertError({
-      previousLocation: this.state.certDetails.get('previousLocation'),
-      frameKey: this.state.certDetails.get('frameKey')
+    aboutActions.dispatchWindowAction({
+      actionType: WindowConstants.WINDOW_SET_URL,
+      location: this.state.previousLocation,
```
diracdeltas: if `this.state.previousLocation` is not defined, it throws an error later when we try to parse it. should be initialize to `about:blank` in the constructor.

js/about/certerror.js
```diff
@@ -3,39 +3,36 @@
...
-      }
-    })
   }
 
   onAccept () {
-    aboutActions.acceptCertError(this.state.certDetails.get('url'))
+    aboutActions.acceptCertError(this.state.url)
+    aboutActions.dispatchWindowAction({
+      actionType: WindowConstants.WINDOW_SET_URL,
+      location: this.state.url,
```
diracdeltas: if this.state.url is not defined, it throws an error later when we try to parse it. should be initialize to about:blank in the constructor.

js/about/certerror.js
```diff
@@ -50,8 +47,8 @@ class CertErrorPage extends React.Component {
       </svg>
       <div className='certErrorText'>
         <span data-l10n-id='certErrorText'></span>&nbsp;
-        <span className='errorUrl'>{this.state.certDetails.get('url') || ''}</span>
-        <span className='errorText'>{this.state.certDetails.get('error') || ''}</span>
+        <span className='errorUrl'>{this.state.url || ''}</span>
+        <span className='errorText'>{this.state.error || ''}</span>
```
diracdeltas: the error text color is a bit hard to read

js/about/certerror.js
diracdeltas: `previousLocation` should default to `about:blank`; otherwise the page just stays on `about:certerror` if there is no previous location and the button is clicked

js/stores/windowStore.js
```diff
@@ -175,6 +177,21 @@ const doAction = (action) => {
       // Since this value is bound we need to notify the control sync
       windowStore.emitChanges()
       return
+    case WindowConstants.WINDOW_SET_FRAME_TAB_ID:
+      windowState = windowState.mergeIn(['frames', FrameStateUtil.getFramePropsIndex(windowState.get('frames'), action.frameProps)], {
+        tabId: action.tabId
```
diracdeltas: `tabId` should be added to docs/state.md

js/about/certerror.js
bridiver: should the button even be there if there is no previous location? Not sure how that could happen except from an old saved state

js/stores/windowStore.js
```diff
@@ -175,6 +177,21 @@ const doAction = (action) => {
...
+      })
+      break
+    case WindowConstants.WINDOW_SET_FRAME_ERROR:
+      const frameKey = action.key || windowState.get('activeFrameKey')
+      windowState = windowState.mergeIn(['frames', FrameStateUtil.getFramePropsIndex(windowState.get('frames'), action.frameProps)], {
```
diracdeltas: `aboutDetails` should also be added to docs/state.md

js/about/certerror.js
bridiver: I noticed that too, but I didn't make any changes it so I'd leave that with @bradleyrichter

js/stores/windowStore.js
bridiver: actually the tabId shouldn't persisted so I'll delete it in the cleanup

js/about/certerror.js
diracdeltas: i opened a new tab, navigated to `https://lenovo.com`, and clicked the `back to safety` button. it didn't do anything, so i think `previousLocation` does not exist in that case. the `back to safety` button should always be there, UX-wise.

js/stores/windowStore.js
diracdeltas: i think `docs/state.md` is supposed to be all app state, regardless of whether it's persistent

docs/state.md
```diff
@@ -296,6 +288,7 @@ WindowStore
     maxHeight: number, // the maximum height of the popup window
     src: string, // the src for the popup window webview
   },
-  cleanedOnShutdown: boolean // whether app data was successfully cleared on shutdown
+  cleanedOnShutdown: boolean, // whether app data was successfully cleared on shutdown
+  history: array // navigation history
```
diracdeltas: this should be documented as a frame-level prop, not an app-level prop

bbondy: @bridiver  ping

PR #1810, opened 2016-05-19
Bravery panel

bbondy:


bbondy:
merging but will do review comments as followups.

less/forms.less
```diff
@@ -54,6 +54,167 @@
   }
 }
 
+.braveryPanelContainer {
```
diracdeltas: If i make my window as small as possible, the bravery panel is partially hidden because there's no vertical scrollbar

diracdeltas:
Bravery panel controls don't work in private tabs.

(also private tabs are opening `about:error` instead of `about:newtab` for me but might be unrelated)

js/commonMenu.js
```diff
@@ -314,90 +306,3 @@ module.exports.aboutBraveMenuItem = () => {
     }
   }
 }
-
-module.exports.buildBraveryMenu = function (settings, init) {
```
diracdeltas: I think we should leave this in for now, at least in the app menu. Otherwise users have no way to change their global settings. Ex: a user who has noscript on will have no way to turn it off for every site except to go to the site and manually turn off each one.

app/filtering.js
```diff
@@ -468,7 +468,33 @@ module.exports.init = () => {
   })
 }
 
-module.exports.isResourceEnabled = (resourceName) => {
+module.exports.isResourceEnabled = (resourceName, url) => {
```
diracdeltas: i think this needs to have some sort of parameter indicating whether this is being called for a private partition (see `registerPermissionHandler` for example) so that it can check `temporarySiteSettings` instead of `siteSettings` if necessary.

the general principles i was using:
* siteSettings apply to all tabs, including private
* temporarySiteSettings apply to only private tabs
* siteSettings is persistent
* temporarySiteSettings is cleared on browser close

app/filtering.js
```diff
@@ -468,7 +468,33 @@ module.exports.init = () => {
...
+  }
+
+  // Check the resource vs the cookie setting
+  if (resourceName === appConfig.resourceNames.COOKIEBLOCK &&
+      settings && settings.get('cookieControl') === 'allowAllCookies') {
+    return false
+  }
+
+  // If the particular resource we're checking is disabled then don't enable
+  if (settings && settings.get(resourceName) !== undefined) {
```
diracdeltas: safer to check `typeof settings.get(resourceName)` is boolean

js/commonMenu.js
diracdeltas: especially since there is actually no way to get to a site that is blocked by safebrowsing using per-site Bravery settings (since the blocked page redirects immediately to `about:safebrowsing`).

js/components/braveryPanel.js
```diff
@@ -0,0 +1,240 @@
...
+    this.onToggleSafeBrowsing = this.onToggleSiteSetting.bind(this, 'safeBrowsing')
+    this.onToggleNoScript = this.onToggleSiteSetting.bind(this, 'noScript')
+    this.onToggleCookieControl = this.onToggleSiteSetting.bind(this, 'cookieControl')
+    this.onToggleHTTPSE = this.onToggleSiteSetting.bind(this, 'httpsEverywhere')
+    this.onReload = this.onReload.bind(this)
+  }
+  get isExtendedValidation () {
+    return this.props.frameProps.getIn(['security', 'isExtendedValidation'])
+  }
+  get isSecure () {
```
diracdeltas: i think `isSecure` and `isExtendedValidation` are unnecessary here

js/components/braveryPanel.js
```diff
@@ -0,0 +1,240 @@
...
+    return this.props.frameProps.getIn(['security', 'isSecure'])
+  }
+  get isBlockingTrackedContent () {
+    return this.blockedByTrackingList && this.blockedByTrackingList.size > 0
+  }
+  get isMixedContent () {
```
diracdeltas: also not needed

js/components/braveryPanel.js
```diff
@@ -0,0 +1,240 @@
...
+  }
+  get redirectedResources () {
+    return this.props.frameProps.get('httpsEverywhere')
+  }
+  get isRedirectingResources () {
+    return this.redirectedResources && this.redirectedResources.size > 0
+  }
+  get origin () {
+    const parsedUrl = urlParse(this.props.frameProps.get('location'))
+    return `${parsedUrl.protocol}//${parsedUrl.host}`
```
diracdeltas: this should check `parsedUrl.slashes` to determine whether the double slashes are necessary. actually it could just use  `siteUtil.getOrigin`

app/filtering.js
bbondy: ya safe browsing still needs some work.

js/commonMenu.js
bbondy: ya safe browsing still needs some work.

bbondy: See the commit comment, the global options are coming in preferences instead later today.

js/components/braveryPanel.js
```diff
@@ -0,0 +1,240 @@
...
+                  <div className={cx({
+                    braverySelectTitle: true,
+                    disabled: !shieldsUp
+                  })} data-l10n-id='adControl' />
+                  <select value={this.getSiteSetting('adControl', 'showBraveAds')} onChange={this.onToggleAdControl} disabled={!shieldsUp}>
+                    <option data-l10n-id='showBraveAds' value='showBraveAds' />
+                    <option data-l10n-id='blockAds' value='blockAds' />
+                    <option data-l10n-id='allowAdsAndTracking' value='allowAdsAndTracking' />
+                  </select>
+                  <SwitchControl onClick={this.onToggleSafeBrowsing} rightl10nId='safeBrowsing' checkedOn={this.getSiteSetting('safeBrowsing', true)} disabled={!shieldsUp} />
```
diracdeltas: I would move the safebrowsing switch last, since it's the one that is least useful

js/components/main.js
```diff
@@ -359,6 +379,16 @@ class Main extends ImmutableComponent {
   }
 
   get enableNoScript () {
+    if (this.activeSiteSettings) {
+      if (this.activeSiteSettings.get('shieldsUp') === false) {
+        return false
+      }
+
+      if (this.activeSiteSettings.get('noScript') !== undefined) {
```
diracdeltas: better to check that `typeof` this setting is boolean

js/components/main.js
```diff
@@ -429,6 +459,28 @@ class Main extends ImmutableComponent {
...
+    if (!activeFrame) {
+      return undefined
+    }
+    return siteSettings.getSiteSettingsForURL(this.allSiteSettings, activeFrame.get('location'))
+  }
+
+  get braveShieldsDisabled () {
+    const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
+    const parsedUrl = urlParse(activeFrame && activeFrame.get('location') || '')
+    return !(parsedUrl.protocol || '').startsWith('http')
```
diracdeltas: highly unlikely, but this would return true for custom schemes that happen to start with `http`

diracdeltas:
other than not working in private tabs, the bravery panel is working great! :fireworks:

app/filtering.js
bbondy: done

js/components/braveryPanel.js
bbondy: done

js/components/braveryPanel.js
bbondy: done

js/components/braveryPanel.js
bbondy: that should be moved to urlUtil because siteUtil is only for dealing with site state. I'll use it for now though.

js/components/braveryPanel.js
bbondy: done

js/components/main.js
bbondy: done

js/components/main.js
bbondy: done

less/forms.less
bbondy: done

app/filtering.js
bbondy: I don't have a good way to do this yet, I'm going to disable it for now in private tabs.

bbondy: and post a follow up

PR #1812, opened 2016-05-19
Redesign preferences sidebar

bennyschmidt:
• Restyled `.preferencesTopBar` (`TopBar`) to be a sidebar to reflect latest mockups, and renamed it to `.prefAside`

• Moved `HelpfulHints` out of `.prefBody` (`AboutPreferences`) and into `.prefAside` (`TopBar`)

• Added new variables: `@carotRadius`, `@sideBarWidth`, `@gray`, `@mediumGray`, `@darkGray`, `@gray30`, `@black25`, `@white60`, `@insetShadow` 

Two tests related to locating `.notificationBar` are currently failing, but it seems unrelated to the changes: 
screen shot 2016-05-19 at 12 34 47 pm

### DIFF
--- a/js/about/preferences.js
+++ b/js/about/preferences.js
@@ -317,9 +317,26 @@ class TopBarButton extends ImmutableComponent {
   }
 }
 
+class HelpfulHints extends ImmutableComponent {
+  render () {
+    return <div className='helpfulHints'>
+      <span className='hintsTitleContainer'>
+        <span data-l10n-id='hintsTitle' />
+        <span className='hintsRefresh fa fa-refresh'
+          onClick={this.props.refreshHint} />
+      </span>
+      <div data-l10n-id={`hint${this.props.hintNumber}`} />
+      <div className='helpfulHintsBottom'>
+        <a data-l10n-id='sendUsFeedback' href={appConfig.contactUrl} />
+      </div>
+    </div>
+  }
+}
+
 class TopBar extends ImmutableComponent {
   render () {
-    return <div className='preferencesTopBar'>
+    return <div className='prefAside'>
+      <div className='prefAsideHeading'>Browser Settings</div>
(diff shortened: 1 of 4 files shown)


js/about/preferences.js
```diff
@@ -317,9 +317,26 @@ class TopBarButton extends ImmutableComponent {
...
+      </div>
+    </div>
+  }
+}
+
 class TopBar extends ImmutableComponent {
   render () {
-    return <div className='preferencesTopBar'>
+    return <div className='prefAside'>
+      <div className='prefAsideHeading'>Browser Settings</div>
```
bbondy: Please mark up this string like the others are.  See data-l10n-id in the same file.

bbondy:
Otherwise lgtm!

js/about/preferences.js
bbondy: You can `git grep sendUsFeedback` and follow the same.  Note that you would only add it to the en_US locale and not the others. The others are auto generated.

bennyschmidt:
ok cool, made that change in https://github.com/brave/browser-laptop/pull/1812/commits/a1a987196bb0c620c238b85fefe183258dc22221

bbondy:
lgtm, merging!

diracdeltas:
wrt the tests, do they pass if you just run:
```
npm run test -- --grep='notificationBar'
```

They still pass for me after this change but take a long time (sometimes 2+ seconds per test), so I'm not surprised if they are timing out after 5s.

bennyschmidt:
same result as last time for me. i am system prompted to store creds in my apple keychain when this test runs, idk if that has anything to do with it

screen shot 2016-05-19 at 5 26 05 pm

diracdeltas:
the system prompt is probably why it's taking so long. may pass if you tell it to allow always for Brave

PR #1839, opened 2016-05-20
restyled hamburger menu

bradleyrichter:
@bbondy

### DIFF
--- a/less/contextMenu.less
+++ b/less/contextMenu.less
@@ -10,7 +10,7 @@
   color: black;
   cursor: default;
   display: flex;
-  font-size: 11px;
+  font-size: 14px;
   &.contextMenuScrollable {
     overflow-y: scroll;
   }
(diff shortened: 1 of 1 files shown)


bsclifton:
This looks great to me! Just so everyone else can see:

## Before
burger-before

## After
burger-after

bbondy:
The 3px gap needs to be fixed, I think there's also some alignment issues on the bookmarks toolbar folders last I checked.

bradleyrichter:
Ideally this menu would overlap the tabs toolbar but I was unable to get that to work.

Chrome example:

image


(ours can come from the top right corner in order to stay within the browser window frame. (unfortunate but current limitation)

bbondy:
I'll merge this in and work on fixes.

luixxiul:
I think this change broke dragging items in the bookmark folder.

luixxiul:
I filed an issue for it, #1975.

luixxiul:
Icon size or padding should be fixed.

Issue #1724, opened 2016-05-13
Middle click on links in bookmarks toolbar should open them in a new tab


bsclifton:
Doing a middle click currently turns the curson into a scroll mode icon and does not open the link

Here's a screen cap of the bookmarks toolbar with a link highlighted that I should be able to middle click
![2016-05-13](https://cloud.githubusercontent.com/assets/4733304/15252656/ce024f92-18e3-11e6-9ca8-ccb1388b417b.png)

bbondy:
fairly sure I added support for this a while back but I don't have a mouse connected at the moment to test. Possibly a newer issue or perhaps only with some mice?

bsclifton:
Could very well be a newer issue, I don't remember if it used to work.  Can anyone else confirm? cc: @luixxiul

bbondy:
just tried in a vm with a mouse attached and it's working for me.

bsclifton:
Issue may be specific to Windows 10; will need to dig in more to understand

luixxiul:
Yes, confirmed on Windows 10.

bbondy:
ok this was probably always broken on windows 10 but when I fixed it only worked n pre win10 (possibly just for win7)

luixxiul:
FYI this seems to be specific to Windows 10 as this is not producible on Ubuntu 14.04.

bsclifton:
Checking this out now... it works fine on links, but not when they're nested in a folder

PR #1861, opened 2016-05-22
Properly handle middle mouse button click on context menus

bsclifton:
This PR fixes https://github.com/brave/browser-laptop/issues/1724

However, I'd like get other folks feedback before anyone would consider this for a merge. It does work, but I feel like a bad person. Here's the executive summary:
* If you have bookmarks toolbar enabled and you have a folder with items inside, you cannot open those sub-items by doing a middle click (CTRL + click works fine, though)
* Middle click is never handled for ContextMenu object (js/components/contextMenu.js)
 * ContextMenu control derives from ImmutableComponent which derives from React.Component
 * onClick properly handles left clicks
 * onContextMenu properly handles right clicks
 * since onClick was not being triggered for middle clicks, I created a new handler for onMouseDown

Is this a bug w/ React or is that expected? I'm not sure. Here's a link to the relevant mouse event documentation, which is sparse:
https://facebook.github.io/react/docs/events.html#mouse-events

I'm also curious if this happens on any other platform, other than Windows 10. Windows responds fine when doing a middle click on the bookmarks toolbar (BTB), just not the context menus that are created by clicking a folder contained inside the BTB.

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -454,6 +454,13 @@ class Main extends ImmutableComponent {
     while (node) {
       if (node.classList &&
           (node.classList.contains('popupWindow') || node.classList.contains('contextMenu'))) {
+
+        // Middle click (on context menu) needs to fire the click event.
+        // We need to prevent the default "Auto-Scrolling" behavior.
+        if (node.classList.contains('contextMenu') && e.button === 1) {
+          e.preventDefault()
+        }
+
         return
       }
       node = node.parentNode


bbondy:
onClick should be fired for middle clicks too, for example see the bookmark toolbar item handler that checks for a secondary action via `eventUtil.isForSecondaryAction(e)`.

Checkout the code here it might need some tweaking:
https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L447
https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L461

bsclifton:
Awesome, will check those out tonite :smile:

bsclifton:
Thanks for the heads up on where to look @bbondy, that was exactly where I needed to be :smile: 

I backed out the [old busted commit](https://github.com/brave/browser-laptop/pull/1861/commits/e51f34a171ee5732ca1288d75dc222c3e2860ed0) and replaced it with a nice, simple, clean commit

**Please note that besides fixing the click inside a Bookmarks Toolbar folder, this will also start sending the onClick() event when middle clicking any contextMenu object**. 

This does not affect the context menus which pop up on the page when right clicking, it only affects the ones inheriting from React.Component (for example, the items in the burger menu). Before, it would turn the cursor into the Auto-Scroll icon (which is also confusing). After this change, the click is propagated which is the same as a left click.

If that behavior isn't acceptable, I can look at restricting this further (only calling e.preventDefault if node has classes matching the bookmarks toolbar, for example).

Ready for review!

bbondy:
the fix seems pretty harmless, thanks!

Issue #1743, opened 2016-05-14
If bookmark is 3 characters or less do not collapse to ellipsis 


bbondy:
Sometimes 3 characters or less bookmarks collapse to ...

cezaraugusto:
@bbondy with few bookmarks it doesn't collapses (I don't have many). Is it still happening?

bbondy:
yes you need lots to repro

kqpham:
Hi there, is this issue fixed or can I attempt to tackle this issue ?

PR #1879, opened 2016-05-23
Fix #1848 (URL tooltips for bookmarks) v2

### LINKED ISSUE #1848: URL tooltip for items on bookmarks toolbar
Each item on the bookmarks toolbar could have a tooltip that shows the URL of the bookmark. Here is an example from Firefox:
image

Liunkae:
- [ ] Ran `git rebase -i` to squash commits if needed.

Adds a tooltip that shows the bookmark's URL.

Fixes #1848 (URL tooltip for items on bookmarks toolbar).

### DIFF
--- a/js/components/bookmarksToolbar.js
+++ b/js/components/bookmarksToolbar.js
@@ -152,6 +152,9 @@ class BookmarkToolbarButton extends ImmutableComponent {
       })}
       draggable
       ref={(node) => { this.bookmarkNode = node }}
+      {
+        !this.isFolder ? title={this.props.bookmark.get('location')}
+      }
       onClick={this.onClick}
       onDragStart={this.onDragStart}
       onDragEnd={this.onDragEnd}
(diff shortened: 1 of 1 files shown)


luixxiul:
This could address partially #1743

js/components/bookmarksToolbar.js
```diff
@@ -152,6 +152,9 @@ class BookmarkToolbarButton extends ImmutableComponent {
       })}
       draggable
       ref={(node) => { this.bookmarkNode = node }}
+      {
+        !this.isFolder ? title={this.props.bookmark.get('location')}
+      }
```
bbondy: This isn't valid syntax and will cause the browser to not start. 
I think you want something like:
`title={this.isFolder ? '' : this.props.bookmark.get('location')}`

it would be good to also have the bookmark title there though too like chrome.

bbondy: If you make changes to your branch you can `git add -u` and then `git commit --amend` and then `git push -f` and the pull request will be updated.

Liunkae:
See #1926.

PR #1882, opened 2016-05-23
Contribution graph

luixxiul:
https://waffle.io/brave/browser-laptop/metrics/throughput

"Throughput shows the amount of work completed in a certain time period. For teams, it's useful to determine a cadence, to help predict the amount of work the team is capable of doing."

I just thought the graph would encourage the dev team :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ Some platforms are available as pre-configured VMs. See the [readme](https://git
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


bsclifton:
Since there hasn't been any comments yet, I wanted to break the ice :smile: 

Looks good to me... although, it might be more appropriate if we had a dev portal (instead of in the README) OR if put this content in as a wiki page.  [Similar to the timeline](https://github.com/brave/browser-laptop/wiki/Brave-Development-Timeline)

I'm curious what others think?

bbondy:
I'm wondering if a wiki page would be more appropriate for this.

PR #1896, opened 2016-05-24
transition to frame search queries

bridiver:
move logic from actions to store
add frame queries

@bbondy


bbondy:
nice cleanup, thanks.

PR #1898, opened 2016-05-24
specify browserWindow icon for Ubuntu

### LINKED ISSUE #1209: No icon for Brave in Cinnamon panel (Linux Mint)
Running 0.8.2 on Linux Mint 17.3:

image

### LINKED ISSUE #1235: Brave Web Browser .Desktop File
Please make the installer add .desktop file during installation.

In the mid-time, everyone who wants a workaround, make your own Brave.desktop file in `/usr/share/applications` with the following content:

But first download icon from  [Brave icon](https://www.brave.com/favicon.ico) and make it .png with [Iconvert Icons](https://iconverticons.com/online/)

```
[Desktop Entry]
Name=Brave Web Browser
GenericName=Web Browser
Keywords=Internet;WWW;Browser;Web;Explorer
Exec=/opt/Brave-linux-x64/brave
Terminal=false
X-MultipleArgs=false
Type=Application
Icon=/usr/share/icons/Brave.png
Categories=GNOME;GTK;Network;WebBrowser;
MimeType=text/html;text/xml;application/xhtml+xml;application/xml;application/rss+xml;application/rdf+xml;image/gif;image/jpeg;image/png;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;x-scheme-handler/chrome;video/webm;application/x-xpinstall;
StartupNotify=true
```

### LINKED ISSUE #1282: browserWindow has no icon
browserWindow needs an icon for brave to show an icon on Ubuntu or you get a grey icon:
selection_001

This should be added here: https://github.com/brave/browser-laptop/blob/75c788aeb00b82957b7be58473332df5da13ac80/js/stores/appStore.js#L111 and it should be a png.

luixxiul:

Fix #1282, Fix #1235, Fix #1209

@bbondy, @majomuchi123, @ham1, Would you please test this one? Thanks.

### DIFF
--- a/js/stores/appStore.js
+++ b/js/stores/appStore.js
@@ -113,6 +113,7 @@ const createWindow = (browserOpts, defaults) => {
   const autoHideMenuBarSetting = isDarwin || getSetting(settings.AUTO_HIDE_MENU)
 
   let mainWindow = new BrowserWindow(Object.assign({
+    icon: __dirname + '/res/app.png',
     // smaller min size for "modal" windows
     minWidth,
     minHeight,


luixxiul:
Confirmed on Ubuntu 14.10. Not sure on Linux Mint.

bbondy:
great!

bbondy:
Sorry I thought you had confirmed it wasn't working before this fix and that it was working after. But I think the path is specified wrong and also it should probably only be set for linux since the docs say Windows should use an icon file.    I tested and it had no effect on Ubuntu 14.0.4 but maybe it breaks other things so it's safer to revert.

luixxiul:
Oh I see, sorry :-(

luixxiul:
Then I'm quite confused why on my machine the icon is displayed correctly :S

bbondy:
but does it show the icon without this commit too?
Note paths should also be joined using path.join from the require('path') module.

You can try to console.log(__dirname + '/res/app.png') and see if the path exists, let me know but I think __dirname is relative to the current file.  Maybe the path was correct not sure.  In any case though it should be joined via path.join and only for process.platform being linux

luixxiul:
Yeah, the path was completely wrong: ``/home/test/browser-laptop/app/extensions/brave/res/app.png``

luixxiul:
How about this?
````
const path = require('path')

icon: path.join(__dirname, '..', '..', 'res', 'app.png'),
````
But do you think this does not work on Windows right?

**Comment by bbondy** [comment]
I think that will work but I think you need to have:

```
const windowProps = {
    // smaller min size for "modal" windows		      // smaller min size for "modal" windows
    minWidth,		      minWidth,
    minHeight,		      minHeight,
...
}

if (process.platform === 'linux') {
  windowProps.icon = ...
}

 let mainWindow = new BrowserWindow(Object.assign(windowProps))
```

Please test both for packaged builds `CHANNEL=dev npm run build-package` and non pacakged builds `npm start`.  Packaged builds are in the Brave-* subfolder after building.

luixxiul:
Thanks, I tested it based on your suggestion and submitted a PR for it #1904.

This time it will be ok since I tested with another png file and that was applied as it should be. Will you please double check it for me, @bbondy?

PR #1909, opened 2016-05-24
centralize dispatching

bridiver:
@bbondy


bridiver:
all manual tests pass except for safe browsing which is also broken in master

bbondy:
Looks great, thanks!

PR #1918, opened 2016-05-25
Add font-awesome folder icon for folders on bookmarks toolbar and menus

### LINKED ISSUE #1469: Use font-awesome folder icon for folders on bookmarks toolbar and menus

luixxiul:

Fix #1469 

Image 1:

Please make sure that ``@bookmarksIconSize: 16px;`` is based on ``const iconSize = 16`` [(here)](https://github.com/luixxiul/browser-laptop/blob/b2c4775606a2384bdff7d80ddb7f58343c27f33b/js/components/bookmarksToolbar.js#L126).

I like the white icon better than ``fa-folder``, which is a black icon.
http://fontawesome.io/icon/folder/

### DIFF
--- a/js/components/contextMenu.js
+++ b/js/components/contextMenu.js
@@ -16,6 +16,9 @@ class ContextMenuItem extends ImmutableComponent {
   get hasSubmenu () {
     return this.submenu && this.submenu.size > 0
   }
+  get isBookmarksToolbarSubmenu () {
+    return this.hasSubmenu && this.props.contextMenuItem.get('bookmark')
+  }
   onClick (clickAction, shouldHide, e) {
     e.stopPropagation()
     if (clickAction) {
@@ -155,6 +158,11 @@ class ContextMenuItem extends ImmutableComponent {
             [faIcon]: !!faIcon
           })}
             style={iconStyle}></span>
+          : null
+      }
+      {
+        this.isBookmarksToolbarSubmenu
+          ? <span className='submenuIndicator contextMenuIcon bookmarkIcon fa fa-folder-o' style={iconStyle} />
           : null
       }
       <span className='contextMenuItemText'
--- a/less/contextMenu.less
+++ b/less/contextMenu.less
@@ -27,6 +27,11 @@
   -webkit-user-select: none;
   &.reverseExpand {
     flex-direction: row-reverse;
+    .fa {
+      font-size: 0;
+      width: 0;
+      margin: 0;
+    }
   }
 
   .contextMenuSingle {
@@ -133,5 +138,9 @@
       float: right;
       margin: auto 0 auto 5px;
     }
+  }
+
+  .fa {
+    font-size: @bookmarksIconSize;
   }
 }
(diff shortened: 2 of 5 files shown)


luixxiul:
As the submenu is global, some work is needed to fix this:

clipboard01

js/components/contextMenu.js
```diff
@@ -157,6 +157,11 @@ class ContextMenuItem extends ImmutableComponent {
             style={iconStyle}></span>
           : null
       }
+      {
+        this.hasSubmenu
+        ? <span className='submenuIndicator contextMenuIcon bookmarkIcon fa fa-folder-o' style={iconStyle} />
```
bbondy: The word bookmark shouldn't be in this file since it's a general context menu file.

less/contextMenu.less
```diff
@@ -27,6 +27,11 @@
   -webkit-user-select: none;
   &.reverseExpand {
     flex-direction: row-reverse;
+    .bookmarkIcon {
```
bbondy: ditto

less/contextMenu.less
```diff
@@ -134,4 +139,8 @@
       margin: auto 0 auto 5px;
     }
   }
+  
+  .bookmarkIcon {
```
bbondy: ditto

js/components/contextMenu.js
bbondy: I suggest to modify `js/contextMenu.js` and look in `bookmarkItemsInit`. That's where icons are set. If you find it is a folder you can add a new property like faIcon and then if it is specified inside `js/compnents/contextMenu.js` you can use it.

less/contextMenu.less
bbondy: You can just check for &.fa I think

js/components/contextMenu.js
luixxiul: OK I will

luixxiul: Is it a bad idea to do like this?

````
      {
        this.hasSubmenu && (icon || faIcon)
        ? <span className='fa fa-folder-o' />
        : null
      }
````

luixxiul:
I'm trying to add bookmark icons to submenu, but can't :/

bsclifton:
You were really close! Here's what I did that made it work:
https://github.com/bsclifton/browser-laptop/commit/98e06cfcfe6d08f43d6bcfab44fb07f5204517f6

Feel free to either make the commit yourself or add my remote and then cherry pick this hash (I can help if you don't know how to do that)

:smile:

luixxiul:
Thanks! It was a great help :-)

luixxiul:
OK I think it's ready.

bsclifton:
lgtm! :+1:

luixxiul:
I think ``.bookmarkToolbar`` should be set to 28px, in order to avoid this:

clipboard01

Should I send another PR to fix this?

luixxiul:
Hm, restart fixed it.

bsclifton:
Interesting.. what platform are you on? Here's how it looks for me (Windows).



Do the fonts get offset differently because you have unicode characters in the bookmarks toolbar?

luixxiul:
It seems that the difference of padding settings causes it. If you set ``padding: 0px 10px`` in ``.bookmarksToolbar`` they are aligned well.

Either changing the folder icon size or hiding the folder in the hide favicon mode would also work.

PR #1919, opened 2016-05-25
Filtering localization

aekeus:
Initialized localization strings in session.setPermissionRequestHandler


app/index.js
```diff
@@ -366,8 +368,8 @@ app.on('ready', () => {
       initiateSessionStateSave()
     })
 
-    Extensions.init()
     Filtering.init()
+    Extensions.init()
```
bbondy: Does this need to be re-ordered?

app/filtering.js
```diff
@@ -235,31 +236,31 @@ function registerForHeadersReceived (session) {
...
-    fullscreen: {
-      action: 'use fullscreen mode'
-    },
-    openExternal: {
-      action: 'open an external application'
-    }
-  }
+  let permissions = {}
   session.setPermissionRequestHandler((webContents, permission, cb) => {
+    permissions = {
```
bbondy: prefer to only set this once if not already set, because it will be a callback called a lot.

bbondy:
If you remove the uneeded reordering and then change to only set the strings once in the permission callback then please go ahead and merge. Thanks!

app/index.js
aekeus: Nope, mistake. Will fix.

app/filtering.js
aekeus: Ok, will cache after first call.

PR #1947, opened 2016-05-27
status bar position fix

### LINKED ISSUE #1712: change "status bar" so that it can not be covered by downloads bar

luixxiul:

Close #1712

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -380,7 +380,8 @@ class Frame extends ImmutableComponent {
         case messages.LINK_HOVERED:
           method = (href, position) => {
             position = position || {}
-            let nearBottom = position.y > (window.innerHeight - 150) // todo: magic number
+            const downloadsBarHeight = 50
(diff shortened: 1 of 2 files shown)


js/components/frame.js
```diff
@@ -380,7 +380,8 @@ class Frame extends ImmutableComponent {
         case messages.LINK_HOVERED:
           method = (href, position) => {
             position = position || {}
-            let nearBottom = position.y > (window.innerHeight - 150) // todo: magic number
```
bbondy: I don't think the changes in frame.js is necessary is it?

luixxiul: Without this the status bar appears always left, even though the downloads bar is displayed. The value of the const is based on [this](https://github.com/brave/browser-laptop/blob/master/less/downloadBar.less#L21).

bbondy:
Thanks for taking this, great work :)

PR #1960, opened 2016-05-28
Fix typo in permRequestHeader

Liunkae:


bbondy:
sorry I didn't notice this until after correcting from a comment on transifex!

PR #1976, opened 2016-05-30
Fix changelog links

Liunkae:


luixxiul:
Also the links on the release note should be fixed in the next release cc: @bbondy

bbondy:
Thanks, I noticed this too and was going to fix on the next change, but this is better :)

PR #2021, opened 2016-06-01
enable escape for exit of webview fullscreen

bridiver:
@bbondy

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -202,6 +202,17 @@ class Frame extends ImmutableComponent {
       if (this.props.isActive && !prevProps.isActive) {
         this.webview.focus()
       }
+
+      // make sure the webview content updates to
+      // match the fullscreen state of the frame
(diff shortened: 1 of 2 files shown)


js/components/frame.js
```diff
@@ -521,7 +532,6 @@ class Frame extends ImmutableComponent {
     })
     this.webview.addEventListener('enter-html-full-screen', () => {
       windowActions.setFullScreen(this.props.frame, true, true)
-      setTimeout(windowActions.setFullScreen.bind(this, this.props.frame, undefined, false), 5000)
```
bridiver: didn't realize what this was for, but it looks like it turns off the fullscreen warning message so I'll put it back in

bbondy:
I think the logic can be tweaked a bit in electron instead but going with this for now.
Was going to debug but my xcode needs to be reinstaleld or something because it keeps crashing :(

PR #2031, opened 2016-06-02
Fixed Spaces in a URL perform a google search

### LINKED ISSUE #2002: Spaces in a URL perform a google search
Yes

When a URL contains a space in the address bar and you hit the enter key, it opens a search instead of navigating to the URL.

**Expected behavior:**
You should be taken to the URL without performing a search, like other browsers.
- Platform (Win7, 8, 10? OS X? Linux distro?):
  Win7
- Brave Version:
  0.10.0
- Steps to reproduce:
  1. Type a URL in the address bar that contains a space
  2. Hit the enter key

If you need an example, here's a few URLs you can copy and paste in your address bars : 
`https://twitter.com/search?q=kitty cats`
`http://help.forumotion.com/st/ange tuteur`

You can replace the space with %20 or + and it'll navigate to the URL, but people who aren't familiar with that will have this problem. It's probably not common for someone to type the keywords in the URL, but I figured I'd report it because lazy me does that. :stuck_out_tongue_winking_eye:  
- Screenshot if needed:
  N/A
- Any related issues:
  [This](https://github.com/brave/browser-laptop/issues/519) was sorta related, but it concerned trailing and preceding white space.

dlion:

Close #2002

### DIFF
--- a/js/lib/appUrlUtil.js
+++ b/js/lib/appUrlUtil.js
@@ -6,6 +6,7 @@ const Immutable = require('immutable')
 const path = require('path')
 const UrlUtil = require('./urlutil')
 const config = require('../constants/config')
+const urlParse = require('url').parse
 
 /**
  * Determines the path of a relative URL from the hosted app
@@ -112,6 +113,9 @@ module.exports.isTargetAboutUrl = function (input) {
  */
 module.exports.isUrl = function (input) {
   input = input.trim()
+  if (urlParse(input).protocol) {
+    input = encodeURI(input)
+  }
   return UrlUtil.isURL(input) && !input.includes(' ')
 }
 
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -9,6 +9,7 @@ const rscheme = /^(?:[a-z\u00a1-\uffff0-9-+]+)(?::(\/\/)?)/i
 const defaultScheme = 'http://'
 const fileScheme = 'file://'
 const os = require('os')
+const urlParse = require('url').parse
 
 /**
  * A simple class for parsing and dealing with URLs.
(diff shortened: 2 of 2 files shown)


bbondy:
In general we want spaces in the url bar to cause a search. But perhaps not when a protocol is specified or similar.  Please check chrome and try for similar functionality. The fix should be in `js/lib/urlutil.js`

bbondy:
see `isNotURL` in that file.

dlion:
Fixed!

Before:
screenshot - 06022016 - 05 43 32 pm
After:
screenshot - 06022016 - 05 43 56 pm
Before:
screenshot - 06022016 - 05 44 18 pm
After:
screenshot - 06022016 - 05 44 35 pm

Checking if the url has the protocol and if is it encode it.
What do you think about it ?

bbondy:
Looking better thx, but pls change in the same file I mentioned above instead so all URL handling is handled the same in the app.

dlion:
@bbondy I have to let it here because the condition `UrlUtil.isURL(input) && !input.includes(' ')` returns false if the input includes spaces so I need to encode the url first.
Obviously I add the condition on the `isNotURL` function too.

bbondy:
oh ok I see, that's ok then.

bbondy:
Please put back to the last fix you had because this will cause various problems. Like if you paste a URL with %20 in it it'll re-encode the %.

dlion:
@bbondy 
Can you give me an example ?
I use `http://help.forumotion.com/st/ange%20tuteur` and it works fine.
Same thing for search: `Hello%20My%20Friend` string works very well.

bbondy:
I'm not sure if there's a bug or not but it seems wrong because you'd be passing in a URL which is not the one the user is visiting to `return UrlUtil.isURL(input) && !input.includes(' ')`

dlion:
But it's a problem, how can I recognize if %20 is a encoded space or just my url ?

bbondy:
I'm pretty sure the best fix is to do the `input.includes(' ')` check inside `isNotURL` and check for protocol before that as you do.

bbondy:
I think the only reason `input.includes(' ')` was added there was because we used to use a third party module for urlUtil `isNotURL` and we forked it.  Should have been within there to begin with.

bbondy:
going to close this for now but please do another PR if you implement as per my last message. Thanks!

PR #2077, opened 2016-06-04
Make .bookmarkFolderList sticky

### LINKED ISSUE #2076: Make .bookmarkFolderList on about:bookmarks sticky
I'm pushing a PR with https://github.com/yahoo/react-stickynode

luixxiul:

close #2076 
I'm not quite sure about the thing related to licencing. If there is a problem with using the library, I'll close this request.

### DIFF
--- a/js/about/bookmarks.js
+++ b/js/about/bookmarks.js
@@ -5,6 +5,7 @@
 // Note that these are webpack requires, not CommonJS node requiring requires
 const React = require('react')
 const Immutable = require('immutable')
+const Sticky = require('react-stickynode')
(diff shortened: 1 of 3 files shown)


bbondy:
The license looks ok.
I prefer not to add deps if not needed but this fixes  the UX problem nicely so lets just go with it.

It can also be accomplished by making `bookmarkFolderList` and `bookmarkList` `overflow-y: auto` and the parent container with a `height` specified.

If we don't use it in other places then we can remove the dep later, if we do we can just keep it.

luixxiul:
This can be removed when ``position:sticky`` will be implemented anyway.

Issue #2096, opened 2016-06-06
Avoid using display:table with pixels on about:preferences


luixxiul:
**Describe the issue you encountered:** From a viewpoint of l10n, it is not a good idea to combine ``display: table`` with ``width: px`` because it does not solve a bug on about:preferences like this:

![](https://cloud.githubusercontent.com/assets/3362943/15811204/187d2dc0-2be1-11e6-8ce7-3ae3bc879f60.jpg)

It was fixed with #2042 once, but just as a workaround, not in principle.

**Expected behavior:** I think the better solution is to use ``display: flex`` with percent. Great part of ``about:preferences`` would need rewriting in this case.

- Brave Version: Master Branch

Edit: There will be an overhaul of about pages.

luixxiul:
I think this has been solved in 0.11.1 about:preferences. Closing.

PR #2098, opened 2016-06-06
Removed height of input and select in settingItem

luixxiul:
Related with #2096

Fix this
clipboard01
Into
clipboard02


bbondy:
thanks!

PR #2117, opened 2016-06-07
send settings to all iframes

### LINKED ISSUE #2041: ipcRenderer events are not received inside iframes
messages sent by `this.webview.send` are not received by `brave-default.js` running inside an iframe. as a result, ads are not replaced in https://zyan.scripts.mit.edu/brave/slashdot.html and fingerprinting protection will not always work inside iframes.

i would expect the top-level instance of `brave-default.js` to forward the messages onto subframes.

blocks #313 somewhat, but can be worked around

cc @bridiver

bridiver:

requires https://github.com/brave/electron/commit/42ffb5c19830ea73f29eb9a32d78a9a527d81f1e
fixes #2041 

@bbondy @diracdeltas


bridiver:
this currently crashes with private tabs

bridiver:
fixed with https://github.com/brave/electron/commit/0b58df348733014baa3b955dcdddca1f5a1ad732

app/extensions.js
```diff
@@ -15,23 +17,51 @@ let generateBraveManifest = () => {
       {
         run_at: 'document_start',
         all_frames: true,
-        matches: ['<all_urls>'],
+        matches: ['http://*/*', 'https://*/*', 'file://*'],
```
diracdeltas: also `data:` and probably `blob:`?

diracdeltas: pretty sure they are needed at least for `data:` URLs because those can be HTML documents

bridiver: ok, should be pretty easy change

diracdeltas:
this looks much cleaner, thanks.

bridiver:
I think there is still a lot we can do to clean things up using the new background page functionality, but I didn't want to do too much in one PR

bbondy:
Is this ready to go? I think you can merge after fixing rebase.

bridiver:
it requires an updated electron build, but I do want to get this merged asap

diracdeltas:
i'm fine with that. lmk when upstream electron build is available over npm.

bbondy:
We're at 1.2.2 now, merging.

PR #2158, opened 2016-06-10
add event store for ledger client

bridiver:
cc: @bbondy @mrose17 

This provides initial support for the event data required for the ledger client. It does not currently include persistence, but the ledger alpha client can add an event listener to collect the information it needs.


bbondy:
cc @aekeus in case it conflicts with stuff he's doing.

Issue #1576, opened 2016-05-05
'Block all HTTP' mode


diracdeltas:
HTTPS Everywhere has a global 'Block all HTTP' mode where no HTTP requests are allowed. It was pretty well-received and easy to add so we should add it

luixxiul:
Let's set up "Block all trackings/images" mode too. cf. https://github.com/brave/browser-laptop/commit/4c30cd08af9dda75a0e3d6ef634abbbdc2a39c74

luixxiul:
Definitely I'd like to see this implemented near in the future. Sites like http://weather.goo.ne.jp/ and http://map.goo.ne.jp/ accept ONLY plain http requests, and tons of beacons track your navigations there :astonished:

diracdeltas:
Setting a milestone in light of recent attention on https://www.theguardian.com/technology/2017/mar/28/privacy-protection-sell-web-browsing-history-data in the US

Liunkae:
Related: #7559 (make http links more obvious to highlight their insecurity)

bsclifton:
Tentatively moving this to 0.14.3 but please feel free to pull back if you have a fix ready 😄

diracdeltas:
@evq working on this

bsclifton:
Closing in favor of https://github.com/brave/brave-browser/issues/779

PR #2160, opened 2016-06-11
Gray fa-unlock-alt in http mode

luixxiul:

Related #1576

### DIFF
--- a/less/navigationBar.less
+++ b/less/navigationBar.less
@@ -340,20 +340,6 @@
     opacity: 1;
   }
 
-  .urlbarIcon.fa-unlock-alt {
-    margin-top: 4px;
-    font-size: 16px;
-    min-height: 10px;
-    min-width: 16px;
-  }
-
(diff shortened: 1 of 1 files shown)


bsclifton:
I love the idea of showing a different icon for HTTP... however, I personally think the icon is too close to the lock icon (some browsers show standard SSL as gray lock and EV SSL as green).

I'm curious what others would think about changing this to yellow or orange and then using a more obvious icon to represent HTTP. Something like:
fa-exclamation-triangle
fa-times-circle

or for comedy purposes:
frown-o

luixxiul:
@bsclifton Thanks for the feedback. I thought about it too, but yellow or ``@BraveOrange`` should be reserved for worse scenarios until #1576 will be merged IMO.

diracdeltas:
I like how this makes HTTP and HTTPS look more different from each other, so +1. Agree that we should consider making HTTP look scarier as @bsclifton suggested.

Issue #2157, opened 2016-06-10
script blocking and canvas fp blocking no longer working


diracdeltas:
possibly broken by the frame-settings merge

diracdeltas:
actually it looks like content scripts in general aren't running

diracdeltas:
i think i failed to update electron correctly; closing for now

PR #2166, opened 2016-06-12
re-enable noScript after https://github.com/brave/browser-laptop/comm…

bridiver:

…it/76e956b2a6c09bbd14d6d4b244f998e2b39205ab

partial fix for #2157
@diracdeltas @bbondy

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -234,7 +234,7 @@ function registerForHeadersReceived (session) {
  * @param {string} partition name of the partition
  */
 function registerPermissionHandler (session, partition) {
-  const isPrivate = !partition.startsWith('persist:') && partition !== ''
+  const isPrivate = !partition.startsWith('persist:') && partition !== '' && partition !== 'main-1'
   // Keep track of per-site permissions granted for this session.
   let permissions = null
   session.setPermissionRequestHandler((webContents, permission, cb) => {
@@ -403,10 +403,7 @@ function registerSession (partition, fn) {
 function initForPartition (partition) {
   let fns = [registerForBeforeRequest, registerForBeforeRedirect,
     registerForBeforeSendHeaders, registerPermissionHandler]
-  if (partition !== '') {
-    // Don't block scripts in the background page
-    fns.push(registerForHeadersReceived)
-  }
+  fns.push(registerForHeadersReceived)
 
   fns.forEach(registerSession.bind(this, partition))
 }
@@ -428,7 +425,7 @@ module.exports.init = () => {
   setTimeout(() => {
     registerForDownloadListener(session.defaultSession)
   }, 1000)
-  ;[''].forEach((partition) => {
+  ;['', 'main-1'].forEach((partition) => {
     initForPartition(partition)
   })
   let initializedPartitions = {}
--- a/app/noScript.js
+++ b/app/noScript.js
@@ -10,6 +10,7 @@ const siteSettings = require('../js/state/siteSettings')
 const siteUtil = require('../js/state/siteUtil')
 const ipcMain = require('electron').ipcMain
 const messages = require('../js/constants/messages')
+const { getIndexHTML } = require('../js/lib/appUrlUtil')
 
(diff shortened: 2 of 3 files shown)


app/filtering.js
```diff
@@ -428,7 +425,7 @@ module.exports.init = () => {
   setTimeout(() => {
     registerForDownloadListener(session.defaultSession)
   }, 1000)
-  ;[''].forEach((partition) => {
+  ;['', 'main-1'].forEach((partition) => {
```
bridiver: why do we initialize the partition for the main window?

bbondy: @diracdeltas

diracdeltas: mainly because HTTPS Everywhere should apply to favicon loads, I think.

js/stores/appStore.js
```diff
@@ -239,7 +239,7 @@ function windowDefaults () {
     windowOffset: 20,
     webPreferences: {
       sharedWorker: true,
-      partition: ''
+      partition: 'main-1'
```
diracdeltas: why was it necessary to re-add `main-1`?

bridiver: some confusion with electron. Normally partitions that are not prefixed with `persist:` are in-memory, but apparently '' is a special case that returns a persistent partition. It's particularly annoying because there is no way to get a matching '' in-memory partition which is needed for `spanning` extensions. At some point we will either have to change the default partition used by normal tabs or add a method that allows creating an in-memory '' partition.

bridiver: to make everything work correctly with private tabs and extensions, we should have to identically named partitions with one in-memory and one persistent. The logic in electron takes care of matching them together. Private tabs should use the in-memory '' partition and the browser process should load favicons through the partition that matches the one used by the tab. Not a big deal right now because the extensions we are using don't really need the spanning support, but eventually it will be an issue

diracdeltas: ok, thanks for the explanation!

bridiver: going to bring it up in slack so everyone is aware and understands the details

app/noScript.js
```diff
@@ -77,7 +81,7 @@ function onHeadersReceived (details) {
   }
 
   let csp = details.responseHeaders[cspHeaderName] || []
-  csp.push("script-src 'none'")
+  csp.push('script-src chrome-extension://*')
```
diracdeltas: decided in slack this is probably not needed

js/stores/appStore.js
diracdeltas: > Private tabs should use the in-memory '' partition and the browser process should load favicons through the partition that matches the one used by the tab.

The same logic would apply to search autocomplete queries. They aren't strictly associated with a tab, but I guess they should go through the partition of the active tab when the query is entered.

bridiver: are the autocomplete queries cacheable? That's really the only issue here because otherwise the in-memory main-1 partition is fine. I wonder if private partitions can use the cache from their corresponding persistent partition? Rather than being a data leak, that would actually improve privacy because fewer remote requests will be made.

diracdeltas: i think they are cacheable in theory, but they are small and relatively infrequent so i doubt it's much of a performance issue if not

app/noScript.js
diracdeltas: it also breaks noscript UI in the bravery panel (hard coded to count for console errors involving script-src none). removing for now

bridiver: I have some changes for that in a PR I'm working on. I'll verify that everything works without chrome-extension:// and if not, I'll check the noscript ui.

bridiver: actually I'll remove it from this one either way so we can merge

diracdeltas:
for some reason, the number of scripts blocked in the Bravery panel doesn't show up until 'Block scripts' is toggled off then on again. i'll look into it

diracdeltas:
merging for now, @bridiver to fix settings/siteSettings issue separately

PR #2170, opened 2016-06-12
add extension actions and initial support for location -> publisher m…

bridiver:
cc @bbondy @aekeus @mrose17 

the xpath/dom selector info could be stored in appConfig or appState and passed to publisherIdentification. This just provides the basic functionality and some simple sample code.

### DIFF
--- a/js/stores/eventStore.js
+++ b/js/stores/eventStore.js
@@ -11,7 +11,8 @@ const { isSourceAboutUrl } = require('../lib/appUrlUtil')
 
 let eventState = Immutable.fromJS({
   page_load: [],
-  page_view: []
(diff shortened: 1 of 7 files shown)


js/stores/eventStore.js
```diff
@@ -11,7 +11,8 @@ const { isSourceAboutUrl } = require('../lib/appUrlUtil')
 
 let eventState = Immutable.fromJS({
   page_load: [],
-  page_view: []
+  page_view: [],
+  publisher_map: {}
```
bridiver: not sure if this is the best way to store it because we may want to consider things like embedded youtube videos

PR #2179, opened 2016-06-14
sync up private and regular tab partition names for extensions

bridiver:
@bbondy

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -121,9 +121,11 @@ class Frame extends ImmutableComponent {
 
     let partition
     if (this.props.frame.get('isPrivate')) {
-      partition = 'private-1'
+      partition = 'default'
     } else if (this.props.frame.get('partitionNumber')) {
       partition = `persist:partition-${this.props.frame.get('partitionNumber')}`
+    } else {
+      partition = 'persist:default'
     }
     if (partition) {
       ipc.send(messages.INITIALIZE_PARTITION, partition)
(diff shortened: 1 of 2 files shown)


bridiver:
this is not ready to merge, there is an issue with it

js/components/frame.js
```diff
@@ -121,9 +121,11 @@ class Frame extends ImmutableComponent {
 
     let partition
     if (this.props.frame.get('isPrivate')) {
-      partition = 'private-1'
+      partition = 'default'
     } else if (this.props.frame.get('partitionNumber')) {
       partition = `persist:partition-${this.props.frame.get('partitionNumber')}`
+    } else {
+      partition = 'persist:default'
```
bbondy: Does this cause a different partition name for the default partition? In particular people that have a gmail account setup for example will still have that setup right?

bridiver: is there something other than cookies that gets stored locally for the partition?

bbondy: I think local storage, cache, cookies, etc.

bridiver: that will definitely be lost as discussed in slack. Everyone seemed ok with it, but if not I can go another route. There is also currently an issue with ad insertion that this fixes for some reason.

bbondy: I don't think it's ok to make people re-setup all user accounts.

bbondy: @bridiver said he verified and it's considered the same partition. Merging.

Issue #2135, opened 2016-06-08
Remember full screen mode in session storage


bbondy:
A user from support was asking why we didn't remember the full screen setting as he always wanted it to start on his page full screen.

cezaraugusto:
Would like to try implementing it, can I?

bbondy:
yes it seems similar to this one though, so take note there for review comments:
https://github.com/brave/browser-laptop/pull/2029

bbondy:
cc @bsclifton @bridiver in case they have any comments.

bridiver:
I think we can roll all of this into a general window size restore. Maximized and Fullscreen could be represented with -1 and -2 size or something similar

cezaraugusto:
@bridiver Is @bsclifton PR on hold due to this? If so, think he can handle it better. Anyway waiting for his thoughts :)

bridiver:
the PR is on hold because it will maximize every window on restore if any single window is maximized

bridiver:
but if we're going to remember window size, we might as well handle all of them together

bsclifton:
Will definitely adjust the code per the comments :smile: 

I think this issue is a dupe of one of the following:
https://github.com/brave/browser-laptop/issues/928
https://github.com/brave/browser-laptop/issues/146

bsclifton:
@cezaraugusto this issue is ready if you're up for it :smile:  The PR I had open has been merged:
https://github.com/brave/browser-laptop/pull/2029

cezaraugusto:
@bsclifton, thanks but I'm going to leave it for now, lots of work here. Feel free to keep going :+1:,  Anyway I'm following closely this issue. :eyes:

bsclifton:
Awesome, I'll go ahead and grab this

PR #2181, opened 2016-06-14
Fullscreen status for all windows is now saved.

bsclifton:

Fullscreen status for all windows is now saved.
Fixes https://github.com/brave/browser-laptop/issues/2135

### DIFF
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -755,7 +755,7 @@ const windowActions = {
    * Sets the fullscreen state of the window
    * @param {boolean} isFullScreen - true if window is fullscreen
    */
-  setFullscreenState: function (isFullScreen) {
+  setWindowFullScreen: function (isFullScreen) {
     dispatch({
       actionType: WindowConstants.WINDOW_SET_FULLSCREEN_STATE,
       isFullScreen
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -752,6 +752,17 @@ const windowActions = {
   },
 
   /**
+   * Sets the fullscreen state of the window
+   * @param {boolean} isFullScreen - true if window is fullscreen
+   */
+  setFullscreenState: function (isFullScreen) {
+    dispatch({
+      actionType: WindowConstants.WINDOW_SET_FULLSCREEN_STATE,
+      isFullScreen
+    })
+  },
+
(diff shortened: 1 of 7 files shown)


bsclifton:
Ready for review :smile: I did some manual testing with several windows and it worked great; would be awesome to try it on Ubuntu and Mac

luixxiul:
I tried it on Ubuntu 14.04 with 2 windows and it worked fine :-)

luixxiul:
On Ubuntu 14.04 if one window is fullscreen and another is not, restarting does not restore the fullscreen status of the first one.

js/actions/windowActions.js
```diff
@@ -738,6 +738,17 @@ const windowActions = {
   },
 
   /**
+   * Sets the fullscreen state of the window
+   * @param {boolean} isFullScreen - true if window is fullscreen
+   */
+  setFullscreenState: function (isFullScreen) {
```
bridiver: I think we need to do something to better differentiate webview and window fullscreen. The webview fullscreen state is triggered by the electron `enter-html-full-screen` event and windowActions.setFullScreen updates the `isFullScreen` property of the frame. Maybe we could have `setWebviewFullScreen' and `setWindowFullScreen` to make it more clear?

bridiver: actually `setFrameFullScreen` is probably more consistent with other webview actions

bsclifton:
Updated per your feedback, @bridiver :smile: 

I also fixed a nasty unintentional behavior I introduced w/ capturing the window position. It was creating hundreds of window actions per second if you're moving the window. Fix was to add a sliding one second timer to the event. Much better (and noticeably faster)

bbondy:
Thanks, merging!

js/components/main.js
```diff
@@ -408,14 +408,31 @@ class Main extends ImmutableComponent {
...
     })
+
     win.on('unmaximize', function () {
       windowActions.setMaximizeState(false)
     })
+
+    let moveTimeout = null
     win.on('move', function (event) {
-      windowActions.savePosition(event.sender.getPosition())
+      if (moveTimeout) {
```
bridiver: fyi - there is a `debounce` method that we also use in other places

bridiver:
++

PR #2155, opened 2016-06-10
Make about error pages scrollable

luixxiul:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Ran `git rebase -i` to squash commits if needed.

Closes #2152

This commit makes the about error pages, about:certerror, about:error and about:safebrowsing, scrollable.

Overflow of the retry button will be fixed by merging #2149

PR #2205, opened 2016-06-16
Fix "retry the URL" button width

luixxiul:

addresses #2155 

This fixes the width of the retry button into this, which is the current one.

Also would anyone please tell me the case where ``showBackButton()`` would be fired, which is ``this.state.previousLocation !== this.state.url``? I need an example to style the back button, since currently the same className is attached to the two buttons.


bbondy:
thanks!

PR #2244, opened 2016-06-20
Fix search selection menu item length

### LINKED ISSUE #2240: Search selection doesn’t truncate early enough
Yes

The “Search selection” contextual menu item only works well for shorter words.

**Expected behavior:**
The contextual “Search selection” menu is a reasonable length.

**Platform:**
All

**Brave Version:**
All

**Steps to reproduce:**

1. Select a long string of text ([anything in German is an especially good choice](https://theweek.com/articles/463500/8-favorite-ridiculously-long-german-words)).
2. Right click
3. Notice the contextual menu is long, very long, and that maybe the OS has even truncated the menu item for us (shown here on OS X): brave-contextual-menu-search-selection

neeklamy:
- [x] Submitted a ticket for my issue if one did not already exist.
- [x] Used GitHub auto-closing keywords in the commit message.

**Existing approach:** truncate text in menu item to three “words”, with any additional words replaced by an ellipsis.

**Improved approach:** When invoking the contextual menu on selected text, truncate the “Search for” menu item for text longer than 25 characters at the 20th character and add an ellipse.

**For example:**
<tt>[25] immunoelectrophoretically &emsp; immunoelectrophoretically</tt>
<tt>[26] zusammengehörigkeitsgefühl &emsp;zusammengehörigkeits...</tt>

This also has the added benefit of not limiting selections with shorter “words”, for example, if we were counting “words”:
<tt>[25] a b c d e f g h i j k l m &emsp; a b c...</tt>

Counting characters:
<tt>[25] a b c d e f g h i j k l m &emsp; a b c d e f g h i j ...</tt>

Closes #2240

### DIFF
--- a/js/contextMenus.js
+++ b/js/contextMenus.js
@@ -649,7 +649,7 @@ const copyEmailAddressMenuItem = (location) => {
 }
 
 const searchSelectionMenuItem = (location) => {
-  var searchText = textUtils.ellipse(location, 3)
+  var searchText = textUtils.ellipse(location)
   return {
     label: locale.translation('openSearch').replace(/{{\s*selectedVariable\s*}}/, searchText),
     click: (item, focusedWindow) => {
--- a/js/lib/text.js
+++ b/js/lib/text.js
@@ -2,12 +2,21 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// Add ... characters and limit length of a string to a specific size
-module.exports.ellipse = (text, length) => {
-  var tokens = text.split(/\s+/)
-  if (tokens.length > length) {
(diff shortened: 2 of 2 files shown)


bbondy:
I think there should be an attempt to not break up words, but in worst case if we hit a limit of number of characters then just split it and add the ...

In almost all cases it should break on words, except for a very large string with no spaces.   See Chrome as an example.

neeklamy:
Often times we’re not searching on words, but on strings of text that can be of very variable lengths, programming related for example.

Google Chrome’s approach isn’t that great, it’s better, but it’s truncating the text later than it needs to:

Google Chrome’s contextual menu, search selected text

That’s serving no purpose at all, we know what text is selected, the menu item is a simple aide-memoire.

bbondy:
ok that sounds reasonable, thanks for the PR!

PR #2249, opened 2016-06-21
User prefs

bridiver:

requires https://github.com/brave/electron/pull/22

cc @bbondy @diracdeltas


js/state/contentSettings.js
```diff
@@ -0,0 +1,112 @@
...
+  ]
+}
+
+const getContentSettingsFromSiteSettings = (appState) => {
+  let braveryDefaults = siteSettings.braveryDefaults(appState, appConfig)
+
+  let contentSettings = {
+    cookies: braveryDefaults.cookieControl === 'block3rdPartyCookie' ? getBlock3rdPartyStorage() : getAllowAllCookies(),
+    javascript: [],
+    canvasFingerprinting: [{
```
diracdeltas: If you change this setting globally in about:preferences, it doesn't take effect until browser restart.

bridiver: works in https://github.com/brave/browser-laptop/pull/2256
we should change canvas fingerprinting to a resource like the others so they all work consistently

diracdeltas: the site setting works immediately though

bridiver: all the site settings work the same way, but the other resources are updated globally with APP_SET_RESOURCE_ENABLED and canvas fp is updated with APP_CHANGE_SETTING

bridiver: if you try again in master it should work

diracdeltas: ++, posted last comment before seeing your comments

diracdeltas: follow-up bug: disabling shield does not override the global fingerprinting setting.
screen shot 2016-06-21 at 2 25 45 pm

bridiver: yea, that makes sense and would apply to ads as well. contentSettings translate the siteSettings into rules so I can use the activeSettings (which will eventually be replaced by contentSettings), but I can add an override for now

bridiver: fixed here https://github.com/brave/browser-laptop/commit/9751f28f10edf4acb08e45e600ca77336e38b4d7

PR #2258, opened 2016-06-22
Add 'Esc' shortcut to stop loading for Linux and Windows

neeklamy:
- [x] Submitted a ticket for my issue if one did not already exist.

Web browsers on Windows and Linux tend to use the `Escape` key to stop page loading.


bbondy:
looks good, thanks.

PR #2290, opened 2016-06-24
git rid of those pesky ublockable ads

bridiver:
cc @bbondy @BrendanEich


js/data/siteHacks.js
```diff
@@ -11,6 +11,26 @@ const googleTagManagerRedirect = 'data:application/javascript;base64,' + base64E
...
+      return {
+        redirectURL: 'data:application/javascript;base64,MA=='
+      }
+    }
+  },
+  'nypost.com': {
+    enableForAdblock: true,
+    enableForTrackingProtection: true,
+    onBeforeRequest: function(details) {
+      if (urlParse(details.url).pathname !== '/content-control.html') {
```
bbondy: This one should just move to the abp-filter-parser-cpp data list, I'll just make a note to remove it here when I'm working on cosmetic filterds.

bbondy:
++

Issue #2298, opened 2016-06-24
1password auto-start stopped working


bridiver:
**Describe the issue you encountered:**
With 1password enabled and 1password mini not running in the background, trying to use 1password in Brave fails to start 1password as it should. On OSX it opens the startup page, but never finds and starts 1password. On Windows it crashes

**Expected behavior:**
Should open the startup page and start 1password mini

- Platform (Win7, 8, 10? macOS? Linux distro?):

- Brave Version:
0.10.4dev

- Steps to reproduce:
    1. Quit 1password and 1password mini
    2. Enable 1password in Brave preferences
    3. Select the 1password context menu item

brettholz:
I have this same issue also. I am running Windows 7 and Brave 0.10.4

brettholz:
I followed your steps to reproduce and it crashed Brave. I restarted my computer and now my 1Password is back to acting like described in issue #2308

bridiver:
Finishing up a fix now and will hopefully put out a bug fix release early this week

PR #2311, opened 2016-06-27
always use the correct partition for extensions

bridiver:
requires https://github.com/brave/electron/commit/9832990e70bffac193c692aa242e90f3a8bf17b6
The main-1 partition was creating extra extension background pages
@bbondy


bridiver:
fixes #2298

bbondy:
We're on updated electron now, so merging.  Changes look good.

PR #2316, opened 2016-06-27
add LastPass

bridiver:
@bbondy 
requires https://github.com/brave/browser-laptop/pull/2311


bbondy:
yay!

NikosEfthias:
how we can download the version with lastpass i dont see any link is it a secret ?

bridiver:
it will be available in the next release

roqstr:
wasn't 0.11.0 planned for 07/01/16, so yesterday?

bbondy:
Look at github milestones, they are more exact dates.  So Tuesday is planned.  Those dates on the timeline are just estimations and actual ones can vary by a week or two, and factors like surrounding holidays also affect it.

rosenpin:
So how can we install and use the Lastpass extension?

NikosEfthias:
though it has problems with save pass dialog and other dialogs that lastpass opens or searching through your vault within the extension

PR #2356, opened 2016-07-01
Add downloadInProgressUnknownTotal to download.properties

### LINKED ISSUE #2351: downloadInProgressUnknown shows in downloads bar
When downloading a file for a filesize that is unknown downloads toolbar shows the l10n ID instead of the localized string.

An example of this is the github releases page downloading a source code bundle.

It should show the localized string there.

ludable:

Copied the text from each locale's downloadInProgress property and
replaced the progress percent with a U+2026 (Horizontal Ellipsis)
character. Note that encoding of ellipsis is not consistent across
property files in the project: some use three dots, some use the
single unicode char, with a prevalence of three dots. Despite that,
I opted for the unicode char as the space is tight in the downloads
popup window. In Wikipedia (https://en.wikipedia.org/wiki/Ellipsis#In_different_languages),
I read that in Japanese the ellipsis is actually six dots, but I saw
that in other properties the localizer had used only three so I stuck
to that. Tested only in U.S. English locale in OS X.

Fix #2351

### DIFF
--- a/app/extensions/brave/content/scripts/blockCanvasFingerprinting.js
+++ b/app/extensions/brave/content/scripts/blockCanvasFingerprinting.js
@@ -78,7 +78,6 @@ if (chrome.contentSettings.canvasFingerprinting == 'block') {
 
     // Block the read from occuring; send info to background page instead
     chrome.ipc.sendToHost('got-canvas-fingerprinting', msg)
-    throw 'access denied'
   }
 
   /**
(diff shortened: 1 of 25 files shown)


luixxiul:
The translation files are generated via transifex, so you just need the English property I think.

ludable:
@luixxiul Thanks! Should have guessed.

ludable:
Closing and reissuing clean without the localizations.

Issue #2374, opened 2016-07-02
Search terms are still highlighted after closing the find panel


Liunkae:
**Did you search for similar issues before submitting this one?**
Yes.

**Describe the issue you encountered:**
Search results are still highlighted on a page after clicking the close button in the find panel.

**Expected behavior:**
The search results should be unhighlighted when the close button is clicked.

- Platform (Win7, 8, 10? macOS? Linux distro?):
macOS 10.11.5

- Brave Version:
0.10.4

- Steps to reproduce:
    1. Open a webpage (e.g. [https://duckduckgo.com](https://duckduckgo.com))
    2. Start the Find on Page feature by clicking Edit > Find on Page... or Ctrl/Cmd+F.
    3. Type a term that is on the page (e.g. "search engine")
    4. Click the close button in the find panel in the top-right corner

- Screenshot if needed:
![screenshot](https://cloud.githubusercontent.com/assets/19424103/16542527/f0bd937e-4071-11e6-8b42-19ef2278fd74.png)

- Any related issues:
None.

ajdinhedzic:
Fixed, about to submit a PR for this.

bbondy:
https://github.com/brave/browser-laptop/pull/2381

PR #2381, opened 2016-07-03
Closing Findbar removes text search highlighting from page

ajdinhedzic:
This commit is for #2374.  

I can work on adding more tests to the Findbar actions so something like this doesn't happen again.


Liunkae:
Could we include this in 0.11.1?

bbondy:
Sorry for the delay in merging this!

bbondy:
Thanks for the patch, merging.

PR #2401, opened 2016-07-05
Auto-open bookMarksToolbar when page is bookmarked. Closes #1487

### LINKED ISSUE #1487: auto-open bookmarks bar the first time user presses bookmark button (star)
If already open, leave open.

later - add some animation similar to Firefox (mac)

cezaraugusto:

### DIFF
--- a/js/components/navigationBar.js
+++ b/js/components/navigationBar.js
@@ -40,6 +40,7 @@ class NavigationBar extends ImmutableComponent {
       appActions.removeSite(siteUtil.getDetailFromFrame(this.props.activeFrame, siteTags.BOOKMARK), siteTags.BOOKMARK)
     } else {
       appActions.addSite(siteUtil.getDetailFromFrame(this.props.activeFrame, siteTags.BOOKMARK), siteTags.BOOKMARK)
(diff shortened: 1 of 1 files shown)


bsclifton:
Is this behavior always wanted? I love the bookmarks toolbar personally, but some folks might prefer to only use the Bookmarks Manager page

bradleyrichter:
@bsclifton actually, we may change things up since adding BMs to the bar is not even always wanted.

Bookmarks is a big, multi-piece project and we are getting there in pieces. Let's keep this on the back burner for now.

bbondy:
thanks @cezaraugusto, I talked to @bradleyrichter and we can merge this but could you do a couple changes:

i) As implemented, if a user tries to hide the bar because they hate bookmark toolbars, then it will keep popping up after each bookmark. Could you only do this when there are no bookmarks and no bookmark folders?

ii) Could you add a test for it appearing on first bookmark and not re-appearing after 2nd bookmark if it's not shown?

cezaraugusto:
@bbondy all good, will do it.

cezaraugusto:
Closing it in favor of #2571

PR #2402, opened 2016-07-05
Drag and drop should not be allowed to pin a tab for about:blank and about:newtab tabs

### LINKED ISSUE #2395: Drag and drop should not be allowed to pin a tab for about:blank and about:newtab tabs
Currently you can still pin a tab of those types by dragging a tab into the pinnable area (which only appears when there are at least one other pinned tab)

CC @darkdh

darkdh:

fix #2395


bbondy:
is it possible to not even enable the drop zone so a user doesn't think he can drag it onto the pinned tabs section?

darkdh:
Sounds like a better solution and it's more intuitive for users. I will try to figure out how to achieve it.

darkdh:
c12010c has the effect of disabling drop zone. Please take a look.

bbondy:
lgtm, thanks!

PR #2428, opened 2016-07-09
Add a way to view TLS certificates when error

### LINKED ISSUE #1057: add a way to view TLS certificates
There should be some way to at least view/download the raw TLS certificate on an HTTPS site. However, AFAICT the cert is only exposed to javascript when there is a certificate error event. 

For sites where there is a cert error, the frame security state includes 
```
       cert: {
            data: Uint8Array,
            issuer: string
        }
```

This should be visible somewhere in the UI - either by clicking on the lock icon in the URL bar, or as an advanced option on `about:certerror`.

darkdh:

Fix #1057 

Problems needs to be solved:
~~~- [ ] `[52870:0709/161617:ERROR:browser_plugin_guest.cc(419)] Attempting to require callback on nonexistent surface` keeps emerging when click `Show Certificate`
<img width="257" alt="screen shot 2016-07-09 at 16 42 07" src="https://cloud.githubusercontent.com/assets/11330831/16707020/75a32432-45f4-11e6-8425-96f5c442e4de.png">~~~

~~~- [ ] Waiting for [`UMD` problem](https://github.com/digitalbazaar/forge/issues/198) of [node-forge](https://github.com/digitalbazaar/forge) to be [fixed](https://github.com/digitalbazaar/forge/pull/357) then we can use this library to parse certificate info~~~

- [x] Waiting for brave/electron#25 being merged
- [x] Waiting for brave/electron#27 being merged

### DIFF
--- a/js/about/certerror.js
+++ b/js/about/certerror.js
@@ -6,17 +6,67 @@ const React = require('react')
 const Button = require('../components/button')
 const aboutActions = require('./aboutActions')
 const WindowConstants = require('../constants/windowConstants')
+const messages = require('../constants/messages')
+const ipc = window.chrome.ipc
 
 require('../../less/button.less')
 require('../../less/window.less')
 require('../../less/about/error.less')
 
+function toHexString (byteArray) {
+  return byteArray.map(function (byte) {
+    return ('0' + (byte & 0xFF).toString(16)).slice(-2)
+  }).join(':')
+}
+
+function toByteArray (str) {
+  var bytes = []
+  for (var i = 0; i < str.length; ++i) {
+    bytes.push(str.charCodeAt(i))
+  }
+  return bytes
+}
+
+function seperateHex (hexStr) {
+  var result = []
+  for (var i = 0; i < hexStr.length; ++i) {
+    result += hexStr[i]
+    if (i % 2 && i !== hexStr.length - 1) {
+      result += ':'
+    }
+  }
+  return result
+}
+
 class CertErrorPage extends React.Component {
   constructor () {
     super()
     this.state = {
-      advanced: false
+      advanced: false,
+      certDetail: false,
+      certIssuerName: null,
+      certSubjectName: null,
+      certSerialNumber: null,
+      certValidStart: null,
+      certValidExpiry: null,
+      certFingerprint: null
     }
+
+    ipc.on(messages.SET_CERT_ERROR_DETAIL, (e, detail) => {
+      var validStart = new Date()
+      var validExpiry = new Date()
+      validStart.setTime(detail.validStart * 1000)
+      validExpiry.setTime(detail.validExpiry * 1000)
+      this.setState({
+        certDetail: true,
+        certIssuerName: detail.issuerName,
+        certSubjectName: detail.subjectName,
+        certSerialNumber: seperateHex(detail.serialNumber),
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
     "keytar": "^3.0.0",
     "l20n": "^3.5.1",
     "lru_cache": "^1.0.0",
+    "node-forge": "^0.6.4",
```
diracdeltas: can be taken out?

diracdeltas: NVM, just saw your comment

js/about/certerror.js
```diff
@@ -49,12 +63,30 @@ class CertErrorPage extends React.Component {
...
         <span className='errorText'>{this.state.error || ''}</span>
+        {this.state.cert
+          ? (<div>
+            <span className='certErrorText'>{'Issued To'}</span>
+            <span className='errorText'>{'Common Name (CN): '}</span>
+            <span className='errorText'>{'Organization (O): '}</span>
+            <span className='errorText'>{'Organization Unit(OU): '}</span>
+            <span className='errorText'>{'Serial Number: '}</span>
+            <span className='certErrorText'>{'Issued By'}</span>
+            <span className='errorText'>{'Common Name (CN): ' + this.state.cert.issuerName}</span>
```
diracdeltas: I just verified that although sending the entire cert object via IPC causes a renderer crash, sending only the `issuerName` works fine. So maybe to avoid the crash, just display the issuerName in the UI and provide a link for the cert data to be saved directly as a file.

diracdeltas:
great start, thanks

darkdh:
The screenshot of 77c534a
screen shot 2016-07-13 at 01 11 14

package.json
darkdh: I did some little hack on [electron.](brave/electron#25) for 77c534a. I think maybe we can make some changes on electron to extract the detail attributes from certificate. It may be a long term and better solution and we can reduce the effect to integrate with `node-forge`

diracdeltas: That seems good for now

js/about/certerror.js
darkdh: @diracdeltas do we have to submit a issue which is sending the entire cert object via IPC causes a renderer crash?
And do we still need a link for user to download raw certificate after showing more [detail](https://github.com/brave/browser-laptop/pull/2428#issuecomment-232114982)

darkdh:
The value of `Serial Number` we show is not correct. I'm working on it.

darkdh:
The serial number issue has been fixed with brave/electron#27
screen shot 2016-07-16 at 01 55 26

diracdeltas:
Is the serialnumber fix in the latest electron release?

bbondy:
I'll be doing new builds son, maybe today.

bbondy:
It is available with npm install now and will be in beta3

diracdeltas:
++, thanks

diracdeltas:
Actually there is a problem where the serial number is undefined if it's above a certain length. Ex: https://expired.badssl.com. will push a fix to Electron

darkdh:
This PR(https://github.com/brave/electron/pull/28) will fixed the issue you described. I just tested it with this commit and it works fine.

screen shot 2016-07-20 at 13 41 21

PR #2432, opened 2016-07-09
webcontents is no longer available in the permission callback

bridiver:
requires https://github.com/brave/electron/pull/23


bbondy:
++

PR #2450, opened 2016-07-12
use fetch for non-webview url requests

bridiver:
requires https://github.com/brave/electron/pull/24

cc @bbondy @diracdeltas

### DIFF
--- a/js/lib/request.js
+++ b/js/lib/request.js
@@ -5,56 +5,34 @@
 'use strict'
 
 const electron = require('electron')
-const BrowserWindow = electron.BrowserWindow
-const ipcMain = electron.ipcMain
-const messages = require('../constants/messages')
-
-const getWebContents = () => {
-  try {
-    return BrowserWindow.getAllWindows()[0].webContents
-  } catch (e) {
-    return null
-  }
-}
-
-var nonce = 0
+const session = electron.session
 
 /**
  * Sends a network request using Chromium's networks stack instead of Node's.
  * Depends on there being a loaded browser window available.
  * @param {string} url - the url to load
- * @param {function} callback - callback to call with the response metadata and
- *   body
  */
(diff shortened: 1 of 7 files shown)


js/lib/request.js
```diff
@@ -5,56 +5,34 @@
...
  * Depends on there being a loaded browser window available.
  * @param {string} url - the url to load
- * @param {function} callback - callback to call with the response metadata and
- *   body
  */
 module.exports.request = (url, callback) => {
-  const webContents = getWebContents()
-  if (!webContents) {
-    callback(new Error('Request failed, no webContents available'))
+  let defaultSession = session.defaultSession
```
diracdeltas: is defaultSession persistent?

diracdeltas:
haven't tried with brave/electron#24 but ++ for the change

js/lib/request.js
bridiver: yes

diracdeltas: i guess that's fine for now. in the future, if `request` is handling GET requests that contains info about private browsing (ex: for the ledger), it would be better to not write to disk

bridiver: it would be easy to change to the non-persistent store

diracdeltas:
lmk when new electron build is ready; i can check that this does fetch with proxy settings

bridiver:
it uses the request context from the browser context, but I guess it doesn't hurt to double-check

js/lib/request.js
bridiver: the ledger doesn't record visits from private sessions

bbondy:
++

@diracdeltas this is ready now w/ brave/electron 1.2.7 which is the default for when you npm install now.

diracdeltas:
looks like it's using system proxy. 👍

Issue #2473, opened 2016-07-13
frame 'src' does not update on renderer initiated navigations


bridiver:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When navigating using the urlbar, both `src` and `location` are updated in the window state. When navigating by clicking on links only `location` is updated

**Expected behavior:**
`src` and `location` should be the same regardless of the navigation method

- Platform (Win7, 8, 10? macOS? Linux distro?):
All

- Brave Version:
0.11.0

- Steps to reproduce:
    1. Open a page and click on any link that opens a different page

`src` will still be the original page, but `location` will be the url of the current page

diracdeltas:
if `src` is to be updated on navigation, then i can't think of a reason why we can't just combine `src` and `location` into one. @bbondy ?

diracdeltas:
^ i mean get rid of `src` in frame state if all it's doing is tracking what the webview `src` is

bbondy:
there was a reason it had to be only set initially back in the gecko browser days, It might have had to do with inter-page nav too, but I don't recall now.  We could try it and see if anything obvious breaks. It should be obvious if there's a problem.

bridiver:
I'm going to make a smaller change to fix a few things that I can see are broken and we can try removing it after this release

bridiver:
related from windowStore.js, but I wonder if this is just a workaround for the fact that they are no synced up
```
      // If the user is changing back to the original src and they already navigated away then we need to
      // explicitly set a new location via webview.loadURL.
        let activeShortcut
        if (frame.get('location') !== action.location &&
            frame.get('src') === action.location &&
            !isIntermediateAboutPage(action.location)) {
          activeShortcut = 'explicitLoadURL'
        }
```

bridiver:
appears that it is not so simple

PR #2475, opened 2016-07-13
use location because src is not always updated

bridiver:

partial fix for #2473

check for test errors before merging
@bbondy @diracdeltas

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -137,21 +137,21 @@ class Frame extends ImmutableComponent {
     this.expireFlash(this.origin)
   }
 
-  updateWebview (cb, newSrc) {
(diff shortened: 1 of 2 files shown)


bbondy:
I think there's a good chance this could cause some regressions, can we wait until after 0.11.1 release to merge in?

diracdeltas:
this does seem to cause a bunch of test errors

diracdeltas:
pages don't load for me at all with this change

bridiver:
yep, I was actually going to update this because I realized I missed a few things

bbondy:
sorry this bitrot since before, if we should still add this please update the PR.

PR #2510, opened 2016-07-15
add a small delay to mutation observer registration to avoid "update …

bridiver:
cc @diracdeltas @bbondy

### DIFF
--- a/app/extensions/brave/content/scripts/flashListener.js
+++ b/app/extensions/brave/content/scripts/flashListener.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const placeholderUrl = chrome.extension.getURL('about-flash.html')
+
 function isAdobeLink (href) {
   if (typeof href !== 'string') {
     return false
@@ -20,17 +22,17 @@ function isAdobeLink (href) {
       }
     })
   }
-  var observer = new window.MutationObserver(function (mutations) {
+  setTimeout(() => {
+    var observer = new window.MutationObserver(function (mutations) {
+      replaceAdobeLinks()
+    })
     replaceAdobeLinks()
-  })
-  replaceAdobeLinks()
-  observer.observe(document.documentElement, {
-    childList: true,
-    subtree: true
-  })
+    observer.observe(document.documentElement, {
+      childList: true,
+      subtree: true
+    })
+  }, 1000)
 })()
-
-const placeholderUrl = 'chrome-extension://mnojpmjdmbbfmejpflffifhffcmidifd/about-flash.html'
 
 /**
  * Whether a src is a .swf file.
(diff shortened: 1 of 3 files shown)


bridiver:
this should be double-checked to make sure the delay doesn't break any edge cases

app/extensions/brave/content/scripts/flashListener.js
```diff
@@ -20,18 +22,18 @@ function isAdobeLink (href) {
       }
     })
   }
-  var observer = new window.MutationObserver(function (mutations) {
+  setTimeout(() => {
+    var observer = new window.MutationObserver(function (mutations) {
+      replaceAdobeLinks()
+    })
     replaceAdobeLinks()
```
diracdeltas: i think this call to `replaceAdobeLink` should happen immediately unless that is also causing perf issues

app/extensions/brave/content/scripts/flashListener.js
```diff
@@ -150,15 +152,17 @@ function insertFlashPlaceholders (elem) {
...
-    subtree: true
+setTimeout(() => {
+  var observer = new window.MutationObserver(function (mutations) {
+    insertFlashPlaceholders(document.documentElement)
   })
-}
+
+  if (chrome.contentSettings.flashActive != 'allow' ||
+      chrome.contentSettings.flashEnabled != 'allow') {
+    insertFlashPlaceholders(document.documentElement)
```
diracdeltas: same as above, `insertFlashPlaceholders` should be called immediately on document end

app/extensions/brave/content/scripts/flashListener.js
bridiver: it definitely helps to defer it and I don't think anyone will really notice the 1s delay

diracdeltas: true; the flash placeholder one will be noticeable though

bridiver: @bbondy ?

bbondy: I prefer perf over better flash placeholder, might be worth filing an issue though to investigate further.

diracdeltas: how about do a `document.querySelectorAll('object')` at load end and only run insertFlashPlaceholders immediately if it returns nonempty? this might get rid of the delay on most sites while keeping perf good for sites that don't load flash

bbondy:
Going to merge for now but please see @diracdeltas comment for a follow up.

app/extensions/brave/content/scripts/flashListener.js
bridiver: seems reasonable, I can try it and see if there is any performance impact

bridiver: the check didn't really make any difference and looking at the code, that makes sense because the worst case scenario for `document.querySelector('object')` is no flash and in that case it has to scan the entire DOM just like `elem.getElementsByTagName('embed')` does. However, just adding `insertFlashPlaceholders` back in at dom load doesn't seem to effect performance in any measurable way so I'll push out that change

bridiver: https://github.com/brave/browser-laptop/commit/ad8fb6d1e9aa0aa93ee451900eeb5451fd9538de

PR #2511, opened 2016-07-15
Update app.l20n

Tanvir-rahman:
I think these words are more appropriate and meaningful than previous.


luixxiul:
Translation is done here: https://www.transifex.com/brave/brave-laptop

PR #2514, opened 2016-07-16
Bookmarks search

### LINKED ISSUE #995: Add a search box on the about:bookmarks page
It should filter as you type.
It should search across all folders.

aekeus:

  * Searches existing bookmarks on character input
  * Cancelled when search text is cleared
  * Closes #995

### DIFF
--- a/js/about/bookmarks.js
+++ b/js/about/bookmarks.js
@@ -180,14 +180,28 @@ class BookmarksList extends ImmutableComponent {
   }
 }
 
+class SearchResults extends React.Component {
+  render () {
+    return (
+      <list className='bookmarkList'>
+      {
+        this.props.bookmarks.map((bookmark) => <BookmarkItem bookmark={bookmark} />)
+      }
+      </list>
+    )
+  }
+}
+
 class AboutBookmarks extends React.Component {
   constructor () {
     super()
     this.onChangeSelectedFolder = this.onChangeSelectedFolder.bind(this)
+    this.onChangeSearch = this.onChangeSearch.bind(this)
     this.state = {
       bookmarks: Immutable.Map(),
       bookmarkFolders: Immutable.Map(),
-      selectedFolderId: 0
+      selectedFolderId: 0,
+      search: null
     }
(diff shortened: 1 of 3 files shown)


js/about/bookmarks.js
```diff
@@ -201,9 +215,22 @@ class AboutBookmarks extends React.Component {
       selectedFolderId: id
     })
   }
+  onChangeSearch (evt) {
+    console.log('search term changed')
```
bbondy: Pls remove this one

js/about/bookmarks.js
```diff
@@ -201,9 +215,22 @@ class AboutBookmarks extends React.Component {
...
+    this.setState({
+      search: evt.target.value
+    })
+  }
+  searchedBookmarks (searchTerm, bookmarks) {
+    return bookmarks.filter((bookmark) => {
+      const title = bookmark.get('customTitle') || bookmark.get('title')
+      return title.match(new RegExp(searchTerm, 'gi'))
```
bbondy: Pls search all of title, customTitle and location.  For example if someone does www.google.com it should find a match

js/about/bookmarks.js
```diff
@@ -180,14 +180,28 @@ class BookmarksList extends ImmutableComponent {
...
+      }
+      </list>
+    )
+  }
+}
+
 class AboutBookmarks extends React.Component {
   constructor () {
     super()
     this.onChangeSelectedFolder = this.onChangeSelectedFolder.bind(this)
```
bbondy: When selected folder changes please clear out the current search term and show the folder contents instead

bbondy: You could not clear but just search only within the selected folder and descendants would be fine too.  If you do the search all though maybe we should show a selected virtual item like chrome does for search terms so there's no ambiguity.

bbondy:
This is looking awesome, can't wait to get this functionality in. 

It would be great to have a new component for an input with a x to clear in it.  I won't block this PR on that though.

PR #2518, opened 2016-07-16
right click to "open in new tab" should open tab to the right of curr…

### LINKED ISSUE #1423: right click to "open in new tab" should open tab to the right of current tab, not the end of tabs
Related with #2453

mpizza:
fix #1423


bbondy:
Thanks for the patch. I think this is already working though and the issue was just not closed at the time it was fixed.  Is there still a case where the problem happens?


The way I am testing:
i) have 3 blank tabs
ii) load some site in the middle tab (digg.com)
iii) right click and open new tab on a link. (Video button in the header)

It opens as the new 3rd tab pushing the old 3rd blank tab to 4th position

mpizza:
It works good on master, I'll close my PR and find next good first bug.
Thanks.

PR #2527, opened 2016-07-17
Fix extension ID for PDFJS

bbondy:
This was failing to load for me I think because the PDFJS_ORIGIN ID was different from the one installed. It's unclear to me if the new one that I have that is in use by my instance will be the same for everyone.  This works concistently for me though and keeps the 2 IDs in sync

Auditors: @bridiver, @diracdeltas

### DIFF
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -132,7 +132,6 @@ let generateBraveManifest = () => {
 }
 
 let defaultExtensions = {
-  PDFJS: 'oemmndcbldboiebfnladdacbdfmadadm',
   OnePassword: 'aomjjhallfgjeglblehebfpbcfeobpgk',
   Dashlane: 'fdjamakpfbbddfjaooikfcpapjohcfmg',
   LastPass: 'hdokiejnpimakedhajhdlcegeplioahd'
@@ -181,10 +180,10 @@ module.exports.init = () => {
     installExtension(config.braveExtensionId, getExtensionsPath('brave'), {manifest_location: 'component', manifest: generateBraveManifest()})
 
     if (getSetting(settings.PDFJS_ENABLED)) {
-      installExtension(defaultExtensions.PDFJS, getExtensionsPath('pdfjs'))
-      enableExtension(defaultExtensions.PDFJS)
+      installExtension(config.PDFJSExtensionId, getExtensionsPath('pdfjs'))
+      enableExtension(config.PDFJSExtensionId)
     } else {
-      disableExtension(defaultExtensions.PDFJS)
+      disableExtension(config.PDFJSExtensionId)
     }
 
     if (getSetting(settings.ONE_PASSWORD_ENABLED)) {
--- a/js/constants/config.js
+++ b/js/constants/config.js
@@ -40,5 +40,6 @@ module.exports = {
     authUrl: (userId) => `${vaultHost}/v1/users/${userId}`,
     replacementUrl: adHost
   },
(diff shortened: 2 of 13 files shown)


js/stores/windowStore.js
```diff
@@ -36,7 +36,7 @@ let windowState = Immutable.fromJS({
 let lastEmittedState
 
 const CHANGE_EVENT = 'change'
-const PDFJS_ORIGIN = 'chrome-extension://adnmjfhcejodgpaljdmlmjoclihpcfka/'
```
bbondy: This is the main bug that here it is using `adnmjfhcejodgpaljdmlmjoclihpcfka` but above it installs as `oemmndcbldboiebfnladdacbdfmadadm`

js/constants/config.js
```diff
@@ -40,5 +40,6 @@ module.exports = {
     authUrl: (userId) => `${vaultHost}/v1/users/${userId}`,
     replacementUrl: adHost
   },
-  braveExtensionId: 'mnojpmjdmbbfmejpflffifhffcmidifd'
+  braveExtensionId: 'mnojpmjdmbbfmejpflffifhffcmidifd',
+  PDFJSExtensionId: 'ckglimbbooghmggmkikhbolcfmgniemo'
```
diracdeltas: where did you get this ID from?

app/extensions.js
```diff
@@ -132,7 +132,6 @@ let generateBraveManifest = () => {
 }
 
 let defaultExtensions = {
-  PDFJS: 'oemmndcbldboiebfnladdacbdfmadadm',
```
diracdeltas: this ID is hard-coded in pdfjs's telemetry.js module, but luckily we don't use that at all

js/constants/config.js
diracdeltas: with this change, pdfjs doesn't load for me at all (same thing that was happening for you before this change). i.e., chrome-extension://ckglimbbooghmggmkikhbolcfmgniemo/https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf is blank but chrome-extension://adnmjfhcejodgpaljdmlmjoclihpcfka/https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf loads fine. maybe this origin is unique per install?

diracdeltas: in chrome, the PDF viewer origin is `chrome-extension://oemmndcbldboiebfnladdacbdfmadadm` which makes more sense

diracdeltas: appears that `chrome.extension.getURL` is returning an arbitrary value in https://github.com/brave/browser-laptop/blob/06a2aa180a54bc4dabafd201203dd75b8748c9f8/app/extensions/pdfjs/extension-router.js#L21 instead of `chrome-extension://<extension_id>`. this does not seem to be an issue in chrome. @bridiver ideas?

diracdeltas:
Just pushed a fix for the extension URL inconsistency issue so now it works for me. @bbondy please check that it works for you.

app/extensions.js
bbondy: Yep I noticed it there but we didn't want telemetry anyway.

bbondy:
Excellent, works perfectly! Thanks!

js/constants/config.js
bbondy: Looks like this is all worked out in the follow up commit by the way by Yan. I got the ID from the one that is used when opening PDFs from a link.

PR #2537, opened 2016-07-17
Remove 'chrome-extension://*' prefix from URL bar

### LINKED ISSUE #2534: Remove prefix from URL bar of PDF files
Remove the prefix ``chrome-extension://jkpcbbnkhhgoelcjhhgknpjcopjeande/`` from the URL of a tab which opens a PDF file with the html5 reader. It must be annoying for users when they want to copy the URL.

- Platform (Win7, 8, 10? macOS? Linux distro?): OS X

- Brave Version: master branch

- Screenshot if needed:
screenshot 2016-07-17 13 54 04

- Any related issues: #2513 #2527 #2530

darkdh:
- [ ] Ran `git rebase -i` to squash commits if needed.

fix #2534

### DIFF
--- a/js/components/urlBar.js
+++ b/js/components/urlBar.js
@@ -15,8 +15,11 @@ const ipc = global.require('electron').ipcRenderer
 const UrlBarSuggestions = require('./urlBarSuggestions.js')
 const messages = require('../constants/messages')
 const dragTypes = require('../constants/dragTypes')
+const { getSetting } = require('../settings')
+const settings = require('../constants/settings')
 const contextMenus = require('../contextMenus')
 const dndData = require('../dndData')
+const appStoreRenderer = require('../stores/appStoreRenderer')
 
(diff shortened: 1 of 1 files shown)


bbondy:
Thanks for the patch, we should go with the more generalized `startsWith` then substring exact check though because there are some extension URLs we don't want to truncate.  One example is if you enable dashlane it will open a sign in chrome extension page and make available a dashlane preferences page which is a chrome-extension URL.

darkdh:
You are right. We cannot always truncate all the chrome-extension URL. 
888c5fa will be more flexible.

diracdeltas:
This should probably only happen if PDFJS is enabled in prefs. Otherwise LGTM!

darkdh:
Thanks @diracdeltas. Checking prefs is in bba8abf.

diracdeltas:
lgtm! please squash commits and i'll merge

PR #2546, opened 2016-07-17
prefs that require browser restart should show prompt for browser restart

### LINKED ISSUE #2543: prefs that require browser restart should show prompt for browser restart
There are a bunch of prefs in `about:preferences` that require browser restart. If they are changed, Brave should show a notification bar that says, 'Do you want to restart now?' and restart when 'Yes' is clicked.

darkdh:

fix #2543

### DIFF
--- a/app/extensions/brave/locales/en-US/app.properties
+++ b/app/extensions/brave/locales/en-US/app.properties
@@ -149,3 +149,7 @@ flashTitle=Flash Object Blocked
 flashRightClick=Right-click to run Adobe Flash
 flashSubtext=from {{source}} on {{site}}.
 flashExpirationText=Approvals reset 7 days after last visit.
+prefsRestart=Do you want to restart now?
+yes=Yes
+no=No
+neverForThisSite=Never for this site
--- a/app/index.js
+++ b/app/index.js
@@ -63,6 +63,8 @@ let throttleKeytar = false
 // Map of password notification bar messages to their callbacks
 const passwordCallbacks = {}
 
+const prefsRestartCallbacks = {}
+
 /**
  * Gets the master key for encrypting login credentials from the OS keyring.
  */
@@ -381,6 +383,26 @@ app.on('ready', () => {
       app.quit()
     })
 
+    ipcMain.on(messages.PREFS_RESTART, () => {
+      var message = locale.translation('prefsRestart')
+
+      appActions.showMessageBox({
+        buttons: [locale.translation('yes'), locale.translation('no')],
+        options: {
+          persist: false
+        },
+        message
+      })
+      prefsRestartCallbacks[message] = (buttonIndex, persist) => {
+        if (buttonIndex === 0) {
+          app.relaunch({args: process.argv.slice(1) + ['--relaunch']})
+          app.exit(0)
+        } else {
+          appActions.hideMessageBox(message)
+        }
+      }
+    })
+
     ipcMain.on(messages.UPDATE_APP_MENU, (e, args) => {
       Menu.init(AppStore.getState().get('settings'), args)
     })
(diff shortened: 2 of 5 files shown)


app/index.js
```diff
@@ -381,6 +383,26 @@ app.on('ready', () => {
       app.quit()
     })
 
+    ipcMain.on(messages.PREFS_RESTART, () => {
+      var message = 'Do you want to restart now ?'
```
luixxiul: This should be translatable.

darkdh: Updated in 69bd164c2160a1ae2323d56b8cb8de40ebe930d6, thanks.

luixxiul: Thanks, and will you please include ``?`` too?

darkdh:
Sure. Will do.

app/index.js
```diff
@@ -381,6 +383,26 @@ app.on('ready', () => {
       app.quit()
     })
 
+    ipcMain.on(messages.PREFS_RESTART, () => {
+      var message = locale.translation('prefsRestart')
+
+      appActions.showMessageBox({
+        buttons: ['Yes', 'No'],
```
luixxiul: Should this also be translatable?

darkdh: Yes, and also update button of password prompt in 94d28c7

app/extensions/brave/locales/en-US/app.properties
```diff
@@ -149,3 +149,7 @@ flashTitle=Flash Object Blocked
 flashRightClick=Right-click to run Adobe Flash
 flashSubtext=from {{source}} on {{site}}.
 flashExpirationText=Approvals reset 7 days after last visit.
+prefsRestart=Do you want to restart now ?
```
diracdeltas: There's an extra space after 'now'

darkdh: Thanks for reminding, @diracdeltas. Fixed!

bbondy:
I think it would be nice to remove the Requires a restart text from the affected preferences as well because the text is too long and repetitive that way.  Maybe also say this: 

> This change requires a restart to come into effect.  Would you want to restart now?

bbondy:
cc @bradleyrichter

darkdh:
I came up with one scenario, what if the user made change to the pref required restart and ignore the notification bar and keep changing other prefs. Then this user may not know which pref required restart specifically.

bbondy:
oh ok I see, we could do an electron MessageBox which is modal instead but we try to avoid those because it locks up everything completely.  Will defer to Brad.

js/about/preferences.js
```diff
@@ -591,6 +592,10 @@ class AboutPreferences extends React.Component {
       settings: this.state.settings.set(key, value)
     })
     aboutActions.changeSetting(key, value)
+    if (key === settings.DO_NOT_TRACK || key === settings.HARDWARE_ACCELERATION_ENABLED ||
+      key === settings.PDFJS_ENABLED) {
```
diracdeltas: @bbondy was there a reason you added the restart requirement for PDFJS in e962ef99cb6ad70ec743f8911abb7be304a4b1bc? it seems to work fine when i turn it on/off and don't restart

bbondy: I'm not sure how to reproduce but sometimes even having the option on gives a save prompt. It's something related to flipping pref though. It works reliably with a restart without the bug.

diracdeltas:
cherry-picked onto master, thanks

app/index.js
```diff
@@ -381,6 +383,26 @@ app.on('ready', () => {
...
+
+      appActions.showMessageBox({
+        buttons: [locale.translation('yes'), locale.translation('no')],
+        options: {
+          persist: false
+        },
+        message
+      })
+      prefsRestartCallbacks[message] = (buttonIndex, persist) => {
+        if (buttonIndex === 0) {
```
diracdeltas: I think this is sometimes being called before the pref is actually saved. if i have flash enabled and disable in prefs then click 'yes' to restart, it is still enabled when Brave relaunches.

PR #2572, opened 2016-07-19
Fixes #2444: Pressing down on URL bar should open autosuggest

### LINKED ISSUE #2444: Pressing down on URL bar should open autosuggest
Expected:
When there are other suggestions based on the current URL load, pressing down doesn't help bring up the autocomplete.

Actual:
When there are other suggestions based on the current URL load, pressing down does nothing.

willy-b:
- Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.

### DIFF
--- a/js/components/urlBar.js
+++ b/js/components/urlBar.js
@@ -174,6 +174,12 @@ class UrlBar extends ImmutableComponent {
     if (this.isSelected()) {
       windowActions.setUrlBarSelected(false)
       windowActions.setUrlBarActive(true)
+
+      windowActions.setUrlBarAutocompleteEnabled(true)
+      if (!this.urlBarSuggestions.suggestionList) {
+        this.urlBarSuggestions.suggestionList = this.urlBarSuggestions.getNewSuggestionList()
+      }
+      this.urlBarSuggestions.updateSuggestions(null)
     }
   }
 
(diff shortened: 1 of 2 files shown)


willy-b:
Oops, it looks like some checks have failed.
However, the AppVeyor errors seem to be in an early part of the build process (npm preinstall/install ; installing bloom-filter-cpp?) which was not modified. I didn't change any dependencies, so confused how the 'npm install' step got broken. Looking into it.

See below:
appveyor_builderror_1
appveyor_builderror_2

bbondy:
We don't have AppVeyor fully setup yet so you can ignore that. Also Travis will intermittently fail so no worries there.

willy-b:
OK, consistent with that, the Travis test failures are all timeouts:
travisintegrationtests

I'll still look into those tests locally and see if these timeouts might be reflecting a UI slowdown from the changes.

bbondy:
This is a little different behaviour than what I was expecting and what chrome/firefox does.
Could you only show when there is an onKeyDown event?

willy-b:
Whoops, thought the issue was asking for dropdown on mouse-down! :-).

bbondy:
perfect, thanks for the patch!

PR #2593, opened 2016-07-19
add clone tab support

bridiver:
requires https://github.com/brave/electron/commit/683ea441eba072aee131d735726009cf41eb98cf
cc @bbondy


js/state/frameStateUtil.js
```diff
@@ -233,27 +213,49 @@ function getPartition (frameOpts) {
   }
   return partition
 }
+
+function cloneFrame (frameOpts, guestInstanceId) {
```
bridiver: I think this actually needs to selectively include properties instead of exclude them. Unlike a webview that is detached and reattached, a cloned webview won't have the same audio playback for instance

bsclifton:
Awesome PR! This should enable the "move tab to new window" functionality 😄

js/state/frameStateUtil.js
bridiver: fixed

bridiver:
yea, this was intended to be a first step towards "move tab to new window" which I'm working on right now

js/state/frameStateUtil.js
```diff
@@ -233,27 +213,63 @@ function getPartition (frameOpts) {
...
+  const cloneableAttributes = [
+    'audioMuted',
+    'canGoBack',
+    'canGoForward',
+    'icon',
+    'title',
+    'isPrivate',
+    'partitionNumber',
+    'themeColor',
+    'computedThemeColor'
```
bbondy: what about `provisionalLocation` and maybe `src` and `location`.
provisionalLocation I can see maybe being useful on a about:certerror type of page.

js/contextMenus.js
```diff
@@ -834,6 +834,15 @@ function mainTemplateInit (nodeProps, frame) {
               }
             }
           },
+          {
+            label: locale.translation('cloneTab'),
```
bbondy: Please move this to the tab context menu instead of the in page one.

bbondy:
merging for now, seems like mainly the only thing to do in a follow up is cloneTab in the tab and not the page context menu.

bbondy:
Great work! :)

js/state/frameStateUtil.js
bridiver: it won't clone about pages

js/components/frame.js
```diff
@@ -334,6 +334,14 @@ class Frame extends ImmutableComponent {
         }
         this.webview.reloadIgnoringCache()
         break
+      case 'clone':
+        if (this.isAboutPage()) {
```
bbondy: Pls also add a similar check on the context menu to not even show the option to avid people thinking it's a bug.
Is it possible to use appUrlUtil's `isNavigatableAboutPage` here and there instead?

PR #2624, opened 2016-07-21
Transition between title mode and address entry in URL bar. Closes #1192

### LINKED ISSUE #1192: Make "URL / Title" event have more graceful transitions
I find the current transition from `URL` to `Title` to be slightly jaring, it happens a bit too quickly and harshly. I'd wager adding a 25 ms delay to the transition, as well as maybe a 50 - 100 ms "fade" would make Brave feel a bit more solid and polished!

This also "feels" in the territory of the the "hover on tab for preview" feature, which I disabled and found to be highly jaring. Might be worth doing some quantitaitve user testing to choose ideal default? :smile:

cezaraugusto:

**Worth notes**:

* I had to use `keyframes` instead of opacity because `titleBar` is a child of `urlbarForm`, and setting `urlbarForm` to zero has lead its entire children to inherit the property.
* The animation duration was arbitrary and after testing for a while I must have given way more or way less animation time than it needed.

I think @bradleyrichter would also like to take a look on it, so I did a GIF:

brave-urlbar

### DIFF
--- a/less/navigationBar.less
+++ b/less/navigationBar.less
@@ -8,6 +8,16 @@
 
 @import "variables.less";
 
+// Fade In animation for URL bar transitions
+@keyframes fadeIn {
+  from {
+    opacity: 0
+  }
+  to {
+    opacity: 1
+  }
+}
+
 // On MacOS we need to keep a padding left to avoid overlapping
 // with the window buttons to close/maximize/minimize the window.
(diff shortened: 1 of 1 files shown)


bbondy:
This looks sooo good, I'm going to merge it in. Brad might want to tweak the variables after but no need to wait on merging.

bbondy:
@bradleyrichter

PR #2644, opened 2016-07-21
animation for tab previews. Closes #2641

### LINKED ISSUE #2641: Animation for tab previews
Please add a cross-fade between tab previews. 0.2 seconds should be enough.
cc @cezaraugusto

cezaraugusto:

Preview:

brave-pr2

### DIFF
--- a/less/window.less
+++ b/less/window.less
@@ -100,6 +100,7 @@ html,
     &.isPreview {
       background: gray;
       z-index: @zindexWindowIsPreview;
+      transition: background @transitionDuration ease-in;
     }
 
     // Webviews can cause flickers w/ Chrome  v49 if left visible


bbondy:
😮 👍

bradleyrichter:
nice! Let me pull and test drive...

bsclifton:
Very well done! 😄

bradleyrichter:
Looking great @cezaraugusto. This mostly reduces the somewhat jarring effect when traveling across the tabs.

I did find one issue when it is fading between tabs. The current active tabs appears in the middle. I suspect this is because it is showing through. 

Is there any way we could set it up so this doesn't happen? (see screen shot)

image

cezaraugusto:
Hey @bradleyrichter, good catch. 

About the current active tab appearing in the middle, there's an issue I'm currently working on, here #142. 

About the new tab preview popping, the fade only happens for previews. Maybe tweaking with animation duration and timing-function could fix it. It's currently set as ease-in, maybe ease-out would be better. Will play with to see which fits best 😋  and ask you for feedback, thanks.

PR #2650, opened 2016-07-21
Hide scrollbar in tab toolbar

### LINKED ISSUE #2599: Arch linux: Scroll bar appears over tabs when they fill the tab bar
My search skills may have failed me, but I did not find this bug.  I apologize if this is a duplicate.

Problem:  After opening enough tabs (depends on window size) to fill the tab bar, a horizontal scroll bar appears that conceals the lower ~75% of the tabs.  I am still barely able to switch tabs by clicking on the very top of the tab, but it is difficult to see which tab is which.  The scroll bar is always "full", meaning it cannot actually scroll left or right no matter how many tabs there are.  Closing or pinning tabs until they no longer fill the available space causes the scroll bar to disappear.

Expected behavior: The scroll functionality should not conceal the tabs, and should effectively scroll to tabs or tab groups that are not currently visible.

Platform:  Arch Linux

Brave Version: v0.11.0

Steps to reproduce:
    1.  Open brave
    2. Open multiple tabs until they fill the available space in the tab bar.
    3. Observe the scroll bar covering the tabs.
    4. Close tabs until the scroll bar disappears.

2016-07-19-212108_1148x1041_scrot

dangerginger:
Changed `overflow-x` to `hidden` for `.tabStripContainer` to prevent a scrollbar from appearing over the tabs when they fill the tab bar.

Fixes #2599

### DIFF
--- a/less/tabs.less
+++ b/less/tabs.less
@@ -21,7 +21,7 @@
   display: flex;
   flex: 1;
   z-index: @zindexTabs;
-  overflow-x: auto;
+  overflow-x: hidden;
   overflow-y: hidden;
   &.allowDragging {
     -webkit-app-region: drag;


bbondy:
great stuff, thanks!

PR #2663, opened 2016-07-22
don't allow scaling or zooming of the ui

bridiver:
cc @bbondy 
requires https://github.com/brave/electron/commit/1e6ba023190fdcf9738e05d0ca2049173a76e576#commitcomment-18359328


bbondy:
arg sorry I merged this too early.  Made a note to re-add it when I have the new electron uploaded.

PR #2667, opened 2016-07-22
enable file access from main window

bridiver:
required changes for https://github.com/brave/electron/commit/f0bd45b6e8984d1448d459c08799adc2411436f8

### DIFF
--- a/js/stores/appStore.js
+++ b/js/stores/appStore.js
@@ -244,7 +244,9 @@ function windowDefaults () {
     windowOffset: 20,
     webPreferences: {
       sharedWorker: true,
-      partition: 'default'
+      partition: 'default',
+      allowFileAccessFromFileUrls: true,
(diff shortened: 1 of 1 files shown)


bridiver:
cc @bbondy

bbondy:
Could you explain more what the options do exactly and why the motivation for the change? thanks.

bridiver:
do be honest it wasn't clear to me exactly what the options do, but they were previously enabled by default in electron and disabled by default in Chrome. https://github.com/brave/electron/commit/f0bd45b6e8984d1448d459c08799adc2411436f8 disables them be default in electron, but they appear to be required for node and disabling them breaks some things in the main window. I tested file urls and they still work fine and other file resources with a problem.

bbondy:
k thx, matching chrome is best as long as it still works.

PR #2691, opened 2016-07-24
Fix #2682: Make getBaseDomain use LRU cache

### LINKED ISSUE #2682: Make getBaseDomain use LRU cache
@diracdeltas mentions this function is slow and we call it during filtering so we should be using an LRU cache for the lookups.

willy-b:
- [x ] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.

Should the refererExceptions array also be updated to include the base domains that keep popping up? (E.g. facebook.com, naver.net, etc.)

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -196,7 +196,7 @@ function registerForBeforeSendHeaders (session) {
           requestHeaders['Cookie'] = undefined
         }
         if (requestHeaders['Referer'] &&
-            !refererExceptions.includes(parsedUrl.hostname) && !refererExceptions.includes(getBaseDomain(parsedUrl.hostname))) {
+            !refererExceptions.includes(parsedUrl.hostname)) {
           requestHeaders['Referer'] = undefined
         }
       }
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -196,7 +196,7 @@ function registerForBeforeSendHeaders (session) {
           requestHeaders['Cookie'] = undefined
         }
         if (requestHeaders['Referer'] &&
-            !refererExceptions.includes(parsedUrl.hostname)) {
+            !refererExceptions.includes(parsedUrl.hostname) && !refererExceptions.includes(getBaseDomain(parsedUrl.hostname))) {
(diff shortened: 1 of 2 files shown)


js/lib/baseDomain.js
```diff
@@ -44,5 +54,8 @@ module.exports.getBaseDomain = function (hostname) {
     tld--
   }
 
-  return curDomain
+  baseDomain = curDomain
```
bbondy: not a big deal and super nit but you can probably just remove this line and use curDomain for the below 2 lines.

app/filtering.js
```diff
@@ -196,7 +196,7 @@ function registerForBeforeSendHeaders (session) {
           requestHeaders['Cookie'] = undefined
         }
         if (requestHeaders['Referer'] &&
-            !refererExceptions.includes(parsedUrl.hostname)) {
+            !refererExceptions.includes(parsedUrl.hostname) && !refererExceptions.includes(getBaseDomain(parsedUrl.hostname))) {
```
bbondy: cc @diracdeltas This task is good on its own without this. We don't need both checks here.  Would you prefer we update `refererExceptions` to the base domains and use `getBaseDomain` here or would you prefer we use the more narrow exceptions with not modifying `refererExceptions` and not using `getBaseDomain`?

bbondy:
Thanks for doing this! Just have a question for Yan first and then this will need a slight modification one way or the other and then we can land.

app/filtering.js
diracdeltas: i think the way it was before is fine

willy-b:
Sweet. Thanks for running a project that makes it so easy to contribute! 

Once I know whether to use refererExceptions or getBaseDomain I'll make that change and address the nit you pointed out earlier.

willy-b:
filtering.js is reverted per @diracdeltas  and a redundant line is avoided per @bbondy

bbondy:
++ thanks!

luixxiul:
In 0.11.2?

PR #2696, opened 2016-07-24
 Enable swipe gesture only when cursor is in content

### LINKED ISSUE #2684: Trackpad Scrolls Browser Chrome
Yes.

Tabs and browser window itself can be scrolled using trackpad. The location bar can become very long and extend beyond, and is clipped by the scrollable window. Scrolling sideways to get to the back button results in history navigation.

**Expected behavior:**
Tabs should not scroll up. The window should not have an outer invisible scrollable area.

Mac OS

- Brave Version:

- Steps to reproduce:
I do not know what causes this bug.

- Screenshot if needed:

- Any related issues:
The "trackpad scrolls history" issue.

darkdh:

fix #2684

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -438,6 +438,12 @@ class Frame extends ImmutableComponent {
       }
     })
     this.webview.addEventListener('focus', this.onFocus)
+    this.webview.addEventListener('mouseenter', (e) => {
+      remote.getCurrentWindow().webContents.send(messages.ENABLE_SWIPE_GESTURE)
+    })
+    this.webview.addEventListener('mouseleave', (e) => {
+      remote.getCurrentWindow().webContents.send(messages.DISABLE_SWIPE_GESTURE)
+    })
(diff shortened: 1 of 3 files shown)


bbondy:
We should work like chrome here, I think that they disable the gesture but not sure how only when the mouse pointers is on top of the urlbar .

bbondy:
This doesn't seem like a huge thing btw, if there's not an easy solution we could just sit on the issue.

darkdh:
Yeah, focus seems not enough. When cursor is on the urlbar, swipe back/forward can still happen when scrolling.

darkdh:
We should consider only enable gesture when cursor is in content window like Chrome does

bbondy:
Looks great but could you call it enable/disable swipe gesture? start/stop sounds to me like the motion of the swipe gesture is starting.

darkdh:
Sorry for the naming. Already fixed.

bbondy:
++ thanks

PR #2705, opened 2016-07-25
Add tab-preview timer for non-previewmode. Closes #1434

### LINKED ISSUE #1434: add tab-preview timer reset when entering and exiting tab bar to remove unwanted previewing
User returns to the URL bar constantly. This mouse travel invokes the tab preview timer because the mouse moves over the tabs on the way to the URL bar. Because the tab bar is re-entered before the timer ends, the preview is shown on the way back to the page.

cezaraugusto:

I talked with Brad about this but I don't think this PR solves the problem fully, but adding a small delay could make it better. And by making the delay short it also doesn't make tabs lagging when slowly switching between them. 

Edit: Previously I added an enhancement [as requested by Brad here](https://github.com/brave/browser-laptop/pull/2644#issuecomment-234577026), but found a better solution that will be more suitable for another commit.

### DIFF
--- a/js/components/tab.js
+++ b/js/components/tab.js
@@ -111,7 +111,7 @@ class Tab extends ImmutableComponent {
     const previewMode = new Date().getTime() - this.lastPreviewClearTime < 1500
     window.clearTimeout(this.hoverClearTimeout)
     this.hoverTimeout =
-      window.setTimeout(windowActions.setPreviewFrame.bind(null, this.props.frameProps), previewMode ? 0 : 400)
+      window.setTimeout(windowActions.setPreviewFrame.bind(null, this.props.frameProps), previewMode ? 200 : 400)
   }
 
   onClickTab (e) {


bbondy:
@bradleyrichter it's not clear to me, do you want this merged?

bradleyrichter:
need comment from @cezaraugusto so please wait.

cezaraugusto:
@bbondy @bradleyrichter. Supposing the following scenario:

User tries to access urlBar from a non active tab, which is right below urlBar.

**Notice that without the delay the tab sometimes have an unwanted preview:**

current

**This commit makes a slight delay to prevent the preview activation:**

enhancement

bradleyrichter:
This looks like a solution to me. Are there any side effects? If not, let's keep making headway like this and merge it in.

bbondy:
Sorry it's not clear, is there still work to do before I should merge?

bradleyrichter:
Can you merge and I will test in the master branch? Just need to hunt for side effects.

cezaraugusto:
@bbondy actually this is all. It only prevents tabs from flashing if user tries to hit the url bar from a non active tab.

I checked for drawbacks (lagging maybe) but hadn't saw any big deal. Would be great if others could test it.

I can close this PR and point to another branch so Brad could play with and still keep `master` away from a possible mistake, if that's fits better.

bbondy:
merged, thanks, please test Brad.

PR #2706, opened 2016-07-25
Better tab transition

cezaraugusto:
This commit is a fix for an unwanted behaviour pointed by Brad [here](https://github.com/brave/browser-laptop/pull/2644#issuecomment-234577026)

Preview:

ezgif-3071481646


bbondy:
++!

PR #2718, opened 2016-07-26
#323 - differentiate between external and internal navigation when hiding findbar

### LINKED ISSUE #323: Change when findbar is hidden
Inside `js/components/findbar.js`
```
if (this.props.frame.get('location') !== prevProps.frame.get('location')) {
  this.props.onFindHide()
```

Yan writes:
"I don't think this should happen for inter-page navigation. probably need to listen for webview will-navigate or did-navigate"

I agree.

david-gang:
differentiate between external and internal navigation when hiding findbar

closes #323 
The code was moved from findbar.js to frame.js, as it is more logic for the frame to decide who closes him
I wanted originally use will-navigate as it is better to close the find dialog beforehands, but for some reason the event did not fire when using   this.webview.addEventListener('will-navigate'
So i used did-navigate. This means that the search box will be closed after the user navigated.

### DIFF
--- a/js/components/findbar.js
+++ b/js/components/findbar.js
@@ -69,9 +69,6 @@ class FindBar extends ImmutableComponent {
       // Redo search if details have changed
       this.onFindFirst()
     }
-    if (this.props.frame.get('location') !== prevProps.frame.get('location')) {
-      this.props.onFindHide()
-    }
   }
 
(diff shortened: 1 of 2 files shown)


bbondy:
Sweet, thanks so much! Great fix.

PR #2726, opened 2016-07-26
Search Settings Panel: Search Suggestions (Part 1)

### LINKED ISSUE #2694: Search Settings Panel: Search Suggestions (Part 1)
- Add Search settings to the preferences with more search engines listed:
(Qwant and Yandex are explained in separate issues - #2701, #2703)

- Add shortcut key prefix string support in the location bar so you can search specific engines at any time to override your default engine

- use favicons provided by each search engine, pointing to them so they remain current over time

- engine order is alphabetical

- default is google (regional defaults may be added later)

image

darkdh:

fix #2694


darkdh:
ibwiwmffpi

bbondy:
wow this is looking really awesome so far.  The functionality is a major improvement over our current search experience. 

Here's a few things I noticed from testing:
- When I go to Search preferences, it is not showing me my default from 0.11.1
- It might already be the case but please make sure the default from 0.11.1 is preserved
- When you detect a prefix in the urlbar please show the favicon to the left in the urlbar. 
- The titles of the search preferences page (Default, Search Engines, Engine Go Key (type first)), seem to underline and look like they are clickable but they don’t do anything. Maybe just remove the underlining for them for now.

docs/state.md
```diff
@@ -134,7 +134,7 @@ AppStore
     'general.downloads.default-save-path': string, // default path for saving files
     'general.autohide-menu': boolean, // true if the Windows menu should be autohidden
     'general.disable-title-mode': boolean, // true if title mode should always be disabled
-    'search.default-search-engine': string, // path to the open search XML
+    'search.default-search-engine': string, // name of search engine, from js/data/searchProviders.js
```
bbondy: This will need to be set inside app.sessionStore based on the user's old selected xml engine.  And only done once, after that it should jus preserve the setting.  Also some users won't have an old xml setting if it is their first time install.

bbondy: It is the same setting name as before so just check for special case values of content/search/google.xml and content/search/duckduckgo.xml when loading the session state and migrate accordingly.

js/about/preferences.js
```diff
@@ -161,19 +163,73 @@ class GeneralTab extends ImmutableComponent {
...
-        <SettingItem dataL10nId='defaultSearchEngine'>
-          <select value={getSetting(settings.DEFAULT_SEARCH_ENGINE, this.props.settings)}
-            onChange={changeSetting.bind(null, this.props.onChangeSetting, settings.DEFAULT_SEARCH_ENGINE)}>
-            <option value='content/search/google.xml'>Google</option>
-            <option value='content/search/duckduckgo.xml'>DuckDuckGo</option>
-          </select>
-        </SettingItem>
-      </SettingsList>
+      <SettingItem dataL10nId='defaultSearchEngine'>
+        <SortableTable headings={['Default', 'Search Engines', 'Engine Go Key (type first)']} rows={this.searchProviders}
```
bbondy: these strings should be localized pls

bbondy: I suggest to make sortableTable take in a list of l10n ids alternately to display those.

bbondy:
ok made a few minor comments but this is close to being able to be merged, this might still make 0.11.2.

bbondy:
Pls also add some tests for the shortcut prefix text for this as well.

bbondy:
thanks!

darkdh:
Ok, working on it

bbondy:
There's a couple things off about the urlbar icon image. It seems like not enough padding on the right and also some of the favicons just look bad.   cc @bradleyrichter please take a look locally before you work more on that aspect @darkdh

app/sessionStore.js
```diff
@@ -292,6 +292,13 @@ module.exports.loadAppState = () => {
 
     try {
       data = Object.assign(module.exports.defaultAppState(), JSON.parse(data))
+      // xml migration
+      if (data.settings['search.default-search-engine'] === 'content/search/google.xml') {
+        data.settings['search.default-search-engine'] = 'Google'
+      }
+      if (data.settings['search.default-search-engine'] === 'contnt/search/duckduckgo.xml') {
+        data.settings['search.default-search-engine'] = 'DuckDuckGo'
+      }
```
bbondy: great!

bradleyrichter:
ok. I'll run the PR and reply with suggestions if needed.

So far from the clip above, it looks great!

js/components/urlBar.js
```diff
@@ -207,6 +222,28 @@ class UrlBar extends ImmutableComponent {
...
+    let location = this.props.urlbar.get('location')
+    if (location !== null && location.length !== 0) {
+      const isLocationUrl = isUrl(location)
+      if (!isLocationUrl) {
+        let entries = searchProviders.providers
+        entries.forEach((entry) => {
+          const searchRE = new RegExp('^' + entry.shortcut + ' .*', 'g')
+          if (searchRE.test(location)) {
+            const iconSize = 16
+            this.searchFaviconStyle = {
```
bbondy: Style should really only be calculated inside render, here you'd set state which indicated the active engine, something like `frame.x.navbar.urlbar.activeSearchEngine`.   I think we should also show the favicon as soon as you put searchTerm + ' ' and you don't need an extra char after that. I think it will be fixed by what I'm suggesting here.  
You might want to wait for Brad to weigh in first to make sure we still want to have the favicon in the urlbar because I mentioned some of the favicons are looking bad as is, like yahoo and duck duck go seems pixelated.

bbondy:
It's a super great feature, can't wait to get it in.  My concerns are with some favicon replacement in the urlbar after the shortcut is entered, in particular I noticed duckduckgo looks small and pixelated and yahoo icon looks huge and ugly.  Text doesn't seem to be fully aligned and missing padding as small tweaks that would help.  Maybe we need a smaller than 16x16 image here.

darkdh:
I noticed that not just urlbar but in search engine table got the same issue. Maybe we should replace the icon image source

luixxiul:
It would be nice if svg images can be used here

bradleyrichter:
1. There is a mystery to solve as to why they appear different in the URL bar and tab...

image

@darkdh Can you check if the same favicon path is being used in both cases?

2. Can you set the URL bar icon to appear after the space is typed rather than waiting for the first letter?

3. It looks like the URL bar favicon needs to move down 2 px.

4. Make sure it is hidden when the URL bar changes to Title Mode.

image


5. oops. the wrong favicon is sticking around in different URLS.

image

bradleyrichter:
@darkdh I restyled the panel a bit but I won't commit until your are done with the other things.

Please remove the title under "Search Settings" in your code though...

image

luixxiul:
Are favicons cached here?

darkdh:
Thanks for all your valuable feedback.
There is one problem remain that cause the different icon between preference and urlbar. 
Which is I cannot use favicon url in preference due to CSP.
screen shot 2016-07-30 at 16 45 06
So in 3efd6dc there are two image source of search providers in `js/data/searchProviders.js`. Static data uri is for preference used and favicon url can be used in urlbar.

bbondy:
cc @diracdeltas for advise.

diracdeltas:
feel free to change the `app/extensions.js` CSP (i assume that's where the error is from?) to `* data:`

darkdh:
Thanks @diracdeltas. It works fine now.

bradleyrichter:
@darkdh Can you please use this fontawesome icon in place of the "X" to indicate the default choice?

http://fontawesome.io/icon/check-square/

16x16px

Color it as @braveOrange or #FF5000.

darkdh:
@bradleyrichter Ok, I will replace it with fontawesome in next commit.

darkdh:
@bradleyrichter , please take a look at screenshot of b9567d6
screen shot 2016-08-02 at 10 54 09

bradleyrichter:
looks great! thanks...

bbondy:
thanks for all your hard work on this!

PR #2733, opened 2016-07-27
Extra polish for showing history on back/forward long press

bsclifton:
- Back/Forward buttons now consider secondary actions (holding meta key, including honoring shift).
- History will attempt to load icons from appState if possible before defaulting favicon URL.
- Refactored getHistory method in Frame
- Organized unit test better; tried to write Frame unit tests but ran into complications :frowning:

### DIFF
--- a/js/contextMenus.js
+++ b/js/contextMenus.js
@@ -31,6 +31,7 @@ const textUtils = require('./lib/text')
 const {isIntermediateAboutPage, isUrl} = require('./lib/appUrlUtil')
 const {getBase64FromImageUrl} = require('./lib/imageUtil')
 const urlParse = require('url').parse
+const eventUtil = require('./lib/eventUtil')
 
 const isDarwin = process.platform === 'darwin'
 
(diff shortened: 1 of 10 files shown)


luixxiul:
I would like to see all of the title when ``contextMenuItem`` is hovered.

js/contextMenus.js
```diff
@@ -1088,11 +1098,20 @@ function onForwardButtonHistoryMenu (activeFrame, history, rect) {
...
       menuTemplate.push({
         label: history.entries[index].display,
         icon: history.entries[index].icon,
-        click: (item, focusedWindow) => {
-          activeFrame.goToIndex(index)
+        click: (e, focusedWindow) => {
+          if (eventUtil.isForSecondaryAction(e)) {
+            windowActions.newFrame({
+              location: url,
+              partitionNumber: activeFrame.props.frame.get('partitionNumber')
```
bbondy: thanks for catching partition number too 👍

bbondy:
++ awesome

PR #2774, opened 2016-07-29
added shadow to the nub and reduced off-state grey

bradleyrichter:
cc @bbondy 

please wife-test if possible but this should solve it.


bbondy:
++

Issue #2062, opened 2016-06-03
Flickering active wrapper showing tab previews on hover


luixxiul:
**Describe the issue you encountered:** Active wrapper flickers showing tab previews on hover

**Expected behavior:** In previewing tabs the wrapper should not flickr

- Platform (Win7, 8, 10? OS X? Linux distro?): Ubuntu, Windows 10

- Brave Version: Master branch

PR #2780, opened 2016-07-29
Tweak preview tabs timeout & transition. Fixes #142

### LINKED ISSUE #142: Tweak preview tabs timeout

cezaraugusto:
## [WIP]

**Notes on this commit:**

* I had to change the transition animation, because setting opacity to zero for non active frames was also causing the issue. May require another polish.
* Current preview mode assumes a mouse route from bottom to top, as if user is trying to reach urlBar.

I think that with this commit #1434 should be fixed without drawbacks and possibly #2641 / #2706 would need a new makeover.

### DIFF
--- a/js/components/tab.js
+++ b/js/components/tab.js
@@ -100,16 +100,17 @@ class Tab extends ImmutableComponent {
 
   onMouseLeave () {
     window.clearTimeout(this.hoverTimeout)
-    this.lastPreviewClearTime = new Date().getTime()
     windowActions.setPreviewFrame(null)
   }
 
-  onMouseEnter () {
-    // If a user has recently seen a preview they likely are scrolling through
-    // previews.  If we're not in preview mode we add a bit of hover time
-    // before doing a preview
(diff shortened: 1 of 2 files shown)


luixxiul:
In the latest beta build I noticed #2062 started happenning again.

cezaraugusto:
Hi Suguru, yes, latest commits just smoothed transitions but actually didn't solved the problem. Hope this one will.

bbondy:
the PR is labeled WIP, is it ready to merge or are you still wokring on it?

cezaraugusto:
@bbondy yep, it's working as is but I think animations can improve a bit so I'm digging it.

bbondy:
@bradleyrichter please take a look on master and we'll go from there.

PR #2814, opened 2016-07-30
Trim for long strings inside Bravery Panel. Fix #2813

### LINKED ISSUE #2813: Longer domains may break Bravery Panel UI
### Description: ###
When expanding the *Bravery Panel* while on a site with a very long domain name, the user may find the panel UI to be broken.

### Expected Behavior: ###
The user expects the *Bravery Panel* to be irrespective of the domain they happen to be visiting.

- Platform: Windows 10, build 10586
- Brave Version: 0.11.1
- Steps to reproduce:
    1. Navigate to [a URL with a long domain name][1]
    2. Open the *Bravery Panel*
    3. Note the ability to scroll laterally

### Screenshot: ###
Below is an animated GIF of the issue:

brave-long-address

### Related Issues: ###

A similar issue exists with Brave on iOS, however the URL is negatively impacted rather than the surrounding space.

image

 [1]: http://www.thelongestdomainnameintheworldandthensomeandthensomemoreandmore.com/

cezaraugusto:


bbondy:
great, thanks. I think it would be good to add a title attribute though so you can hover for the full name.  Please do in a follow up.

PR #2832, opened 2016-07-31
Fix #2831. Remove intermediate pages from about page URLs

### LINKED ISSUE #2831: about:about should not list intermediate about pages
If you navigate to about:about it lists the about pages. 

But some of the pages that are showing shouldn't show up there. For example `about:error`

`git grep isIntermediateAboutPage` for usage to check if it is an intermediate about page to filter them out.

Just before the map in `js/about/about.js` there should be a `.filter(...)` which filters out the intermediate pages.

itwalton:

-- Filter intermediate pages from about page URLs

### DIFF
--- a/js/about/about.js
+++ b/js/about/about.js
@@ -4,7 +4,7 @@
 
 const React = require('react')
 const ImmutableComponent = require('../components/immutableComponent')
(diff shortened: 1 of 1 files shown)


bbondy:
awesome thanks for the PR!

PR #2842, opened 2016-08-01
fix flickering effect between tabs. closes #2817

### LINKED ISSUE #2817: Tab transitions now cause flicker
Yes

When hovering over multiple tabs, the 'peeking' functionality leaps back to the active tab causing a flickering effect.

**Expected behavior:**
This is a UX call, but my opinion would be to persist the view of the peeked tab until we're sure they are no longer scanning across tabs. Perhaps track mouse dx and infer behavior?

All

- Brave Version:
0.11.2

- Steps to reproduce:
    1. Open multiple tabs for different sites
    2. Move the mouse across the tabs

- Screenshot if needed:
tab_switching

- Any related issues:
https://github.com/brave/browser-laptop/issues/2765

cezaraugusto:
**Preview:**

tab_preview_fix

### DIFF
--- a/js/components/tab.js
+++ b/js/components/tab.js
@@ -107,12 +107,11 @@ class Tab extends ImmutableComponent {
     // relatedTarget inside mouseenter checks which element before this event was the pointer on
     // if this element has a tab-like class, then it's likely that the user was previewing
     // a sequency of tabs. Called here as previewMode.
-    const previewMode = /tab/i.test(e.relatedTarget.classList)
+    const previewMode = /tab(?!pages)/i.test(e.relatedTarget.classList)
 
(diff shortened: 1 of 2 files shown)


cezaraugusto:
I noticed now that's better to add a gutter to first non-pinned tab instead of last pinned one. Working on it.

cezaraugusto:
Ready for review :+1:

cezaraugusto:
Squashed a commit improving experience while going from/to urlBar.

Current delay for non-previews was changed to 200ms. Could be altered if appropriate.

**Preview:** (CC @alexwykoff)

flickening

bbondy:
++

luixxiul:
in 0.11.2?

PR #2845, opened 2016-08-01
match titlemode pref on win. fix #2301

### LINKED ISSUE #2301: Upon install, URL bar always shows but doesn't match preferences
yes

I tried out the RC today and noticed the URL bar was always showing. I went to about:preferences and then saw that the option for this was un-checked (which means it should NOT always be showing- it should auto-hide).  I toggled the setting and from that point on, it worked as expected.

**Expected behavior:**
URL bar should auto-hide by default like it did before until the user changes the preference
  Windows 7
- Brave Version:
  0.10.4 RC3
- Steps to reproduce:
  1. Don't have Brave installed (clear the personal folder which has session, etc)
  2. Install Brave on Windows
  3. Open Brave and notice URL bar is always showing (even when mouse is not over that area) and that preferences have the "Always show URL bar" option unchecked.
- Screenshot if needed:
- Any related issues:

cezaraugusto:
Auditors: @bsclifton


bsclifton:
Let's keep digging here and understand the root cause. This patch is basically saying for Windows, go off of the check state (which defaults to false), which is why the fix would work for Windows.  But there's an underlying reason

I can help you dig into this- hit me up if you get stuck :smile:

bbondy:
I'll close this PR for now since it seems like something else is the root cause, please re-submit later when fixed again.

PR #2847, opened 2016-08-02
created new variables to connect loose switches

bradleyrichter:
and turn them back to orange. Also added separate BG color for big
switch when disabled.

cc @bbondy


bbondy:
++

luixxiul:
in 0.11.2?

PR #2849, opened 2016-08-02
Perf

### LINKED ISSUE #2736: Typing in URL bar fast can be laggy
I don't reproduce this myself but Brendan does. If you type in the URL bar it can lag. This could probably be fixed by profiling and just optimizing things around keydown in the urlbar.

bbondy:
See each commit for details of the perf change
Mainly this is react changes around fixing perf when typing in the URLbar. 
Fix #2736


bbondy:
merging but @bridiver pls review when you can.

PR #2903, opened 2016-08-03
Quick fix for nasdaq.com; it's unusable w/ adblocker enabled at the moment

bsclifton:


bbondy:
awesome, thanks for taking this.

PR #2923, opened 2016-08-03
Fixes #2913: Closing an inactive tab should not change the active tab

### LINKED ISSUE #2913: Closing an inactive tab should not change the active tab
Yes

When an inactive tab is closed by clicking the X button, the active tab changes to the closed tab's opener.

**Expected behavior:**
The active tab should remain the same.

- Brave Version: 0.11.1

- Steps to reproduce:
    1. Open a page that has 2 or more links on it
    2. Open at least 2 links, each in a new tab
    3. Switch to one of the opened tabs
    4. Close the *other* tab by hovering the tab, waiting for the X button to appear, and clicking the X button.

Expected: The active tab remains the same

Actual: The first tab becomes the active tab

willy-b:

### DIFF
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -13,6 +13,7 @@ const siteTags = require('../constants/siteTags')
 const siteUtil = require('../state/siteUtil')
 const UrlUtil = require('../lib/urlutil')
 const currentWindow = require('../../app/renderer/currentWindow')
+const windowStore = require('../stores/windowStore')
 
 function dispatch (action) {
   AppDispatcher.dispatch(action)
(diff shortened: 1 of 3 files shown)


bsclifton:
This does fix the case outlined in the issue... However, I was able to reproduce a scenario where it doesn't work though. Can you help look into why?

Steps:
1) Open https://www.wikipedia.org/ in tab 1
2) Follow any link on that page, using middle click to create a new tab (but not sending focus there). This is tab 2
3) Open a new tab, tab 3, by clicking the plus and then navigating to whatever site you like
4) Set the active focus to tab 1
5) Close tab 3
6) Focus is now on tab 2 :(

willy-b:
Thanks @bsclifton for catching that.

 I just pushed another commit which should fix the case you pointed out, and the separate case where the 3rd tab is the child of the 2nd tab.

bsclifton:
Works great! Thanks for taking the time to dig in and get the parentFrameKey fix in there :smile:

PR #2924, opened 2016-08-04
Preferences now only allow for one active password manager

bsclifton:
- setting is now a drop-down instead of seperate check-boxes
- includes all logic to backfill existing customers (determine the new setting based on existing settings)
- context menu is now dynamic (right click on page, where you select 1Password, etc)
- small refactors to eliminate duplication (single extensionId, etc)
- PW manager is not autodetected (yet)

All intended logic should be reinforced by the included tests :smile: 

Also includes 2 small clean ups:
- removes unintentional console.log from popupWindow.js
- moves constants/settings tests under unit tests

cc: @bbondy @bradleyrichter


bsclifton:
Some screenshots here (needs review- @bradleyrichter)

Here's the selection (on a Mac) when drop-down is clicked:
screen shot 2016-08-03 at 5 08 00 pm

Here's the default look and feel. Notice "Manage Passwords..." is in Orange.
screen shot 2016-08-03 at 5 08 28 pm

bsclifton:
In the process of adding this, per feedback from @bradleyrichter:
image

js/about/preferences.js
```diff
@@ -435,27 +436,25 @@ class SecurityTab extends ImmutableComponent {
...
+            <option data-l10n-id='builtInPasswordManager' value={passwordManagers.BUILT_IN} />
+            <option data-l10n-id='onePassword' value={passwordManagers.ONE_PASSWORD} />
+            <option data-l10n-id='dashlane' value={passwordManagers.DASHLANE} />
+            <option data-l10n-id='lastPass' value={passwordManagers.LAST_PASS} />
+            <option data-l10n-id='doNotManageMyPasswords' value={passwordManagers.UNMANAGED} />
+          </select>
+        </SettingItem>
+        {
+          getSetting(settings.ACTIVE_PASSWORD_MANAGER, this.props.settings) === passwordManagers.BUILT_IN
+          ? <label className='linkTextSmall' data-l10n-id='managePasswords'
```
bbondy: Last pass used to have a preferences link that I think got removed in this by accident. Pls re-add

bbondy:
Since there is potential for problems around upgrades, please add steps to test. I added the label for this `qa-steps-specified`.

bbondy:
lgtm otherwise, thx for the tests too.
Please be sure to try the manual upgrade case from old session store data.

bsclifton:
Checking this out right now...

js/about/preferences.js
bsclifton: Great catch! Fixed :smile:

bsclifton:
(this comment is a WIP- saving so I don't lose it)
# Testing notes

Currently, you can select any (or all) of following in the preferences screen:
- Brave's manager (built in)
- 1Password
- Dashlane
- LastPass

It looks like this with all enabled:
image

Moving to this new release, you can only choose 1 (via dropdown) and you're also given the option to have no password manager.  When the upgrade happens, here is how the code will determine which password is your "active" password manager:

bbondy:
++

PR #2971, opened 2016-08-05
remove null site setting and don't annoy users with javascript error …

bridiver:
@bbondy

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -5,6 +5,14 @@
 
 'use strict'
 
+// TODO(bridiver) - this should also send a notification to Brave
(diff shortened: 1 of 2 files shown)


app/index.js
```diff
@@ -5,6 +5,14 @@
 
 'use strict'
 
+// TODO(bridiver) - this should also send a notification to Brave
+process.on('uncaughtException', function (error) {
+  var message, ref, stack
```
bridiver: this is just the code from Electron without the annoying popup

bbondy:
++ thanks

Issue #100, opened 2015-12-26
Tabs don't resize with window


bridiver:
## Test plan

https://github.com/brave/browser-laptop/pull/6900#issue-203781822

<hr>

Steps to reproduce:
1. Open Brave and maximize it
2. Keep creating tabs until you reach 6
3. Tabs will finally resize for the window

Not sure what the correct behavior should be here with a fixed number of tabs. Currently they only expand to fit the window size when there are 6 in the tab group. Maybe they should always fit the window width and get smaller until they max out at 6? @bradleyrichter ?

psimyn:
Can alternatively set tabArea maxWidth programmatically with `calc()` - does mean that on narrow viewports with only one or two tabs they look very narrow though


```
style={{maxWidth: `calc(${100 / this.props.tabsPerTabPage}% - ${addTabButtonWidth / this.props.tabsPerTabPage}px)`}}
```

![tabstretch](https://cloud.githubusercontent.com/assets/1731922/13031299/db8476a2-d31c-11e5-83e6-b3287fcb59e5.gif)

bridiver:
@bradleyrichter and I can up with a formula to keep them from growing too large or too small, we just need to implement it. Maybe he can post the details here?

cezaraugusto:
I'm late here but the UI seems good for me at this moment and the conversation here hadn't been concluded. Can we close it?

bsclifton:
I've confirmed this still is an issue and it does seem a bit weird.

STR that make it obvious:
1) Maximize window
2) Go to prefs; set Tabs => Tabs per tab page to 6
3) Keep hitting the + next to tabs until you have 5 tabs
4) Hit the + one more time- notice that all the tabs resize themselves, becoming much wider
5) Close the 6th tab- notice all the tabs shrink back to their previous size

This problem seems to be masked because the browser (at least in my case) defaults to 10 tabs per page. That is enough tabs that it does properly fill out the window and it's not noticeable.

@bradleyrichter: are you happy with the current behavior?

bradleyrichter:
The new 10-tab default reduced the problem as you said @bsclifton but ideally we would implement a solution. I am game for trying @psimyn 's suggestion above because we are already doing this when reducing the window width.

bridiver:
@bradleyrichter and I came up with a good sizing algorithm, but I don't think it ever got implemented. Do you still have the notes?

bridiver:
[...]

PR #2980, opened 2016-08-05
Fix bookmarks toolbar button text clipping

ayumi:
Before:
screen shot 2016-08-05 at 01 44 58

After:
screen shot 2016-08-05 at 01 46 52

Fixes brave/browser-laptop/issues/2979


bsclifton:
Reproduced before and after, this patch works great!

Thank you :smile:

bsclifton:
BTW- We have a somewhat similar issue (except with tabs) you may or may not be interested in :smiley: 
https://github.com/brave/browser-laptop/issues/100

PR #2996, opened 2016-08-05
Update crash URL to point to Heroku endpoint

aekeus:
* This change will point the crash client directly at the Heroku instance recording crashes.
  * Tested on macOS and Windows.

### DIFF
--- a/js/constants/appConfig.js
+++ b/js/constants/appConfig.js
@@ -6,7 +6,7 @@ const { getTargetAboutUrl } = require('../lib/appUrlUtil')
 // BRAVE_UPDATE_HOST should be set to the host name for the auto-updater server
 const updateHost = process.env.BRAVE_UPDATE_HOST || 'https://brave-laptop-updates.global.ssl.fastly.net'
 const winUpdateHost = process.env.BRAVE_WIN_UPDATE_HOST || 'https://brave-download.global.ssl.fastly.net'
-const crashURL = process.env.BRAVE_CRASH_URL || 'https://laptop-updates.brave.com/1/crashes'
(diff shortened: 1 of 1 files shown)


bsclifton:
++

js/constants/appConfig.js
```diff
@@ -6,7 +6,7 @@ const { getTargetAboutUrl } = require('../lib/appUrlUtil')
 // BRAVE_UPDATE_HOST should be set to the host name for the auto-updater server
 const updateHost = process.env.BRAVE_UPDATE_HOST || 'https://brave-laptop-updates.global.ssl.fastly.net'
 const winUpdateHost = process.env.BRAVE_WIN_UPDATE_HOST || 'https://brave-download.global.ssl.fastly.net'
-const crashURL = process.env.BRAVE_CRASH_URL || 'https://laptop-updates.brave.com/1/crashes'
+const crashURL = process.env.BRAVE_CRASH_URL || 'https://brave-laptop-updates.herokuapp.com/1/crashes'
```
diracdeltas: doesn't this need to be added to the CSP at `browser-laptop/app/extensions/brave/index{-dev}.html`?

aekeus: I don't think it needs to be added to the CSP as the HTTP request is issued by breakpad (separate process). But it probably wouldn't hurt to add it.

bbondy: better to not add if not needed

aekeus: I checked the current list and it does not contain the previous crash URL. So not adding should be fine. This was tested operating in dev (shows the crash menu item).

PR #2997, opened 2016-08-05
kill brave if the app fails to load due to an uncaught exception

bridiver:
cc @bbondy

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -5,12 +5,19 @@
 
 'use strict'
 
+let ready = false
 // TODO(bridiver) - this should also send a notification to Brave
 process.on('uncaughtException', function (error) {
   var message, ref, stack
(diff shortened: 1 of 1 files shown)


app/index.js
```diff
@@ -5,12 +5,19 @@
...
   var message, ref, stack
   stack = (ref = error.stack) != null ? ref : error.name + ': ' + error.message
   message = 'Uncaught Exception:\n' + stack
   console.error('An uncaught exception occurred in the main process ' + message)
+  setTimeout(() => {
+    if (!ready) {
+      console.error('Process failed to load within 60 seconds')
+      process.exit(1)
+    }
+  }, 60 * 1000)
```
bbondy: ++ since the timeout is only happens on uncaught exceptions which is rare

PR #3051, opened 2016-08-09
Clear browser data

### LINKED ISSUE #2471: security settings panel: clear browsing data now button
Browsing data can be cleared immediately (vs automatically on browser close) by using the "Clear Browsing Data Now..." button.

A modal (to the page) dialog opens and provides the same switches for a batch clearing action.

The button action can be canceled if desired. (for users who are only exploring the functionality)

image

bbondy:

Fix #2471


bbondy:
Auditors: @diracdeltas

bbondy:
Going to merge this but please review and I'll do followups.

app/menu.js
```diff
@@ -393,18 +392,17 @@ const init = (settingsState, args) => {
           accelerator: 'Shift+CmdOrCtrl+Delete',
           enabled: siteUtil.hasNoTagSites(appStore.getState().get('sites')),
```
diracdeltas: I think the menu needs to be rebuilt when the first tagless site is added; otherwise this shows up as disabled even though i have visited some sites.
STR:
clear history
visit a site
open History menu, the 'clear history' item is disabled

app/menu.js
```diff
@@ -393,18 +392,17 @@ const init = (settingsState, args) => {
...
+          click: function (item, focusedWindow) {
+            CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_OPEN_CLEAR_BROWSING_DATA_PANEL, {cachedImagesAndFiles: true}])
           }
         }, {
           label: locale.translation('clearSiteData'),
-          click: function () {
-            Filtering.clearStorageData()
-            Filtering.clearCache()
+          click: function (item, focusedWindow) {
+            CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_OPEN_CLEAR_BROWSING_DATA_PANEL, {allSiteCookies: true}])
```
diracdeltas: seems like this option should also clear cache

js/stores/appStore.js
```diff
@@ -503,6 +503,26 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_CLEAR_DATA:
+      if (action.clearDataDetail.get('browserHistory')) {
+        handleAppAction({actionType: AppConstants.APP_CLEAR_SITES_WITHOUT_TAGS})
+      }
+      if (action.clearDataDetail.get('downloadHistory')) {
+        handleAppAction({actionType: AppConstants.APP_CLEAR_COMPLETED_DOWNLOADS})
```
diracdeltas: this was throwing an error that prevented data from being cleared when there were no downloads; fixed in 034e47619bbdefb94b020b6e2bc994304fadc11a

js/stores/appStore.js
```diff
@@ -503,6 +503,26 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_CLEAR_DATA:
+      if (action.clearDataDetail.get('browserHistory')) {
+        handleAppAction({actionType: AppConstants.APP_CLEAR_SITES_WITHOUT_TAGS})
```
diracdeltas: the browser history is still available in the `history` property on the frame. after i clear browser history, my session-store-1-dev file has entries like:
```
"frames":[{"canGoBack":false,"canGoForward":false,"computedThemeColor":"rgb(51, 51, 51)","endLoadTime":1470784542553,"findDetail":{"caseSensitivity":false,"searchString":""},"fingerprintingProtection":{},"history":["http://samy.pl/evercookie/","https://www.google.com/#q=abcdyahoo","https://www.google.com/#q=abcdyahoo","https://www.bing.com/"], ...
```

diracdeltas: the history is also in `closedFrames`; should probably just delete all of those when history is cleared

app/menu.js
bbondy: Actually I'll just simplify this to not disable the item, there is no need to disable the functionality even if it does no-op because already cleared.  And disabling adds confusion for why it is disabled.

app/menu.js
bbondy: it will either way, but I'll add that so it is more accurate.

js/stores/appStore.js
bbondy: This is the same as how we did it before but I posted in this issue to handle that in next release after this one. 
https://github.com/brave/browser-laptop/issues/3065

js/stores/appStore.js
bbondy: thanks!

PR #3091, opened 2016-08-10
Fixes #3068 Implements 'Switch to new tabs immediately' setting'

### LINKED ISSUE #3068: New Tab Always gets focus when "Switch to new tabs immediately" is disabled
Every time open a new tab it takes focus. Even when "Switch to new tabs immediately" is disabled.

Expected behavior:
I expect the new tab not to get focus when "Switch to new tabs immediately" is disabled. 
- Platform: 
  OSX 10.11.6 
- Brave Version: 
  0.11.3
- Steps to reproduce:
  1. make sure "Switch to new tabs immediately" is disabled
  2. ctrl + t

Autokrator:

1. Added functionality for the 'Switch to new tab immediately' setting
2. Applied setting to work for redirects from about:bookmarks page
3. Added condition to ignore pages starting with 'about:' except 'about:newtab'


bbondy:
I think there are a lot of calls of `newFrame` not updated with the new value.  Also SWITCH_TO_NEW_TABS was meant to work like Firefox originally. Could you compare the existing and your change to it and let me know more specifically how it changes it? Thanks!

bsclifton:
@Autokrator could you rebase against current master and also check out @bbondy's comment above? Thanks :smile:

Autokrator:
My apologies, will try to get this fixed up today.

Autokrator:
@bbondy I looked how firefox handles this setting. Here are some of the things I noticed:
- [ ] Switch immediately to new tab
- "about" pages are prioritized and ignore this setting
- bookmarks and history links open in the currently active tab
- Ctrl/Cmd + click obeys preference value
- using Ctrl/Cmd + Shift + click overrides the preference and opens tab as active frame

- [x] Switch immediately to new tab
- Ctrl/Cmd + click obeys preference value
- using Ctrl/Cmd + Shift + click overrides the preference and opens tab as background frame

So the most interesting behaviour is the Ctrl/Cmd + Shift + click override which basically does the opposite of the current value specified in the preferences. If I remove my check for "about:newtab" we will basically have the same behaviour as firefox.

Regarding the calls of `newFrame` not being updated, I saw these were previously defaulting to true, I'm not sure which calls lead to this, it's been there before this update. Let me know in more detail what you would like regarding this though, I can look into it. Meanwhile I'll get the code rebased with the new master.

PR #3105, opened 2016-08-11
On new tab don't clear URL bar

### LINKED ISSUE #2197: Disappearing keystrokes in the address bar
Yup
Start typing in the address bar, and about half the time, they all will disappear after a second or two, and you will have to start over.
**Expected behavior:**
Not disappearing!
- Platform (Win7, 8, 10? OS X? Linux distro?):
Win 8.1 64bit
- Brave Version:
0.10.3
- Steps to reproduce:
    1. Just start typing and watch
    2.
    3.

- Screenshot if needed:

- Any related issues:

### LINKED ISSUE #2812: Whatever I type in URL bar immediately after creating new tab gets erased

I did a few issue searches, hopefully this isn't a duplicate!

I'm pretty fast on the keyboard. If I cmd + t to open a new tab and immediately start typing, whatever I've typed gets erased about half a second after I start.

**Expected behavior:**

The URL I've started typing should persist.

macOS

- Brave Version:

Brave: 0.11.1
Electron: 1.2.7
libchromiumcontent: 51.0.2704.103
V8: 5.1.281.65
Node.js: 6.1.0
Update Channel: dev

- Steps to reproduce:
    1. cmd + t to open a new tab
    2. Start typing immediately
    3. Notice after ~.5 seconds it gets deleted

- Screenshot if needed:

brave3

- Any related issues:

ayumi:
Fix #2197
Fix #2812

Adds a check in windowStore.js WINDOW_SET_NAVIGATED for new tab.

I'm not sure if this is a legit way to check for a new tab, but it fixes the problem for me :)

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -302,7 +302,12 @@ const doAction = (action) => {
           title: ''
         })
       }
-      updateNavBarInput(action.location, frameStatePath(key))
+
+      // Update nav bar unless when spawning a new tab. The user might have
+      // typed in the URL bar while we were navigating -- we should preserve it.
+      if (!(action.location === 'about:newtab' && !FrameStateUtil.getActiveFrame(windowState).get('canGoForward'))) {
+        updateNavBarInput(action.location, frameStatePath(key))
+      }
       break
     case WindowConstants.WINDOW_SET_NAVBAR_INPUT:
       updateNavBarInput(action.location)


Autokrator:
I just tried this solution and it works most of the time, except extremely fast cases of `CrtlorCmd + T + keypress`

bridiver:
there is definitely a larger problem that needs to be addressed because this happens in other places as well due to input/state change race conditions, but I think this is a good workaround. It's actually possible to detect this using `NavigationController::IsInitialBlankNavigation`, but we don't currently expose that from electron. I'm wondering if it might be better to make isInitialBlankNavigation a frameProp which then controls whether or not `updateNavBarInput` gets called or not? That would make it somewhat testable because we can unit test `FrameStateUtil.addFrame` for the `isInitialBlankNavigation` property (should be false for cloned and restored tabs as well as frames created with location). We can always refactor later too, but the actual issue is very hard to write a test for so just thinking about some way that we can have some testing around it.

bbondy:
I posted this to track the deeper fix:
https://github.com/brave/browser-laptop/issues/3162

And I'll merge this for now since it should be an improvement from what we have now.

Issue #3036, opened 2016-08-08
Titlebar on Windows has a lot of wasted space


bsclifton:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
Brave on Windows has a lot of titlebar bloat. Here are some screenshot demonstrating that:

Windows 10
![screen shot 2016-08-08 at 8 04 52 am](https://cloud.githubusercontent.com/assets/4733304/17484476/e253f776-5d3e-11e6-8cbe-93193a8d6fe8.png)

Windows 7
![screen shot 2016-08-08 at 8 17 49 am](https://cloud.githubusercontent.com/assets/4733304/17484881/cb8951ba-5d40-11e6-828a-60427ab8680c.png)

In order to show the minimize/maximize/close buttons, Windows requires you to add the WS_CAPTION style to the window, which also adds the titlebar and the icon.

Electron has a [frameless mode](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md) but unfortunately that doesn't have the minimize/maximize/close buttons (and it doesn't support menus). While [working on a proof of concept](https://github.com/bsclifton/BrowserWindowStyles), I discovered the possibility exists to manually render these elements using Windows API, but there are complications.

The resolution for this issue needs to:

* Create the BrowserWindow object as frameless
* Manually render the caption buttons
* Manually render the menu
* Only update Windows; no changes are desired on Mac/Linux

**Expected behavior:**

Windows 10 (and 8) UI layout:

![image](https://cloud.githubusercontent.com/assets/13509546/18258104/40c3c1be-7384-11e6-9201-8a9b35b0b873.png)

Windows 7 UI layout:

![image](https://cloud.githubusercontent.com/assets/13509546/18258108/6a8ceb56-7384-11e6-9244-626000bb0fa9.png)

- Platform (Win7, 8, 10? macOS? Linux distro?):
All Windows platforms

- Any related issues:
https://github.com/brave/browser-laptop/issues/2386

KillianKemps:
I would say the issue is not Windows-only. Here it is the same on Linux.

Screenshot on Arch Linux with Gnome: 
![screenshot from 2016-08-10 09-38-01](https://cloud.githubusercontent.com/assets/6884630/17545571/17ac7858-5ede-11e6-920d-bcfad62ef98c.png)

Screenshot on elementaryOS (based on Ubuntu):
[...]

PR #3133, opened 2016-08-11
Add custom title bar Fixes #3126 #3036

### LINKED ISSUE #3126: Titlebar on some Linux distros has a lot of wasted space
Yes

Brave has a double titlebar on some Linux distributions (one from the desktop environment and one from Brave's Electron) which is useless and it wastes space. Some screenshots to illustrate:

Screenshot on Arch Linux with Gnome: 
screenshot from 2016-08-10 09-38-01

Screenshot on elementaryOS (based on Ubuntu):
screenshot

**Expected behavior:**
As written for the similar issue for Windows #3036, it would be nice to only have Brave's titlebar without the one of the desktop environment.
  Arch Linux with GNOME
  elementaryOS (based on Ubuntu)
- Brave Version:
  0.11.3
- Any related issues:
  #3036

Sh1d0w:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist. #3126 #3036 

Reviewers: @bbondy @luixxiul @bradleyrichter 

Preview is available here: http://webm.land/media/g4DR.webm


bsclifton:
After digging in more, maybe we can work together to accept this :smile: The electron route I was looking at is a very long-term kind of change.

I don't think it would be too hard to save the native windowing environment's bitmaps and then have this code reference the image.  I'm working on some issues related to History, but afterwards I'd love to take your commits and see if we can make a great solution together :smile:

bbondy:
@bsclifton I think we could just draw it with css instead of using bitmaps directly. It might be better for dpi scaling reasons if we do it manually without a bitmap.  Tweaking the styles if needed on what we have in this PR.

package.json
```diff
@@ -71,6 +71,7 @@
   "homepage": "https://www.brave.com/",
   "dependencies": {
     "abp-filter-parser-cpp": "1.1.x",
+    "electron-drag": "^1.2.0",
```
bbondy: do we need this since we don't use it already for macOS?

package.json
```diff
@@ -83,6 +84,7 @@
     "keytar": "^3.0.0",
     "l20n": "^3.5.1",
     "lru_cache": "^1.0.0",
+    "npm": "^3.10.5",
```
bbondy: why was this added as a dep here?

js/components/main.js
```diff
@@ -577,6 +586,25 @@ class Main extends ImmutableComponent {
     }
   }
 
+  onMinimizeClick (e) {
+    const window = electron.remote.getCurrentWindow();
```
bbondy: These getCurrentWindow calls are bad for perf in general, it's better to have the current window cached in the component via using 
`require('../app/renderer/currentWindow')`


See windowActions.js for an example

Sh1d0w:
@bbondy Thanks. I removed not needed deps and tweaked the code as you suggested. Will rebase once we agree this PR is ready for merge.

@bsclifton Cool man. Let me know once you are ready to work on the styling of the buttons.

Meanwhile if you just hint me how I can get the OS theme styles for the buttons I can do it by myslef. I did a lot of digging but so far I could not find how I can achieve it.

bsclifton:
@Sh1d0w starting off with the stock look and feel would be a great start :smile:  

Windows 7 would have this kind of styling (similar to stock, except for colors):
image

Windows 10 would be stock. Here's an example of how that looks:
79fd43dc-5d46-11e6-9550-c4d41179e733


I have Windows 7 and 10 VMs ready to go, let me know if you need screenshots to help w/ the styling (how each looks with the mouse over, dimensions of each, etc).

Getting actual OS themes represented via CSS is a much larger task which I think is safe to defer.

bsclifton:
Here is how you can check for the OS version. The process.platform checks are already in place, which return 'win32'.  To narrow down which version of Windows, you can use the os module:

```
const os = require('os')
const isWindows10 = /10.0./.test(os.release())
const isWindows7 = /6.1./.test(os.release())
```

Sh1d0w:
@bsclifton @bbondy  I could really use your help, if you can provide me with a codepen with the windows styled buttons, so I can just put them in the PR? I am not familiar with the animations of the windows environment, etc.

I have made some example styles:

Default look on Linux:

http://webm.land/media/n4G2.webm

Default look on Win7 and Win10:

http://webm.land/media/0bzf.webm

I believe everything other than the styling is finished. In the future for Linux platfrom we can further customize the look and feel for Unity, Gnome, XFCE and KDE looks, but for now I think mac os look fit very good for Linux. Several other electon apps like Nylas, Kitematic use it as default look. 

Of course if you provide any other default design you wish to put there I can change it :)

luixxiul:
@Sh1d0w it looks nice for Windows! I would like to see where and how the menubar will appear.

Sh1d0w:
@luixxiul Thanks man. I am really excited as well about this compact look. Very polished and professional looking.

Good you mentioned menus. I havent thinked of it. Since we hide the native window frame, we have no more the native menus as well, just noticed that.

I guess I will have to do custom component for menus as well, with support for keyboard shortcuts. 

Once I implement it will let you know guys.

bbondy:
I think that's why Chrome doesn't do a top menu. If we had everything in the hamburger then I think that would be ok but I don't think we have everything today. CC'ing @bradleyrichter for thoughts.   

BTW about the bitmap stuff, whatever is easiest, just note that if you use bitmaps then we'll need different dpi images.

Sh1d0w:
@bbondy Yes that was exactly what I was thinking. I can add all menu elements in the native menu to the hamburger menu (these that are not currently presented there)? Let me know if you want me to do that.

@bradleyrichter Can you create some SVGs for Windows 7 and Windows 10, that I can use in the styling, please? :smile:

bsclifton:
This is looking great; Awesome job, @Sh1d0w!

Here's something I put together real quick with Windows 10 (note: the blue outlines are my theme's color- even with a frameless window, we already get that)
image

The first 7 rows are when the window is not maximized. 
Min/max/close are all the same size at 45 pixels wide by 29 pixels tall
* row 1 is default
* row 2, 3, 4 show hover state
* row 5, 6, 7 show click state

The next two rows are when the window is maximized.
Min/max/close are still the same size, but this time 45 wide x 21 tall
Note that the maximize icon changes

Sh1d0w:
@bsclifton Thanks man, really helpful! I have tinkered a little bit with your image and converted the buttons to SVGs, then applied some CSS. Here is the final result:

https://gfycat.com/RequiredHappygoluckyDuckbillcat

Let me know what you think :smile:

cc @bbondy @luixxiul

bradleyrichter:
This is looking great guys. 

Windows 7 will need something a bit different in order to blend in. I'll work on that today. 

Also, we are moving the brave button and hamburger so let's time this carefully so that we don't end up with several changes on the way to 1.0.

Sh1d0w:
@bradleyrichter Thanks. The new buttons are implemented as single component, which can be moved wherever you wish in the template, so I don't think that will cause any conflicts with the changs you are already implementing :smile: 

No one mentioned the Linux version look, so I guess we will go with the macOS like buttons that I proposed then?

bsclifton:
The Linux version will be tricky. Let's leave that alone for now until we have a more clear strategy. We can treat that as a separate issue, because of the design considerations

bsclifton:
BTW- awesome job wrapping this up in a component :smile: This will make enabling on Linux easy once we figure out how we'd like it

One of the challenges we'll have to solve before accepting the PR is allowing for the menu bar. Mac is already fine, but Windows and Linux (Unity excluded) need to have the menu at the top. Electron doesn't support frameless with a menu yet (that I know of). I'll start looking into this today

Sh1d0w:
@bsclifton One possible solution is to implement custom component for the top menu. But I was left with the impression that we will go Chrome's way with only a hamburger menu from the last comment from @bbondy ?

About the Linux styling, we can try to detect the Desktop environmnet of the current session on startup with executing simple bash script like this one (eigther from whiting node or a bash):

    if [ "$XDG_CURRENT_DESKTOP" = "" ]
    then
      desktop=$(echo "$XDG_DATA_DIRS" | sed 's/.*\(xfce\|kde\|gnome\).*/\1/')
    else
      desktop=$XDG_CURRENT_DESKTOP
    fi

    desktop=${desktop,,}  # convert to lower case
    echo "$desktop"

Once we know the DE used I can write button theme for each DE, like KDE, Unity, Gnome, XFCE etc.

I think this problem can be fully solved only when custom theming is implemented in Brave, so it will allow the user to pick from a list of predefined CSS themes to match your Desktop Environment styles on Linux for example. You can have buttons theme for KDE, Unity, Gnome etc.

bradleyrichter:
There seems to be quite a bit of backlash against Chrome's removal of the traditional menu bar. The hamburger is also becoming less popular over time, especially for web sites, and this has a trickle-down effect for general app UI. 

A hamburger menu which contains all of the menu items can be cumbersome to use because it requires 3 levels. The burger is meant to be filled with a collection of only the most used menu items from all menus, using minimal hierarchy.

So what to do? Ideally we can find a way to have both menus like Firefox. 

Is there a way to estimate the task length to add the normal menu bar that appears with the Alt key?

bsclifton:
I'll dig into the frameless window with menu later today- I think that one is doable but will require an electron change

Sh1d0w:
Hey guys. Just quick update regarding Linux platform.

From the Electron API docs it seems we can use `ELECTRON_FORCE_WINDOW_MENU_BAR` so we can force to show the global menu, even if we have frame set to `false`. Do you think this will be a suitable solution for Linux users? Also will worth trying it unde different desktop envrinments like KDE, Gnome and Unity so I can confirm the behaviour.

@bsclifton He man, any info from your side? Let me know if you find something that we can use to complete this PR :smile: 

cc @bbondy @bradleyrichter

bsclifton:
For now, we should only consider Windows (exploring what's available on Linux is good though)

The option you mentioned is a little different- we want to have the menu work like it should natively. For example, I think Unity on Ubuntu uses a system menu, but other distros don't.

The bad news: it looks like adding a menu via calls to electron is not possible. I'm digging into electron and chromium to understand what options we have (other than faking a menu, like we did with hamburger menu)

bsclifton:
I noticed there are similar libraries, possibly one was renamed.  For example:
`addressinput_util.lib` is missing, but there IS a file `libaddressinput_util.lib`

I tried to resolve that like so
```
cd "C:\Users\brian\Documents\GitHub\upstream\electron\vendor\brightray\vendor\download\libchromiumcontent\shared_library"
cp libaddressinput_util.lib addressinput_util.lib
```

When you resolve that one issue, another one pops up for `phonenumber.lib` and `phonenumber_without_metadata.lib`. I followed suit with those and then the linker moves along.  It then runs into this:
```
os_crypt.lib(os_crypt.os_crypt_win.obj) : error LNK2019: unresolved external symbol CryptProtectData referenced in function "public: static bool __cdecl OSCrypt::EncryptString(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > *)" (?EncryptString@OSCrypt@@SA_NAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@PEAV23@@Z)
os_crypt.lib(os_crypt.os_crypt_win.obj) : error LNK2019: unresolved external symbol CryptUnprotectData referenced in function "public: static bool __cdecl OSCrypt::DecryptString16(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class std::basic_string<wchar_t,struct std::char_traits<wchar_t>,class std::allocator<wchar_t> > *)" (?DecryptString16@OSCrypt@@SA_NAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@PEAV?$basic_string@_WU?$char_traits@_W@std@@V?$allocator@_W@2@@3@@Z)
brave.exe : fatal error LNK1120: 2 unresolved externals
```

Sh1d0w:
@bsclifton Good job, man! Let me know when you figure that out we can do the necessary changes to the PR if needed :)

Just merged master into my branch to keep up with the recent fixes and features.

bsclifton:
The issues I was experienced have been resolved with:
https://github.com/brave/electron/commit/67506440be0140cba7943d56aa1702dbce27439f
https://github.com/brave/electron/pull/45

I'll start looking at this tomorrow morning :smile:

Sh1d0w:
@bsclifton Hey man, any info on this menu stuff? Let me know if I can help you with something :)

bsclifton:
The problem is that with electron updated, it will show the menu above the titlebar. Here's a screenshot comparing this PR with menu enabled (top) with Firefox (bottom).
screen shot 2016-08-30 at 4 38 30 pm

Windows carves out the area with the menu bar as non-client area, which means we can't push the client area min/max/close buttons into there. I spent a lot of time trying different configurations with the raw win32 calls using a [simple test program](https://github.com/bsclifton/BrowserWindowStyles).

* I tried an approach of taking over the non-client area and claiming it as client area. This functionally works great- the menu bar is overlaid on top of the client area. The problem is that it won't render (it gets overwritten by rendering in the client area). If you click where it's supposed to be, Windows will pop up the menu, rendering it. I cannot find any documentation to manually render the menu bar.

* I am currently working through a different approach- defining a non-client area at the top of the window which encapsulates both the menu bar and taking ownership of this. This approach allows us to manually render the min/max/close buttons. I am seeing progress with this approach and I do think it can be applied to electron. This result (when complete) would look similar to Firefox. Here's a current screenshot:
screen shot 2016-09-02 at 1 29 50 am

I'm going to keep digging on this one; I had opened an issue [upstream in electron](https://github.com/electron/electron/issues/6823) and will likely share details there too as I start trying to implement this.

For the time being, I'm going to close the PR. Thanks again for not only providing the change but rebasing and checking on it. I wish I had better news :cry:

PR #3134, opened 2016-08-11
Cloned tabs appear next to the original one

### LINKED ISSUE #2779: Cloned tab should appear next to the original one
**Describe the issue you encountered:** I would like to see a cloned tab appears next to the original one, not the last opened one. This could be of my preference.

**Expected behavior:**

- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 7 64 Bit

- Brave Version: 0.11.2 Beta1

- Steps to reproduce:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues: #2768

ayumi:
Added test

Fix #2779

### DIFF
--- a/test/components/frameTest.js
+++ b/test/components/frameTest.js
@@ -150,6 +150,47 @@ describe('clone tab', function () {
     })
   })
 
+  describe('index', function () {
+    Brave.beforeAll(this)
+
+    before(function * () {
+      this.clickWithTargetPage = Brave.server.url('click_with_target.html')
+      this.page1 = Brave.server.url('page1.html')
+
+      yield setup(this.app.client)
+      yield this.app.client
+        .waitUntilWindowLoaded()
+        .windowByUrl(Brave.browserWindowUrl)
+        .waitForUrl(Brave.newTabUrl)
+        .url(this.clickWithTargetPage)
+        .waitForVisible('#name')
+        .click('#name')
+      yield this.app.client
+        .windowByUrl(Brave.browserWindowUrl)
+        .waitForExist('.tab[data-frame-key="2"]')
+      yield this.app.client
+        .ipcSend('shortcut-set-active-frame-by-index', 0)
+        .windowByUrl(Brave.browserWindowUrl)
+        .ipcSend(messages.SHORTCUT_ACTIVE_FRAME_CLONE)
+        .waitUntil(function () {
+          return this.getTabCount().then((count) => {
+            return count === 3
+          })
+        })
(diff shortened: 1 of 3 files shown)


test/components/frameTest.js
```diff
@@ -150,6 +150,44 @@ describe('clone tab', function () {
...
+        .waitForVisible('#name')
+        .click('#name')
+      yield this.app.client
+        .windowByUrl(Brave.browserWindowUrl)
+        .waitForExist('.tab[data-frame-key="2"]')
+      yield this.app.client
+        .ipcSend('shortcut-set-active-frame-by-index', 0)
+        .windowByUrl(Brave.browserWindowUrl)
+        .ipcSend(messages.SHORTCUT_ACTIVE_FRAME_CLONE)
+        .windowByUrl(Brave.browserWindowUrl)
```
bridiver: this is one of the places where you really have to be careful. There is a race condition between cloning the frame and selecting the window. If the actual frame is created after the windowByUrl call, but before the waitForExist call, the waitForExist will fail because it's actually looking for the selector the new tab

bridiver: one option is to wait until `getTabCount` returns the incremented count

test/components/frameTest.js
```diff
@@ -150,6 +150,44 @@ describe('clone tab', function () {
...
+        .waitForExist('.tab[data-frame-key="3"]')
+    })
+
+    it('inserts after the tab to clone', function * () {
+      var clickWithTargetPage = this.clickWithTargetPage
+      yield this.app.client
+        .windowByUrl(Brave.browserWindowUrl)
+        .ipcSend('shortcut-set-active-frame-by-index', 1)
```
bridiver: I think it might be better in this case to use webview:nth-child to check the webview src. Might even be useful to check all 3 of them and make sure the framekey => nth-child index is what you expect for all three

test/components/frameTest.js
ayumi: thanks, that is helpful – I decided to check `getTabCount`

test/components/frameTest.js
ayumi: I like checking each tab.
Cloning and webview src seems to be tested earlier here (line 133) so I think checking each `tab:nth-child` has the correct `[data-frame-key]` would be good.

ayumi:
@bridiver ty for the feedback. Have updated tests – can you re-review?

bridiver:
++

PR #3155, opened 2016-08-12
Padding next to Brave logo- gives extra drag area on smaller windows

bsclifton:
Added spacing left of Brave icon, per Brad's suggestion. Adds more draggable area when the window is smaller (when the URL bar takes up most of the window)


bsclifton:
Here's a demo:
screen shot 2016-08-12 at 9 27 50 am

PR #3168, opened 2016-08-13
Whitespace in url fixes #3167

### LINKED ISSUE #3167: Brave can't handle URLs with whitespace in it

Yes.

If you try to visit url, which has whitespace in it, Brave will assume you want to search via your default engine, not matter that the url starts with http or https. How I discovered this was just simply uploading an image with whitespace in it to postimage for another issue here. The generated url was this:

`https://s10.postimg.org/snp1s3aah/Screenshot+from 2016-08-13 17-50-27.png`

Now if you past this url in Chrome or FF they will actually visit this url, in Brave I get my default search engine queried.

I have rised a PR to fix this behaviour.

**Expected behavior:**
When I paste this url `https://s10.postimg.org/snp1s3aah/Screenshot+from 2016-08-13 17-50-27.png` in the url bar I expect to see the image.

Every platform.

- Brave Version:
0.11.4
- Steps to reproduce:
    1.Paste `https://s10.postimg.org/snp1s3aah/Screenshot+from 2016-08-13 17-50-27.png` in the url bar
    2. Hit eneter

- Screenshot if needed:

- Any related issues:

Sh1d0w:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist. #3167 

cc @bbondy @bridiver @bsclifton 

This PR allows to open valid URLs whith whitespace in it such as 

`https://s10.postimg.org/snp1s3aah/Screenshot+from 2016-08-13 17-50-27.png`

The original behaviour to query the search engine is preserved if you type a word or set of words. But if the string starts with scheme, no matter it has whitespace in it it will try to load the url.

What is the behaviour now?

| String typed in the url bar        | Before                  | Now                     |
|-----------------------------------|-------------------------|-------------------------|
| google.com                        | open https://google.com | open https://google.com |
| test                              | searches for test       | searches for test       |
| test test                         | searches for test test  | searches for test test  |
| `http://example.com/test image.jpg` | searches for `http://example.com/test image.jpg` via default search engine         | opens `http://example.com/test image.jpg`           |

### DIFF
--- a/js/components/urlBar.js
+++ b/js/components/urlBar.js
@@ -106,9 +106,7 @@ class UrlBar extends ImmutableComponent {
           // Filter javascript URLs to prevent self-XSS
           location = location.replace(/^(\s*javascript:)+/i, '')
           const isLocationUrl = isUrl(location)
-          // If control key is pressed and input has no space in it add www. as a prefix and .com as a suffix.
-          // For whitepsace we want a search no matter what.
-          if (!isLocationUrl && !/\s/g.test(location) && e.ctrlKey) {
+          if (!isLocationUrl && e.ctrlKey) {
             windowActions.loadUrl(this.activeFrame, `www.${location}.com`)
           } else if (this.shouldRenderUrlBarSuggestions && (this.urlBarSuggestions.activeIndex > 0 || this.props.locationValueSuffix)) {
             // Hack to make alt enter open a new tab for url bar suggestions when hitting enter on them.
--- a/js/lib/appUrlUtil.js
+++ b/js/lib/appUrlUtil.js
@@ -130,7 +130,7 @@ module.exports.isTargetAboutUrl = function (input) {
  */
 module.exports.isUrl = function (input) {
   input = input.trim()
-  return UrlUtil.isURL(input) && !/\s/g.test(input)
+  return UrlUtil.isURL(input)
 }
 
 /**
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -91,28 +91,38 @@ const UrlUtil = {
       return true
     }
 
+    // for cases where we have scheme and we dont want spaces in domain names
+    const caseDomain = /^[\w]{2,5}:\/\/[^\s\/]+\//
     // for cases, quoted strings
     const case1Reg = /^".*"$/
     // for cases, ?abc and "a? b" which should searching query
     const case2Reg = /^(\?)|(\?.+\s)/
     // for cases, pure string
     const case3Reg = /[\?\.\/\s:]/
-    // for cases, data:uri and view-source:uri
-    const case4Reg = /^\w+:.*/
+    // for cases, data:uri, view-source:uri and about
+    const case4Reg = /^data|view-source|about|chrome-extension:.*/
 
     let str = input.trim()
+    let scheme = this.getScheme(str)
+
     if (str.toLowerCase() === 'localhost') {
       return false
     }
+
     if (case1Reg.test(str)) {
       return true
     }
+
     if (case2Reg.test(str) || !case3Reg.test(str) ||
-        this.getScheme(str) === str) {
+        (scheme === undefined && /\s/g.test(str))) {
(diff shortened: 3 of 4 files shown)


js/components/urlBar.js
```diff
@@ -108,7 +108,7 @@ class UrlBar extends ImmutableComponent {
           const isLocationUrl = isUrl(location)
           // If control key is pressed and input has no space in it add www. as a prefix and .com as a suffix.
           // For whitepsace we want a search no matter what.
```
neeklamy: Should probably update this comment about whitespace and searching too.

js/lib/appUrlUtil.js
```diff
@@ -130,7 +130,7 @@ module.exports.isTargetAboutUrl = function (input) {
  */
 module.exports.isUrl = function (input) {
```
bridiver: @bbondy we should just get rid of this method and use UrlUtil

bridiver:
Thanks for the PR! Can you add some unit tests for this in `urlUtilTest.js`? I'm also not sure this is 100% correct because while spaces are legal in the path (encoded as `%20`), they are not legal in a domain name or the scheme. `my protocol://blah`, `test domain.com`, `http: //domain.com`, etc..  should return `false` for `isUrl`

js/lib/appUrlUtil.js
bbondy: yes I agree it was only ever added here at some point because urlUtil used to be an external module we couldn't modify.

Sh1d0w:
@bridiver You are completely right. I have fixed this behaviour and added tests as you requested.

js/lib/urlutil.js
```diff
@@ -91,30 +91,40 @@ const UrlUtil = {
...
     // for cases, quoted strings
     const case1Reg = /^".*"$/
     // for cases, ?abc and "a? b" which should searching query
     const case2Reg = /^(\?)|(\?.+\s)/
     // for cases, pure string
     const case3Reg = /[\?\.\/\s:]/
-    // for cases, data:uri and view-source:uri
-    const case4Reg = /^\w+:.*/
+    // for cases, data:uri, view-source:uri and about
+    const case4Reg = /^data|view-source|about\:.*/
```
bridiver: I think those might just be examples and not an exhaustive list. We also have registered and app protocols like bitcoin, 1password, etc... @bbondy ?

bridiver:
Thanks! Can you add just a few more tests to cover all the changes? The other cases I can think of are
1. multiple words
2. space in schema
3. an arbitrary registered protocol
if those pass I think this is good to merge

bridiver:
can you do an `npm run lint` and fix any issues? The CI tests are failing on some lint issues

Sh1d0w:
@bridiver I think linting is failing generally even on master, it is not a fault on my branch.

Can you please tell me what are the registered protocols for lastpass, bitcoint ect, so I can add them to the test as well? I believe they should invoke `chrome-extension://` protocol, no?

bridiver:
at least some of the lint errors appear to be related to these changes https://travis-ci.org/brave/browser-laptop/builds/152446285
don't worry about any lint changes that are not related to you changes, but feel free to fix them :) 
I wouldn't worry too much about the exact protocols, but maybe add `chrome-extension` just to verify that `-` is valid for an extension. Maybe also check that `chrome extension://` fails?

bridiver:
lint normally runs as a precommit hook on other platforms, but doesn't run on windows. I'm switching it to a pre-push hook that should work everywhere

Sh1d0w:
@bridiver Fixed linting. Added few more test cases.

Let me know if you find anything else that seems not ok.

bridiver:
Looks good! Just waiting for CI to finish and I'll merge if everything looks good there. I know some tests are broken in master right now so I'm only looking for yours to pass.

Sh1d0w:
@bridiver Cool, thanks man!

bridiver:
++

Issue #2894, opened 2016-08-02
Improve UI/UX of bookmark popup


bradleyrichter:
The current bookmark popup appears in the center of the BM bar and has that "not finished" look.

![image](https://cloud.githubusercontent.com/assets/13509546/17345985/8c1970e6-58bd-11e6-861c-40f5b6cede2f.png)

Let's improve the styling and UX with this design:

![image](https://cloud.githubusercontent.com/assets/13509546/17354901/781d60cc-5901-11e6-99e1-d3ee201f89fd.png)

- The editable field for the path should not be necessary as the user can either edit it before pressing the BM button, or edit it later. (we can gauge feedback or add a method later)

- the name of the BM should be taken from the page title (if existing) and open as selected so typing can begin

- the bookmark is created when the Add button is pressed, but this dialog allows the user to append changes in the form of the text or location
- if the location is changed, the BM must actually be moved
- remove will remove the BM, uncheck the BM icon and close the dialog
- Done accepts and closed the dialog with info as shown
- close-button = done
- the default folder is the Bookmarks toolbar
- a "View all bookmarks" link/button on the bottom provides a shortcut to the bookmarks manager. (about:bookmarks)
- the popup should open at the location of the bookmark button

- address keyboard navigation flow as described below by @neeklamy 
- recall last use folder after initial default to Bookmarks Bar

(note: the UI behind the popup in this image will be specified in a separate issue)

Part 2 of this UI task is to extend the component for use as the bookmark edit dialog. It should appear at the actual bookmark which is being edited on the BM bar.

![image](https://cloud.githubusercontent.com/assets/13509546/17796934/5a5c4128-6579-11e6-8441-4f7947f897bb.png)

- size extends to make room for the URL field
- title changes to "Edit Bookmark"
- button text changes to "Cancel/Save"

cc @bsclifton

gulkily:
I disagree that the editable field for the path is not necessary. I clean up the URL I'm adding quite often, and get quite annoyed at browsers that don't allow this in the add-bookmark dialog.

If the field is not there, then I have to break my flow to find the bookmark, wherever it may be, open the edit dialog, and only then edit and save it.

bradleyrichter:
@gulkily Thanks for the honest feedback. It does seem that browsers vary in this regard.

@bbondy Is there any way to allow the URL field to stay editable while the bookmark hanger is open? I am trying to avoid the complete duplication of the URL string, and keep the popup minimal and easy.

[...]

PR #3190, opened 2016-08-15
Validate bookmark name before creation Fixes #3188

### LINKED ISSUE #3188: Users can create bookmark folders with no name
Yes

On the Bookmarks toolbar, if I choose to create a new folder, I do not have to provide a folder name. A folder with no name is created.

**Expected behavior:**
The system should prevent me from creating a folder with no name.

Tested on OS X

- Brave Version:
Discovered in 0.11.5 RC1, but repro'd in 0.11.2. It's probably always been there.

- Steps to reproduce:
    1. Make sure the bookmarks toolbar is visible
    2. Right-click on the toolbar and select 'Add Folder...'
    3. Click Save

- Screenshot if needed:
blank_bookmark_folder

- Any related issues:

Sh1d0w:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist. #3188 

### DIFF
--- a/js/components/addEditBookmark.js
+++ b/js/components/addEditBookmark.js
@@ -24,9 +24,15 @@ class AddEditBookmark extends ImmutableComponent {
     this.onSave = this.onSave.bind(this)
     this.onRemoveBookmark = this.onRemoveBookmark.bind(this)
   }
+
   get isBlankTab () {
     return ['about:blank', 'about:newtab'].includes(this.props.currentDetail.get('location'))
   }
+
+  get bookmarkNameValid () {
+    return (this.props.currentDetail.get('title') || this.props.currentDetail.get('customTitle'))
+  }
+
   get isFolder () {
     return siteUtil.isFolder(this.props.currentDetail)
   }
@@ -68,6 +74,7 @@ class AddEditBookmark extends ImmutableComponent {
     } else {
       currentDetail = currentDetail.set('customTitle', e.target.value)
     }
+
     windowActions.setBookmarkDetail(currentDetail, this.props.originalDetail, this.props.destinationDetail)
   }
   onLocationChange (e) {
@@ -79,6 +86,11 @@ class AddEditBookmark extends ImmutableComponent {
     windowActions.setBookmarkDetail(currentDetail, this.props.originalDetail, this.props.destinationDetail)
   }
   onSave () {
+    // First check if the title of the currentDetail is set
+    if (!this.bookmarkNameValid) {
+      return false
+    }
+
     const tag = this.isFolder ? siteTags.BOOKMARK_FOLDER : siteTags.BOOKMARK
     appActions.addSite(this.props.currentDetail, tag, this.props.originalDetail, this.props.destinationDetail)
     this.onClose()
@@ -91,7 +103,7 @@ class AddEditBookmark extends ImmutableComponent {
     if (this.props.currentDetail.get('customTitle') !== undefined) {
       return this.props.currentDetail.get('customTitle')
     }
-    return this.props.currentDetail.get('title')
+    return this.props.currentDetail.get('title') || ''
   }
   render () {
     return <Dialog onHide={this.onClose} isClickDismiss>
@@ -125,7 +137,7 @@ class AddEditBookmark extends ImmutableComponent {
               ? <a data-l10n-id='delete' className='removeBookmarkLink link' onClick={this.onRemoveBookmark} />
               : null
             }
-            <Button l10nId='save' className='primaryButton' onClick={this.onSave} />
+            <Button l10nId='save' className={this.bookmarkNameValid ? 'primaryButton' : 'primaryButton disabled'} onClick={this.onSave} />
           </div>
         </div>
       </div>
--- a/less/addEditBookmark.less
+++ b/less/addEditBookmark.less
@@ -5,3 +5,7 @@
 .removeBookmarkLink {
   margin-right: auto;
 }
+
+.disabled {
+    background-color: #ccc !important;
+}


Sh1d0w:
Reviewers: @bbondy @bridiver

js/components/addEditBookmark.js
```diff
@@ -12,7 +11,7 @@ const KeyCodes = require('../constants/keyCodes')
 const siteTags = require('../constants/siteTags')
 const siteUtil = require('../state/siteUtil')
 
-class AddEditBookmark extends ImmutableComponent {
+class AddEditBookmark extends React.Component {
```
bbondy: Thanks for the patch. 

Please keep this as ImmutableComponent, and then just determine in render if it's valid or not and render accordingly.  Also I think this only styles that it's invalid currently and not stops it from submitting it. 

For any visual changes also pls include a screenshot, thanks.

less/addEditBookmark.less
```diff
@@ -5,3 +5,7 @@
 .removeBookmarkLink {
   margin-right: auto;
 }
+
+.invalid-input {
+    border: 1px solid #f2075d;
```
bbondy: pls create a variable for this border value in variables.less and use it here.

Sh1d0w:
@bbondy Styles are updated and the class as well. The only visual change that I introduce and has to be re rendered is when the state changes to invalid:



Let me know if I have to tweak something else.

bbondy:
Sorry it doesn't look disabled to me, @bradleyrichter pls provide styling suggestions.

Sh1d0w:
@bbondy  Oh well I am only applying css style to the input, not on the button. 

What would you say to gray it out like this, untilthe name is empty?

bradleyrichter:
@Sh1d0w thanks for working on this.

I would prefer to only use the full-browser grey-out for full-modal dialogs. (needs to be spec'd)

I think your greyed button above combined with the red focus indication will be fine. 

This dialog should also have a cancel button so you can give up and start over/leave with no changes.

Later we can style it more like the new BM door hanger: #2894

Sh1d0w:
@bradleyrichter Thanks. No problems, I like the Brave's idea and I am happy to contribute with what I can.

I just didn't get should I implement the cancel button and should I make any further changes in order the PR to be accepted or it is good to go like this? It is just for fixing the currenct functionality, not to intorduce new one at first place.

The designs look very cool, btw!

bbondy:
I'll add a few more follow up changes, we really appreciate the PR!

js/components/addEditBookmark.js
```diff
@@ -23,7 +23,17 @@ class AddEditBookmark extends ImmutableComponent {
     this.onClick = this.onClick.bind(this)
     this.onSave = this.onSave.bind(this)
     this.onRemoveBookmark = this.onRemoveBookmark.bind(this)
+    this.state = {
+      bookmarkNameValid: true,
+      saveButtonDisabled: true
+    };
```
bbondy: I'd recommend that you have an accessor defined for this instead:

```
get bookmarkNameValid () {
...
}
```

Note that there is an onChange handler on the bookmark name so the site data is always known, so you can just check it there in `this.props.currentDetail`.

I think you don't need to know whether the bookmark name is valid separately from if the save button should be disabled. In particular the save button should be disabled when the bookmark name is invalid.

js/components/addEditBookmark.js
```diff
@@ -23,7 +23,17 @@ class AddEditBookmark extends ImmutableComponent {
...
   }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return (Object.keys(nextState).some((prop) => nextState[prop] !== this.state[prop]) || 
+      Object.keys(nextProps).some((prop) => nextProps[prop] !== this.props[prop]))
+  }
```
bbondy: And remove this `shouldComponentUpdate`, no immutable component should need to ever override it.

js/components/addEditBookmark.js
```diff
@@ -68,6 +78,8 @@ class AddEditBookmark extends ImmutableComponent {
     } else {
       currentDetail = currentDetail.set('customTitle', e.target.value)
     }
+
+    this.setState({saveButtonDisabled: (e.target.value.length === 0)})
```
bbondy: ImmutableComponents would not have setState either

js/components/addEditBookmark.js
```diff
@@ -79,6 +91,11 @@ class AddEditBookmark extends ImmutableComponent {
     windowActions.setBookmarkDetail(currentDetail, this.props.originalDetail, this.props.destinationDetail)
   }
   onSave () {
+    // First check if the title of the currentDetail is set
+    if (!this.props.currentDetail.get('title') && !this.props.currentDetail.get('customTitle')) {
+      return this.setState({bookmarkNameValid: false});
+    }
```
bbondy: Here you'd check 
if (!this.bookmarkNameValid) {
  return false
}

js/components/addEditBookmark.js
```diff
@@ -91,15 +108,15 @@ class AddEditBookmark extends ImmutableComponent {
...
+    return this.props.currentDetail.get('title') || ''
   }
   render () {
     return <Dialog onHide={this.onClose} isClickDismiss>
       <div className='genericForm' onClick={this.onClick}>
         <div className='genericFormTable'>
           <div id='bookmarkName' className='formRow'>
             <label data-l10n-id='nameField' htmlFor='bookmarkName' />
-            <input spellCheck='false' onKeyDown={this.onKeyDown} onChange={this.onNameChange} value={this.displayBookmarkName} ref={(bookmarkName) => { this.bookmarkName = bookmarkName }} />
+            <input spellCheck='false' className={this.state.bookmarkNameValid ? '' : 'invalid-input'} onKeyDown={this.onKeyDown} onChange={this.onNameChange} value={this.displayBookmarkName} ref={(bookmarkName) => { this.bookmarkName = bookmarkName }} />
```
bbondy: `this.bookmarkNameValid` after the above changes

less/addEditBookmark.less
```diff
@@ -2,6 +2,16 @@
...
 .removeBookmarkLink {
   margin-right: auto;
 }
+
+.invalid-input {
+    border: @errorBorderSize solid @errorBorderColor;
+}
+
+.disabled {
+    background-color: #ccc!important;
```
bbondy: nit: add a space before  `!important` or just be more specific and then you can bypass the need for it altogether.
background-color: #ccc !important;

bbondy:
OK submitted the review feedback, pls also shoot me an email at brian@brave.com and we'll get you on the community slack channel if you want.

js/components/addEditBookmark.js
Sh1d0w: The only reason I did this is because if I use only one porperty to check for both the button and the input, is that the input will be highlighted in read intially, which I think should happen only if you submit empty string. However with the disabling of the button if it is empty it is not necessary, so I will refactor.

Thanks for the feedback.

Sh1d0w:
@bbondy Thanks for the great feedback. I have refactored the code according your advices.

bbondy:
super clean, thanks!

js/components/addEditBookmark.js
```diff
@@ -24,9 +24,15 @@ class AddEditBookmark extends ImmutableComponent {
     this.onSave = this.onSave.bind(this)
     this.onRemoveBookmark = this.onRemoveBookmark.bind(this)
   }
+
   get isBlankTab () {
     return ['about:blank', 'about:newtab'].includes(this.props.currentDetail.get('location'))
   }
+
+  get bookmarkNameValid () {
+    return (this.props.currentDetail.get('title') || this.props.currentDetail.get('customTitle'))
```
bbondy: Actually I'll ask for a small follow up if you don't mind, I think no title is valid for bookmarks but not bookmark folders. In that case we display as just the URL.
So the follow up PR would just have something like `return !this.isFolder || ...`

Sh1d0w: Actually in case of Bookmark `this.props.currentDetail.get('title')` will always be set as the website's title, so even if you have empty customTItle, you should not see grayed button and if you submit the title will fallback to website's title.

bbondy: oh ok great then

PR #3202, opened 2016-08-16
Update tab title on back/forward fixes #3200

### LINKED ISSUE #3200: After Back/Forward navigation, tab shows page URL instead of page title
Yes

When navigating using the Back button, the tab's text changes to the page's URL instead of the title. The Forward button is also affected by this issue.

**Expected behavior:**
Tab should display page title

- Brave Version: 0.11.5

- Steps to reproduce:
    1. Open a page with a title and at least one link. The link should not open in a new window.
    2. Ensure that the page has a title by looking at the tab text.
    3. Click the link to go to another page.
    4. Go back to the original page by pressing the Back button

Expected: The page's title is displayed in the tab text in both steps 1 and 4

Actual: In step 4, the page title changes to the page URL

- Screenshot if needed:

image

#2798

Sh1d0w:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist. #3200 

The `page-title-updated` event on webview is not triggered when invoking back/forward actions. This PR adds ipc message that makes sure window data such as title is always sent back to electron, so it can be properly updated when the page changes.

Reviewers: @bbondy @bridiver @bsclifton @diracdeltas

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -764,6 +764,7 @@ class Frame extends ImmutableComponent {
       if (this.props.findbarShown) {
         this.onFindHide()
       }
+
       for (let message in this.notificationCallbacks) {
         appActions.hideMessageBox(message)
       }
@@ -773,6 +774,10 @@ class Frame extends ImmutableComponent {
         this.webview.focus()
       }
       windowActions.setNavigated(e.url, this.props.frameKey, false)
+      // After navigating to the URL, set correct frame title
+      let webContents = this.webview.getWebContents()
+      let title = webContents.getTitleAtIndex(webContents.getCurrentEntryIndex())
+      windowActions.setFrameTitle(this.frame, title)
     })
     this.webview.addEventListener('crashed', (e) => {
       windowActions.setFrameError(this.frame, {
--- a/test/components/tabTest.js
+++ b/test/components/tabTest.js
@@ -3,7 +3,7 @@
 const Brave = require('../lib/brave')
 const messages = require('../../js/constants/messages')
 const settings = require('../../js/constants/settings')
-const {urlInput} = require('../lib/selectors')
+const {urlInput, backButton, forwardButton, activeTabTitle} = require('../lib/selectors')
 
 describe('tabs', function () {
   function * setup (client) {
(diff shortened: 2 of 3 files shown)


bbondy:
@bridiver pls review

app/extensions/brave/content/scripts/windowData.js
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function () {
+  var ipcRenderer = chrome.ipc;
+
+  var data = {
+    "title": document.title,
+    "url": window.location.href
```
bridiver: do we need the url here? The title is already set to the url if `page-title-updated` isn't fired. I can't articulate a particular threat, but I just feel like it would be generally safer to use the url from frames.js rather than sending back `window.location.href` @diracdeltas ?

test/components/tabTest.js
```diff
@@ -15,6 +16,53 @@ describe('tabs', function () {
...
+        .windowByUrl(Brave.browserWindowUrl)
+        .waitForExist('.tab.active')
+        .getText('.tab.active .tabTitle')
+        .then(function(text) {
+          assert.equal('Page 1', text)
+        })
+        .tabByIndex(0)
+        .loadUrl(page2)
+        .windowByUrl(Brave.browserWindowUrl)
+        .waitForExist('.tab.active')
```
bridiver: I think you have a race condition with these checks because `.tab.active` already exists from the previous load. You could put the getText and assert equal in a `waitUntil` block and that should be more reliable

bridiver:
Thanks! This is definitely a reasonable change from the browser side and I think it's a good workaround, but I think we really want to retrieve the title from the navigation controller entry (not currently exposed in electron). I can open a separate ticket to make that electron change and we can merge this for now after comments.

Sh1d0w:
@bridiver Just simplified the tests. I have also fixed the ipc event to take the url from the frame if no title is provided.

Let me know if there is something else :smile:

bridiver:
actually I was wrong, @bsclifton added the method that is required for this in https://github.com/brave/electron/commit/11b8318758f348c6da1b53e8e18354c47e2a2693
`getTitleAtIndex` and `getCurrentEntryIndex` will give you the title in frames.js

Sh1d0w:
@bridiver Thanks, I was looking for this in official electron docs, obviously I should have looked in another place :smile: 

I have changed the way I set the title, since now we don't get it from the web view, do you think I should get rid of `windowData.js` and just put the updating of the title in `THEME_COLOR_COMPUTED` event in frame.js?

bridiver:
I think you can remove `windowData.js` completely and set the title from the navigation controller in `did-navigate`

Sh1d0w:
@bridiver PR is updated. Let me know if all ok. Thank you very much for your feedback :)

js/components/frame.js
```diff
@@ -773,6 +774,10 @@ class Frame extends ImmutableComponent {
         this.webview.focus()
       }
       windowActions.setNavigated(e.url, this.props.frameKey, false)
+      // After navigating to the URL, set correct frame title
+      let webContents = this.webview.getWebContents()
+      let title = webContents.getTitleAtIndex(webContents.getCurrentEntryIndex())
```
bridiver: @bsclifton - if you add these methods to https://github.com/brave/electron/blob/master/lib/renderer/web-view/web-view.js#L391 they can be called directly on the webview

bridiver:
++ Thanks!

Issue #2889, opened 2016-08-02
Back button history does not hint at options beyond a certain point.


alexwykoff:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
After navigating to a lot of different pages (ex. clicking https://www.reddit.com/r/random/ a lot of times), long click on the back button displays history for the tab. If you scroll, more history is shown, but there is no UI hint that this history exists. 
 
**Expected behavior:**
If a list item is not showing all of the elements, there should be a hint that more options exist and are not displayed.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Tested on OS X

- Brave Version:
0.11.2 beta 2

- Steps to reproduce:
    1. Navigate to a lot of different pages.
    2. Long click on the back button.
    3. Check to see if all of the history is displayed.

- Screenshot if needed:
![history_scroll](https://cloud.githubusercontent.com/assets/1434040/17343274/69435d20-58c2-11e6-915b-5423753c4ec9.gif)

- Any related issues:

bbondy:
I think this should probably be limited to 12

bbondy:
cc @bradleyrichter

bradleyrichter:
I added a note about this to the original spec'd issue in #2737. 



@bsclifton @bbondy

luixxiul:
or scrollable?

bradleyrichter:
I usually find scrolling menus to be a bit tricky to navigate but perhaps we can add a "more" button at the bottom which would refresh the list with the next 10? That may need to be a lower priority feature.

bsclifton:
Will make sure to revisit this w/ the history issues I'm looking at :smile:

Issue #1993, opened 2016-05-30
Bookmarks Menu should list the bookmarks like other browsers do


garvankeeley:
(support request)

bridiver:
when working on this ticket keep in mind that we will eventually want to manage the bookmarks with https://developer.chrome.com/extensions/bookmarks

bbondy:
Here's a WIP for this task.
From here it needs:
- Rebase to work on current master tip
- Add folder support (recursive traversal of bookmarks)
- Favicons
- Apply the same thing for recent history (maybe 10 sites or so, whatever chrome does there is fine).

```
From: secret <void@void.void>
Date: Fri, 15 Jul 2016 11:36:46 -0700
Subject: [PATCH] bookmarks: first pass at showing bookmarks in menu

---
 app/menu.js          |  5 +++--
 js/commonMenu.js     | 24 ++++++++++++++++++++++++
 js/state/siteUtil.js |  9 +++++++++
 3 files changed, 36 insertions(+), 2 deletions(-)

diff --git a/app/menu.js b/app/menu.js
index bb3f3b3..cc00ecf 100644
--- a/app/menu.js
+++ b/app/menu.js
@@ -426,8 +426,9 @@ const init = (settingsState, args) => {
         CommonMenu.bookmarksMenuItem(),
         CommonMenu.bookmarksToolbarMenuItem(),
         CommonMenu.separatorMenuItem,
-        CommonMenu.importBookmarksMenuItem()
-      ]
+        CommonMenu.importBookmarksMenuItem(),
+        CommonMenu.separatorMenuItem
+      ].concat(CommonMenu.createLinks(siteUtil.getBookmarks(appStore.getState().get('sites'))))
     }, {
       label: locale.translation('window'),
       role: 'window',
diff --git a/js/commonMenu.js b/js/commonMenu.js
index 1e4a007..75f9f68 100644
--- a/js/commonMenu.js
+++ b/js/commonMenu.js
@@ -250,6 +250,30 @@ module.exports.importBookmarksMenuItem = () => {
   */
 }

+module.exports.createLinks = (sites) => {
+  var payload = []
+  sites.forEach((site) => {
+    // TODO implement bookmark folders
+    // this if statement will currently ignore folders
+    // we should include folders as submenus populated with
+    // their bookmarks
+    if (site.get('location')) {
+      payload.push({
+        label: site.get('customTitle') || site.get('title'),
+        click: (item, focusedWindow) => {
[...]

Issue #444, opened 2016-01-29
Implement history and Ctrl +Y to bring it up


Tombsmen:
Ctrl +Y does not bring up history in windows version of Brave, it literaly does nothing.

hasanhuseyincakir:
Ctrl + H doesn't work either.

Tombsmen:
hasanhuseyincakir Ctrl + Y is the key set for Brave Ctrl + H is for other browsers.

opyh:
Is there an ETA for that? That's really the biggest thing that prevents me from being able to try Brave as my main browser…

dunn:
Based on the forward- and back-button context menus, it looks like each BrowserWindow has its own history.  Does this mean that Electron has no concept of global history and it'll have to be implemented from scratch in Brave?

Duranis:
this seems like a downer.  Received a Brave - UPDATE message.  Went ahead with it thinking either 1) would reopen with the tabs I currently had open...no it did not (yes I did have that preference turned off - but when the browser prompts you to perform an update and that it needs to restart perhaps either bypass the preference and restore pages on restart or give the option to override your preference at that time) that did kinda irritate me because I had forgotten I had that option off.
But 2) History : I should be able to see my history and restore the tabs that A) I had opened last (...Firefox - "Restore Previous Session") or B) at least see the history of pages I have opened over the course of today, yesterday, etc...  Sadly I cannot see any history, no history log at all and the "Reopen Last Closed Window" does squat.
Windows 7 Ultimate
Brave 0.11.4

Issue #2702, opened 2016-07-24
Re-open closed tabs should include history


echosa:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When I reopen the last closed tab, it opens and loads the page it was on. However, any history for that tab is lost. Once I reopen the last closed tab, I'm unable to use its back button go to back to previous pages viewed in that tab prior to closing it. Safari does this, and it is a very nice and convenient feature.

**Expected behavior:**
I expect to be able to reopen the last closed tab and not only see the page it was on when it was closed but also be able to traverse the tab's history with the back button.
- Platform (Win7, 8, 10? macOS? Linux distro?):
  OS X 10.11.5
- Brave Version:
  0.11.1
- Steps to reproduce:
  1. Open a tab in Brave.
  2. Visit a few pages.
  3. Close the tab.
  4. Reopen the tab with the "Reopen Last Closed Tab" action.
  5. Note that you can't use the back button.

#963

luixxiul:
Duping to #889 and #444

bsclifton:
#444 was closed but did not include this functionality. Re-opening

jkup:
Another request for this feature https://community.brave.com/t/maintain-tab-history-after-restore/198

jamonholmgren:
Adding my +1 to this, and adding that this actually impacts my workflow quite a lot. Safari, Chrome, and Firefox all support restoring tab history on reopening the tab.

srirambv:
+1 from community https://community.brave.com/t/ctrl-shift-t-does-not-restore-history/1342

ayumi:
+2

patrick99e99:
+1

Brampyr:
+1

hsribei:
+1

luixxiul:
+1 from https://community.brave.com/t/all-tab-history-lost-after-brave-restarts/7840?u=suguru

CC @zentagonist for setting a milestone.

NumDeP:
+1

Ratselblitz:
This has happened to me repeatedly - browser updates and I lose all tabs *and* history!  One more time and I'll stop using Brave.

luixxiul:
You mean tabs are lost or tabs' histories are lost?

Ratselblitz:
Both…  Mac OXS if that helps…

Alino:
why is this open for a year? :\
[...]

PR #3206, opened 2016-08-16
Adds about:history, sets back/forward nav limit, more efficient dynamic menus

bsclifton:

- Brand new about:history page. It's very minimal but works (you can double click entries to launch)
- History menu now has a "Show Full History" which pulls up about:history. Fixes https://github.com/brave/browser-laptop/issues/444
- History in back/forward nav buttons is limited to 10. Fixes https://github.com/brave/browser-laptop/issues/2889
- History menu shows recently closed frames (based on WindowState)
- Broke methods out from menu.js to a helper file (menuUtil.js)
- menuUtil tests now use mockery (and work properly), stubbing electron (we can also unit test other electron dependent code like this).
- history and bookmarks menus are now dynamic and uses clear/insert calls instead of rebuilding all menu items. Fixes https://github.com/brave/browser-laptop/issues/1993
- history menu now considers modifier keys (similar to bookmarks menu)

@bridiver I could definitely use your help profiling the menu calls :smile:

For the dynamic menu, I ended up not going with removeItem...[I had made electron changes](https://github.com/bsclifton/electron/commit/13341f439ba8237316dc624805c0bf2a0ce16182), but when using the new method, electron does NOT actually remove the item from the menu object... it only hides it from the menu. Additionally, I was seeing memory leaks when I profiled this, even with electron.Menu.destroy (MenuItem does not have a destroy)

I'd also like to cc @bbondy @darkdh @diracdeltas for feedback on the changes and @alexwykoff as a heads up (I can create QA steps which outline the changes)


app/menu.js
```diff
@@ -625,35 +570,75 @@ const createMenu = (CommonMenu) => {
...
+const updateMenu = (CommonMenu, appState, windowState) => {
+  const updated = menuUtil.checkForUpdate(appState, windowState)
+  if (updated.nothingUpdated) {
+    return
+  }
+
+  // Only update menu when necessary
+
+  if (updated.settings || updated.closedFrames) {
+    console.log('rebuilding history menu')
```
bridiver: logs from debugging? Do we need/want these? We really need to add our own log method with more control over output granularity

bsclifton: Good catch :smile: Removed!

bbondy:
mentioned via DM, seems to be a bug with clean session when there are no sites. After navigating gto a single site it works for all future restarts.

js/constants/messages.js
```diff
@@ -110,6 +110,9 @@ const messages = {
   RESPONSE_WINDOW_STATE: _,
   LAST_WINDOW_STATE: _,
   UNDO_CLOSED_WINDOW: _,
+  // Menu rebuilding
+  REQUEST_WINDOW_STATE_FOR_MENU: _,
+  RESPONSE_WINDOW_STATE_FOR_MENU: _,
```
bbondy: If you're only using closedFrames from windowState then I think you should call this:
REQUEST_WINDOW_CLOSED_FRAMES
RESPONSE_WINDOW_CLOSED_FRAMES

And then update accordingly.

bbondy: actually call it something like request menu data for window, and only include closed frames in it.

app/index.js
```diff
@@ -525,7 +535,11 @@ app.on('ready', () => {
     // save app state every 5 minutes regardless of update frequency
     setInterval(initiateSessionStateSave, 1000 * 60 * 5)
     AppStore.addChangeListener(() => {
-      Menu.init(AppStore.getState().get('settings'), AppStore.getState().get('sites'))
+      if (BrowserWindow.getFocusedWindow()) {
+        BrowserWindow.getFocusedWindow().webContents.send(messages.REQUEST_WINDOW_STATE_FOR_MENU)
```
bbondy: I think a better place for requesting this data is in a `mainWindow.on('focus', ...` event of the `BrowserWindow` in the appStore.

app/extensions/brave/locales/en-US/menu.properties
```diff
@@ -43,10 +43,12 @@ home=Home
 back=Back
 forward=Forward
 reopenLastClosedWindow=Reopen Last Closed Window
-showAllHistory=Show All History
+showAllHistory=Show Full History
```
bbondy: For now just leave out the word `Full` since app state sites data is truncated to most recent 500.

app/index.js
```diff
@@ -295,6 +295,13 @@ app.on('ready', () => {
     saveIfAllCollected()
   })
 
+  // Window state must be fetched from main process; this is fired once it's retrieved
+  ipcMain.on(messages.RESPONSE_WINDOW_STATE_FOR_MENU, (wnd, data) => {
```
bbondy: pls name `data` more specific here, maybe something like `windowLevelData`

js/constants/messages.js
bbondy: Just noticed we also have `UPDATE_APP_MENU` you should rename that to be more specific to be about bookmarks, and then maybe that is an indication that you should be more specific for perf reasons in the message names as I originally suggested in the first message in this thread.

js/contextMenus.js
```diff
@@ -954,6 +954,8 @@ function onHamburgerMenu (location, e) {
 function onMainContextMenu (nodeProps, frame, contextMenuType) {
   if (contextMenuType === 'bookmark' || contextMenuType === 'bookmark-folder') {
     onBookmarkContextMenu(Immutable.fromJS(nodeProps), Immutable.fromJS({ location: '', title: '', partitionNumber: frame.get('partitionNumber') }))
+  } else if (contextMenuType === 'history') {
+    // TODO: add new onHistoryContextMenu() and associated methods.
```
bbondy: Please do this in another commit in this same PR.

js/contextMenus.js
```diff
@@ -1072,6 +1075,18 @@ function onBackButtonHistoryMenu (activeFrame, history, rect) {
         }
       })
     }
+
+    if (stopIndex !== -1) {
+      menuTemplate.push(
+        CommonMenu.separatorMenuItem,
```
bbondy: I think this separator should be added conditionally on if there is at least 1 item before it. I think it's the case there could be 0.

js/contextMenus.js
```diff
@@ -1103,6 +1119,18 @@ function onForwardButtonHistoryMenu (activeFrame, history, rect) {
         }
       })
     }
+
+    if (stopIndex !== history.entries.length) {
+      menuTemplate.push(
+        CommonMenu.separatorMenuItem,
```
bbondy: ditto double check on separator here.

js/entry.js
```diff
@@ -53,6 +53,10 @@ ipc.on(messages.REQUEST_WINDOW_STATE, () => {
   ipc.send(messages.RESPONSE_WINDOW_STATE, windowStore.getState().toJS())
 })
 
+ipc.on(messages.REQUEST_WINDOW_STATE_FOR_MENU, () => {
+  ipc.send(messages.RESPONSE_WINDOW_STATE_FOR_MENU, windowStore.getState().toJS())
```
bbondy: per before this is best to be more specific pls.

js/lib/appUrlUtil.js
```diff
@@ -75,7 +75,7 @@ module.exports.isIntermediateAboutPage = (location) =>
   ['about:safebrowsing', 'about:error', 'about:certerror'].includes(getBaseUrl(location))
 
 module.exports.isNotImplementedAboutPage = (location) =>
-  ['about:config', 'about:history'].includes(getBaseUrl(location))
+  ['about:config'].includes(getBaseUrl(location))
```
bbondy: yay!

test/unit/lib/menuUtilTest.js
```diff
@@ -0,0 +1,255 @@
+/* global describe, before, after, it */
```
bbondy: great work on these tests.

js/contextMenus.js
```diff
@@ -1103,6 +1119,18 @@ function onForwardButtonHistoryMenu (activeFrame, history, rect) {
...
+        {
+          label: locale.translation('showAllHistory'),
```
bbondy: Pls include the same item on the back and forward click and hold menus.

less/about/history.less
```diff
@@ -0,0 +1,47 @@
...
+      padding-top: 10px;
+    }
+  }
+}
+
+.searchInput {
+    float: right;
+    padding: 5px;
+    margin-top: -35px;
+}
```
bbondy: there might be an opportunity to refactor some common styles with bookmarks page, if so please take the time to do that now before we merge this.

js/about/history.js
```diff
@@ -0,0 +1,128 @@
...
+    return <div role='listitem'
+      className={className}
+      onContextMenu={aboutActions.contextMenu.bind(this, this.props.history.toJS(), 'history')}
+      data-context-menu-disable
+      onDoubleClick={this.navigate.bind(this)}>
+    {
+      this.props.history.get('customTitle') || this.props.history.get('title')
+      ? <span className='aboutListItem' title={this.props.history.get('location')}>
+        <span className='aboutItemDate'>{new Date(this.props.history.get('lastAccessedTime')).toLocaleDateString()}</span>
+        <span className='aboutItemTitle'>{this.props.history.get('customTitle') || this.props.history.get('title')}</span>
```
bbondy: There is a chance the title is blank, pls test and format correctly for that case.

bbondy: correctly means just with the url in that case.

js/about/history.js
```diff
@@ -0,0 +1,128 @@
...
+          <HistoryList history={this.state.history.filter((site) => site.get('tags').isEmpty())}
+            onChangeSelectedEntry={this.onChangeSelectedEntry}
+            selectedEntry={this.state.selectedEntry} />
+        </Sticky>
+      </div>
+    </div>
+  }
+}
+
+module.exports = <AboutHistory />
```
bbondy: If you can pull out some common controls with bookmarks it would help because we plan to do things like multi select in the future. It'll be very soon work in any case so you can save the refactor for that if you prefer.

js/lib/menuUtil.js
```diff
@@ -0,0 +1,166 @@
...
+    if (site.get('tags').includes(siteTags.BOOKMARK) && site.get('location')) {
+      payload.push({
+        // TODO include label made from favicon. It needs to be of type NativeImage
+        // which can be made using a Buffer / DataURL / local image
+        // the image will likely need to be included in the site data
+        // there was potentially concern about the size of the app state
+        // and as such there may need to be another mechanism or cache
+        //
+        // see: https://github.com/brave/browser-laptop/issues/3050
+        label: site.get('customTitle') || site.get('title'),
```
bbondy: || site.get('location')

js/lib/menuUtil.js
```diff
@@ -0,0 +1,166 @@
...
+          } else {
+            CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_ACTIVE_FRAME_LOAD_URL, site.get('location')])
+          }
+        }
+      })
+    } else if (siteUtil.isFolder(site)) {
+      const folderId = site.get('folderId')
+      const submenuItems = bookmarks.filter((bookmark) => bookmark.get('parentFolderId') === folderId)
+      payload.push({
+        label: site.get('customTitle') || site.get('title'),
```
bbondy: || site.get('location')

bbondy:
Looking great, I can't wait to get this in!
I made a bunch of small asks but I think we can get this in soon. Probably after the next commit.

js/lib/menuUtil.js
bsclifton: This one is a little weird, since it's a folder (and doesn't have a location). But I did implement location for the one above :smile:

bsclifton:
OK most of the feedback implemented :smile: Next up- focusing on the concerns with the History page (breaking code into a control, implementing context menu, etc)

js/constants/messages.js
```diff
@@ -111,8 +110,9 @@ const messages = {
   LAST_WINDOW_STATE: _,
   UNDO_CLOSED_WINDOW: _,
   // Menu rebuilding
-  REQUEST_WINDOW_STATE_FOR_MENU: _,
-  RESPONSE_WINDOW_STATE_FOR_MENU: _,
+  REQUEST_MENU_DATA_FOR_WINDOW: _,
+  RESPONSE_MENU_DATA_FOR_WINDOW: _,
+  UPDATE_MENU_BOOKMARKED_STATUS: _, /** @arg {Object} currently only has a boolean "bookmarked" */
```
bbondy: nit documentation bug, I think this arg is actually a boolean type and not an Object

js/entry.js
```diff
@@ -53,6 +53,10 @@ ipc.on(messages.REQUEST_WINDOW_STATE, () => {
   ipc.send(messages.RESPONSE_WINDOW_STATE, windowStore.getState().toJS())
 })
 
+ipc.on(messages.REQUEST_MENU_DATA_FOR_WINDOW, () => {
+  ipc.send(messages.RESPONSE_MENU_DATA_FOR_WINDOW, windowStore.getState().toJS())
```
bbondy: Pls send in a smaller subset here instead in a follow up

bridiver: I'd really like to avoid this kind of state transfer in general and I don't think it's going to be correctly synced anyway because this updates on appState changes, not windowState changes. This will be much easier once we move webContents management to the main process and if the only purpose of this is to get close frames I would just leave it out for now. If we really need it now then I think appState should track WINDOW_CLOSE_FRAME, WINDOW_UNDO_CLOSED_FRAME. That will mean some state duplication between windowState and appState, but still better than relaying windowState imo. There is also likely a performance issue doing this on every app state change.

js/components/navigationBar.js
```diff
@@ -51,8 +51,12 @@ class NavigationBar extends ImmutableComponent {
     if (!isBookmarked) {
       appActions.addSite(siteDetail, siteTags.BOOKMARK)
     }
+    // Show bookmarks toolbar after first bookmark is saved
     appActions.changeSetting(settings.SHOW_BOOKMARKS_TOOLBAR, !hasBookmark || showBookmarksToolbar)
+    // trigger the AddEditBookmark modal
     windowActions.setBookmarkDetail(siteDetail, siteDetail)
+    // Update checked/unchecked status in the Bookmarks menu
+    ipc.send(messages.UPDATE_MENU_BOOKMARKED_STATUS, this.bookmarked)
```
bridiver: similar to the issue with DATA_FOR_WINDOW, we're using ipc messages in a lot of places where we could (and should I think) just be using actions. The windowActions are already sent to the appStore via the dispatcher, we're just not handling them there. Ultimately this seems a little backwards to me anyway because bookmarks are part of appState, not windowState and the bookmarked state of the current url should come from there. I realize there are legacy reasons for all of this, but it would really be nice to clean some of it up.

js/lib/menuUtil.js
```diff
@@ -0,0 +1,166 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public * License, v. 2.0. If a copy of the MPL was not distributed with this file,
```
bridiver: new files should use the new directory structure - see https://github.com/brave/browser-laptop/blob/master/docs/directoryStructure.md

bbondy:
++

js/lib/menuUtil.js
bsclifton: sweet, didn't know about this- thanks :smile:

bridiver: yea, I guess @bbondy and I talked about it and then didn't tell anyone :)

echosa:
@bsclifton Does this PR also make reopened tabs have the same history they had when closed? https://github.com/brave/browser-laptop/issues/2702

bsclifton:
@echosa great catch- no, it does not.  It seems that #444 was used to not only track an about:history page but also restoring the history for a reopened tab.  I went ahead and re-opened https://github.com/brave/browser-laptop/issues/2702 and can take a look at that while I'm in the history code :smile:

echosa:
@bsclifton Great! Thanks! This is one of my (ever-dwindling) checklist items of "must be implemented before Brave becomes my default browser". :-D Looking forward to it! :+1:

bridiver:
it does not, but I was going to add in the necessary methods in electron to make that functionality possible so we can add it in the next few releases

PR #3211, opened 2016-08-16
Temporarily remove insertions

bbondy:
To be investigated later probably without the use of ad-div info

The branch for ad replacements will be maintained here:
https://github.com/brave/browser-laptop/tree/adReplacements

Auditors: @bridiver


bbondy:
CC @alexwykoff 
Please test upgrade case both with global settings and site specific setting.

bridiver:
my only concern was upgrade path and sounds like you're already looking into that, so ++

bbondy:
upgrade path already worked for me but will let @alexwykoff confirm.

PR #3215, opened 2016-08-16
Make targets.host UNIQUE, which auto-creates index

garvankeeley:
Done in regards to: https://github.com/brave/browser-ios/issues/310
Creating the index massively boosts performance for equality queries


garvankeeley:
Next step will be to get a new HTTPS-E file,
@diracdeltas is this a hassle?

diracdeltas:
Awesome. Making new HTTPS-E file won't be a problem; biggest hassle is I don't have my AWS 2factor device with me right now

garvankeeley:
a few days wait == sgtm! Bug has been around for many weeks, isn't super time-sensitive

PR #3128, opened 2016-08-11
URL bar autocomplete mouseover does not interfere with typed URL (fixes #3012)

willy-b:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Ran `git rebase -i` to squash commits if needed.

Fixes #3012: Autofill should not happen on URL bar unless item in autocomplete list is clicked

PR #3225, opened 2016-08-17
 URL bar autocomplete mouseover does not interfere with typed URL (fixes #3012) (take 2)

### LINKED ISSUE #3012: Autofill should not happen on URL bar unless item in autocomplete list is clicked
From support:
> When I try to paste a hyperlink or type something in I need to ensure the cursor is far away from the drop-down history in the URL address bar, otherwise it auto-fills with one of those links and then starts appending what I'm typing onto it. The auto-fill shouldn't happen unless I click on the hyperlink from the history when the drop down box appears.

Sometimes I notice this too.

willy-b:

Fixes #3012: Autofill should not happen on URL bar unless item in autocomplete list is clicked
This replaces the previous PR which was not rebased correctly and lacked a test (PR #3128 )


js/lib/eventUtil.js
```diff
@@ -3,7 +3,28 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const isDarwin = process.platform === 'darwin'
-module.exports.isForSecondaryAction = (e) =>
+module.exports.isForSecondaryAction = (e) => {
```
bbondy: nit for future ref, skip braces when not needed is preferred

bbondy:
Wow this feels a lot better, thanks so much for the great work here.

bsclifton:
@willy-b great job w/ this! The mouse always got in the way for me, this is a very noticeable fix :smile:

PR #3241, opened 2016-08-18
Add Autofill support

### LINKED ISSUE #860: Form fill (Autofill)
Possibly by adding support for WebAutofillClient to electron

darkdh:

fix #860

auditors: @bridiver, @bbondy


bbondy:
I think this needs a rebase pls. I'll wait to review until we have new chromium and electron builds avail.

bridiver:
I don't think this is actually ready to merge yet. I'm still working with @darkdh on a few issues on the electron side and I'm pretty sure he is still working on the browser side as well.

darkdh:
@bridiver I just solved the locking issue in electron by https://github.com/brave/electron/pull/40/commits/d1991b734eca89c4ba7ca9ae6d295e3fb68b79c5. And I think browser side can be reviewed for its function aspect while I'm working on making it looks better :)

docs/state.md
```diff
@@ -169,6 +169,27 @@ AppStore
...
+      streetAddress: string,
+      city: string,
+      state: string,
+      postalCode: string,
+      country: string,
+      phone: string,
+      email: string,
+      guid: Object.<string, <string>> // map of (partition, id) used to access the autofill entry in database
+    }],
+    creditCards: [{
```
bridiver: we can't store plain text CC data in the state. It's already encrypted on the electron side so we should either store the guid and read from there or encrypt (base64 encoded) here. Preferably the former I think because we don't really need two copies.

bridiver: just FYI - this is a blocker for merge because it's a major security issue

bridiver:
Can we add some tests for this? A few things that should be covered:
1. autofill data doesn't leak from regular to private
2. add/update and delete with good input and bad input

js/contextMenus.js
```diff
@@ -374,6 +374,35 @@ function usernameTemplateInit (usernames, origin, action) {
...
+    } else if (frontendId === -5) { // POPUP_ITEM_ID_AUTOFILL_OPTIONS
+      value = 'Autofill Settings'
+    }
+    if (frontendId === -3) { // POPUP_ITEM_ID_SEPARATOR
+      items.push(CommonMenu.separatorMenuItem)
+    } else {
+      items.push({
+        label: value,
+        click: (item, focusedWindow) => {
+          ipc.send('autofill-selection-clicked', frame.get('tabId'), value, frontendId, i)
```
bridiver: can we get rid of the ipc message and do a window action here that calls `webContents.autofillSelect` in frames.js? Maybe track the autofill status in the frame state and show/hide the context menu based on that? Could definitely be done in a follow-up commit, but we will have to refactor the ipc eventually anyway when moving webContents mgmt to the browser process

docs/state.md
darkdh: Fixed in 90f62be6c22934e55142a1e2b6a0981854e58a1c. Thanks.

bbondy:
@bridiver please review again but it seems like the credit card thing is resolved so I'm going to merge so we can get 0.11.6 testing in on it for the beta3 build.

darkdh:
Tests added in ec95567 and a116bbb and I will discard ipc in https://github.com/brave/browser-laptop/issues/3358

PR #3253, opened 2016-08-18
History follow up

bsclifton:
Follow up for https://github.com/brave/browser-laptop/pull/3206

cc: @bridiver @bbondy 

The only thing not addressed are the IPC calls. I definitely agree we can move:
- closed frame info
- active frame location

and more into AppStore (and then we'd just use actions). Those changes are "bigger than a breadbox" so I left them out of this PR.

Here's what's fixed:
- about:history now supports context menu (shares code w/ bookmarks)
- added isBookmark to siteUtil and updated code to use it
- fixed comment on UPDATE_MENU_BOOKMARKED_STATUS
- Delete history item now works
- Broke out common styles (between history and bookmarks about pages) to siteDetail.less, updated both to use it
- Sending menu update (via IPC) only sends what's needed; needs to be refactored to live in appState
- Updated directory structure for menu, menuUtil, and commonMenu per https://github.com/brave/browser-laptop/blob/master/docs/directoryStructure.md

### DIFF
--- a/app/browser/lib/menuUtil.js
+++ b/app/browser/lib/menuUtil.js
@@ -3,12 +3,12 @@
 
 'use strict'
 
-const CommonMenu = require('../commonMenu')
-const messages = require('../constants/messages')
-const siteTags = require('../constants/siteTags')
-const eventUtil = require('./eventUtil')
-const siteUtil = require('../state/siteUtil')
-const locale = require('../../app/locale')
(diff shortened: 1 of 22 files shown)


bbondy:
would you mind rebasing this? It looks like there is a conflict to merge.  I still haven't reviewed though btw.

bsclifton:
@bbondy done :smile:

bbondy:
++

PR #3264, opened 2016-08-19
Ledger pre beta updates

mrose17:
1. consolidate code for setting bravery properties
2. turn off ledger code when disabled
3. remove `enable` as an entry point, it's redundant!

Auditor @diracdeltas


app/ledger.js
```diff
@@ -209,18 +209,8 @@ if (ipc) {
...
-      var amount = parseInt(value, 10)
-      var bravery = client.getBraveryProperties()
-
-      if (isNaN(amount) || (amount <= 0)) return
-
-      underscore.extend(bravery.fee, { amount: amount })
-      client.setBraveryProperties(bravery, callback)
-      getPaymentInfo()
-    }
+    if (key === settings.PAYMENTS_ENABLED) return initialize(value)
```
diracdeltas: these lines in app/index.js should be removed, right?
```
    ipcMain.on(messages.LEDGER_ENABLE, (e, enabled) => {
      ledger.enable(enabled)
    })
```

otherwise

app/ledger.js
```diff
@@ -354,8 +344,14 @@ eventStore.addChangeListener(() => {
  * module initialization
  */
 
-var initialize = () => {
-  enable(getSetting(settings.PAYMENTS_ENABLED))
+var initialize = (onoff) => {
+  enable(onoff)
```
diracdeltas: is there any reason to keep `enable` as a module entry point? when would it get called independently of initialize?

app/ledger.js
mrose17: agreed. i'll remove that.

app/ledger.js
mrose17: not at all! good catch.

diracdeltas:
thanks

PR #3273, opened 2016-08-20
guard against race condition

mrose17:
`run` uses `setTimeout` to invoke `client.sync()` in the future.

but if the user disables the ledger before then, hilarity ensues.

this patch prevents that from happening…

auditor: @diracdeltas or @bbondy

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -812,7 +812,7 @@ var roundtrip = (params, options, callback) => {
 var run = (delayTime) => {
   if ((typeof delayTime === 'undefined') || (!client)) return
 
-  var state
+  var active, state
   var ballots = client.ballots()
   var siteSettings = appStore.getState().get('siteSettings')
   var winners = ((synopsis) && (ballots > 0) && (synopsis.winners(ballots))) || []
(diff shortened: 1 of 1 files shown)


app/ledger.js
```diff
@@ -838,7 +838,10 @@ var run = (delayTime) => {
     delayTime = client.timeUntilReconcile()
     if (delayTime === false) delayTime = 0
   }
-  if (delayTime > 0) return setTimeout(() => { if (client.sync(callback) === true) return run(0) }, delayTime)
+  if (delayTime > 0) {
+    active = client
```
diracdeltas: i don't think `active` is necessary. you can just check `if (client && client.sync) ...`

mrose17: the original bug was caused by going from 'on -> off' and then having the timer fire.

if someone does 'on -> off -> on', then client has changed, and when the timer fires, the old client is used..

that's why there's a check that the client instance hasn't changed.  your thinking?

diracdeltas: makes sense. does this still need to check that `client` is not null in the timeout?

mrose17:
I don't think it is possible for `active` to be non-`null`

diracdeltas:
i'd still rather that check be there just to be safe, since the error crashed the browser for me

PR #3281, opened 2016-08-20
Updates to history page

bsclifton:

There is a bare-bones history page in place. This PR updates the styles to make it look more "Brave-like" (borrowing from about:preferences), adds the ability to search history, and groups
history items by date. All strings are localized and ready for translation.

cc: @bradleyrichter @bbondy 

Specific changes from commit log:
- Style refresh
- Page now breaks history items down by date (sorted desc)
- shortened location shown to "domain.tld"
- page is now localized (including formatting dates)
- Added working "Clear browsing data button"
- History is now searchable

Fixes https://github.com/brave/browser-laptop/issues/441

### DIFF
--- a/app/extensions/brave/about-history.html
+++ b/app/extensions/brave/about-history.html
@@ -7,7 +7,7 @@
     <meta charset="utf-8">
     <meta name="availableLanguages" content="">
     <meta name="defaultLanguage" content="en-US">
-    <meta name='theme-color' content='#035500'>
+    <meta name='theme-color' content='#565656'>
     <link rel="shortcut icon"type="image/x-icon" href="data:image/x-icon;,">
     <title data-l10n-id="historyTitle"></title>
     <script src='js/about.js'></script>
--- a/app/extensions/brave/locales/en-US/history.properties
+++ b/app/extensions/brave/locales/en-US/history.properties
@@ -1,2 +1,8 @@
 historyTitle=History
 history=History
+clearBrowsingDataNow=Clear browsing data
+removeSelectedItems=Remove selected items
+historySearch=Search history
+time=Time
+title=Title
+domain=Domain
--- a/js/about/history.js
+++ b/js/about/history.js
@@ -4,11 +4,15 @@
 
 // Note that these are webpack requires, not CommonJS node requiring requires
 const React = require('react')
-const Immutable = require('immutable')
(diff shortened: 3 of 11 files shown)


bsclifton:
Now for some screenshots :smile: 

Please review and give me feedback; I can update the styles and take care of any asks you may have.

Here's a shot of the regular screen:
screen shot 2016-08-20 at 1 39 05 pm

When you click `clear browsing data`, it pops up the modal:
screen shot 2016-08-20 at 1 39 19 pm

You can search the results; one thing to note, it shows the time next to each entry, but does NOT group it by day. I need to fix this and would like to hear recommendations about how to show results. For sure, we'd want to say "X results found" (which is not currently shown)
screen shot 2016-08-20 at 1 39 40 pm

Highlight color is no longer BLUE (like the bookmarks page). Instead, it borrows from SortableTable (see preferences > search for an example).
screen shot 2016-08-20 at 1 40 12 pm

Context menu works great. Anything else we'd like to see here?
screen shot 2016-08-20 at 1 40 30 pm

bsclifton:
Still working through some look and feel updates per notes from @bradleyrichter. Here is the proposed mockup:
screen shot 2016-08-21 at 12 23 32 pm

This PR is fairly close to that look and feel right now.  The content for each day is now stored in  SortableTable control. This control is also used for the default search engines in preferences, so we need to confirm that still works as expected.  

My next step: look into creating tests for this new functionality.

The remaining work before it matches 100% (which can be broken out to separate issues):
- delete icon needs to be added to table header (for time column)
- hover style needs to be updated to match spec
- hover style needs to show the X to delete history entries
- style updates for SortableTable header (gradient, orange column stripe, up/down arrow)
- "jump" functionality (shown on the day view)

bsclifton:
Ready for re-review :smile:

bradleyrichter:
Of course, styles can be refined. I'll poke around a bit. But looking good so far!

The one question I have for now is if we can live with only one column header at the top? This would become the "sort captain" and all sub-tables would follow. This should cut down the redundancy and still provide the intended function.

bsclifton:
@bradleyrichter a single table header... that's an interesting idea.  It would take some extra work, but I think it's feasible

bbondy:
going to open off of a shared branch

Issue #3001, opened 2016-08-05
New Tab Design: Overview with subtasks (about:newtab)


bradleyrichter:
When the user makes a new tab, a set of useful information will be displayed which includes their Top-Sites, a digital clock showing their local time, and a cumulative readout of ads/trackers blocked over time. This info is presented over photo backgrounds which are sourced from a creative commons image database of earth surface imagery.

First version features:
- Image background, selected at random from source cache which is loaded from a remote server at run time. (random, but different than currently displayed image)
- top-sites large-image thumbnails with BG preview feature
- top sites grid-toggle button to show 6, 12, or 18 squares
- Cumulative Bravery interesting data - Trackers, Ads blocked, and estimated loading time saved
- Time clock that shows the current regional time for the user
- "toolbar" with shortcuts to several Brave about:pages including prefs, bookmarks and history

Later features:
- ability to select from 3-5 background themes of different images https://github.com/brave/browser-laptop/issues/5306
- ability to select a local volume as the BG image source https://github.com/brave/browser-laptop/issues/5305
- ability to customize the top-sites area to show source sites from a bookmark folder https://github.com/brave/browser-laptop/issues/5304
- ability to change the view from thumbnail squares to a list view https://github.com/brave/browser-laptop/issues/5303
- add local temp/weather to time display https://github.com/brave/browser-laptop/issues/5302

![image](https://cloud.githubusercontent.com/assets/13509546/17451103/126e495c-5b19-11e6-838a-9e4742156942.png)

1.0 ENG tasks:
- build data collector for "Bravery blocking reports", and add time estimation
- build high-res favicon collector mechanism, pulled from page header at domain root page
  - create local cache for large favicons 
  - add fall-back mechanism when large favicon is not present for a given site
- build the top-sites list creation mechanism using a blend of freceny and other means which may be adjusted over time to improve the results
- build the BG images retrieval mechanism, with smart randomizer to eliminate the same image twice in a row (images will be stored and pulled from an AWS volume)
- interface with tab-preview mechanism to be utilized for previewing the topsites on mouse-over. (replaces BG image during mouseover and uses the same delay approach as tab-previews)
- build data connector for local-time display
- connect new-tab to prefs

1.0 Front-end tasks
- arrange BG image in CSS to zoom-to-width or height with above-centered anchor point
- create separate black gradient mask in layer over BG image
- create adjustable table layout for top-site buttons with high-res favicons
- create CSS for Bravery info display
- create CSS for live Clock display
[...]

PR #3309, opened 2016-08-22
Add *.jpg extension to webpack file-loader

cezaraugusto:
Currently I can't call `*.jpg` images directly inside `*.less` files. 

Adding this will be necessary in a near future when [background-preview](https://github.com/brave/browser-laptop/issues/3001#issue-169700482) is implemented.


bbondy:
++

PR #3315, opened 2016-08-22
Improved UI/UX of find bar

### LINKED ISSUE #3159: Improved UI/UX of find bar

jkup:

Fix #3159

Sorry, I got my git history in a really bad state. This is the same UI/UX PR as #3255 but I fixed the bugs outlined by @bbondy.

The last thing I know to need work is the design of the buttons when tab color is in effect. Screenshots posted below. @bradleyrichter can you give me a little feedback on the buttons?
screen shot 2016-08-22 at 1 58 07 pm
screen shot 2016-08-22 at 1 58 16 pm

### DIFF
--- a/js/actions/webviewActions.js
+++ b/js/actions/webviewActions.js
@@ -83,7 +83,7 @@ const webviewActions = {
     }
   },
 
-  findInPage (searchString, caseSensitivity, forward, webview) {
+  findInPage: function (searchString, caseSensitivity, forward, webview) {
     webview = webview || getWebview()
     if (!webview) {
       return
@@ -100,7 +100,7 @@ const webviewActions = {
     }
   },
 
-  stopFindInPage (webview) {
+  stopFindInPage: function (webview) {
     webview = webview || getWebview()
     if (!webview) {
       return
--- a/js/actions/webviewActions.js
+++ b/js/actions/webviewActions.js
@@ -81,6 +81,31 @@ const webviewActions = {
         webview.executeJavaScript('document.webkitRequestFullscreen()')
       }
     }
+  },
+
+  findInPage (searchString, caseSensitivity, forward, webview) {
+    webview = webview || getWebview()
+    if (!webview) {
+      return
+    }
+
+    if (searchString) {
+      webview.findInPage(searchString, {
+        matchCase: caseSensitivity,
+        forward: forward !== undefined ? forward : true,
+        findNext: forward !== undefined
+      })
+    } else {
+      webview.stopFindInPage('clearSelection')
+    }
+  },
+
+  stopFindInPage (webview) {
+    webview = webview || getWebview()
+    if (!webview) {
+      return
+    }
+    webview.stopFindInPage('keepSelection')
   }
 }
 
--- a/js/actions/webviewActions.js
+++ b/js/actions/webviewActions.js
@@ -83,7 +83,7 @@ const webviewActions = {
     }
   },
 
-  findInPage: function (searchString, caseSensitivity, forward, webview) {
+  findInPage: function (searchString, caseSensitivity, forward, findNext, webview) {
     webview = webview || getWebview()
     if (!webview) {
       return
@@ -92,8 +92,8 @@ const webviewActions = {
     if (searchString) {
       webview.findInPage(searchString, {
         matchCase: caseSensitivity,
-        forward: forward !== undefined ? forward : true,
-        findNext: forward !== undefined
+        forward,
+        findNext
       })
     } else {
       webview.stopFindInPage('clearSelection')
--- a/js/components/findbar.js
+++ b/js/components/findbar.js
@@ -7,13 +7,16 @@ const ImmutableComponent = require('./immutableComponent')
 const Immutable = require('immutable')
 const keyCodes = require('../constants/keyCodes')
 const Button = require('./button.js')
+const SwitchControl = require('../components/switchControl')
 const windowActions = require('../actions/windowActions')
 const windowStore = require('../stores/windowStore')
+const {getTextColorForBackground} = require('../lib/color')
 
 class FindBar extends ImmutableComponent {
   constructor () {
     super()
     this.onBlur = this.onBlur.bind(this)
+    this.onClear = this.onClear.bind(this)
     this.onKeyDown = this.onKeyDown.bind(this)
     this.onChange = this.onChange.bind(this)
     this.onFindPrev = this.onFindPrev.bind(this)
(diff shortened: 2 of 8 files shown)


bradleyrichter:
@jkup Getting close. Here is a spec vs code snap shot and a short list of changes that should get us there:

- increase bar height to make room for taller search box 
- increase search box height to 24px
- change prev/next buttons to show outline on hover only
- set hover outline height for prev/next to 24 and align to search box
- set all radii to 4px

for colored tab theming:
- set hover outline color, prev/next arrow color, and label color to #DCDCDC

image

bbondy:
let me know when I can take another look @jkup

jkup:
Cool I'm just gonna get these last design tweaks and then I'll ping you!

bbondy:
This is form my original feedback but the text is glued to the focus ring, could we add some padding to avoid that?  I see some black text touching the blue.

screenshot 2016-08-23 00 18 29

bradleyrichter:
Actually, the blue highlight is in the wrong place. It should be changing the outer box border style, not the inside text field.

jkup:
@bbondy I had fixed that but there was a typo in my CSS. The padding should be there now!

bbondy:
It looks like the count not updating isn't fixed yet either. Steps to reproduce was in my previous feedback. (Open search, count updates, close, open again search, count doesn't update)

bbondy:
I think it would be nice to also use getTextColorForBackground for the previous/next button.  On some themes I don't think it looks right (example brianbondy.com)
screenshot 2016-08-23 00 26 36
Both for the outline and the arrow

bradleyrichter:
we can do that instead of what I proposed.

But the outlines are only for mouse over. But in general, everything can use the GetTextColor for the theme-color tab style.

bbondy:
the right x button as well, please use getTextColorForBackground

jkup:
On it!

js/components/findbar.js
```diff
@@ -60,6 +63,18 @@ class FindBar extends ImmutableComponent {
     input.select()
   }
 
+  componentWillMount () {
+    let stylesheet = document.getElementById('findBarStylesheet')
```
bbondy: Honestly I'd probably get around this problem just by changing the hover state, it's probably not worth having this to spec for the amount of time it can take to do it properly.

However if you want to, I think the proper way is to handle mouseEnter and moueLeave and set state in the `findDetail` for if it is hovered or not.  Then render based on that state.

bbondy: Please do get rid of this though ;)

js/actions/webviewActions.js
```diff
@@ -81,6 +81,31 @@ const webviewActions = {
...
+        matchCase: caseSensitivity,
+        forward: forward !== undefined ? forward : true,
+        findNext: forward !== undefined
+      })
+    } else {
+      webview.stopFindInPage('clearSelection')
+    }
+  },
+
+  stopFindInPage (webview) {
```
bridiver: does anything actually call this?

jkup: It looks like that's called by onFind in frame.js https://github.com/brave/browser-laptop/blob/master/js/components/frame.js#L1024

bridiver: wouldn't this be `webviewActions.stopFindInPage` ?

bbondy: I didn't refactor that way because `webview.stopFindInPage` takes a different parameter currently and I wanted to keep it the same as what it was.

bridiver: that makes sense, I'm just trying to figure out where this actually gets called and I'm not seeing a call to `webviewActions.stopFindInPage`

bridiver: fyi @bbondy - the problem with this turned out to be the result of calling stopFindInPage on hide. `keepSelection` doesn't actually keep any state in the webcontents, it just keeps the selection highlighted. If you want to keep the find state you can't call `stopFindInPage`

js/actions/webviewActions.js
```diff
@@ -81,6 +81,31 @@ const webviewActions = {
         webview.executeJavaScript('document.webkitRequestFullscreen()')
       }
     }
+  },
+
+  findInPage (searchString, caseSensitivity, forward, webview) {
```
bridiver: I actually didn't even know this was valid syntax, but for consistency (and syntax highlighting) I think we should keep the `:`

bbondy: that syntax wasn't intentional, I'll fix.

js/actions/webviewActions.js
bbondy: webviewActions.stopFindInPage isn't needed and not called, will remove.

js/components/main.js
```diff
@@ -620,6 +623,16 @@ class Main extends ImmutableComponent {
     windowActions.setUrlBarActive(false)
   }
 
+  onFindHide () {
+    const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
+    windowActions.setFindbarShown(activeFrame, false)
+    webviewActions.stopFindInPage()
```
bbondy: @bridiver this had to be called here to retain the existing behaviour. 
`stopFindInPage` needs to be called to clear the highlight but to keep the current match in case you `command+g`.

bridiver: the problem with this is that calling `stopFindInPage` resets the webcontents find state so you can't get back to the same search results

bbondy: but it must keep it because it remembers the next position?

bbondy: I think btw the webviewAction was used at one point and was committed out.

bridiver: it doesn't remember the next position, that was the bug. I started a slack conversation to explain

bbondy: We can't remove the call because when we do the yellow highlight stays forever

bbondy:
@alexwykoff I'm going to merge this after all since it's ready.

PR #3336, opened 2016-08-23
Bookmarked history entries now show context menu in about:history

bsclifton:

Fixes https://github.com/brave/browser-laptop/issues/3330

### DIFF
--- a/js/state/siteUtil.js
+++ b/js/state/siteUtil.js
@@ -155,12 +155,15 @@ module.exports.removeSite = function (sites, siteDetail, tag) {
     return sites
   }
   const tags = sites.getIn([index, 'tags'])
-  // If there are no tags and the removeSite call was called without a specific tag
-  // then remove the entry completely.
   if (tags.size === 0 && !tag) {
-    sites = sites.splice(index, 1)
-    return sites
+    // If called without tags and entry has no tags, remove the entry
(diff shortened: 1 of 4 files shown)


js/state/siteUtil.js
```diff
@@ -155,12 +155,15 @@ module.exports.removeSite = function (sites, siteDetail, tag) {
...
   if (tags.size === 0 && !tag) {
-    sites = sites.splice(index, 1)
-    return sites
-  }
+    // If called without tags and entry has no tags, remove the entry
+    return sites.splice(index, 1)
+  } else if (tags.size > 0 && !tag) {
+    // If called without tags BUT entry has tags, null out lastAccessedTime.
+    // This is a bookmark entry that we want to clear history for (but NOT delete/untag bookmark)
+    return sites.setIn([index, 'lastAccessedTime'], null)
```
bbondy: hah I was going to suggest that as the fix, thanks.

bbondy:
++

bridiver:
I don't think this is a complete fix. We are sending everything in `sites` to `about:history` and that seems incorrect to me. We should only be sending actual history items.

bbondy:
The sites are the last 500 history we keep, when last access time is defined, but there's a follow up for it as a micro optimization to exclude those.

PR #3345, opened 2016-08-23
create leveldb for httpse

bridiver:
cc @diracdeltas

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,8 @@
     "jsonfile": "^2.2.3",
     "less": "^2.5.3",
     "less-loader": "^2.2.1",
+    "level": "^1.4.0",
+    "leveldown": "^1.4.6",
     "mkdirp": "^0.5.1",
     "mocha": "^2.3.4",
     "mockery": "^1.7.0",
--- a/preload-httpse.js
+++ b/preload-httpse.js
@@ -7,8 +7,10 @@
 
 'use strict'
 var fs = require('fs')
+var path = require('path')
 var parseString = require('xml2js').parseString
 var sqlite3 = require('sqlite3')
+var levelup = require('level')
 
 // Manually exclude sites that are broken until they are fixed in the next
 // HTTPS Everywhere release.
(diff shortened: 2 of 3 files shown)


preload-httpse.js
```diff
@@ -72,3 +74,77 @@ var db = new sqlite3.Database('httpse.sqlite', function (err) {
...
+console.log('creating httpse.leveldb')
+rmDir('./httpse.leveldb')
+
+const httpseLevelDB = levelup('httpse.leveldb', {compression: false, errorIfExists: true})
+
+const ruleSets = {}
+for (var id in rulesets.rulesetStrings) {
+  ruleSets[id] = ruleSets[id] || []
+  ruleSets[id].push(rulesets.rulesetStrings[id])
+}
```
diracdeltas: would `ruleSets[id]` ever have length > 1?

preload-httpse.js
```diff
@@ -72,3 +74,77 @@ var db = new sqlite3.Database('httpse.sqlite', function (err) {
...
+      } else {
+        rmDir(filePath)
+      }
+    }
+  }
+  fs.rmdirSync(dirPath)
+}
+
+console.log('creating httpse.leveldb')
+rmDir('./httpse.leveldb')
```
diracdeltas: please add httpse.leveldb to .gitignore

preload-httpse.js
```diff
@@ -72,3 +74,77 @@ var db = new sqlite3.Database('httpse.sqlite', function (err) {
...
+for (var id in rulesets.rulesetStrings) {
+  ruleSets[id] = ruleSets[id] || []
+  ruleSets[id].push(rulesets.rulesetStrings[id])
+}
+
+let batch = httpseLevelDB.batch()
+for (var target in rulesets.targets) {
+  let targetRuleSets = []
+  rulesets.targets[target].forEach((id) => {
+    ruleSets[id].forEach((ruleset) => {
```
diracdeltas: lgtm! eventually the block below should be merged into https://github.com/brave/browser-laptop/blob/3de38bfeec25758f98c556f73c5839a95171c2de/preload-httpse.js#L33-L42 so that the perf benefits apply to the JSON version of the db as well

preload-httpse.js
bridiver: some IDs appear to have multiple entries, but I can double-check

diracdeltas: i thought `rulesets.rulesetStrings` was just a map of ID to a string

bridiver: yea, I think I just got something mixed up here

package.json
```diff
@@ -133,6 +133,8 @@
     "jsonfile": "^2.2.3",
     "less": "^2.5.3",
     "less-loader": "^2.2.1",
+    "level": "^1.4.0",
+    "leveldown": "^1.4.6",
```
diracdeltas: is leveldown needed here? i thought level installed both levelup and leveldown

bridiver: https://github.com/Level/levelup#leveldown

diracdeltas:
++ thanks

package.json
diracdeltas: > The level package is available as an alternative installation mechanism. Install it instead to automatically get both LevelUP & LevelDOWN.

bridiver: I see now, I didn't realize there was a `level`, `levelup` and `leveldown`

PR #3351, opened 2016-08-23
update for emscripten and ledger

mrose17:
auditor: @diracdeltas 

1. build for node-anonize2-relic-emscripten to work with asar packaging

2. poll for unconfirmed balance and better synchronization

3. use clientOptions.verboseP for some calls to console.log

4. call updatePublisherInfo at the correct places


diracdeltas:
for future reference, please put a summary of the commit in the first line and auditor in the last line

mrose17:
sorry, i messed that up in the GUI...

PR #3357, opened 2016-08-24
different paths for MacOS v. Linux/Windows

mrose17:
1. use different install paths for MacOS and Linux/Windows for the `anonize2.js.mem` file

2. if `bitcoin:` URI handler not defined, then don't define `ledgerInfo.paymentURL`

auditor: @bbondy

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -711,9 +711,7 @@ var updateLedgerInfo = () => {
     underscore.extend(ledgerInfo, ledgerInfo._internal.cache || {})
 
     if (typeof protocolHandler.isNavigatorProtocolHandled === 'function') {
-/* YAN: this comment is temporary until the preferences panel for payments can handle only one of these properties defined
-      delete ledgerInfo[protocolHandler.isNavigatorProtocolHandled('', 'bitcoin') ? 'buyURL' : 'paymentURL']
- */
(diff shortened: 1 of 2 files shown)


bbondy:
@diracdeltas pls review, but I'm going to merge it in anyway because packaging is broken on master.

app/ledger.js
```diff
@@ -711,9 +711,7 @@ var updateLedgerInfo = () => {
     underscore.extend(ledgerInfo, ledgerInfo._internal.cache || {})
 
     if (typeof protocolHandler.isNavigatorProtocolHandled === 'function') {
-/* YAN: this comment is temporary until the preferences panel for payments can handle only one of these properties defined
-      delete ledgerInfo[protocolHandler.isNavigatorProtocolHandled('', 'bitcoin') ? 'buyURL' : 'paymentURL']
- */
+      if (!protocolHandler.isNavigatorProtocolHandled('', 'bitcoin')) delete ledgerInfo.paymentURL
```
diracdeltas: isn't paymentURL useful even if Brave isn't the bitcoin handler? ex: your bitcoin handler is another desktop app.

diracdeltas: if it's not useful then why show the QR code?

mrose17: some people keep the wallet on another device and use the camera on that device...

willy-b: I scan the desktop bitcoin QRs from my phone usually...

PR #3383, opened 2016-08-24
Wednesday edits

mrose17:
found that if the network was *really* slow, `run` might get invoked before the wallet got created.

auditor: @diracdeltas

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -867,7 +867,11 @@ var run = (delayTime) => {
   }
 
   if (delayTime === 0) {
-    delayTime = client.timeUntilReconcile()
+    try {
+      delayTime = client.timeUntilReconcile()
+    } catch (ex) {
+      delayTime = random.randomInt({ min: 1, max: 10 * msecs.minute })
+    }
     if (delayTime === false) delayTime = 0
   }
   if (delayTime > 0) {


app/ledger.js
```diff
@@ -867,7 +867,11 @@ var run = (delayTime) => {
   }
 
   if (delayTime === 0) {
-    delayTime = client.timeUntilReconcile()
+    try {
+      delayTime = client.timeUntilReconcile()
+    } catch (ex) {
+      delayTime = random.randomInt({ min: 1, max: 10 * msecs.minute })
```
diracdeltas: if this random delay happens to be small enough, is it still possible for `run` to be invoked before wallet initialization? i'm not understanding how this change fixes the problem stated in the PR

mrose17:
The patch catches the error that may be thrown and then endures a non-zero wait before the next call to sync, which ultimately results in a call to run

diracdeltas:
> endures a non-zero wait before the next call to sync, which ultimately results in a call to run

say that the wait is 1ms. will that cause a race condition problem given slow network speeds?

mrose17:
Race condition? No.  Inefficient? Yes.

Tell me the lower bound that you like... Thanks!

diracdeltas:
this is fine as long as it's not going to throw another error. for future reference, please put a longer explanation in the commit message itself (so someone looking through git log can understand the change without having to reference PRs).

mrose17:
+1

diracdeltas:
cherry-picked

PR #3393, opened 2016-08-25
don't send unnecessary ipc messages

bridiver:
auditors: @bbondy

### DIFF
--- a/js/about/aboutActions.js
+++ b/js/about/aboutActions.js
@@ -4,6 +4,8 @@
 
 const messages = require('../constants/messages')
 const serializer = require('../dispatcher/serializer')
+const WindowConstants = require('../constants/windowConstants')
+const AppConstants = require('../constants/appConstants')
 const ipc = window.chrome.ipc
 
 const AboutActions = {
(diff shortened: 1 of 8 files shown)


js/about/aboutActions.js
```diff
@@ -173,7 +214,10 @@ const AboutActions = {
    * @param {object} card - credit card to remove as per doc/state.md's autofillCreditCardDetail
    */
   removeAutofillCreditCard: function (card) {
-    ipc.send(messages.REMOVE_AUTOFILL_CREDIT_CARD, card)
+    AboutActions.dispatchAction({
+      actionType: AppConstants.APP_REMOVE_CREDIT_CARD,
```
darkdh: This should be `APP_REMOVE_AUTOFILL_CREDIT_CARD`

bridiver:
this takes care of a lot of the low-hanging fruit, but there is still a lot more cleanup to do

alexwykoff:
from #testers : 

> yea, we should test that the functionality still works correctly. Basically all the pink sections in app/index.js
> setting change, site setting change, resource enabled, move bookmark, delete password and remove autofill address/credit card
> I checked most of them, but only on mac
> change setting would be most preferences. site setting would be brave menu settings
> I believe resource enabled would be things like ad blocking at the global level (edited)
> others should be self explanatory

Changes from app/index.js:

1. messages.CHANGE_SETTING
2. messages.CHANGE_SITE_SETTING
3. messages.REMOVE_SITE_SETTING
4. messages.SET_RESOURCE_ENABLED
5. messages.MOVE_SITE
6. messages.DELETE_PASSWORD
7. messages.DELETE_PASSWORD_SITE
8. messages.CLEAR_PASSWORDS
9. messages.REMOVE_AUTOFILL_ADDRESS
10. messages.REMOVE_AUTOFILL_CREDIT_CARD

Issue #3376, opened 2016-08-24
Preferences : Payments : Create wallet control text  and status flickers when disabling then enabling brave payments.


alexwykoff:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
After creating a wallet, the 'create wallet' button text changes to 'add funds'. If I toggle the 'Enable' control, the text on the control and status flickers when the wallet is re-enabled.

**Expected behavior:**
The text should remain 'add funds' and neither the control or status text should flicker.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Tested on OS X

- Brave Version:
0.11.6 pre-beta3

- Steps to reproduce:
    1. Enable the Brave wallet.
    2. Create a Brave wallet, observe the text on the create wallet control text.
    3. Disable the wallet, observe the text on the add funds control.
    4. Enable the wallet, observe the text on the create wallet/add funds control.

- Screenshot if needed:
![create_wallet_flicker](https://cloud.githubusercontent.com/assets/1434040/17937068/7d7a94d6-69e6-11e6-85f2-045bda399024.gif)

- Any related issues:

bridiver:
polish? Is that related?

bridiver:
reopen after reverting commit

alexwykoff:
has this been unreverted? in the latest master it appears to be functioning appropriately.
![ledger_latest_master](https://cloud.githubusercontent.com/assets/1434040/17982786/181a098e-6acf-11e6-9607-7cf5dcc8a712.gif)

mrose17:
it was reverted. i think the issue is timing. the fix that got reverted "could not work"... i will consult with @diracdeltas on what can be done.

bridiver:
the commit addressed the flickering during wallet creation and correctly transitioned from "create wallet" to "creating wallet..." to "add funds", but is there another problem as well? Your screenshot does not look correct because it stays on "create wallet" after clicking

diracdeltas:
Just to clarify, there are two separate (but probably related) issues that got merged into this one. The one in Alex's first screenshot (wallet button flicker on creation) i could not repro. The one in Alex's second screenshot, which is the one @bridiver fixed and then reverted, happens consistently. I'm leaving this issue open to address the latter.

bridiver:
the behavior is actually still inconsistent with my change. Sometimes it works correctly and sometimes it doesn't

mrose17:
[...]

PR #3432, opened 2016-08-25
fix 'Create Wallet' flicker

mrose17:
as reported in https://github.com/brave/browser-laptop/issues/3376

auditor: @diracdeltas


diracdeltas:
+1

PR #3444, opened 2016-08-26
Fixes extra separator in about:history context menu

bsclifton:
image

For some reason it was only showing on Windows

Auditors: @luixxiul


bbondy:
macOS suppresses dupes

PR #3445, opened 2016-08-26
Set bookmark url if no title present fixes #3442

### LINKED ISSUE #3442: If a site does not have a title, the bookmark should default to the url instead.
Yes

When attempting to bookmark http://alexwykoff.com/2/demo/ the url did populate the appropriate entry in the modal, however the Save button was not enabled. 

**Expected behavior:**
If the modal contains sufficient data, the Save button should be enabled. 

Tested on OS X

- Brave Version:
0.11.6 pre-beta4

- Steps to reproduce:
    1. Open the latest build
    2. Visit a site with no title, like http://alexwykoff.com/2/demo/
    3. Attempt to bookmark the page, observe the modal.
    4. Click on the Save button.

- Screenshot if needed:

- Any related issues:
https://github.com/brave/browser-laptop/issues/3344

Sh1d0w:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist. #3442 

cc @bridiver

### DIFF
--- a/js/components/addEditBookmark.js
+++ b/js/components/addEditBookmark.js
@@ -32,33 +32,18 @@ class AddEditBookmark extends ImmutableComponent {
   get isBlankTab () {
     return ['about:blank', 'about:newtab'].includes(this.props.currentDetail.get('location'))
   }
-
   get bookmarkNameValid () {
     let title = (this.props.currentDetail.get('title') || this.props.currentDetail.get('customTitle'))
     return ((typeof title === 'string') && title.trim().length > 0)
   }
-
   get isFolder () {
     return siteUtil.isFolder(this.props.currentDetail)
   }
   updateFolders (props) {
     this.folders = siteUtil.getFolders(this.props.sites, props.currentDetail.get('folderId'))
   }
-  setDefaultTitle (props) {
-    // Set the default bookmark title to be the url
-    // if the current title is empty during load of the component
-    let title = props.currentDetail.get('title')
-    let location = props.currentDetail.get('location')
-
-    if ((typeof title === 'string') && title.trim().length === 0) {
-      this.onNameChange({target: {
-        value: location
-      }})
-    }
-  }
   componentWillMount () {
     this.updateFolders(this.props)
-    this.setDefaultTitle(this.props)
   }
   componentWillUpdate (nextProps) {
     if (this.props.sites !== nextProps.sites) {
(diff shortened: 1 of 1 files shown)


bsclifton:
I think it's important to keep the title as empty. Other parts of the code reference that (about:history, long pressing on the back button to show history).

Our bookmarks code is smart enough to try to render using (in this order):
1) customTitle
2) title
3) location (URL)

Would we be able to tweak this instead to set the URL as a placeholder on the the input box (and then make OK button enabled?)  I think this would be the best fix

bridiver:
chrome uses the url as the title, but I definitely see the problem now with my change. If you try to create an empty title it will revert to the url

bridiver:
cc @bradleyrichter

Sh1d0w:
@bridiver @bsclifton I have changed the logic there. Now the button will be disabled only if you try to create folder with an empty name.

For bookmarks, it will default to the website title or url as a placeholder text just like @bsclifton  suggested.

There is one small problem tho and it is in this file:

https://github.com/brave/browser-laptop/blob/master/js/state/siteUtil.js#L112

As you can see, once bookmarked forever bookmarked, so if you bookmarked one website with `customTitle` then delete it and try to bookmark it again, the placeholder text may be different from what will be saved as default. How do you prefer to handle this? Do you want me to check for old custom title to display it in the placeholder text?

bbondy:
When I add a new bookmark now it shows this:
screenshot 2016-08-27 00 10 32

That's not ideal because it shows the title as placeholder text and doesn't allow you to edit it.

I think we should just leave it blank and close this. Having the url as placeholder text kind of makes sense because that's the value you'll see when it's created, but it is kind of confusing too because it's like you are suggesting a url as the title.

Sh1d0w:
@bbondy Ok I agree, but there is a bug introduced by @bridiver with this commit

https://github.com/brave/browser-laptop/commit/fd60bac1fa0d3282a964f02471e199d8d62e8c10#diff-92301445f9beebf63f7a7ea1cc1a541fR106

Can you please look at my comment here https://github.com/brave/browser-laptop/issues/3442

And let me know if you want me to revert this so it can work like it was working before?

bbondy:
yep sorry I was DM'ing with @bridiver on Slack about that, he's reverting that too but allowing save with a blank title.

bsclifton:
Being able to save with a blank title is ideal, since the renderer can choose what to show

Thanks for working through this one with us, @Sh1d0w :smile:

PR #3447, opened 2016-08-26
Add option to allow mixed content

### LINKED ISSUE #3443: Add an option to temporarily disable mixed content protection
Currently we cannot access resource like flash video of mixed content site (ex. https://forgetfulbc.blogspot.com/2016/08/lastweek-t3.html) and it will be upgraded to https by https everywhere so downgrade to http for this site is not working.

Firefox and Chrome provide an option to disable the protection temporarily
---
Firefox
screen shot 2016-08-26 at 13 15 34
screen shot 2016-08-26 at 13 16 48

Chrome
screen shot 2016-08-26 at 13 22 31
screen shot 2016-08-26 at 13 31 34

cc @diracdeltas

darkdh:

1. Add option to allow mixed content
2. Change icon of http and mixed content to fa-unlock

fix #3443 

Detect blocked mixed content
screen shot 2016-08-26 at 18 05 33

After allowing mixed content
screen shot 2016-08-26 at 18 13 59

Pure http site
screen shot 2016-08-26 at 18 13 45


js/components/frame.js
```diff
@@ -557,6 +558,8 @@ class Frame extends ImmutableComponent {
     this.webview.addEventListener('content-blocked', (e) => {
       if (e.details[0] === 'javascript') {
         windowActions.setBlockedBy(this.frame, 'noScript', e.details[1])
+      } else if (!this.allowRunningInsecureContent()) {
```
bridiver: we can now manage running and displaying insecure content through the `displayInsecureContent` and `runInsecureContent` content settings which will allow us to treat them like any other site settings. I know you already did a lot of work using the old method and I'm sorry I didn't catch this sooner, but setting these through the webview is problematic because it restarts the renderer process and will potentially break some sites. We had a small number of exceptions before, but if we're opening this up as a general setting we should transition to the contentSetting

darkdh: Thanks for clarification. Changes has made in 	dd8f9b4.

js/state/contentSettings.js
```diff
@@ -123,6 +127,10 @@ const getContentSettingsFromSiteSettings = (appState) => {
       addContentSettings(contentSettings.javascript, hostPattern, '*',
         hostSetting.noScript ? 'block' : 'allow')
     }
+    if (typeof hostSetting.allowActiveMixedContent === 'boolean') {
```
bridiver: I think it would be better to name this setting `runInsecureContext`. It's easier if we don't have to translate names across various parts of the app

js/stores/windowStore.js
```diff
@@ -742,6 +742,10 @@ const doAction = (action) => {
         windowState = windowState.setIn(path.concat(['security', 'isSecure']),
                                         action.securityState.secure)
       }
+      if (action.securityState.mixedContent !== undefined) {
```
bridiver: same thing as above. If this applies to both display and run then `insecureContent` would be ok, but if this is only for running then we should stick with `runInsecureContent`

js/components/frame.js
```diff
@@ -156,13 +156,13 @@ class Frame extends ImmutableComponent {
...
   shouldCreateWebview () {
-    return !this.webview || this.webview.allowRunningInsecureContent !== this.allowRunningInsecureContent() ||
-      !!this.webview.allowRunningPlugins !== this.allowRunningPlugins()
+    return !this.webview || !!this.webview.allowRunningPlugins !== this.allowRunningPlugins()
   }
 
-  allowRunningInsecureContent () {
-    let hack = siteHacks[urlParse(this.props.location).hostname]
-    return !!(hack && hack.allowRunningInsecureContent)
+  allowActiveMixedContent () {
```
bridiver: do we still need this method? I think it was only for setting the webview attribute

js/components/frame.js
```diff
@@ -574,6 +570,15 @@ class Frame extends ImmutableComponent {
         windowActions.setBlockedBy(this.frame, 'noScript', e.details[1])
       }
     })
+    this.webview.addEventListener('did-block-run-insecure-content', (e) => {
+      const hack = siteHacks[urlParse(this.props.location).hostname]
```
bridiver: I think the site hacks need to be translated into default rules for contentSettings.js. This will apply the setting after the page has already loaded and the content is blocked

darkdh: Move it to load start in cecb967

js/stores/windowStore.js
darkdh: made naming consistent in cecb967

js/state/contentSettings.js
darkdh: you're right. it will make the code more readable. addressed it in cecb967

docs/state.md
```diff
@@ -267,8 +267,8 @@ WindowStore
         realm: string
       },
       isExtendedValidation: boolean, // is using https ev
-      activeMixedContent: boolean, // has active mixed content
-      passiveMixedContent: boolean, // has passive mixed content
+      isRunInsecureConent: boolean, // has active mixed content
+      blockedRunInsecureConent: string, // source of blocked active mixed content
```
bridiver: blockedRunInsecureConent  => blockedRunInsecureContent

bridiver: also, isn't this an array of strings?

docs/state.md
```diff
@@ -267,8 +267,8 @@ WindowStore
         realm: string
       },
       isExtendedValidation: boolean, // is using https ev
-      activeMixedContent: boolean, // has active mixed content
-      passiveMixedContent: boolean, // has passive mixed content
+      isRunInsecureConent: boolean, // has active mixed content
```
bridiver: isRunInsecureContent => runInsecureContent

js/components/frame.js
```diff
@@ -768,6 +770,9 @@ class Frame extends ImmutableComponent {
       if (this.doNotTrack) {
         this.webview.executeJavaScript('Navigator.prototype.__defineGetter__("doNotTrack", () => {return 1});')
       }
+      if (hack && hack.allowRunningInsecureContent) {
+        appActions.changeSiteSetting(this.origin, 'runInsecureContent', true)
```
bridiver: this still isn't going to work correctly because the site setting won't update in time. We should remove the siteHacks for insecure content and translate them to site setting defaults

js/components/siteInfo.js
```diff
@@ -3,34 +3,51 @@
...
+const appActions = require('../actions/appActions')
+const messages = require('../constants/messages')
+const siteUtil = require('../state/siteUtil')
 
 class SiteInfo extends ImmutableComponent {
+  constructor () {
+    super()
+    this.onAllowMixedContent = this.onAllowMixedContent.bind(this)
+  }
+  onAllowMixedContent () {
```
bridiver: MixedContent => RunInsecureContent

js/components/siteInfo.js
```diff
@@ -3,34 +3,51 @@
...
   get isExtendedValidation () {
     return this.props.frameProps.getIn(['security', 'isExtendedValidation'])
   }
   get isSecure () {
     return this.props.frameProps.getIn(['security', 'isSecure'])
   }
-  get isMixedContent () {
-    return this.props.frameProps.getIn(['security', 'isMixedContent'])
+  get isRunInsecureContent () {
+    return this.props.frameProps.getIn(['security', 'isRunInsecureContent'])
```
bridiver: isRunInsecureContent => runInsecureContent

js/components/siteInfo.js
```diff
@@ -46,6 +63,12 @@ class SiteInfo extends ImmutableComponent {
         <span data-l10n-args={JSON.stringify(l10nArgs)} data-l10n-id='sessionInfo' /></li>
     }
 
+    let allowMixedContentButton
```
bridiver: MixedContent => RunInsecureContent

js/state/contentSettings.js
```diff
@@ -111,6 +111,10 @@ const getContentSettingsFromSiteSettings = (appState) => {
     flashActive: [{
       setting: 'block',
       primaryPattern: '*'
+    }],
+    runInsecureContent: [{
+      setting: 'block',
```
bridiver: this is where the site hacks should go for running insecure content

js/stores/windowStore.js
```diff
@@ -742,6 +742,10 @@ const doAction = (action) => {
         windowState = windowState.setIn(path.concat(['security', 'isSecure']),
                                         action.securityState.secure)
       }
+      if (action.securityState.runInsecureContent !== undefined) {
+        windowState = windowState.setIn(path.concat(['security', 'isRunInsecureContent']),
```
bridiver: isRunInsecureContent  => runInsecureContent

bridiver:
can you add some tests for this? There is some inconsistent property naming that looks like it should be causing issues

bridiver:
just want to get @bradleyrichter @diracdeltas feedback on the  popup. I think we should have something like

"Insecure (http) content has been blocked from running on this page. If you allow this content to run it will not be encrypted and it may transmit unencrypted data to other sites."

button - "Run Insecure Content"

bridiver:
warning

bradleyrichter:
Is it modal? If not, no need for the darkening.

bridiver:
closed with https://github.com/brave/browser-laptop/commit/707868014f5d8a330c300cf29bb81dc0aff7c3d8

Issue #3453, opened 2016-08-26
Bookmark folders are appearing in browser history


alexwykoff:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When creating a bookmark folder, it appears in browser history and cannot be cleared. 

**Expected behavior:**
1. Bookmarks and folders should not appear in history.
2. If a user decides to clear history, all items should be cleared. 

- Platform (Win7, 8, 10? macOS? Linux distro?):
Tested on OS X

- Brave Version:
0.11.6 pre-beta4

- Steps to reproduce:
    1. Open the about:history page.
    2. Create a bookmark folder.
    3. Observe the history, attempt to clear the history.

- Screenshot if needed:
![bookmark_folder_in_history](https://cloud.githubusercontent.com/assets/1434040/18008831/596c2cdc-6b6f-11e6-85e6-c0486bfff4e8.gif)

- Any related issues:

bsclifton:
Great catch! :smile: Working on a fix...

bsclifton:
Fixed w/ https://github.com/brave/browser-laptop/pull/3459

Issue #3366, opened 2016-08-24
Bookmarked about pages appear in the history


luixxiul:
When I bookmark about pages, they appear on ``about:history``.

- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 7 64 bit

- Brave Version: 0.11.6 beta3

- Steps to reproduce:
    1. Open ``about:history`` from url bar
    2. bookmark it with CTRL+D and clicking "Save" on the dialog
    3. Open a new tab
    4. Go to github.com
    5. Select the about:history tab

The same thing happens when I open and bookmark ``about:passwords``.

- Screenshot if needed:
![clipboard01](https://cloud.githubusercontent.com/assets/3362943/17921651/e0f1a4da-6a15-11e6-830b-20811dc404c8.png)

I am not sure if this is intended and it works as it should or not.

cc: @bsclifton 

#3362

bbondy:
I think Chrome filters out all about pages, please verify and if so do the same here.

bsclifton:
Fixed w/ https://github.com/brave/browser-laptop/pull/3459

PR #3459, opened 2016-08-26
History bugfixes

bsclifton:

- Added a new isHistoryEntry method
- Folders are no longer considered history; fixes https://github.com/brave/browser-laptop/issues/3453
- Filters out about pages from history; fixes https://github.com/brave/browser-laptop/issues/3366
- Single history entries are now shown; there was no open defect for this; it would just throw an error

### DIFF
--- a/js/about/history.js
+++ b/js/about/history.js
@@ -13,6 +13,7 @@ const aboutActions = require('./aboutActions')
 const getSetting = require('../settings').getSetting
 const SortableTable = require('../components/sortableTable')
 const Button = require('../components/button')
+const siteUtil = require('../state/siteUtil')
 
 const ipc = window.chrome.ipc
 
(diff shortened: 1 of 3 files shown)


js/about/history.js
```diff
@@ -71,7 +72,11 @@ class GroupedHistoryList extends ImmutableComponent {
       }
       return result
     })
-    if (reduced) return reduced
+    if (reduced) {
+      return Array.isArray(reduced)
+        ? reduced
+        : [{date: this.getDayString(locale, reduced), entries: [reduced]}]
```
bsclifton: This line has the fix for showing single entries. Before, it would just return the object (which fails when trying to do the map() in the render for the HistoryDay control)

test/unit/state/siteUtilTest.js
```diff
@@ -414,6 +414,50 @@ describe('siteUtil', function () {
     })
   })
 
+  describe('isHistoryEntry', function () {
+    it('returns true for a typical history entry', function () {
```
bridiver: not sure if someone else reading this will understand what makes it "typical". Maybe "returns true for site entries with a lastAccessedTime"?

bridiver:
Looks good! I think ideally we'd like to filter the data before it gets to the history page, but we can do a larger refactoring later

Issue #3440, opened 2016-08-26
Existing bitcoin wallet address is not displayed


luixxiul:
**Describe the issue you encountered:**
On Windows, existing bitcoin wallet address is not displayed on the dialog which appears by clicking "Add Funds" on about:preferences#payments. 

**Expected behavior:**
The bitcoin wallet address should be displayed.

- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 7 and 10

- Brave Version: 0.11.6 Beta4

- Steps to reproduce:
    1. Move your brave folder temporarily and start the browser
    2. Open Payments on about:preferences
    3. Enable the payment and click "create wallet"
    4. Click "add funds"

- Screenshot if needed:
![clipboard01](https://cloud.githubusercontent.com/assets/3362943/17995197/c22adf90-6b9a-11e6-886c-b01565bd3bd9.png)

The address is expected to be displayed around "copy this link".

- Any related issues:

Setting a milestone.

alexwykoff:
Confirmed this is isolated to Windows. OS X and Ubuntu displayed the wallet address appropriately.

diracdeltas:
https://github.com/brave/browser-laptop/commit/5e0b6793b26abb513c080fb92a9cf2f2728a4e62 changes it so that an error message is shown if there is no address available. that doesn't solve the underlying issue of why there is no `paymentURL` property on windows though.

diracdeltas:
@mrose17 any idea why `paymentURL` would be undefined only for windows? maybe intermittent failure related to ledger server?

diracdeltas:
@luixxiul could  you open your `session-store-1` file and copypaste the `ledgerInfo` value here? (or DM me on slack)

luixxiul:
It seems to be undefined: ``"ledgerInfo":{}``.

mrose17:
@luixxiul - please find the file `ledger-state.json` and DM it to @diracdeltas ... i think that @bsclifton knows where to find that on windows

luixxiul:
@mrose17 ok, sent it to @diracdeltas and you.

luixxiul:
Confirmed it has been fixed on 0.11.6 Beta5 :-)

PR #3470, opened 2016-08-27
updateLedgerInfo should run before client is assigned if previous state is present

mrose17:
it appears the concern that @diracdeltas  has regarding https://github.com/brave/browser-laptop/blob/master/app/ledger.js#L735 is proven, as it explains https://github.com/brave/browser-laptop/issues/3440

time for an extra-large helping of [humble pie](https://en.wikipedia.org/wiki/Humble_pie)

auditor: @diracdeltas

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -32,6 +32,7 @@ const session = electron.session
 
 const acorn = require('acorn')
 const ledgerBalance = require('ledger-balance')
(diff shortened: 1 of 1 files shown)


bbondy:
@diracdeltas please audit, I'm going to merge in for beta5 though.

diracdeltas:
++

PR #3430, opened 2016-08-25
updateledgerinfo refactor

diracdeltas:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Ran `git rebase -i` to squash commits if needed.

consolidates calls for updating ledger info appstate, so it is less likely that ledgerInfo in appState will get out of sync with ledgerInfo in ledger.js

auditors: @mrose17

PR #3471, opened 2016-08-27
Revert "updateledgerinfo refactor"

bbondy:
Reverts brave/browser-laptop#3430

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -138,11 +138,11 @@ var boot = () => {
     if (err.code !== 'ENOENT') console.log('statePath read error: ' + err.toString())
 
     ledgerInfo.creating = true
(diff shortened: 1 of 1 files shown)


bbondy:
@diracdeltas sorry for merging this too early, I saw a positive agreeing comment on the last PR that  @mrose17 left,  but not a full approval, so I thought it would be a net good to merge into beta5, but apparently not. Reverting.

diracdeltas:
np, this isn't really a necessary change except for the part that @mrose17, though the refactor is nice to have.

mrose17:
let me suggest we wait until after the next beta is out and we can test it against master. i'd like to reduce a few moving parts here...

PR #3472, opened 2016-08-27
use different file name for development version

mrose17:
i was running under both `/Applications` and `npm run watch/npm start`
and had some unfortunate interactions

auditor: @bbondy

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -53,13 +53,13 @@ const eventStore = require('../js/stores/eventStore')
 const rulesolver = require('./extensions/brave/content/scripts/pageInformation.js')
 
 // TBD: remove these post beta [MTR]
-const logPath = path.join(app.getPath('userData'), 'ledger-log.json')
-const publisherPath = path.join(app.getPath('userData'), 'ledger-publisher.json')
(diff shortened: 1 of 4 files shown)


app/ledger.js
```diff
@@ -132,7 +132,7 @@ var boot = () => {
   if ((bootP) || (client)) return
 
   bootP = true
-  fs.access(statePath, fs.FF_OK, (err) => {
+  fs.access(pathName(statePath), fs.FF_OK, (err) => {
```
bbondy: pathName could probably just be called only once again above instead of repeatedly below, downside is that someone could forget to call it when using statePath (which isn't actually a path) later. I'll merge though.

bbondy:
btw if something in a pull request is committed then reverted, then it should just be rebased out completely and not included in the pull request.

bbondy:
You can do that with an interactive rebase for future ref: `git rebase -i master` and then delete out the lines for the commits that are not needed.
If you just want to undo a commit and it is the top-most thing you can do `git reset HEAD^ --hard` or to undo the commit and keep the things staged `git reset HEAD^ --soft`

PR #3474, opened 2016-08-27
Add shortcut to view page source

Sh1d0w:
Added handy shortcut to display page source, just like in Google Chrome.

cc @bbondy @bridiver @bsclifton

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -452,8 +452,7 @@ class Frame extends ImmutableComponent {
         break
       case 'view-source':
         const sourceLocation = UrlUtil.getViewSourceUrlFromUrl(this.webview.getURL())
-        let src = this.frame.get('src')
-        if (UrlUtil.isHttpAddress(src) && !UrlUtil.isImageAddress(src)) {
+        if (sourceLocation !== null) {
           windowActions.newFrame({location: sourceLocation}, true)
         }
         // TODO: Make the URL bar show the view-source: prefix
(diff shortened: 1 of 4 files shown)


js/components/frame.js
```diff
@@ -452,7 +452,10 @@ class Frame extends ImmutableComponent {
         break
       case 'view-source':
         const sourceLocation = UrlUtil.getViewSourceUrlFromUrl(this.webview.getURL())
-        windowActions.newFrame({location: sourceLocation}, true)
+        let src = this.frame.get('src')
+        if (UrlUtil.isHttpAddress(src) && !UrlUtil.isImageAddress(src)) {
```
bridiver: can we split this change out from the shortcut? I think we probably want to make these extra checks in `getViewSourceUrlFromUrl` and maybe return a null location if there is not valid view-source for the url
@bradleyrichter is this shortcut in your list?

Sh1d0w: Good point, moved.

bbondy:
If you type in gmail compose, or any text editable for example, and do command+U to underline it, it'll open the page source in a new tab instead.  

I tweaked it to be the same shortcut as Chrome (with an Alt key too and merged. Thanks!)

Sh1d0w:
@bbondy I think you forgot to put the check for the current address like I did in my PR, so if you click CTRL + ALT + U on new tab frame or on a image you currently opened in a browser you will trigger view-source, which does not happend on other browsers and should not happen in Brave as well. 

Please see my PR for reference how I have checked this, so you can put that too :)

bbondy:
The task was merged so it is all commits
Is there a problem on master?

bbondy:
I only merged your PR and then did a commit on top.

Sh1d0w:
@bbondy Ah never mind my comment I tought you closed my PR without merging it because github stated `Closed with unmerged commits`. It's all ok.

PR #3478, opened 2016-08-27
Fix findbar erros in console

Sh1d0w:
The following piece of code was throttling the console with errors because `this.frame` was undefined.

cc @bbondy @bridiver


bbondy:
++ thanks

Issue #3480, opened 2016-08-27
ledger unconfirmed balance takes a long time to show up


diracdeltas:
after adding funds through Coinbase, the ledger `unconfirmed` balance still does not show up even though it's in https://blockr.io.

bbondy:
PR was merged that said it was a fix for this, so closing.

PR #3495, opened 2016-08-28
Ledger unconfirmed balance fix

mrose17:
the value was being kept in a cache that was being read, but not written… no longer.

also, fix a small typo in `docs/state.md`

this is the fix for https://github.com/brave/browser-laptop/issues/3480

auditor: @diracdeltas

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -116,7 +116,7 @@ const doAction = (action) => {
  */
 var init = () => {
   try {
-    ledgerInfo._internal.debugP = ledgerClient.prototype.boolion(process.env.LEDGER_INFO_DEBUG)
+    ledgerInfo._internal.debugP = ledgerClient.prototype.boolion(process.env.LEDGER_CLIENT_DEBUG)
     publisherInfo._internal.debugP = ledgerClient.prototype.boolion(process.env.LEDGER_PUBLISHER_DEBUG)
 
     appDispatcher.register(doAction)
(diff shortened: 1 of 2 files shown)


app/ledger.js
```diff
@@ -471,7 +471,12 @@ var updatePublisherInfo = () => {
   publisherInfo.synopsis = synopsisNormalizer()
 
   if (publisherInfo._internal.debugP) {
-    console.log('\nupdatePublisherInfo: ' + JSON.stringify(underscore.omit(publisherInfo, [ '_internal' ])))
+    data = []
+    publisherInfo.synopsis.forEach((entry) => {
+      data.push(underscore.extend(underscore.omit(entry, [ 'faviconURL' ]), { faviconURL: entry.faviconURL && '...' }))
```
diracdeltas: why `...`?

mrose17: because the "real" string tends to be a long base64 value. all we really need to know is whether a value is present or not, the `"..."` serves that purpose... otherwise, you get a lot of scrolling in the console window with very little _useful_ content.

app/ledger.js
```diff
@@ -976,12 +981,19 @@ var getBalance = () => {
 
   ledgerBalance.getBalance(ledgerInfo.address, underscore.extend({ balancesP: true }, client.options),
   (err, provider, result) => {
+    var unconfirmed
+    var info = ledgerInfo._internal.paymentInfo
```
diracdeltas: just curious why `unconfirmed` needs to be in `ledgerInfo._internal` if it already exists in `ledgerInfo`

bbondy:
@mrose17 btw if you use  `fix #3480` instead of `fix for #3480` it'll auto close the issue when merged.

mrose17:
@bbondy - true. however, i generally prefer not to auto-close an issue unless i created it... this allows the person who actually crated the issue to verify that it's fixed...

bbondy:
@mrose17 please follow the process of closing issues when they are complete on master. Sometimes there are multiple commits and QA has it's own process. It would be too confusing to leave things open for testing and that is not the process we're following for this project. It doesn't scale well. Thanks.

luixxiul:
@mrose17 please ping me or @alexwykoff in case you need QA ;-)

PR #3507, opened 2016-08-29
Add cibuild script

bbondy:
Auditors: @aekeus

### DIFF
--- a/tools/lib/github.py
+++ b/tools/lib/github.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+import json
+import os
+import re
+import sys
+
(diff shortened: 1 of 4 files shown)


bbondy:
This is to save some of my time from running a few different scripts and manually uploading. This will do it all with 1 command.

I'll update docs separate from this

bbondy:
Merged @aekeus but pls audit.

tools/lib/github.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python
```
bbondy: This file is just copied from electron so no need to review it

PR #3538, opened 2016-08-29
fix browser crash if you close a window was the browser is launching

mrose17:
if you start the browser and before it “settles”, you close a tab/window, then you get a crash. because `WINDOW_SET_FOCUS_FRAME` is generated by `action.frameProps` is empty. in this case, `eventStore.js` falls over taking the entire browser with it.

auditor: @bridiver

### DIFF
--- a/js/actions/windowActions.js
+++ b/js/actions/windowActions.js
@@ -385,10 +385,12 @@ const windowActions = {
    * @param {Object} frameProps - the frame properties for the webview in question.
    */
   setFocusedFrame: function (frameProps) {
-    dispatch({
-      actionType: WindowConstants.WINDOW_SET_FOCUSED_FRAME,
-      frameProps: frameProps
-    })
+    if (frameProps) {
+      dispatch({
+        actionType: WindowConstants.WINDOW_SET_FOCUSED_FRAME,
+        frameProps: frameProps
+      })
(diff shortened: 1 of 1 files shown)


js/stores/eventStore.js
```diff
@@ -98,8 +98,10 @@ const doAction = (action) => {
       eventState = eventState.set('page_load', eventState.get('page_load').push(pageLoadEvent))
       break
     case WindowConstants.WINDOW_SET_FOCUSED_FRAME:
-      addPageView(action.frameProps.get('location'))
-      lastActiveTabId = action.frameProps.get('tabId')
+      if (action.frameProps) {
```
bridiver: this looks like a good fix, but I'm thinking that we should apply it in windowActions so anything else that cares about this event in the future won't have the same problem

mrose17: ok, who should we re-assign it to in order to implement a more general solution?

bridiver: It's the same fix, just in a different place, but I can make it if you want

mrose17: since i am not familiar with the code, and you wrote it, let me defer the changes to you... less likely to have an inadvertent mishap... thanks!

PR #3545, opened 2016-08-29
support new Set(...) in ledger publisher ruleset

mrose17:
this accommodates the new longer exclusion list, cf., https://github.com/brave/browser-laptop/issues/3553

auditor: @BrendanEich


app/extensions/brave/content/scripts/pageInformation.js
```diff
@@ -59,6 +59,14 @@
           expr['arguments'].forEach((argument) => { args.push(traverse(argument)) })
           return value[traverse(op1.property, true)].apply(value, args)
 
+        case 'NewExpression':
+          op1 = expr.callee
+          if (op1.type !== 'Identifier') throw new Error('unexpected callee: ' + op1.type)
+          if (op1.name !== 'Set') throw new Error('only new Set(...) is allowed, not new ' + op1.name)
+          args = []
+          expr['arguments'].forEach((argument) => { args.push(traverse(argument)) })
+          return new Set(args[0])
```
BrendanEich: Don't you want `return new Set(...args)` for full generality? No point collecting the  traversed arguments in args past the first one if you use only the first one.

bbondy:
Is there an issue defined for this?

bbondy:
If not, please post an issue, mark 0.11.6 on the issue and then update the PR.
If so, please link up the issue.

The PR can be updated with:
```
git add -u
git commit --amend
git push -f
```

mrose17:
i will create an issue.

app/extensions/brave/content/scripts/pageInformation.js
mrose17: `Set` takes at most one argument, which is an array. if i were to pass `args` instead of `args[0]`, then i would need to do use `.apply()` as well. i will make it clearer.

PR #3547, opened 2016-08-29
Improve themeColor and add it to siteDetail

cezaraugusto:
This commit aims for adding `themeColor` to `siteDetail` so it can be accessed as other `siteDetail`s properties. 

Plus, adds `computedThemeColor` as fallback when there's no color specified.

Follow-up of #3406 

Auditors: @bbondy


bbondy:
++

Issue #3451, opened 2016-08-26
Significant lag between browser history and ledger listing


alexwykoff:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When visiting sites, they show up in the browser history, but do not appear in the ledger, or appear after a significant delay (10 minutes and counting).

**Expected behavior:**
The visited sites should appear within a reasonable window.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Tested on OS X

- Brave Version:
0.11.6 pre-beta4

- Steps to reproduce:
    1. Have Brave payments turned on.
    2. Visit a lot of sites in quick succession (I use news.ycombinator.com) .
    3. Check the browser history.
    4. Wait for a couple minutes, check the ledger.

- Screenshot if needed:
![ledger_lag](https://cloud.githubusercontent.com/assets/1434040/18008111/73edb09c-6b6c-11e6-9ab6-8931e5404b33.gif)

- Any related issues:

mrose17:
@alexwykoff - before i dig into this, let me verify a few things:

- you have to be on a page for at least 8 seconds before it gets considered a "visit"

- you have to leave the page (sometime) after those 8 seconds before it gets considered a "visit"

can you verify that this is what you're seeing? thanks!

alexwykoff:
I'll post back after the investigation. Thanks!

alexwykoff:
It would appear that ledger listings are not occuring until navigation events.

![payment_time](https://cloud.githubusercontent.com/assets/1434040/18035686/0cdbef8a-6d20-11e6-8e4a-cd7621b44fe2.gif)

mrose17:
@BrendanEich & @bbondy - @alexwykoff and i had a hangout in which we found out that it works for MTR but not for alex. this is true regardless of whether i'm using a build or an `npm start`.  all i can suggest at this point is that we get more people trying this out and seeing if they see the same problem. both @alexwykoff and i tested on a mac. worked for me, not for him. very odd...

bbondy:
@mrose17 can you provide him with a logged build? Or ask him to open a console and check for something? /cc @diracdeltas

BrendanEich:
Need other folks than @alexwykoff to reproduce too (or not). More than two people testing is needed to move forward here, unless code inspection finds a bug. Thanks.

mrose17:
[...]

PR #3554, opened 2016-08-30
a few minor things

mrose17:
1. if `process.env-NODE_ENV === ‘test’` then set the `synopsis` option `minDuration` to 0 msecs

2. if `process.env.LEDGER_PUBLISHER_DEBUG === true` then show how visits are counted -- for https://github.com/brave/browser-laptop/issues/3451

3. better error printing in one case

4. include the `ledger-publisher` code for handling `new Set()`

auditor: @diracdeltas


app/ledger.js
```diff
@@ -467,16 +465,19 @@ var updatePublisherInfo = () => {
   syncWriter(pathName(publisherPath), data, () => {})
   syncWriter(pathName(scoresPath), synopsis.allN(), () => {})
 
+  if (synopsis.options.minDuration === 0) synopsis.options.minDuration = 8 * msecs.second
```
diracdeltas: why is it set to 8s here?

app/extensions/brave/content/scripts/pageInformation.js
```diff
@@ -59,6 +59,15 @@
...
           return value[traverse(op1.property, true)].apply(value, args)
 
+        case 'NewExpression':
+          op1 = expr.callee
+          if (op1.type !== 'Identifier') throw new Error('unexpected callee: ' + op1.type)
+          if (op1.name !== 'Set') throw new Error('only new Set(...) is allowed, not new ' + op1.name)
+          args = []
+          expr['arguments'].forEach((argument) => { args.push(traverse(argument)) })
+          if (args.length > 1) throw new Error('Set(...) takes at most one argument')
+          return new Set(args[0])
```
diracdeltas: wasn't this already reviewed in https://github.com/brave/browser-laptop/pull/3545 by @BrendanEich ?

app/ledger.js
mrose17: because that is the actual default value, and when the synopsis is written out, i don't want it to be zero, so that when someone starts the browser without `process.env.NODE_ENV` being test, they won't get surprised... but, i see an easier fix. please stand by.

app/extensions/brave/content/scripts/pageInformation.js
mrose17: yes, as noted in the PR, it's included here as well.

app/ledger.js
mrose17: @diracdeltas - i think the latest commit simplifies things.

PR #3571, opened 2016-08-30
set ledger column items to right align

### LINKED ISSUE #3567: set ledger columns to right-align if appropriate
every column except site and include, i.e., rank, views, time spent, and %

jkup:

Fixes #3567 

Ok so instead of using fixed widths like we were, I moved them to flex width with padding so we can have some items be right aligned. They look a little condensed right now but I was worried about giving them too much padding in case people end up with really long URLs in their ledger.

Thoughts?

Auditors: @bsclifton @bradleyrichter 

screen shot 2016-08-30 at 11 33 21 am


bbondy:
condensed++ :)

Issue #3594, opened 2016-08-31
Ledger client updates


mrose17:
add a latch to ensure that run(0) from the ledger-client doesn't cascade!

fixed https://github.com/brave/browser-laptop/pull/3595

auditor: @bbondy

mrose17:
using a new strategy

PR #3595, opened 2016-08-31
Ledger client updates

mrose17:
Fixes https://github.com/brave/browser-laptop/issues/3594


bbondy:
@diracdeltas pls review instead

mrose17:
using a new strategy

PR #3630, opened 2016-09-01
add a null page view when all windows lose focus

bridiver:
cc @mrose17


bbondy:
Could you link up the issue? Thanks.

js/stores/eventStore.js
```diff
@@ -101,6 +108,16 @@ const doAction = (action) => {
       addPageView(action.frameProps.get('location'))
       lastActiveTabId = action.frameProps.get('tabId')
       break
+    case AppConstants.APP_WINDOW_BLURRED:
+      windowBlurred(action.appWindowId)
+      break
+    case AppConstants.APP_IDLE_STATE_CHANGED:
+      if (action.idleState !== 'active') {
+        addPageView(null)
```
bsclifton: How does a null page view behave?

bridiver:
need to fix the commit, it should say fixes #3650

js/stores/eventStore.js
bridiver: it just logs the timestamp when the browser became inactive. We use the same thing for closing all windows

bbondy:
Moving to 0.12.1 for now, please merge when reviewed @mrose17 it could make it in a respin still.

mrose17:
@bbondy - i have reviewed this PR and it works well on my system.

@bridiver - we may need to iterate a bit in the future, specifically to deal with mouse enter events and scrolling events; however, i'd like us to drive this version for a while before digging into it again.

many thanks for the fast updates!

PR #3632, opened 2016-09-01
serialize app state saves

### LINKED ISSUE #3543: potential race condition on session state save
saving the state can be triggered by a periodic timer, the user quitting the app or the app quit timeout firing. The actual saving of the state to disk uses async io so the event loop is free to handle other method calls once it starts. This means that another call to save the state could come in and start to save while the previous call is still writing/renaming.

bridiver:

fixes #3543

cc @BrendanEich 

There is still an issue with `initiateSessionStateSave` because it's possible for the calls to overlap and potentially call `saveIfAllCollected` before getting state from all the windows. This is possible because it only checks the count and doesn't check to make sure it has state for each active window id.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -63,6 +63,7 @@ const flash = require('../js/flash')
 const contentSettings = require('../js/state/contentSettings')
 const privacy = require('../js/state/privacy')
 const basicAuth = require('./browser/basicAuth')
+const async = require('async')
 
 // Used to collect the per window state when shutting down the application
 let perWindowState = []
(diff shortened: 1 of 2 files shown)


bridiver:
not really sure what to do for a test plan here because the issue is hard to replicate

bbondy:
test plan wise just verify it doesn't regress anything in saving / loading state.

bridiver:
I tested periodic saves and save on quit. Should also probably test save on upgrade, but there's no reason that would be different than save on quit. I'm going to try setting the periodic timer to a very short interval to force some overlap

bridiver:
set periodic save delay to 1ms and verified that each save/rename block happened sequentially

app/index.js
```diff
@@ -63,6 +63,7 @@ const flash = require('../js/flash')
 const contentSettings = require('../js/state/contentSettings')
 const privacy = require('../js/state/privacy')
 const basicAuth = require('./browser/basicAuth')
+const async = require('async')
```
bbondy: It's better to require only `async/queue` so it doesn't pull in everything, pls do in a follow up, I won't block merging this though on it.

bbondy:
++

Issue #3639, opened 2016-09-01
siteUtils.clearSitesWithoutTags has misleading name


bsclifton:
Feedback per https://github.com/brave/browser-laptop/pull/3501/files/dc5d3b7ad01644d918f792a7456e11a78473cb26#r76535126

siteUtils handles:
- history
- bookmarks
- pinned tabs

The methods have more descriptive names, specifically the clearSitesWithoutTags method (called when you clear history).

PR #3640, opened 2016-09-01
Give more meaningful name to siteUtils method

bsclifton:

clearSitesWithoutTags => clearHistory

Fixes https://github.com/brave/browser-laptop/issues/3639

Auditors: @bbondy @bridiver 

Test Plan:
- Visit about:history
- Use the "Clear browsing data button"
- Ensure "Browsing history" is on and click "OK"
- Go back to about:history
- Verify history is cleared


js/state/siteUtil.js
```diff
@@ -371,10 +371,10 @@ module.exports.filterSitesRelativeTo = function (sites, relSite) {
 }
 
 /**
- * Clears out all sites that have no tags.
+ * Clears history (all sites that have no tags).
  * @param sites The application state's Immutable sites list.
  */
-module.exports.clearSitesWithoutTags = function (sites) {
+module.exports.clearHistory = function (sites) {
```
bridiver: if this clears history for all sites we should use the method here https://github.com/brave/browser-laptop/blob/master/app/sessionStore.js#L258

app/sessionStore.js
```diff
@@ -254,12 +256,7 @@ module.exports.cleanAppData = (data, isShutdown) => {
   if (data.sites) {
     const clearHistory = isShutdown && getSetting(settings.SHUTDOWN_CLEAR_HISTORY) === true
     if (clearHistory) {
-      // TODO - this should the history methods from siteUtils
-      data.sites = data.sites.filter((site) => site && site.tags && site.tags.length)
-      data.sites = data.sites.map((site) => {
-        delete site.lastAccessedTime
-        return site
-      })
+      data.sites = siteUtil.clearHistory(Immutable.fromJS(data.sites)).toJS()
```
bsclifton: @bridiver per your feedback, I updated this to use the siteUtil method :smile: 

I'm curious what you think about the temp conversion to/from Immutable. I originally made the siteUtil method detect if the input is Immutable, but if feels bad introducing more code paths.

bridiver: I don't think it's a big deal here, it's just a single call in an infrequently called method

bridiver:
can you squash and then I think this is good to merge

bsclifton:
@bridiver squashed :smile:

bbondy:
nice cleanup, thanks.

Issue #3284, opened 2016-08-21
Bookmark title is not updated unbookmarking/rebookmarking


luixxiul:
- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 7 64 bit

- Brave Version: 0.11.6 Beta1

- Steps to reproduce:
    1. Move away your brave folder temporarily
    2. Navigate to https://github.com/brave/browser-laptop/issues/3283
    3. Bookmark the page

- Screenshot if needed:
![clipboard01](https://cloud.githubusercontent.com/assets/3362943/17834897/278ff3e8-6790-11e6-9ec2-222e412d40b0.jpg)

- Any related issues: #3283

bbondy:
> iii. Bookmark the page

For steps to reproduce pls be super specific, thanks.  In which way did you bookmark, maybe the star button?

bbondy:
This is working for me

bbondy:
What is the "actual results" and "expected results"?

luixxiul:
I can reproduce the issue by bookmarking this page with pushing <kbd><kbd>CTRL</kbd>+<kbd>D</kbd></kbd> and clicking ``Save``.

The actual result is that the bookmark is saved with the title ``Issues · brave/browser-laptop``, while the expected one is the title ``Bookmark name is wrongly saved on GitHub issue page · Issue #3284 · brave/browser-laptop``.

luixxiul:
I assume that database files were corrupted at some point, but this seems to me not an accident because I experienced the same issue on Windows 10 too. On clean install I cannot reproduce this.

bsclifton:
I believe I understand the issue and why we are not able to reproduce it

Bookmark names are saved after they are changed, even after you remove the bookmark. The only way to truly clear the item out is to clear the history.

Here are some steps you can try to reproduce
1) Bookmark any page
2) Edit the bookmark, give it a custom title of "ABC"
3) Remove the bookmark
4) Re-bookmark the page
5) Notice the name is now "ABC"

bsclifton:
Working on a fix for this now...

BrendanEich:
Dup of #3080?

bsclifton:
We want the behavior of keeping the **custom** title-

I believe after pulling hairs out I have characterized the problem (thanks @BrendanEich!)
- bookmark a page (let's say it has title "old title")
- remove bookmark. You no longer have this item bookmarked.
- [time goes by]
[...]

PR #3641, opened 2016-09-01
Bookmark customTitle from previously bookmarked entry is no longer re-used

bsclifton:

Bookmark customTitle from previously bookmarked entry is no longer re-used when re-bookmarking.

Fixes https://github.com/brave/browser-laptop/issues/3284

Auditors: @bbondy @luixxiul, @bradleyrichter

Test Plan:
- Visit a site which is not bookmarked
- Bookmark the site- when the modal comes up, provide your own title "ABC" and hit enter
- Notice bookmark is using customTitle
- Delete bookmark
- Re-add bookmark- this time don't change the title
- Notice bookmark is now properly using title (not the old customTitle)

### DIFF
--- a/js/state/siteUtil.js
+++ b/js/state/siteUtil.js
@@ -82,13 +82,18 @@ module.exports.getNextFolderId = (sites) => {
 }
 
 /**
- * Adds the specified siteDetail to sites
+ * Adds or updates the specified siteDetail in sites.
+ *
+ * Examples of updating:
+ * - editing bookmark in add/edit modal
+ * - when timestamp is added (history entry)
+ * - moving bookmark to/from a folder
  *
  * @param sites The application state's Immutable site list
- * @param siteDetails The site details to add a tag to
- * @param tag The tag to add for this site.
- *   See siteTags.js for supported types. No tag means just a history item.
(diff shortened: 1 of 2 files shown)


js/state/siteUtil.js
```diff
@@ -109,9 +109,6 @@ module.exports.addSite = function (sites, siteDetail, tag, originalSiteDetail) {
     tags = tags.toSet().add(tag).toList()
   }
 
-  // We don't want bookmarks and other site info being renamed on users if they already exist
-  // The name should remain the same while it is bookmarked forever.
-  const customTitle = typeof siteDetail.get('customTitle') === 'string' ? siteDetail.get('customTitle') : (siteDetail.get('customTitle') || oldSite && oldSite.get('customTitle'))
```
bbondy: did you check to make sure this isn't regressed? Use git blame on the task to see the original bug.

bsclifton:
Updated code after @bbondy left a comment (see above). The original commit I had done re-introduced an old bug. I reverted that and instead solved the issue by removing the custom title when bookmark is removed. Tests were added to reinforce this behavior :smile:

bbondy:
excellent great fix, comment cleanup and tests! 👍

PR #3707, opened 2016-09-04
Add missing context menu shortcut labels

Sh1d0w:
Some of the context menu entries had missing shortcuts labels, which were already present in the top menu entries only.

Reviewers: @bbondy


bbondy:
++

PR #3710, opened 2016-09-04
Add Italian language

lorenzoganni:


bbondy:
Thanks for this work but it has to be done in Transifex. You can request access here:
https://www.transifex.com/brave/brave-laptop/

If you have any trouble just email aubrey@brave.com. Thanks!

PR #3199, opened 2016-08-16
[WIP] New tab page

cezaraugusto:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Ran `git rebase -i` to squash commits if needed.

Still a lot of work to do. **WIP of #3001**

**Current output:** (will be updated as things gets done)

newnewtab
# Tasks:
## 1.0 ENG tasks:
- [x] build data collector for "Bravery blocking reports", and add time estimation
  - [x] add mechanism for when the number excels the column width. i.e. `33k`
- [x] add fall-back mechanism when large favicon is not present for a given site
- [x] ~~build high-res favicon collector mechanism, pulled from page header at domain root page~~ **(not needed; best favicon URL is already fetched when page loads)**
- [x] ~~create local cache for large favicons (base64)~~ **(not needed; favicons are already cached by browser)**
- [ ] create mechanism to retain page theme-color
- [ ] **(ON HOLD)** interface with tab-preview mechanism to be utilized for previewing the topsites on mouse-over. (replaces BG image during mouseover and uses the same delay approach as tab-previews)
- [x] build the top-sites list creation mechanism using a blend of freceny and other means which may be adjusted over time to improve the results
- [x] Translate estimated time strings
- [x] build the BG images retrieval mechanism, with smart randomizer to eliminate the same image twice in a row **(working but needs follow-up)**
- [x] build data connector for local-time display
- [x] connect new-tab to prefs
- [x] add mechanism to remove unwanted tiles
- [x] Add mechanism for Pinned tiles
- [x] Add mechanism for bookmarking tiles
- [x] Add reordering mechanism for tiles' grid
- [x] State changes need to be documented in docs/state.md
- [x] Removing a pinned site would also remove it from pinned list
- [x] Pinned sites should be allowed to be dragged as well
- [x] Add 'undo Pinning' **(working)**
- [x] Add 'restore gridLayout' in a way similar as Chrome **(working)**
- [ ] Make `suggestions.js` ignore bookmarked sites on topSites grid
## 1.0 Front-end tasks
- [x] arrange BG image in CSS to zoom-to-width or height with above-centered anchor point
- [x] create separate black gradient mask in layer over BG image
- [x] create adjustable table layout for top-site buttons with high-res favicons
- [x] create CSS for Bravery info display
[...]

PR #3767, opened 2016-09-06
Add clock component

cezaraugusto:
Just a commit split for the current implemented clock (#3199) for usage in #3001

Auditors: @bbondy

### DIFF
--- a/js/about/clock.js
+++ b/js/about/clock.js
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const React = require('react')
+
+class Clock extends React.Component {
+  constructor () {
+    super()
+    this.state = {}
+  }
+  get currentTime () {
+    const currentdate = new Date()
+    let hours = currentdate.getHours()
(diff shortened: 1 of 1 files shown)


bbondy:
++

bsclifton:
@cezaraugusto should this be using JavaScript's toLocaleTimeString to format the hours and separator? (which also includes the period suffix, like AM/PM)

bbondy:
I think that'd be great, pls do in a follow up

cezaraugusto:
@bsclifton @bbondy there are a few considerations about using `.toLocaleTimeString`:

* Will we support variations in localized time formats? i.e. Korean: `오후 05:00`.
* Current spec displays a different font-size for period suffix. If the above is true, which will be the preferred method to keep it loyal to spec, given that suffix will be set according to locale?

cc'ing @bradleyrichter for any thoughts

bsclifton:
Having the the time formatted according to the browser's locale setting is definitely the ideal case; I'd personally recommend just starting off implementing it using toLocaleTimeString() and then capturing the edge cases as separate issues :smile: 

toLocaleTimeString is used on the history page and it'll be important (the more timestamps, etc we introduce) to be consistent within the app.

bbondy:
yep it's ok to use here and exact look on spec doesn't matter here.

bradleyrichter:
We could start with a combined string. Maybe create a special case for English?

PR #3773, opened 2016-09-06
remove mutation observers from brave content scripts

bridiver:
Test Plan:
Check password capture on submit
1. Go to Facebook.com
2. Enter a username and password
3. Brave password save dialog should appear

1. Go to https://mint.intuit.com/login.event
2. Enter a username and password
3. Brave password save dialog should appear

Check password fill for both immediate and delayed load password forms
1. Go to Facebook.com
2. Password should autofill

1. Go to https://mint.intuit.com/login.event
2. Password should autofill

don't add duplicate submit and click event listeners on every password check
only check for changes while page is visible

cc @bbondy @diracdeltas

### DIFF
--- a/app/extensions/brave/content/scripts/passwordManager.js
+++ b/app/extensions/brave/content/scripts/passwordManager.js
@@ -3,7 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 if (chrome.contentSettings.passwordManager == 'allow') {
-
+  let credentials = {}
   function savePassword(username/*: ?string*/, pw/*: string*/, origin/*: string*/, action/*: string*/) {
     chrome.ipc.send('save-password', username, pw, origin, action)
   }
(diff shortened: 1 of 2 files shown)


bbondy:
++ and no more mutation observers for the future.

app/extensions/brave/content/scripts/passwordManager.js
```diff
@@ -3,7 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 if (chrome.contentSettings.passwordManager == 'allow') {
-
+  let credentials = {}
```
diracdeltas: `tryAutofillForm` no longer needs credentials as an argument

app/extensions/brave/content/scripts/passwordManager.js
```diff
@@ -45,11 +45,10 @@ if (chrome.contentSettings.passwordManager == 'allow') {
     }
 
     if (credentials[action]) {
-      credentials[action].push([passwordElem, usernameElem])
-    } else {
-      credentials[action] = [[passwordElem, usernameElem]]
+      return
```
diracdeltas: if there are mutiple forms on a page with the same action, all of them should be autofilled. i believe this change causes only the first to be autofilled.

diracdeltas: i just made a test case for this so i can fix

bridiver: ah, I assumed each action had a single username/password element. The purpose of this change was to avoid adding another listener on each check so just watch out for that with the fix

PR #3786, opened 2016-09-07
Brave doesn't remember last position and size if homepage is set

Sh1d0w:

Brave doesn't remember last position and size if you set it to start with homepage

When closed Brave does not retain last window position and size as it should be when re-launched.

This happens only if in Settings -> General you set the value for Brave starts with to My homepage. If you select Session and tabs from last time it correctly remembers the position and the size of the window.

Fixes ##3754

Auditors
@bbondy

Test Plan:
1. In settings -> general set Brave starts with to My homepage
2. Launch Brave, move the window around and close
3. Re-launch Brave and compare the last window position

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -59,7 +59,14 @@ module.exports.saveAppState = (payload, isShutdown) => {
         wndPayload.tabs = wndPayload.tabs.filter((tab) => !tab.isPrivate)
       })
     } else {
-      delete payload.perWindowState
+      // Reset the window state except for the ui data
+      payload.perWindowState.forEach((wndPayload) => {
+        Object.keys(wndPayload).forEach((key) => {
+          if (key !== 'ui') {
+            delete wndPayload[key]
+          }
+        })
+      })
     }
 
     try {


app/sessionStore.js
```diff
@@ -59,7 +59,14 @@ module.exports.saveAppState = (payload, isShutdown) => {
         wndPayload.tabs = wndPayload.tabs.filter((tab) => !tab.isPrivate)
       })
     } else {
-      delete payload.perWindowState
+      // Reset the window state except for the ui data
+      payload.perWindowState.forEach((wndPayload) => {
+        Object.keys(wndPayload).forEach((key) => {
+          if (key !== 'ui') {
+            delete wndPayload[key]
```
bbondy: I think this is removing all tabs and frames and etc from the state of the window which I don't think is the intention.

app/sessionStore.js
```diff
@@ -59,7 +59,14 @@ module.exports.saveAppState = (payload, isShutdown) => {
         wndPayload.tabs = wndPayload.tabs.filter((tab) => !tab.isPrivate)
       })
     } else {
-      delete payload.perWindowState
+      // Reset the window state except for the ui data
+      payload.perWindowState.forEach((wndPayload) => {
```
bbondy: I think there could be a case somewhere where payload.perWindowState is empty.

app/sessionStore.js
bbondy: This seems like it's removing all open tabs and frames etc, is this intentional here?

Sh1d0w: Yes, previously it deleted everything now it preserves only the UI position params. If you selected the option to start only with home page you don't want tabs and frames from the previousl launch, no? :)

app/sessionStore.js
Sh1d0w: Since it will be an empty array I don't see any problem if it is not unset, because on the next initialization it will be set to empty array again. Let me know if you want me to add check tho

app/sessionStore.js
bbondy: I see but I'm still concerned we're deleting too much. New state gets added here all the time and stuff not related to windows from last time.

app/sessionStore.js
bbondy: specifically [] is not falsy so maybe the check in the block above is superfluous `if (payload.perWindowState && savePerWindowState) {` but it makes me wonder if it can be undefined and not empty sometimes. It could be an old check.

bbondy:
I think you have to change the default window size in appstate and not window state.

Issue #3629, opened 2016-08-31
bitcoin wallet address overflows on some screen resolutions


diracdeltas:
it should always be on the same line as the copy icon, not like:
 
<img width="845" alt="screen shot 2016-08-31 at 6 07 47 pm" src="https://cloud.githubusercontent.com/assets/549654/18150387/28d1d0b0-6f9b-11e6-8cef-315237222243.png">

diracdeltas:
@jkup i think this will be irrelevant with the new ledger add funds panel design, but just keeping this open in case we do another release before that ships.

bbondy:
Moving to 0.12.1 but please close if it's no longer applicable.

mrose17:
related to https://github.com/brave/browser-laptop/issues/3484

alexwykoff:
Marking this as 'no qa needed' since this particular dialog no longer exists.

Issue #3484, opened 2016-08-27
clipboard icon on add funds popover probably needs to be moved


mrose17:
the clipboard icon should probably go to the left of the bitcoin address with a little extra horizontal size for the pop-over...
![2](https://cloud.githubusercontent.com/assets/400552/18029256/e68d3e2a-6c47-11e6-98b0-da8bbe6d92ba.png)

mrose17:
related to https://github.com/brave/browser-laptop/issues/3629

alexwykoff:
Marking this as 'no qa needed' since this particular dialog no longer exists.

PR #3803, opened 2016-09-08
new dialog design for adding funds

### LINKED ISSUE #3677: New (and much improved) layout for Add funds... dialog
This dialog has many improvements based on feedback from Twitter user reports.

- some users were drawn to the coinbase logo and clicked, expecting to send a payment from their existing CB wallet
- original button text for the CB buy widget was "Buy With Coinbase" and may be misleading people to think this is the button to use their existing CB account
- the large QR code may have pushed users to the first option simply because they assumed it was not useful to them, thus entering the wrong path based on subconscious process of elimination. 
- customers outside of the USA are denied access to the CB buy widget so we want to hide that option for users outside of the USA.

New items:

- hid the QR code under a new sub-dialog
- added app buttons if user does want to use the QR code with their phone
- created clear separation between using the QR code and the button to either copy the brave wallet address or send it through the URI handler
- Changed the title to immediately call out the "Three ways"
- added concise options text
- added footnote text
- reduced CB logo and moved it to the bottom of the dialog
- refined layout and button styles

image

jkup:

Fixes #3677 #3484 #3629

Test Plan:

This is still in progress. Just wanted to push a PR so I can get some feedback.


app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -38,14 +38,15 @@ tableEmptyText=No table data.
 notificationEmptyText=Top publisher visits
 syncEmptyText=Sync settings coming soon.
 bitcoin=Bitcoin
-bitcoinAdd=Use an existing Bitcoin wallet
+bitcoinAdd=Use your existing bitcoin wallet/account
```
diracdeltas: Bitcoin not bitcoin

app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -38,14 +38,15 @@ tableEmptyText=No table data.
...
+bitcoinAddDescription=Use any BTC wallet that can transfer bitcoin to your Brave wallet.
```
diracdeltas: same as above

app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -38,14 +38,15 @@ tableEmptyText=No table data.
...
 bitcoinBuy=Buy Bitcoin
 viewPaymentHistory=View Payment History
 paymentHistoryTitle=Your Payment History
 paymentHistoryFooterText=Your next payment submission is {{reconcileDate}}.
 paymentHistoryOKText=OK
 bitcoinAddress=Your Brave wallet address is:
-bitcoinPaymentURL=Scan the QR code or transfer {{amount}} to:
+bitcoinPaymentURL=Your Brave wallet address
```
diracdeltas: this is no longer the payment URL, so it should be merged into `bitcoinAddress` above

js/about/preferences.js
```diff
@@ -260,9 +260,20 @@ class BitcoinDashboard extends ImmutableComponent {
...
+  get bitcoinOverlayContent () {
     return <iframe src={this.ledgerData.get('buyURL')} />
   }
+  get qrcodeOverlayContent () {
+    return <div>
+      <img src={this.ledgerData.get('paymentIMG')} title='Brave wallet QR code' />
+      <div className='bitcoinQRTitle' data-l10n-id='bitcoinQR' />
+    </div>
+  }
+  get qrcodeOverlayFooter () {
```
diracdeltas: i don't think the QR code overlay should show the coinbase logo since it doesn't depend on coinbase at all

less/modalOverlay.less
```diff
@@ -57,26 +57,56 @@
         overflow: hidden;
       }
     }
+
+    .qrcodeOverlay {
```
diracdeltas: this looks somewhat off-center
screen shot 2016-09-09 at 6 01 12 pm

diracdeltas:
the case where Coinbase isn't available (for funds > $5/month) needs some CSS love:
screen shot 2016-09-09 at 6 06 27 pm

js/about/preferences.js
jkup: @bradleyrichter how does that sound to you? I was just going off your mock https://cloud.githubusercontent.com/assets/13509546/18232727/0a78554c-728b-11e6-9146-e8196a95e39d.png

diracdeltas:
lgtm, thanks

PR #3809, opened 2016-09-08
Change TM symbol to (R) symbol for password managers in preferences

anseljh:
Because "1Password", "Dashlane", and "LastPass" are all *registered* US trademarks, it's best to show the &reg; symbol for registered trademarks instead of the &trade; symbol, which is for unregistered or registration-pending trademarks.

This commit only changes the `en-US` locale.

* LastPass: https://tsdr.uspto.gov/#caseNumber=86059768&caseType=SERIAL_NO&searchType=statusSearch
* 1Password: https://tsdr.uspto.gov/#caseNumber=77872651&caseType=SERIAL_NO&searchType=statusSearch
* Dashlane: https://tsdr.uspto.gov/#caseNumber=85290031&caseType=SERIAL_NO&searchType=statusSearch

Disclaimer: IANYL :)

Test Plan: I don't *think* there are any tests implicated here, but I really don't know that for sure.


bsclifton:
@anseljh thanks for this :smile: cc: @bradleyrichter for awareness on registered trademark vs TM

Issue #3830, opened 2016-09-08
a publisher should get a visit increment only on a new tab...


mrose17:
...not when switching to an existing tab

bradleyrichter:
@alexwykoff this is the source of the new UX issue?

alexwykoff:
I think there's a bit of nuance here versus https://github.com/brave/browser-laptop/issues/4274 . In this case @mrose17 is stating that if there are more than one tab open and a user flips back and forth between them, that should NOT increment the visit count.
However, if the user opens a new tab and points it to a publisher already existing in the previously opened tabs, that first visit counts.

PR #3852, opened 2016-09-09
Harmonize publisher visits

mrose17:
Fixes https://github.com/brave/browser-laptop/issues/3830

test plan:

open two windows. in one window, put the payments pane; in the other two tabs on the same publisher but different URLs. observer the visits and duration counts in the first window and then switch back and forth between tabs (waiting one or two minutes between each switch). the visits should stay constant and the duration should increase.


app/ledger.js
```diff
@@ -470,7 +470,8 @@ var updatePublisherInfo = () => {
     var entries = []
 
     underscore.keys(publishers[publisher]).forEach((location) => {
```
diracdeltas: can just do `underscore.each(publishers, ...)`

diracdeltas:
please include a test plan in the commit message
@bridiver should also audit this

app/ledger.js
mrose17: +1

mrose17:
i will add a test plan. i had a brief conversation with @bridiver about this...

app/ledger.js
mrose17: in this case, no... because in the function i need access to both the key and value, not just the value...

mrose17:
fixes https://github.com/brave/browser-laptop/issues/3830

app/ledger.js
diracdeltas: both key and value are accessible http://underscorejs.org/#each

bridiver:
the eventStore changes look good to me. Discussed the definition of a visit with @mrose17 and that sounds ok too. @diracdeltas I think you are in a better position than I am to review the ledger code.

PR #3854, opened 2016-09-09
Optimize titlebar space on Windows

bsclifton:
Changes done, only a few minor bugs open (see checklist below under known issues)

Fixes https://github.com/brave/browser-laptop/issues/3036

Lots of changes; this PR:
- Menubar and caption button changes only affect Windows
  - removes the standard menubar from the Window
  - custom renders the caption buttons (the minimize / maximize / close buttons)
  - custom renders a menubar (if visible). Can be toggled with ALT (left and right ALT both work).
  - Windows 10, 8.1, 8, or other will use Windows 10 styles
  - Windows 7 will use Windows 7 styles
- Functionality for Linux was removed- this is captured separately with https://github.com/brave/browser-laptop/issues/3126
- Keyboard accelerators are now shown in context menus. This affects Mac and Linux

It also includes:
- remove unused `getParentMenuDetails` method

## Auditors
@bbondy @bridiver @bradleyrichter @Sh1d0w

#### Styles should match the mockup for the respective OS
image
image

#### Custom rendered menubar
- Menubar should be hidden by default with a fresh session
- If enabled, the menu will show all the time
- When toggled, the lion head should not move horizontally
- Menu should be usable in full screen mode (F11)

#### Custom render caption buttons
- Minimize / Maximize / Close
- Tooltip shows if you hover over it
- Buttons should work
- When maximized, maximize button turns into a restore icon
- Going full screen (F11) should hide the caption buttons

#### Using menubar w/ the keyboard
- Pressing any arrow keys should not interfere with the frame (unless menu is triggered using ALT)
- Menu should toggle between shown and hidden with ALT
- Navigation
  - When shown, menu will automatically highlight the "File" menu
  - When menu is "highlighted", you should be able to use arrow keys to move left and right and also up/down to:
      - show the initial context menu (the items under File or Edit, for example)
      - move your selection up/down, including wrapping around choose a different item
  - Once a menu item is highlighted, you should be able to trigger it by pressing ENTER
- menu can be hidden at any time (even if it's popped up) by pressing ALT or ESCAPE
- Menu should be readable/usable with usability options enabled:
win7-magnifier
win10-personalization
- Menu hides after choosing an item (unless auto-hide is disabled)
- Keyboard shortcuts making use of ALT should still work fine (and not toggle the menu)

#### Using menubar w/ the mouse
- You can click any menu and it'll show the associated popup
- If you clicked File, you should be able to move your mouse another item, like Edit, and it should automatically:
  - close the popup associated with File
  - open the popup associated with Edit
  - change the selection behind the menu name
- Menu hides after choosing an item (unless auto-hide is disabled)
- Picking copy or paste from Edit menu should not affect focus of the input box on a given page
- Clicking in the middle of a page should unselect the menu item and hide any popup menus that were showing

#### Other style changes
- window should have a 1 pixel black border around it now
- Context menus now show keyboard shortcuts
image
- Menus should be displayable in other languages without issue
- Window should be able to be made smaller without screwing up the URL bar / caption button area

#### Misc things to try
- toggle the auto-hide menu bar (also can be triggered by right clicking in bookmarks toolbar or tab row and picking "Menu bar")
- try using the long press back button / forward button. These also use our custom context menus
- toggle the magic-titlebar (always show URL) setting. Does the window do anything weird during transition?
- resize the window (small, big, push F11 and go fullscreen)

### Mac / Linux
- Custom titlebar should not be showing
- You should see the keyboard shortcuts now in our custom context menus. For example, the hamburger button.
- mac menu
- Keyboard shortcuts should be in the proper order (shift before command, etc) and should be displayed, regardless of locale
- Keyboard shortcuts should not be affected
  - Alt on Linux, option on Mac
  - Test keyboard shortcuts that use this
- Popup elements should still work correctly
  - clicking on lock for security mode
  - adding / editing bookmark
  - clicking lion head to bring up site specific bravery panel
  - adding new autofill data in preferences

**NOTE**: _besides the min/max/close tests (which aren't runnable), there are no other automated tests. I'd like to create some, but webdriver tests on Windows are not working_  

## Known issues (will be working through these)
Issues towards the top of each respective category are higher priority. Issues which are considered show-stoppers are bolded (please feel free to edit this comment).

Issues with the checkbox are FIXED (see follow up commits for fix)

### functional issues
- [x] **Most menu items don't work when selected; they rely on input being passed to the click handler**
- [x] **The lion icon moves around when menu is toggled**
- [x] Menu needs to auto close/expand when you move your cursor from one menu to another
- [x] Menubar needs to de-select the MenubarItem once context menu option is selected or context menu gets blurred
- [x]  After pressing ALT menu shows- but isn't dismissed on a click that isn't the menu (when autohide menubar = true)
- [x] Menu does NOT respect the "Hide the menu bar by default" setting
- [x] Need to show hotkeys in the menu (shortcut keys, accelerators)
- [x] No resize cursor on top corner of window (both on top and on edge and on exact corner)
- [x] Feature needs a setting to disable, for folks that would have negative accessibility impacts
- [x] There is dead space under the min/max/close buttons that cannot be grabbed. This already has the `-webkit-app-region: drag;` style but isn't taking 100% height :frowning:
- [x] Menus should be navigable by keyboard (alt then arrow keys)
- [x] Right click in certain areas brings up restore/move/etc menu (not the expected context menu). This seems to be related to the CSS drag attribute **edit**: _root cause is that different OSes handle this differently. Windows did not treat these areas as draggable (even with the attribute) until the window became frameless. [See frameless docs for more info](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#user-content-context-menu).
- [x] Some menu items are not being selected when hovered?! **edit**: _after updating to allow menu to work w/ keyboard and fixing drag attributes, I can't reproduce this on Win7 or Win10. Marking as fixed.
- [x] Don't show buttons in full screen mode (or is full screen mode working properly?)
- [x] cut/copy are stealing focus
- [x] cut/ copy not working from menu on URL bar
- [ ] first context menu item has issues (Windows 7 specific? able to fix in dev tools by removing border. Element is positioned absolutely and has transparency in BG color... needs more investigation) 
- [ ] resize handle on top right of window not showing properly
- [ ] keyboard does not work with submenus (ex: bookmarks menu when you have folders)

### style issues
- [x] back/forward buttons are oval shaped (not circle) also causes spacing to be messed up:
image
- [x] space between menu text needs to be 22px
- [x] Font for the menu can be much larger than app, because it uses system size (we can override this and make it a static size, if we'd like)
image
- [x] Windows caption buttons are supposed to show a title of "Maximize" or "Minimize" when mouse is left hovering over them
- [x] Context menu accelerators need restyling (need to be the same size as regular text)
- [x] Needs to show border
    - Windows 10 always puts a 1 pixel border with the users highlight color around each window... instead, a box shadow is drawn (which looks nice, but isn't consistent with OS)
    - Windows 7 doesn't use any effects around the border and when you're on a white page, it can be really confusing- because there is no borders and the white will blend in (making it hard to tell what is in the browser and what isn't)
- [x] Window needs a smaller min-width than 700px; currently, when less than that, the caption buttons get messed up.


bbondy:
btw this needs a rebase on master when you get time

bsclifton:
rebase done :+1: manually resolved conflicts with the new extension code

bsclifton:
OK properly rebased and verified :smile:

bsclifton:
Ready for review- @bbondy @srirambv @darkdh any feedback is appreciated :smile: I'll try to find some Windows contributors who can also try it and give feedback

bsclifton:
Just a heads up- this can be test on Mac, BTW
- To make sure I didn't break anything
- To see the new sweet keyboard accelerators in the context menus (burger menu for example)

srirambv:
@bsclifton The new look is really good and looks neat. 

Some observations:

1. When custom title bar is disabled and enabled it shows two sets of windows buttons
image

2. Block script and Favorite icons are not properly aligned when hide URL bar is enabled 


3. File menu/Hamburger menu text size is very small compared to the current version which could be a problem

4. Window buttons are shown even if the browser is in fullscreen mode 

5. Mouse over highlight doesn't happen on some of the menu items


6. If the window is too small then the Sheilds and Password manager icons are hidden. It is displayed when you are on one of the about pages. The window buttons gets stacked one above the other

bsclifton:
Thanks for testing, @srirambv :smile: 

#1 is a good concern; I'd like to defer to @bbondy and @bradleyrichter. Basically, if you go into settings, I've made this new feature configurable (enabled by default, can be turned off). If you do choose to toggle it, you are prompted by a "Restart needed. Do it now? yes/no". However you're not forced to do this.  What do you all think is a good experience? 
- Force a restart by bringing up an OK/cancel dialog?
- remove it as a configurable option for now?
- other options?

#2 looks to be caused by https://github.com/brave/browser-laptop/issues/4007

#3 I'll defer to @bradleyrichter- the font size is 12px (roughly 9pt) which is the default system size. On bigger monitors, this might be unacceptable.

#4 great catch! :smile: I'll add that to my TODO list

#5 will need to investigate... I definitely did not have that experience

#6 is already captured in my help needed, trying to make the window respond better when below 700px width

srirambv:
@clifton Would suggest a Ok/Cancel dialog than the notification bar. Goes along with the standard windows alert messages. This way the window remains as is and only applies the change after a restart is done.

bradleyrichter:
@srirambv These mini-screen captures are priceless. And addictive!

@bsclifton I'm confused about #1. We never want to see the standard non-custom titlebar. That is the whole point of this effort. Am I missing something?

Font size - we need to match the standard default in all cases - somehow copying the approach that WIN uses for it's view settings. Later we can try to support the user-overrides from the system.

bsclifton:
@bradleyrichter for 1, that is a completely acceptable answer :smile: I made it toggle-able just in case there were folks with accessibility requirements (which the custom menu may not work).  We can always revisit that later though

bbondy:
Looking a lot better than my last pass.

A couple of things I noticed.
- I think people will complain that they can't use the keyboard up / down arrows / right arrow to go to next menu, left for previous menu.  Alt + arrows to navigate to item and hit enter to open. This is probably a bigger thing but we should handle it.   As a side note, it'd be nice to have this for hamburger menu too.
- Items like cut, copy, paste, paste with formatting (possibly others) are taking focus away from the page and then the operation doesn't work.  This is both for if your cursor is in the urlbar or just selecting something on a page.
- Minor thing but lion head looks like it needs a couple px more padding on the right to me.
- The top right corner for resizing the window is really hard to make shot up.  Whereas works reliably for other windows.

bsclifton:
Ready for review :smile: 

@alexwykoff @srirambv @luixxiul (and everyone else too) - please let me know what you think about the QA steps specified above.  Are they clear? @bbondy please check them too, in case I missed anything

Commits were squashed into logical blocks:
- The first rev work by @Sh1d0w 
- Removing the Linux features
- My initial implementation

and each commit after that was squashed on a daily basis

srirambv:
@bsclifton 

New observations in the latest commit

1. Favorite and block script icons not shown in title mode
2. Dashlane widget is not opening even if you are signed into the app. Doesn't show the green dot in icon. 1PW and Lastpass works fine opens up the widget window
3. With small screen the close button moves to next row when title mode changes to URL mode


4. If URL is long then the window buttons and Shields not shown in smaller window size resolution



5. New Private tab opens a new tab with chrome-extension://mnojpmjdmbbfmejpflffifhffcmidifd/about-newtab.html?devServerPort=8080 (Is this specific to this branch running locally?)


6. Mouse over on history entries doesn't work.
7. History search result comes up for non-existent values



Suggestion for History page
Could possibly add sort up/down arrows  next to the section headings.

app/common/lib/formatUtil.js
```diff
@@ -0,0 +1,91 @@
...
+  // also, replace w/ name or symbol
+  if (process.platform === 'darwin') {
+    splitResult.sort(function (left, right) {
+      if (macOrderLookup(left) === macOrderLookup(right)) return 0
+      if (macOrderLookup(left) > macOrderLookup(right)) return 1
+      return -1
+    })
+    // NOTE: these characters might only show properly on Mac
+    result = splitResult.join('')
+    result = result.replace('CmdOrCtrl', '⌘')
```
bbondy: nit: Please also support CommandOrControl, we don't use it yet but we might in the future.

app/common/lib/formatUtil.js
```diff
@@ -0,0 +1,91 @@
...
+    result = result.replace('Control', '^')
+    result = result.replace('Ctrl', '^')
+  } else {
+    splitResult.sort(function (left, right) {
+      if (defaultOrderLookup(left) === defaultOrderLookup(right)) return 0
+      if (defaultOrderLookup(left) > defaultOrderLookup(right)) return 1
+      return -1
+    })
+    result = splitResult.join('+')
+    result = result.replace('CmdOrCtrl', 'Ctrl')
```
bbondy: ditto CommandOrControl

app/common/lib/formatUtil.js
```diff
@@ -0,0 +1,91 @@
...
+  return result
+}
+
+/**
+ * Clamp values down to a given range (min/max).
+ * Value is wrapped when out of bounds. ex:
+ *   min-1 = max
+ *   max+1 = min
+ */
+module.exports.wrappingClamp = (value, min, max) => {
```
bbondy: You could just use `const calmp = require('lodash/clamp')` and remove this

app/sessionStore.js
```diff
@@ -412,6 +412,7 @@ module.exports.defaultAppState = () => {
         guid: [],
         timestamp: 0
       }
-    }
+    },
+    menubar: {}
```
bbondy: shouldn't this be stored in windowState instead? Wouldn't this mirror to all open windows?

less/navigationBar.less
```diff
@@ -29,6 +29,393 @@
...
+            left: 10px;
+            border-radius: 0;
+          }
+          .widget3 { display: none; }
+          .widget4 { display: none; }
+          .widget5 { display: none; }
+        }
+      }
+
+      &.close {
```
bbondy: If you mouse over the close button and slowly move your mouse up past the top of the window, the red stays solid.

bbondy: It's also really hard to get the resize diagonal cursor when you mouse over the top right corner of the window.  The other corners are fine.

app/renderer/components/menubar.js
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * Menubar that can be optionally be displayed at the top of a window (in favor of the system menu).
+ * First intended use is with Windows to enable a slim titlebar.
+ * NOTE: the system menu is still created and used in order to keep the accelerators working.
+ */
+class Menubar extends ImmutableComponent {
+  constructor () {
+    super()
+    this.onKeyDown = this.onKeyDown.bind(this)
+    document.addEventListener('keydown', this.onKeyDown)
```
bbondy: better to make it symmetric with componentWillMount since you do it on unmount.

bsclifton:
This response is a WIP as I reply back to issues found by @srirambv 

1. [This appears to have been intentionally done](https://github.com/brave/browser-laptop/commit/af8b203ea106af8173302b72c6bef0504a714cc0) (and not a result of any changes here)

2. looking at this now...

app/renderer/components/menubar.js
```diff
@@ -0,0 +1,216 @@
...
+    const rect = e.target.getBoundingClientRect()
+    showContextMenu(rect, this.props.submenu, this.props.activeFrame)
+  }
+  onMouseOver (e) {
+    const selected = this.props.menubar.props.selectedLabel
+    if (selected && selected !== this.props.label) {
+      this.onClick(e)
+    }
+  }
+  render () {
```
bbondy: Currently you aren't respecting the `visible: false` for menu templates which makes find next / find prev show up int he Edit menu as items but not work. These should be hidden but the shortcuts should still work.

bsclifton: Really good find! :smile:

js/stores/appStore.js
```diff
@@ -253,7 +257,7 @@ function windowDefaults () {
     show: false,
     width: appState.get('defaultWindowWidth'),
     height: appState.get('defaultWindowHeight'),
-    minWidth: 480,
+    minWidth: isWindows ? 640 : 480, // Windows has caption buttons (min/max/close) which take extra space
```
bsclifton: OK will check those out- I haven't had success running the webdriver tests on Windows (yet)

app/common/lib/formatUtil.js
bsclifton: and doh- can't use because it doesn't wrap

app/sessionStore.js
bsclifton: The original intention for doing this was that there is only one system menu (and each Window would reference that).  But let me revisit this thought- thanks for bringing it up :smile:

bsclifton: will keep it here after talking w/ @bridiver; in the future, we can introduce more window specific state in the appState (deprecating windowStore)

js/stores/appStore.js
bsclifton: change backed out- solution was to (on Windows only) reduce the size of elements so that the window can go as skinny as 480 and still look correct

bbondy:
If you work out the copy / paste focus of items thing then I can merge this

bbondy:
Let's get this merged and do anything else as follow ups. Nice work on all of this 🎉

Issue #3257, opened 2016-08-19
ledger notification should have the style of the update bar


diracdeltas:
per https://github.com/brave/browser-laptop/issues/2407

Alert:
Hello! Your Brave Payments account is waiting for a fresh deposit.

Buttons:
[Later] [ Add funds... ]

Trigger:
1 day, and 5 days after a failed payment

____________________
Alert:
Hello! Are you ready to support the sites you use most?

Buttons:
[Not yet ]  [ Sure, I'll try... ]

Trigger:
10 days after first run of Brave, if user has not yet enabled BP.
_____________________
Alert:
Hello! Thanks for supporting your favorite websites! Your contribution of $15.00 was processed! 

Buttons:
[Ok]

Trigger:
Immediately, upon ledger confirmation of contribution processing.

visual spec:

![image](https://cloud.githubusercontent.com/assets/13509546/18334070/680effc6-7529-11e6-9273-f27fecc52423.png)

diracdeltas:
@ayumi there is a notification bar option i added, `updateStyle`, which is supposed to indicate that the NotificationBar should be styled like an UpdateBar. currently this option does nothing

ayumi:
I implemented custom notification styling in https://github.com/brave/browser-laptop/pull/3856.
To finish this :gift: we need to add the 3 ledger notifications.

mrose17:
@ayumi - who needs to do the additions? is this a design thing, a coding thing, a policy thing? thanks!

ayumi:
Just a coding thing now – adding the triggers and alerts.

mrose17:
_excellent!_

bbondy:
Anything else need to block 0.12.2 for this?

ayumi:
I'm stuck on how to test the "thanks for your contribution" notification – I'm unable to trigger a ledger payment on demand. @mrose17 is there a way to try that?

Here it is implemented pending verification: https://github.com/brave/browser-laptop/compare/master...feature/ledger-notifications-more

mrose17:
sadly, there isn't a way to force the ledger to trigger a contribution on demand, other than to edit `ledger-state*.json` and watch it go. what is the PR for this? i can take a look in about 30m. thanks!

ayumi:
@mrose17 PR #4140

PR #3856, opened 2016-09-10
Style ledger notifications like update notifications

ayumi:

Partially implements #3257

- Extract/refactor updateBar HTML and CSS for use in notificationBar
- Add custom styling of notification bar and buttons

Auditors: @bradleyrichter @diracdeltas 

Test Plan:
1. Change conditional in https://github.com/brave/browser-laptop/blob/feature/ledger-notifications-update-style/app/ledger.js#L1220 to `true`
2. Observe the funding notification is styled like the update bar

Preview:
screen shot 2016-09-09 at 17 26 40

### DIFF
--- a/js/stores/appStore.js
+++ b/js/stores/appStore.js
@@ -360,7 +360,9 @@ const handleAppAction = (action) => {
         })
 
         appActions.showMessageBox({
-          buttons: [locale.translation('ok')],
+          buttons: [
+            {text: locale.translation('ok')}
+          ],
           options: {
             persist: false
           },
--- a/js/stores/appStore.js
+++ b/js/stores/appStore.js
@@ -507,13 +507,34 @@ const handleAppAction = (action) => {
       break
     case AppConstants.APP_SHOW_MESSAGE_BOX:
       let notifications = appState.get('notifications')
-      appState = appState.set('notifications', notifications.filterNot((notification) => {
+      notifications = notifications.filterNot((notification) => {
         let message = notification.get('message')
         // action.detail is a regular mutable object only when running tests
         return action.detail.get
           ? message === action.detail.get('message')
           : message === action.detail['message']
-      }).push(Immutable.fromJS(action.detail)))
+      })
+
+      // Insert notification next to those with the same style, or at the end
+      let insertIndex = notifications.size
+      const style = action.detail.get
+        ? action.detail.get('options').get('style')
+        : action.detail['options']['style']
+      if (style) {
+        const styleIndex = notifications.findLastIndex((notification) => {
+          return notification.get('options').get('style') === style
+        })
+        if (styleIndex > -1) {
+          insertIndex = styleIndex
+        } else {
+          // Insert after the last notification with a style
+          insertIndex = notifications.findLastIndex((notification) => {
+            return typeof notification.get('options').get('style') === 'string'
+          }) + 1
+        }
+      }
+      notifications = notifications.insert(insertIndex, Immutable.fromJS(action.detail))
+      appState = appState.set('notifications', notifications)
       break
     case AppConstants.APP_HIDE_MESSAGE_BOX:
       appState = appState.set('notifications', appState.get('notifications').filterNot((notification) => {
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -352,7 +352,10 @@ function registerPermissionHandler (session, partition) {
     }
 
     appActions.showMessageBox({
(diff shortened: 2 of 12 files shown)


bradleyrichter:
sweet! thanks!

diracdeltas:
i haven't looked into why, but this appears to be applying the style of the first notification bar to all the notification bars below it, regardless of whether they are ledger notifications.

to repro:
1. go to https://www.browserleaks.com/geo, see the geolocation notification bar appear
2. close browser, edit https://github.com/brave/browser-laptop/blob/feature/ledger-notifications-update-style/app/ledger.js#L1220 to true
3. re-open browser
4. observe that geolocation notification is styled like ledger notifications now
screen shot 2016-09-09 at 6 14 22 pm

if you close the first notification so that the geolocation notification is on top, then the ledger notification loses its style
screen shot 2016-09-09 at 6 15 05 pm

otherwise i feel it

ayumi:
I initially tried to style each notification separately, but that looked really ugly due to variations in margin/padding.
So I decided it'd be an adequate fallback to style them all like the first notification.

Alternatively we could sort and group notifications by style, and separate them with borders on top/bottom.

ayumi:
I removed the style override based on the 0-th notification, and added borders. What do you think?

screen shot 2016-09-09 at 19 39 28

diracdeltas:
i feel it even more now

bradleyrichter:
We need to sync up the style on the yellow alert. Can we use 3 buttons instead?

No / yes / always 

Or similar....?

diracdeltas:
> We need to sync up the style on the yellow alert. Can we use 3 buttons instead?

It would have to be four buttons (No, Yes, Always No, Always Yes)

ayumi:
@diracdeltas Notifications are now sorted by style and prioritize styled notifications over the defaults:
screen shot 2016-09-09 at 23 18 18

@bradleyrichter Besides restyling the 2 ledger notifications, this doesn't affect other notifications so I think it's out of scope for this PR.

js/stores/appStore.js
```diff
@@ -507,13 +507,34 @@ const handleAppAction = (action) => {
...
         // action.detail is a regular mutable object only when running tests
         return action.detail.get
           ? message === action.detail.get('message')
           : message === action.detail['message']
-      }).push(Immutable.fromJS(action.detail)))
+      })
+
+      // Insert notification next to those with the same style, or at the end
+      let insertIndex = notifications.size
+      const style = action.detail.get
```
bsclifton: for checking against immutable, you might check for the presence of toJS instead; I know a lot of other code does that (just a consistency nitpick)

js/stores/appStore.js
```diff
@@ -507,13 +507,34 @@ const handleAppAction = (action) => {
...
+      if (style) {
+        const styleIndex = notifications.findLastIndex((notification) => {
+          return notification.get('options').get('style') === style
+        })
+        if (styleIndex > -1) {
+          insertIndex = styleIndex
+        } else {
+          // Insert after the last notification with a style
+          insertIndex = notifications.findLastIndex((notification) => {
+            return notification.get('options').get('style')
```
bsclifton: to be explicit, you could coerce this to a bool, like:
`return !!notification.get('options').get('style')`

But usually when something has that "smell", it's best to just check the type:
`return typeof notification.get('options').get('style') === 'string'`

bsclifton:
Comments left on the last commit :+1:

js/stores/appStore.js
```diff
@@ -505,13 +507,34 @@ const handleAppAction = (action) => {
...
         return action.detail.get
           ? message === action.detail.get('message')
           : message === action.detail['message']
-      }).push(Immutable.fromJS(action.detail)))
+      })
+
+      // Insert notification next to those with the same style, or at the end
+      let insertIndex = notifications.size
+      const style = action.detail.get
+        ? action.detail.get('options').get('style')
```
diracdeltas: for future reference, you can just do `thing.getIn(['options', 'style'])` to avoid it throwing an error if `thing.get('options')` is falsy

diracdeltas:
this seems to cause some of the notification bar tests to fail. https://travis-ci.org/brave/browser-laptop/builds/159031606

otherwise lgtm, feel free to merge!

js/stores/appStore.js
ayumi: Thanks! I'm checking that it's a string now.

ayumi:
@diracdeltas thx, i've updated navigationBar test to use the new css selector

Issue #2839, opened 2016-08-01
Set correct application type under Linux


Sh1d0w:
**Did you search for similar issues before submitting this one?**

Yes

**Describe the issue you encountered:**

Currently the brave.desktop launcher which is generated, is not well formated so Gnome desktop for example could not recognize Brave as a web browser, thus it does not appear in default applications list. When you fix this is will allow the users to set Brave as default browser from the Gnome preferences for default applications

Currently this is what we have:

    cat /usr/share/applications/brave.desktop        

    [Desktop Entry]
    Name=brave
    Comment=Brave laptop and desktop browser
    GenericName=brave
    Exec=brave %U
    Icon=brave
    Type=Application
    StartupNotify=true
    Categories=Network;
    MimeType=text/plain;

**Expected behavior:**

This is what it should look like as a browser application:

    [Desktop Entry]
    Name=brave
    Comment=Brave laptop and desktop browser
    GenericName=brave
    Exec=brave %U
    Icon=brave
    Type=Application
    StartupNotify=true
    Categories=Network;WebBrowser;
    MimeType=text/html;text/xml;application/xhtml_xml;image/webp;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;

- Platform (Win7, 8, 10? macOS? Linux distro?):

Linux, Fedora 24 Gnome, Ubuntu 16.04

- Brave Version:
0.11.1
- Steps to reproduce:
    1. In gnome shell search for Details
    2. Click on Default Applications

- Screenshot if needed:

- Any related issues:
#1336

willy-b:
@Sh1d0w, I just submitted a PR that sets the MimeType field to what you provided. (At least on Debian, as `electron-installer-redhat` doesn't yet allow the MimeType in the `.desktop` file to be configured.)

For comparison, the MimeType entry for Firefox's `.desktop` file looks like:
```
MimeType=text/html;text/xml;application/xhtml+xml;application/xml;application/rss+xml;application/rdf+xml;image/gif;image/jpeg;image/png;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;x-scheme-handler/chrome;video/webm;application/x-xpinstall;
```
(from `firefox-mozilla-build.desktop` in `firefox-mozilla-build_47.0.1-0ubuntu1_i386.deb`)

The WebP format looks awesome. Should we also add image/gif and image/png? Or should we leave that to other apps?

Sh1d0w:
[...]

PR #3905, opened 2016-09-12
Fix Redhat desktop launcher

### LINKED ISSUE #3776: Can’t set as default browser under Linux (.desktop)
The `brave.desktop` file only contains `MimeType=text/plain;` despite that `linuxPackaging.json` indicates that [there should be more](https://github.com/brave/browser-laptop/blob/59ca343b00/res/linuxPackaging.json).

Problem in latest public beta and in the more recent brave-0.12.0-1.amd64.rpm. I’ve not checked the .deb.

At least `text/html;x-scheme-handler/http;x-scheme-handler/https;` is required to set Brave as the default web browser in GNOME.

Sh1d0w:

The problem is that under RHEL the generated desktop launcher does not
have the correct mine types, so Brave can not be set as default browser.

This was reported and fixed here https://github.com/brave/browser-laptop/issues/2839
in both Brave and electron-redhat-installer package. However the fixed version
of the installer package was never used by Brave. This PR bumps the version
so it should work ok now.

Fixes #3776

Auditors:
@bbondy

Test Plan:
1. Install Brave under Fedora 24 and make sure it can be set as default
browser from the Default applications menu

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,7 +163,7 @@
   },
   "optionalDependencies": {
     "electron-installer-debian": "^0.3.0",
-    "electron-installer-redhat": "^0.2.0"
+    "electron-installer-redhat": "^0.3.0"
   },
   "standard": {
     "ignore": [


bbondy:
++

PR #3718, opened 2016-09-05
Address bar is not "active" when it should be

Sh1d0w:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Makes the url bar active when switching to a about:newtab
page, as the user is expecting to be able to enter his url
when he switches back to a new tab page.

Fixes #3712

Auditors:
@bbondy

Test Plan:
1- Open a new tab
2- Switch to the new blank tab
3- Go back to the tab you were browsing before you opened the new tab
4- Go to the blank new tab you opened in step 1

On step 4, the address bar should be "active", so users could be able to write the web address without need to click on it.

PR #3933, opened 2016-09-12
Address bar should be active for about:newtab

### LINKED ISSUE #3712: Address bar should be active for about:newtab

No

Address bar is not "active" when it should be

**Expected behavior:**

Address bar "active", so user could add the web address without need to click on it.

Ubuntu

- Brave Version:  

Brave: 0.11.6

- Steps to reproduce:
    
1- Open a new tab
2- Switch to the new blank tab
3- Go back to the tab you were browsing before you opened the new tab
4- Go to the blank new tab you opened in step 1

On step 4, the address bar should be "active", so users could be able to write the web address without need to click on it.

- Screenshot if needed:

- Any related issues:

Sh1d0w:

This issue was fixed via #3718
but was returned with feedback from the QA.

Now it should properly handle switching to new tab via shortcut or mouse.

Fixes #3712 
Auditors:
@bbondy

Test Plan:
While switching to a new tab eighter via shortcut or mouse click
the url bar should be focused.


bbondy:
thanks would you mind adding some tests for this?

bbondy:
does url bar active mean auto complete can be open? Just making sure we have to treat focused and active separately.  And can we get an accurate view of focused with this state always?

Sh1d0w:
@bbondy 

1. URL_BAR_ACTIVE = true - it is visible and not hidden by the page title
2. URL_BAR_SELECTED = true - means the text you have typed in is selected ([also url bar gains focus](https://github.com/Sh1d0w/browser-laptop/blob/283048910f37e1dc05fc19a2736dd3b337320036/js/stores/windowStore.js#L566))
3. URL_BAR_FOCUSED = true - means the focus is on the url bar 

Now these are 3 different states, and they can lay different results if combined together. For example if you type something in the url bar, blur and trigger focus, the cursor will blink at the end of the text. But if you do the same and instead trigger url bar selected and the focus the text you typed in will be selected and if you start typing you will overwrite it.

I tried to replicate the behaviour of FF and Chrome when you switch tabs.

Will add some more tests today or tomorrow. If this blocks the current release you can move it for the next one.

bbondy:
Thanks, I'll merge but pls add some tests because the interactions can break in the future.

PR #3937, opened 2016-09-12
Button icons are style-wonked

### LINKED ISSUE #3917: Button icons are style-wonked. Must fix.
Reload and BM at least...

image

image

also effects home, noscript per @diracdeltas 
image

jkup:

Fix #3917 

Test Plan:

Make sure that refresh and stop icons are correctly centered.

Ok, so what happened here was we turned the span.browserButton selector into just .browserButton so it could be used as a mixin (https://github.com/brave/browser-laptop/commit/c63165f3a1c5b3d036b52234a6d7ac53c32c2f5b) but the problem is that then .browserButton conflicts with .fa and .fa is cancelling out the .browserButton line-height and font-size and some other things.

For a quick fix I just added back the element specific classes so it can still be used as a mixin but we should figure out a cleaner way to name classes / embed selectors so we don't have to hack around specificity.

Auditors @bradleyrichter @ayumi

### DIFF
--- a/less/button.less
+++ b/less/button.less
@@ -19,6 +19,8 @@ span.buttonSeparator {
   margin: 4px 3px 4px 3px;
 }
 
+a.browserButton,
+span.browserButton,
 .browserButton {
   cursor: default;
   display: inline-block;


bradleyrichter:
I confirmed it works in my local master repo. Thanks Jon!

ayumi:
++ :rocket:

PR #3939, opened 2016-09-12
Fix overflowing notification buttons

### LINKED ISSUE #3920: password notification bar is not aligned well
**Describe the issue you encountered:** password notification bar is not aligned well (on Japanese setting)

- Brave Version: 0.12.1

- Steps to reproduce:
    1. Delete your trac.torproject.org password
    2. Go to https://trac.torproject.org/projects/tor/login
    3. Log in

- Screenshot if needed:
clipboard01

- Any related issues:

ayumi:

Notification options have float: right. I moved them to precede other content in the DOM, so now the other content flows around it rather than "pushing" it away.

Fix #3920

Auditors: @luixxiul @jkup

Test Plan:

1. Open Brave and trigger a notification e.g. visit https://www.browserleaks.com/geo
2. Horizontally resize window so the notification text overflows
3. Confirm that the buttons don't overflow out of the notification

Should now look like:
screen shot 2016-09-12 at 16 34 09
screen shot 2016-09-12 at 16 29 57

### DIFF
--- a/js/components/notificationBar.js
+++ b/js/components/notificationBar.js
@@ -35,19 +35,6 @@ class NotificationItem extends ImmutableComponent {
     const options = this.props.detail.get('options')
     const greeting = this.props.detail.get('greeting')
     return <div className={'notificationItem ' + (options.get('style') || '')}>
-      {
-        greeting
-          ? <span className='greeting'>{greeting}</span>
-          : null
-      }
-      <span className='message'>{this.props.detail.get('message')}</span>
-      <span className='notificationAdvanced'>
-        {
-          options.get('advancedText') && options.get('advancedLink')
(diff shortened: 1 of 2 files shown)


jkup:
Code LGTM! The only thought is it might be better to keep the order the same but just make sure the buttons extend the yellow notification background? My thinking is just that since the message is top left I expect it to be above the top right content on collapse.

ayumi:
@jkup I agree that'd be good, but I can't figure out the CSS to make it look like that. :/

Because this fixes an unsightly bug I think it'd be nice to merge as is.
Afterwards we should improve the styling.

luixxiul:
We need `margin-left` to avoid this:
screenshot 2016-09-13 14 44 24

I'm pushing a PR to fix it.

PR #3942, opened 2016-09-13
update site settings for main frame url and use requesting origin in …

### LINKED ISSUE #3877: Regression : session.setPermissionRequestHandler callback is broken
Yes

When testing http://www.bennish.net/web-notifications.html during a test run, it was noticed that the notifications were not displaying. 

**Expected behavior:**
The notifications should work.

Windows 7 ia-32 and OS X

- Brave Version:
0.12.1

- Steps to reproduce:
    1. Visit http://www.bennish.net/web-notifications.html
    2. Trigger a notification

- Screenshot if needed:

- Any related issues:

bridiver:

Test Plan:

…the permission dialog

fix #3877
auditors: @bbondy @diracdeltas @mrose17

geoip service must be enabled through 
GOOGLE_API_KEY
GOOGLE_API_ENDPOINT
env vars
requires https://github.com/brave/electron/commit/f8902043f283ac4813717f68b6261ec1b2b3f3df

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -276,7 +276,7 @@ function registerPermissionHandler (session, partition) {
   const isPrivate = !partition.startsWith('persist:')
   // Keep track of per-site permissions granted for this session.
   let permissions = null
-  session.setPermissionRequestHandler((origin, embedderOrigin, permission, cb) => {
+  session.setPermissionRequestHandler((origin, mainFrameUrl, permission, cb) => {
     if (!permissions) {
       permissions = {
         media: {
@@ -305,7 +305,7 @@ function registerPermissionHandler (session, partition) {
         }
       }
     }
-    const url = origin || embedderOrigin
+    const url = mainFrameUrl || origin
     // Allow notifications for the main app
     if (url === appUrlUtil.getIndexHTML() && permission === 'notifications' ||
         url.startsWith('chrome-extension://') && permission === 'openExternal') {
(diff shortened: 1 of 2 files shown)


bridiver:
this allows us to use the `navigator.geolocation` with any service compatible with the google geolocation api

bbondy:
++

app/filtering.js
```diff
@@ -357,7 +352,7 @@ function registerPermissionHandler (session, partition) {
         {text: locale.translation('deny')},
         {text: locale.translation('allow')}
       ],
-      frameOrigin,
+      origin,
```
diracdeltas: this should be the mainFrameUrl origin

app/filtering.js
```diff
@@ -372,7 +367,7 @@ function registerPermissionHandler (session, partition) {
       cb(result)
       if (persist) {
         // remember site setting for this host over http(s)
-        appActions.changeSiteSetting(urlOrigin, permission + 'Permission', result, isPrivate)
+        appActions.changeSiteSetting(getOrigin(url), permission + 'Permission', result, isPrivate)
```
diracdeltas: this should be `origin` instead of getOrigin(url) since the message shown to the user was for `origin`

bridiver: but shouldn't the permission be set for the page the user visited? Using the `origin` will set the permission for the requesting url

diracdeltas: actually it looks like this change makes the permissions work by top-level origin instead of the origin asking for the permission; i'll fix to make it consistent with the way it was before.

app/filtering.js
bridiver: I guess I'm confused about how this should work. The `origin` is the origin of the url that is actually requesting the permission. If we used `mainFrameUrl` here it would appear to the user that the main page is requesting permission when it may actually be requested by a frame with a different origin

app/filtering.js
```diff
@@ -305,7 +305,7 @@ function registerPermissionHandler (session, partition) {
         }
       }
     }
-    const url = origin || embedderOrigin
+    const url = mainFrameUrl || origin
```
bridiver: so I think this is incorrect too @diracdeltas because there will always be an origin and that is how we want to lookup the permission

Issue #3962, opened 2016-09-13
resolve origin vs frameOrigin confusion for permission requests


diracdeltas:
opening this issue to track follow-ups to https://github.com/brave/browser-laptop/pull/3942

PR #3965, opened 2016-09-13
use origin for the notification message and for site settings and cho…

### LINKED ISSUE #3877: Regression : session.setPermissionRequestHandler callback is broken

bridiver:

Test Plan:
1. Go to https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/Using_geolocation
2. notification should display for https://mdn.mozillademos.org vs page origin of http://developer.mozilla.org
3. change tabs or open a new tab
4. notification should not appear
5. return to the original tab
6. notification should be displayed

The changes for Brave Browser notifications aren't currently  testable because we don't get the mainFrameUrl for notifications yet, but I have tested them with a local hack

don't allow Brave to display notifications without permission
remove unnecessary password copied notifications
cc @bbondy @diracdeltas
fix #3877

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -305,13 +305,6 @@ function registerPermissionHandler (session, partition) {
         }
       }
     }
-    const url = mainFrameUrl || origin
-    // Allow notifications for the main app
-    if (url === appUrlUtil.getIndexHTML() && permission === 'notifications' ||
-        url.startsWith('chrome-extension://') && permission === 'openExternal') {
-      cb(true)
-      return
-    }
 
     if (!permissions[permission]) {
       console.log('WARNING: got unregistered permission request', permission)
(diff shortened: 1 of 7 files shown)


bridiver:
fix #3962

diracdeltas:
this causes some of the notificationBar tests to fail

app/filtering.js
```diff
@@ -367,7 +374,7 @@ function registerPermissionHandler (session, partition) {
       cb(result)
       if (persist) {
         // remember site setting for this host over http(s)
-        appActions.changeSiteSetting(getOrigin(url), permission + 'Permission', result, isPrivate)
+        appActions.changeSiteSetting(origin, permission + 'Permission', result, isPrivate)
```
diracdeltas: need to strip the trailing `/` from origin for backwards compatibility; i think this is why tests are breaking

diracdeltas: hence why `getOrigin` was necessary before

Issue #3759, opened 2016-09-06
Check page status code before running ledger timer


Aeyoun:
Status code should be in the HTTP 2xx Success range. Anything from 3xx (redirects), 4xx (client error), or 5xx (server error) shouldn’t count against the ledger.

Should cover sites out of bandwidth, crashed sites, CloudFlare stuff, dead links, etc.

...and while we're at it, let's ensure that the page is text/html

mrose17:
@bridiver - this looks like a one-liner! care to point me where to add that line (or add the line yourself)? thanks!

bridiver:
We already cover most of these. Redirects last less than 8 seconds and 5xx errors go to about:error which are already excluded. 304 is a valid and common response that shouldn't be ignored. We can exclude 4xx pages, but it's definitely not a one-liner because we normally don't get the http status code. We either get a load success or load failure, but 4xx pages are not considered load failures by chromium. Electron has a `did-get-response-details` that we could possibly use in the place of the webview load, but we don't currently use it and I need to look at it more closely. With the focus/idle changes this seems like a minor issue at the moment. How long are you going to keep a 404 page focused and active? Also fyi - the milestone is set to 0.12.1dev

PR #4032, opened 2016-09-15
Don't register a visit if page returns a non-successful HTTP status code

bsclifton:

Don't register a visit if page returns a non-successful HTTP status code

Fixes https://github.com/brave/browser-laptop/issues/3759
Auditors:
@mrose17, @bridiver

Test Plan:
1. Visit https://clifton.io
2. Stay on page for a while
3. Confirm time spent is incremented
4. Visit https://clifton.io/404
5. Stay on page for a while
6. Confirm time spent is not incremented


app/common/lib/ledgerUtil.js
```diff
@@ -0,0 +1,51 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+/**
+ * Returns true if HTTP response code is one we want to collect usage for
+ * @param {number} responseCode - HTTP response code to be evaluated
+ * @return {boolean} true if the page should be excluded, false if not
+ */
+module.exports.IsValidResponseCode = (responseCode) => {
```
bridiver: `IsValidResponseCode` => `isValidResponseCode`

js/stores/eventStore.js
```diff
@@ -128,6 +139,9 @@ const doAction = (action) => {
       // retains all past pages, not really sure that's needed... [MTR]
       eventState = eventState.set('page_info', eventState.get('page_info').push(action.pageInfo))
       break
+    case WindowConstants.WINDOW_GOT_RESPONSE_DETAILS:
+      addResponseDetails(action.frameProps.get('location'), action.frameProps.get('tabId'), action.details)
```
bridiver: as we discussed, I think we should just use this for the load event and add the details to it

app/common/lib/ledgerUtil.js
bsclifton: Feedback about the method for checking (and what codes to include) is very much appreciated

js/components/frame.js
```diff
@@ -965,6 +965,11 @@ class Frame extends ImmutableComponent {
         }))
       }
     })
+    this.webview.addEventListener('did-get-response-details', (details) => {
+      if (getSetting(settings.PAYMENTS_ENABLED) && details.originalURL === this.props.location) {
```
bsclifton: NOTE: this event will only get sent if payments is enabled (to prevent unneeded overhead)

app/common/lib/ledgerUtil.js
bridiver: I think you probably want to check for errors vs positive responses because there are other potential successful responses. You also want to ignore things like redirects

bridiver: actually I take that back. 2xx and 304 appear to cover it

bridiver: actually I take that back. We need to further restrict 2xx to 200, 203 and 206 I think

js/components/frame.js
bridiver: I thought I commented on this. I think we should always send it. The event store will be used for more than just ledger

app/common/lib/ledgerUtil.js
bridiver: this also might be better as either a general utility method vs ledger. `shouldTrackView` looks ledger-specific so that's ok

bsclifton: Good call- typically REST APIs will use 201 and 202, which shouldn't be returned for regular content :+1:

test/unit/lib/httpUtilTest.js
```diff
@@ -0,0 +1,34 @@
...
+      it('returns true for a cached response (304)', function () {
+        assert.equal(httpUtil.responseHasContent(304), true)
+      })
+    })
+
+    describe('expected failure codes', function () {
+      it('returns false for non-content success codes (used for REST apis, etc)', function () {
+        assert.equal(httpUtil.responseHasContent(201), false) // created
+        assert.equal(httpUtil.responseHasContent(202), false) // accepted
+      })
```
bridiver: 4xx? Don't necessarily need an exhaustive list, but the list isn't that long https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
can be done in a followup too

bridiver:
++

test/unit/lib/httpUtilTest.js
bsclifton: done :+1:

alexwykoff:
Did not work as expected for 0.12.2 RC2 on OS X
screen shot 2016-09-22 at 6 30 42 pm

PR #3973, opened 2016-09-13
add network connection notifications

bridiver:
notifications will not be fired until electron-prebuilt is updated to https://github.com/brave/electron/commit/b59db7297b2b6b2c6ff34d90743543abfbace747
cc @mrose17

requirement for open ledger issue #??

PR #4038, opened 2016-09-15
add network connection notifications

mrose17:
supercedes https://github.com/brave/browser-laptop/pull/3973 (reduce to 1 commit)

Test Plan:

disconnect the network. reconnect the network. 1 second later the ledger client code will run as will the code that gets the balance.

auditor: @bridiver


luixxiul:
Would someone please specify STR to confirm the issue has been fixed?

bridiver:
there is a test plan above, but I think it needs a better method to verify in the release builds. @mrose17 is there something they can check in the interface? Log output is not readily available in the release builds to confirm that the code is running

mrose17:
@bridiver - you can always get log information, it's just a matter of how you start the browser, e.g., ```LEDGER_LOGGING=true LEDGER_VERBOSE=true /Applications/Brave.app/Contents/MacOS/Brave```

although i agree with the point that it ought to be easier to test it.

@ayumi - what do you think about a follow-on issue in which, on `appConstants.APP_NETWORK_CONNECTED`, we update `ledgerInfo.error` with a string saying "Internet inaccessible, so updates unavailable." ?

ayumi:
@mrose17 Right now there's a message saying "error, can't retrieve data" which covers being offline I think.
Are we going to show general notification of offline status (suggested by @bsclifton https://github.com/brave/browser-laptop/pull/3973#issuecomment-246898888)?
If so the user will know it's because they're offline.

mrose17:
@ayumi - perhaps this is a non-issue, then.

certainly a general notification along the lines of what @bsclifton suggested in a post-1.0 timeframe would suffice.

PR #4049, opened 2016-09-16
Add Importer support

### LINKED ISSUE #428: Import data from other browsers
Is the fetaure unimplemented yet or it should be working and it is a bug?

### LINKED ISSUE #2107: General Settings: "Browser data import" setting
Functionality:
Choosing "Import now..." opens the Import Browser Data dialog which floats above the General Settings panel. (This dialog is modal to all other functionality within the browser settings panels.)

The pull-down menu list is populated by polling the OS for installed browsers that have at least one data type that can be imported. Each browser offers a different set of importable data types. The available toggle switches change dynamically based on the selected browser, becoming disabled to indicate lack of availability.

Safari: Preferences, Bookmarks
Chrome: Cookies, Browsing History, Bookmarks
Firefox:  Browsing History, Bookmarks, Saved Passwords, Autofill form data
Superset: Browsing History, Bookmarks, Cookies, Preferences, Saved Passwords, Autofill form data

Item label:
Browser data import

Button string:
"Import now..."

Spec image: (issue item shown in blue)
image

Spec image with import dialog open:
image

darkdh:

Test Plan:

fix #428
partial fix #2107

requires https://github.com/brave/electron/pull/57

Auditors: @bridiver, @bbondy, @bsclifton


app/browser/menu.js
```diff
@@ -390,9 +390,7 @@ const createBookmarksSubmenu = () => {
     },
     CommonMenu.separatorMenuItem,
     CommonMenu.bookmarksManagerMenuItem(),
-    CommonMenu.bookmarksToolbarMenuItem(),
-    CommonMenu.separatorMenuItem,
-    CommonMenu.importBookmarksMenuItem()
+    CommonMenu.bookmarksToolbarMenuItem()
   ]
```
bsclifton: there is some commented out code here in the `menu.js` that I think you can delete, now that you've added import functionality. Search for `locale.translation('importFrom')`

app/importer.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+    sites.push(site)
+  }
+  appActions.addSite(Immutable.fromJS(sites))
+})
+
+importer.on('add-homepage', (e, detail) => {
+})
+
+importer.on('add-bookmarks', (e, bookmarks, topLevelFolder) => {
```
bsclifton: this handler has a LOT going on in it... Any chance we could break this (or parts of it) out into a method? Looks generic enough (not dependent on electron) to put into a utils class, which means we could eventually unit test it too :smile:

docs/appActions.md
```diff
@@ -349,6 +349,16 @@ Clears the data specified in dataDetail
 
 
 
+### importBrowserData(selected) 
+
+Import browser data specified in selected
+
+**Parameters**
+
+**selected**: `object`, the browser data to import as per doc/state.md's importBroserDataSelected
```
bsclifton: definitely not a big deal, just a misspelling here:
s/importBroserDataSelected/importBrowserDataSelected

docs/state.md
```diff
@@ -505,6 +505,23 @@ WindowStore
...
+    {
+      index: string,
+      type: number,
+      name: string,
+      history: boolean,
+      favorites: boolean,
+      cookies: boolean
+    }
+  ],
+  importBroserDataSelected: {
```
bsclifton: whoops, another one:
s/importBroserDataSelected/importBrowserDataSelected

js/components/importBrowserDataPanel.js
```diff
@@ -0,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
bsclifton: would it be possible to move this to the new directory structure?
https://github.com/brave/browser-laptop/blob/master/docs/directoryStructure.md

So for example, this would move to:
app/renderer/components

Great job on the control :smile:

bsclifton:
Left some comments... great work :smile:

app/importer.js
darkdh: But the handler is for electron dependent `bookmarks` object and whole block is meant to handle it specifically. So I am not sure that we can pull out a generic method from it.

darkdh:
Thanks for your feedback, @bsclifton. I've addressed it in ec45bdf.

app/importer.js
bsclifton: ah- makes sense.  I'm OK w/ it as-is :smile:

app/filtering.js
```diff
@@ -693,3 +693,14 @@ module.exports.clearAutofillData = () => {
     ses.autofill.clearAutofillData()
   }
 }
+
+module.exports.setCookie = (cookie) => {
```
bridiver: this method should be moved out of filtering and only called on the default session

app/common/commonMenu.js
```diff
@@ -243,31 +243,17 @@ module.exports.passwordsMenuItem = () => {
...
     click: function (item, focusedWindow) {
-      if (BrowserWindow.getAllWindows().length === 0) {
-        appActions.newWindow(undefined, undefined, undefined, function () {
-          // The timeout here isn't necessary but giving the window a bit of time to popup
-          // before the modal file picker pops up seems to work nicer.
-          setTimeout(() =>
-            module.exports.sendToFocusedWindow(BrowserWindow.getAllWindows()[0], [messages.IMPORT_BOOKMARKS]), 100)
-        })
-        return
+      if (process.type === 'browser') {
```
bridiver: you can avoid this check by making it an appAction
see https://github.com/brave/browser-laptop/blob/master/app/browser/basicAuth.js

js/components/main.js
```diff
@@ -374,6 +376,11 @@ class Main extends ImmutableComponent {
       windowActions.setContextMenuDetail()
     })
 
+    ipc.on(messages.IMPORTER_LIST, (e, detail) => {
```
bridiver: this would also be better as an appAction. We want to avoid adding more ipc handlers

js/components/main.js
```diff
@@ -374,6 +376,11 @@ class Main extends ImmutableComponent {
...
+      windowActions.setImportBrowserDataDetail(detail)
```
bridiver: ideally actions should not be setters, they should describe the actual action. In this case something like `appAction.supportedBrowsersUpdated` which would then be handled in the windowStore and should call helper methods to update the state (don't call actions from actions).

js/components/main.js
```diff
@@ -532,6 +539,10 @@ class Main extends ImmutableComponent {
     windowActions.setClearBrowsingDataDetail()
   }
 
+  onHideImportBrowserDataPanel () {
+    windowActions.setImportBrowserDataDetail()
```
bridiver: along the same lines as the previous comment this would be something like `windowActions.importDataPanelHidden`

app/filtering.js
darkdh: addressed it in 1dee63d2b030914676811eef44829cbd972907a2.

darkdh:
I will do appAction refactoring in follow-up commit

bridiver:
sounds good! Only the `setCookie` thing is critical for merging

bbondy:
merging!

Issue #3955, opened 2016-09-13
Make sure we include all initial supported languages


bbondy:
@CCBrave got us this list of languages we'll ship 1.0 with.

@aekeus could you please verify that we include all of them, the regional codes may not be fully accurate but just fill in only the language if ones are missing.

Note this task is in 0.12.2 timeline so freeze is Friday.  We mainly want to get this moving to give localizers time.

German
Russian
Chinese
Japanese
French (France)
Bengali (Bangladesh)
Italian
Spanish
Brazilian-Portuguese
Polish
modern Malay (not Jawi)
Ukrainian
Korean
Indonesian
Tamil
Turkish
Slovenian 
Basque
Czech

CCBrave:
The first 10 listed would be a must. If you need to drop any, the last 5
would be it (combo of lots to translate still (Tamil and Turkish), and
small markets for Slovenian/Basque/Czech).

aekeus:
Current list running in master:

<img width="296" alt="screen shot 2016-09-13 at 3 51 49 pm" src="https://cloud.githubusercontent.com/assets/210240/18489110/32b80b60-79ca-11e6-87b5-22d1b69e1222.png">

aekeus:
Translations for the remaining seven languages are at 81.9%. Request for info sent to Transifex, no response back yet.

Issue #1747, opened 2016-05-14
Sort languages in preferences for non english


bbondy:
The placement of the languages follow the English sorting order even for different languages.
So you can have for example Anglais in French in the middle of the list.

luixxiul:
Or we could follow the wikipedia sorting style.

How about this? 

- Automatically apply the translation according to the system preference.
- Let users change the language setting.

This might be better for those who don't understand English at all.

bbondy:
it should already pick the right language depending on the OS language for the default.

If you know that it doesn't do this pls post a diff bug though.

Liunkae:
Making a note of bbondy's comment on https://github.com/brave/browser-laptop/pull/4055#issuecomment-247746311:

> I think doing a sort within the render in react would be better because it would work with all langs

PR #4055, opened 2016-09-16
Use language name for sort order instead of identifier. Fix a few outdated comments

aekeus:
Issue: #3955

Auditors: @bbondy, @alexwykoff

Test Plan:

  i. Open about:preferences
  ii. Ensure languages are in alphabetical order in select

### DIFF
--- a/app/locale.js
+++ b/app/locale.js
@@ -220,8 +220,7 @@ exports.translation = function (token) {
   if (translations[token]) {
     return translations[token]
   } else {
-    // This will return an identifier in upper case enclosed in square brackets
-    // Useful for determining if a translation was not requested in the menu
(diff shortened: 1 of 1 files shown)


Liunkae:
With this change, the languages will be sorted by their _English_ names. This could be a little confusing for non-English users since the language list has no logical order in their language (the language names themselves are localized).

If we go with sorting by language name, I suggest we also sort the list in other languages. (Another issue should be opened for that, though.)

luixxiul:
I strongly agree, How about following Wikipedia?

Liunkae:
@luixxiul How does Wikipedia sort languages?

Also, I found this old issue for my previous point (#1747).

bbondy:
no objection since it doesn't make it worse but I think doing a sort within the render in react would be better because it would work with all langs

PR #4121, opened 2016-09-20
Use custom versions of electron-squirrel-startup and electron-winstaller

aekeus:
* Use forked version of two Windows installation modules
  * Do not re-create shortcuts on the desktop on auto-update

Auditors: @bbondy, @alexwykoff

Test plan:

  This change can only be tested via an auto-update

  Scenario 1

  i)   Install older version of browser
  ii)  Remove shortcut from desktop
  iii) Do an auto-update
  iV)  Ensure the shortcut was not created

  Scenario 2

  i)   Install older version of browser
  ii)  Do NOT remove shortcut from desktop
  iii) Do an auto-update
  iv)  Ensure the shortcut opens the most recent version of the browser

Fixes: #1181


bbondy:
++ thanks for doing this.

PR #4141, opened 2016-09-21
Add notification to suggest users use Brave Payments

### LINKED ISSUE #3257: ledger notification should have the style of the update bar

ayumi:

Fix #3257 (Finishes the remaining task)

- Record firstRunTimestamp into session-state
- 10 days from firstRunTimestamp, show notification suggesting user tries Payments
- If the user enables Payments they won't see the notification

Auditors: @bsclifton @diracdeltas

Test Plan:

1. Edit `ledger.js` to reduce showNotifications delay interval to 3 * msecs.second
2. Run Brave with an existing profile
3. Confirm there's no notification
4. Close Brave
5. Update userData session-store-1 firstRunTimestamp to be way in the past e.g. from '14..' to '13..' (~/Library/Application\ Support/session-store-1)
6. Open Brave
7. Confirm notification


app/ledger.js
```diff
@@ -383,6 +391,9 @@ eventStore.addChangeListener(() => {
 var initialize = (onoff) => {
   enable(onoff)
 
+  // Check if relevant browser notifications should be shown every 15 minutes
+  setInterval(showNotifications, 15 * msecs.minute)
```
diracdeltas: `clearInterval` never gets called on this, so i think a new timer is added every time the user clicks the on/off switch. this may lead to notifications being shown more often than intended or something

diracdeltas: maybe add a global var to keep track of whether `setInterval` has already been called and do nothing here if it has

diracdeltas:
otherwise :+1:

app/ledger.js
ayumi: thanks good call– it was originally there but i removed it (oops).
i added it back

bsclifton:
:+1:

PR #4145, opened 2016-09-21
use main frame url for webrequest first party url

### LINKED ISSUE #4137: unsafe use of firstPartyUrl
we are using it for main frame URL checking in a few places, which is not exactly safe according to https://cs.chromium.org/chromium/src/net/url_request/url_request.h?sq=package:chromium&dr=C&l=257

diracdeltas:

Test Plan: covered by automated test
fix #4137
fix https://github.com/brave/browser-laptop/issues/4300

auditors: @bridiver

### DIFF
--- a/app/adBlock.js
+++ b/app/adBlock.js
@@ -28,7 +28,7 @@ const whitelistHosts = ['disqus.com', 'a.disquscdn.com']
 
 const startAdBlocking = (adblock, resourceName, shouldCheckMainFrame) => {
   Filtering.registerBeforeRequestFilteringCB((details) => {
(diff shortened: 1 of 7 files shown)


bridiver:
this looks good, but can we add some tests for it? The changes look right, but it's hard to know for sure without either automated or manual testing

diracdeltas:
@bridiver thanks, i have rebased and added a test which fails on master but works with this PR. the problem appears to be that `firstPartyUrl` in an cross-origin iframe is `data:` instead of the main frame URL, which is consistent with https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-2.1.

PR #4181, opened 2016-09-22
Fix the check for fullscreen window.

Sh1d0w:
Auditors:
@bbondy @bridiver 

Test Plan:
Press F11 and see if the div#window has class isFullScreen. The top bar should also be hidden in fullscreen mode.

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -842,7 +842,7 @@ class Main extends ImmutableComponent {
 
     return <div id='window'
       className={cx({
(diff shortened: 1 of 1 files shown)


bbondy:
just making sure, was this tested with both html full screen and per window full screen?

bbondy:
++

bbondy:
Sorry I had to revert this.
This is not the behavior other browsers do so we don't want this. We could add an option for the functionality but disabled by default.

Issue #4051, opened 2016-09-16
Idle timer never stopping the ledger timer


Aeyoun:
Left my browser open over night. This added 7 hours to the ledger timer of the website that was left open in Brave. Apparently the idle timer isn’t triggered or is ignored.

Brave version 0.12.1dev, Fedora 24 64-bit.

mrose17:
@bridiver - can you take a quick look at this? thanks!

bridiver:
haven't tested it on Fedora, but I wonder if there is something else preventing the idle notification. The idle status is system-wide and not just for Brave. @mrose17 I assume he should see the timer start and stop with ledger logging enabled? It should kick in after 15 minutes of inactivity or on lock - https://developer.chrome.com/apps/idle

mrose17:
@alexwykoff - could someone internally try to reproduce this? (i'm away from my VMs...) thanks!

bridiver:
I can look later as well, but I need to setup a Fedora VM

mrose17:
@Aeyoun - thanks again for reporting this. i may have found the issue. i will circle back in a bit.

mrose17:
i'm pretty sure that https://github.com/brave/browser-laptop/pull/4186 fixes this. the idle timer is 15m.

note that a future revision may ignore the idle timer if `audioPlaybackActive: true`

mrose17:
Fixed https://github.com/brave/browser-laptop/pull/4194

PR #4186, opened 2016-09-22
when the idle state changes, flush the timer

mrose17:
Fixes https://github.com/brave/browser-laptop/issues/4051

Auditor: @ayumi

Test Plan:

1. invoke the browser via `LEDGER_PUBLISHER_VERBOSE=true ...`

2. go to a publisher site

3. do NOT touch the keyboard and wait *more* than 15m

4. go to another site and browser around, or restart the browser, etc.

5. go to `about:preferences#payments` and verify that the publisher got no more than 15m of update

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -110,7 +110,24 @@ let notificationTimeout = null
 const doAction = (action) => {
   var i, publisher
 
+/* TBD: handle
+
+    { actionType: "window-set-blocked-by"
+    , frameProps:
+      { audioPlaybackActive: true
+        ...
+      }
+    , ...
+    }
+ */
+  if (publisherInfo._internal.verboseP) {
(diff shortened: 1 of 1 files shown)


app/ledger.js
```diff
@@ -110,7 +110,24 @@ let notificationTimeout = null
 const doAction = (action) => {
   var i, publisher
 
+/* TBD: handle
+
```
ayumi: what's this– does it pertain to the idle state

mrose17: yes, at some time in the future, we're not going to count idle when you are on an audio(/video) page... so, i've just put in a comment as a reminder that a check can start there.

mrose17:
superseded by https://github.com/brave/browser-laptop/pull/4194

PR #4193, opened 2016-09-22
Show top panel in fullscreen on mouse over

Sh1d0w:
Currently when you are in fullscreen mode (F11)
There is no way to see all opened tabs or to type new address
in the url bar. The only way to do it is to switch off fullscreen
and type new address/go to other tab and then switch back to
fullscreen.

This PR makes the top bar behave like the top bar in Chrome and FF.
When you hover with the mouse at the top of the page while in fullscreen
it will appear. Same will happen if you focus the URL bar while in
fullscreen mode (CTRL + L).

Auditors:
@bbondy

Test Plan:
1. Go in fullscreen (F11)
2. Hover whit mouse the top of the page
3. Focus the url bar (CTRL + L)


bbondy:
++ thanks

srirambv:
This works intermittently on Windows 10. Doesn't show the top bar every time mouse hover to the top of the screen.

aekeus:
I am getting the Minimize and Close buttons in a frame while navigating to the top of the screen in full-screen mode

Sh1d0w:
@srirambv The area which toggles the bar is exactly the 1px top border, so unless your mouse is exactly at the top of the browser it won't show. Let me know if you think this is wrong and this area have to be expanded?

@aekeus Unfortunately I don't have windows machine. Maybe @bsclifton can take a look, since he tweaked the windows caption buttons? :)

bsclifton:
Taking a look now...

bsclifton:
I think I have reproduced? Please help me understand:

If you turn off the "Hide the menu bar by default" preference, as expected you'll always see the menu. I'm guessing that going fullscreen doesn't look correct with menu shown.

If you have menu hidden and push Ctrl + L, the URL bar pops back up (which includes the menus).  I don't see the caption buttons though?

Sh1d0w:
@bsclifton Actually it works just fine under Linux. I guess it is because of the windows caption buttons modification. While in fullscreen I just show/hide the whole top panel (url bar, bookmarks etc). Since now the caption buttons are part of that top panel they will also show. Maybe we can add CSS while the container have `isWindowFullscreen` class we dont show them?

@aekeus Can you please provide screencast of the problem you experience, so we can understand what actually it is? And also eplain what the correct behaviour should be? Thanks.

bsclifton:
@Sh1d0w can you share a pic of what it looks like for you? (fullscreen w/ URL bar shown). I think that'll help clear things up :smile:

Sh1d0w:
@bsclifton Sure. Here is a video:

http://webm.land/media/jWyx.webm

Electron is correctly hiding the caption buttons in fullscreen.

srirambv:
@Sh1d0w  this is what is happening on Windows machine. I do not see the window buttons when top bar is shown.

topbar

srirambv:
The effect is same on 100% magnification and at 150% magnification is a little better. Top bar is visible more when at 150% than 125% or 100% magnification

aekeus:
In fullscreen mode, moving the cursor to the top of the screen brings down a panel containing the close and resize button instead of the url bar.

screen shot 2016-09-29 at 9 56 33 am

bbondy:
Reverted here:
https://github.com/brave/browser-laptop/pull/4372

It can be re-added once we have a fix for Windows.  Sorry about this but this is holding up RC3 from releasing and we need it out today so it is the best option.

bsclifton:
@srirambv opened issue https://github.com/brave/browser-laptop/issues/4371 to re-track this issue since hover was not working as expexted. Changes were reverted with:
https://github.com/brave/browser-laptop/commit/ff888e9de6f23f988c15c8b52ef07833d0e79588

PR #4231, opened 2016-09-23
Revert "Fix the check for fullscreen window."

bbondy:
Reverts brave/browser-laptop#4181


bbondy:
This is not the behavior other browsers do so we don't want this. We could perhaps add an option for the functionality but disabled by default.

Issue #4232, opened 2016-09-23
Right clicking on bookmarks toolbar gives window menu instead of bookmarks menu


bbondy:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
If you right click on the bookmarks toolbar in an open space (not on top of a bookmark) it shows the windows system menu instead of the bookmarks toolbar menu.

**Expected behavior:**
You should get options like add bookmark etc.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Windows 7

- Brave Version:
master tip

- Steps to reproduce:
    1. Show bookmarks toolbar
    2. Right click on an open space
    3. Notice it shows the window system menu instead of bookmarks context menu.

- Screenshot if needed:
<img width="333" alt="screenshot 2016-09-23 07 51 20" src="https://cloud.githubusercontent.com/assets/831718/18784864/dbe19e8c-8162-11e6-8aa1-ea00227a4d30.png">

- Any related issues:
https://github.com/brave/browser-laptop/issues/3036
#4171

luixxiul:
Basically this seems to be the same issue as #4171

bbondy:
OK I moved that into 0.12.2 too

Issue #4171, opened 2016-09-22
Windows system context menu appears by doing a right click around windowCaptionButtons


luixxiul:
**Describe the issue you encountered:** Windows system context menu appears by doing a right click around ``windowCaptionButtons``

**Expected behavior:** It should appear by doing a right click only on ``windowCaptionButtons``

- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 10

- Brave Version: 0.12.2 RC1

- Steps to reproduce:
    1. Make the browser window smallest
    2. Open as many tabs as tab set buttons appear
    3. Do a right click on the row of the tab set buttons
    4. The system context menu appears

- Screenshot if needed:
![clipboard01](https://cloud.githubusercontent.com/assets/3362943/18737417/dbe412c6-80cb-11e6-9e93-a2f727b41b4a.png)

- Any related issues: #3036

bsclifton:
OK i'm able to repro after a 2nd tabset is opened (clicking in that area)

PR #4243, opened 2016-09-23
0.12.2dev titlebar fixes (round 1)

bsclifton:
Don't set `allowDragging` class for Windows on:
- bookmarks toolbar
- tab pages control
- tabs toolbar

Cleaner fix than marking class as `no-drag !important` w/ CSS

Fixes https://github.com/brave/browser-laptop/issues/4171
Should fix https://github.com/brave/browser-laptop/issues/4232

Auditors: @bbondy (important that you try it- I couldn't reproduce 4232)


bbondy:
++ but it would be great to re-add these areas as draggable in the future. Pls post an issue to investigate how to suppress the system menu for later investigation

bsclifton:
Issue created https://github.com/brave/browser-laptop/issues/4244

PR #4251, opened 2016-09-24
0.12.2dev titlebar fixes (round 3)

bsclifton:
Fix click binding logic in Menubar control. Will now work it's way through nested menus.

Fixes https://github.com/brave/browser-laptop/issues/4226

Auditors: @bbondy


bbondy:
++

PR #4314, opened 2016-09-27
Reset old reconcileStamp when re-enabling Payments

### LINKED ISSUE #4058: Brave payments wrong contribution date
No

I have enabled Brave payments long ago and disabled it for a while. After re-enabling it again the Next contribution date was in the past.

**Expected behavior:**
The next contribution date should never be in the past.

All
- Brave Version:
0.12.1

- Steps to reproduce:
    1.Enable Brave payments
    2. Disable Brave paymetns
    3. Wait several days and re-enable

- Screenshot if needed:

- Any related issues:

ayumi:
Fix #4058

Auditors: @mrose17

Test Plan:
1. Open Brave
2. Enable Payments then disable
2. Close Brave
3. Edit {userData}/ledger-state.json `reconcileStamp` to be way in the past (change 14.. to 13..)
4. Open Brave
5. Enable Payments
6. Note that contribution date is today + 30 days, rather than way in the past

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -129,8 +129,16 @@ const doAction = (action) => {
       break
 
     case appConstants.APP_CHANGE_SETTING:
-      if (action.key === settings.PAYMENTS_ENABLED) return initialize(action.value)
-      if (action.key === settings.PAYMENTS_CONTRIBUTION_AMOUNT) return setPaymentInfo(action.value)
+      switch (action.key) {
+        case settings.PAYMENTS_ENABLED:
+          initialize(action.value, 'changeSettingPaymentsEnabled')
+          break
+        case settings.PAYMENTS_CONTRIBUTION_AMOUNT:
+          setPaymentInfo(action.value)
+          break
+        default:
+          break
+      }
       break
 
     case appConstants.APP_CHANGE_SITE_SETTING:
@@ -407,8 +415,8 @@ eventStore.addChangeListener(() => {
  * module initialization
  */
 
-var initialize = (onoff) => {
-  enable(onoff)
+var initialize = (paymentsEnabled, reason) => {
+  enable(paymentsEnabled)
 
   // Check if relevant browser notifications should be shown every 15 minutes
   if (notificationTimeout) {
@@ -416,7 +424,7 @@ var initialize = (onoff) => {
   }
   notificationTimeout = setInterval(showNotifications, 15 * msecs.minute)
 
-  if (!onoff) {
+  if (!paymentsEnabled) {
     client = null
     return appActions.updateLedgerInfo({})
   }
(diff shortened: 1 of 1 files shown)


app/ledger.js
```diff
@@ -441,9 +448,24 @@ var initialize = (onoff) => {
...
+            state
+          )
+
+          // Scenario: User enables Payments, disables it, waits 30+ days, then
+          // enables it again -> reconcileStamp is in the past.
+          // In this case reset reconcileStamp to the future.
+          if (reason === 'changeSettingPaymentsEnabled') {
+            let timeUntilReconcile = client.timeUntilReconcile()
+            if (typeof timeUntilReconcile === 'number' && timeUntilReconcile < 0) {
+              let futureReconcileStamp = underscore.now() + 30 * msecs.day
```
mrose17: it is better to call `setTimeUntilReconcile` with a `null` timestamp and let the `ledger-client` package figure out the next time.

app/ledger.js
```diff
@@ -441,9 +448,24 @@ var initialize = (onoff) => {
...
+              client.setTimeUntilReconcile(futureReconcileStamp, callback)
```
mrose17: you shouldn't use `callback` here, since that's used for `client.sync` ... look at how the call to `setBraveryProperties` does it!

app/ledger.js
```diff
@@ -129,8 +129,15 @@ const doAction = (action) => {
       break
 
     case appConstants.APP_CHANGE_SETTING:
-      if (action.key === settings.PAYMENTS_ENABLED) return initialize(action.value)
-      if (action.key === settings.PAYMENTS_CONTRIBUTION_AMOUNT) return setPaymentInfo(action.value)
+      switch (action.key) {
+        case settings.PAYMENTS_ENABLED:
+          return initialize(action.value, 'changeSettingPaymentsEnabled')
```
mrose17: probably better to get rid of `return` and add a `break` line -- that would be consistent with everything else in the function.

app/ledger.js
```diff
@@ -129,8 +129,15 @@ const doAction = (action) => {
...
+        case settings.PAYMENTS_CONTRIBUTION_AMOUNT:
+          setPaymentInfo(action.value)
+          break
+        default:
```
mrose17: thank you for adding the `default` case!

app/ledger.js
```diff
@@ -441,9 +448,24 @@ var initialize = (onoff) => {
         }
 
         getStateInfo(state)
+
         try {
-          client = ledgerClient(state.personaId,
-                                              underscore.extend(state.options, { roundtrip: roundtrip }, clientOptions), state)
+          client = ledgerClient(
```
mrose17: why did the indentation change?

i agree that line 446 is indented too much, that must have been a slip of the finger. i thought it would be:

          client = ledgerClient(state.personaId,
                                underscore.extend(state.options, { roundtrip: roundtrip }, clientOptions), state)

ayumi: I think it's clearest to indent the 3 parameters equally

app/ledger.js
ayumi: good idea

app/ledger.js
mrose17: except that this is the only example of that style of indentation in the file...

perhaps we can compromise by putting the first parameter right after the "(", and the ")" right after the final parameter?

ayumi: that sounds fair

app/ledger.js
ayumi: thanks, i've updated it

bbondy:
mrose said it could be merged now, thanks.

PR #4317, opened 2016-09-27
Fix #3068

### LINKED ISSUE #3068: New Tab Always gets focus when "Switch to new tabs immediately" is disabled

rickycodes:
closes: #3068

I now realize this clashes with pull #3729, but I think this fix is perhaps slightly more elegant?

Figured I'd still throw my hat into the ring and see what your thoughts are :)

Please let me know if anything else is required.

**example:**

new_tab

### DIFF
--- a/docs/windowActions.md
+++ b/docs/windowActions.md
@@ -215,13 +215,14 @@ Dispatches a message to the store to create a new frame
 **Parameters**
 
 **frameOpts**: `Object`, An object of frame options such as isPrivate, element, and tab features.
-                 These may not all be hooked up in Electron yet.
+These may not all be hooked up in Electron yet.
 
-**openInForeground**: `boolean`, true if the new frame should become the new active frame
+**openInForeground**: `boolean`, true if the new frame should become the new active frame.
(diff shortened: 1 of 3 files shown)


bsclifton:
I've updated the PR to match the requested feature exactly. I changed the code because we need to respect the application's behavior (instead of hijacking it with the user's settings).  Basically, this will now default the openInForeground if no value is present (which was previously always setting to true).

Unfortunately, after trying this out, I don't think this is behavior that we want.  I'll leave a comment in the original issue

bsclifton:
Closing this PR for the moment.  Comment was left in https://github.com/brave/browser-laptop/issues/3068 and issue was closed as won't fix

Issue #1725, opened 2016-05-13
Moving mouse to other bookmarks toolbar items after click should make them expand


bsclifton:
If you click a folder in the bookmarks toolbar, it expands and shows the links inside the folder. Unfortunately if you move your mouse to another folder while it's open, nothing happens- the original folder stays open.  I would expect the same behavior as Chrome/FF/etc where moving the mouse to another folder closes the original item and opens the newly selected folder

bsclifton:
Here's a demo:

I originally clicked to open the clifton.io folder in the bookmarks toolbar. I then moved mouse over to morgopolis folder and nothing happened.  You can clearly see the hover style activated over the morgopolis icon, since the mouse is over it
![2016-05-13 1](https://cloud.githubusercontent.com/assets/4733304/15252791/6b2382d2-18e4-11e6-84dd-4f34cefc7584.png)

luixxiul:
@bbondy confirmed on Windows 10.

PR #4361, opened 2016-09-28
Fixing bookmarks toolbar "hover after click" behavior on folders

bsclifton:

When context menu is expanded for a bookmark folder, mousing over other folders will auto-expand them (closing your previous folder's context menu).
Matching behavior in Chrome, mousing over non-folders will hide the menu.

Fixes https://github.com/brave/browser-laptop/issues/1725

Also includes:
- Added constant to faviconUtils for default icon size (referenced from bookmarksToolbar component)
- Moved bookmarksToolbar component and faviconUtils to new directory structure

Auditors: @bridiver 

## Test Plan
1. Launch Brave and ensure you have bookmarks toolbar showing with at least 2 folders and 1 non-folder
2. Click one of the folders in bookmarks toolbar and confirm it opens
3. Mouse over the other folder and confirm it auto-expands the menu
4. Mouse over a non-folder (ex: a bookmark) and confirm it hides the menu
5. Mouse back over a folder and confirm it auto-expands it again
6. Click anywhere to dismiss the context menu
7. Drag a bookmark over a bookmark folder and confirm it auto-expands. For bonus points, you can try dragging to a folder inside a folder.


app/renderer/components/bookmarksToolbar.js
```diff
@@ -47,6 +48,21 @@ class BookmarkToolbarButton extends ImmutableComponent {
     }
   }
 
+  onMouseOver (e) {
+    // Behavior when a bookmarks toolbar folder has its list expanded
+    if (this.props.selectedFolderId) {
```
bridiver: I think that ideally this should trigger something like an `onMouseOverBookmarkFolder` action with whatever params are needed to allow the windowstore (browser process eventually) to update the state rather than having the logic here to open context menus and update state. The are two reasons for this:
1. It's more inline with how flux is intended to work. We should have declarative actions instead of imperative
2. We want to take node out of the main window renderers so we want to move electron api calls out of the react components (showBookmarkFolderMenu indirectly calls `electron.remote` methods) in preparation for moving them to the browser process

bsclifton: @bridiver:
1. I can update the code to do this.  It would be a larger change and would put "business logic" into the WindowStore.  For example, if the selection changes from null or -1 to a valid number, the WindowStore would know it needs to create a ContextMenu renderer control, which I guess it would handle by doing another windowAction?  Maybe we can talk this out more

2. This particular method (`showBookmarkFolderMenu`) doesn't use `electron.remote`. However, there are methods in that file (`contextMenus.js`) which do (specifically for creating / showing popup menu templates)

bridiver: I was looking at `this.props.showBookmarkFolderMenu` which at first looked like it called the electron api directly, but it calls `setContextMenuDetails` which eventually triggers a call the remote api. The windowStore really should directly care about any of this, the logic itself should be in a "reducer" similar `extensionState`. In the flux/redux model that is exactly where the logic is supposed to go. The UI should just fire off events and react to state changes. We have a bit of code to untangle because actions shouldn't call other actions, but some of the actions are "setters" and aren't written in a way that makes it easy to reuse the login in them.

test/components/bookmarksToolbarTest.js
```diff
@@ -0,0 +1,61 @@
...
+  // }
+
+  // function * addBookmarkFolder (name) {
+  //   yield this.app.client.moveToObject(bookmarksToolbar)
+  //       .rightClick(bookmarksToolbar)
+  //       .keys('ArrowDown')
+  //       .keys('ArrowDown')
+  //       .keys('ArrowDown')
+  //       .keys('ArrowDown')
+  //       .keys('Enter')
```
darkdh: You can try `.click('.contextMenuItemText=Add Folder')`

bsclifton: @darkdh thanks for checking this out- I did end up looking at it more and unfortunately, the menu that pops up is a system popup (not one of our custom rendered context menus, like autofill :disappointed:). I'll remove the commented out tests for now

bsclifton: Follow up- I did find a way to test this (see commits)

bsclifton:
Chatting w/ ya on Slack now :smile: 
- The sites have to be passed (for now) since they come from the appState. FolderId & bookmark are needed (for now)
- I ran into weird issues w/ passing windowState and assigning. I think it has to do with the emit. Will definitely have to revisit this

bsclifton:
If @bridiver gives the big thumbs up, let's merge this :smile: Assigning to 0.12.5...

bridiver:
++

PR #4369, opened 2016-09-29
Supersede cd3dd7829731f6e36eb70ac1ac4397598e615153 by "scroll-touch-edge"

darkdh:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Auditors: @bbondy

Test Plan:
1. Open google search result
2. Click any link and back to the search result
3. Make the window smaller by open dev console (shift + f8)
4. Make sure you can horizontally scroll the page
5. Make sure you can swipe back/forward when you scroll to the edge

PR #4083, opened 2016-09-17
Swipe behavior fix

darkdh:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

  1. Make swipe distance shorter
  2. Use scroll-touch-edge to solve swipe conflict

requires brave/electron#58

fix #2548, #4064

Test Plan:
  for swipe conflict
   1. Go to https://hootsuite.com
   2. Create timelines as much as you have to scroll horizontally

PR #4383, opened 2016-09-29
fix: update swipe navigation for electron native implementation of swipe events, closes #3299

### LINKED ISSUE #3299: Back/forward gesture does not respect system preferences
**Experienced behavior:**
I've tried to use the forward and backwards swipe gesture in Brave on my Macbook Pro trackpad but it never worked for me. I was quite puzzled because I've looked up if Brave supported macOS trackpad Gestures and read that in theory it should recognize them.

I later discovered that I've diverted from the default two finger "Swipe between pages" option in the Trackpad settings to three fingers.
screen shot 2016-08-22 at 00 10 25

When defaulting to the two finger option Brave oddly recognizes three finger back/forward gestures, but not the selected two finger ones, which doesn't really feel right.

**Expected behavior:**
Brave should adhere to the setting made in macOS and use two and three finger swipes according to the selected option, or at least support 3 finger swipes when they are selected in settings.
- Platform:
  macOS 10.11.6
- Brave Version:
  0.11.5
- Steps to reproduce:
  i. select "Swipe with three fingers" in Tackpad Settings->More Gestures->Swipe between pages
  ii. try to use three finger back/forward swipe gesture in brave
- Related Issue:
  Support MacBook trackpad pinch to zoom #1364
  Two-finger swipe gesture on macOS should have visual indicator #2477
  Support for three finger tap: "Look up & data detectors"
  Simulated 3 Finger Swipe Gestures not working #6587

lucidNTR:
Because swipe events can hardly be tested currently there has been a regression with the recent electron build.

electron now implements swipe events natively with the nice side effect of detecting 3 or 2 finger swipes depending on system settings.

as far as i understand the manual swipe detection is now completely obsolete and was removed

### DIFF
--- a/js/actions/webviewActions.js
+++ b/js/actions/webviewActions.js
@@ -4,8 +4,6 @@
 
 'use strict'
 
-const messages = require('../constants/messages.js')
-
 const getWebview = () =>
   document.querySelector('.frameWrapper.isActive webview')
 
@@ -50,21 +48,6 @@ const webviewActions = {
     const webview = getWebview()
     if (webview) {
       webview.showDefinitionForSelection()
-    }
-  },
-
-  /**
-   * Check two-finger gesture swipe back/forward ability
-   * @param {bool} back - true for back, false for forward
-   */
-  checkSwipe: function (back) {
-    const webview = getWebview()
-    if (webview) {
-      if (back) {
-        webview.send(messages.CHECK_SWIPE_BACK)
-      } else {
-        webview.send(messages.CHECK_SWIPE_FORWARD)
-      }
     }
   },
 
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -7,7 +7,6 @@ const ImmutableComponent = require('./immutableComponent')
 const Immutable = require('immutable')
 const electron = global.require('electron')
 const ipc = electron.ipcRenderer
-const systemPreferences = electron.remote.systemPreferences
 
 // Actions
 const windowActions = require('../actions/windowActions')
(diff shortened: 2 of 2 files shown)


Liunkae:
This would also fix #3299.

Liunkae:
This seems similar to PR #4369.

darkdh:
[swipe](http://electron.atom.io/docs/all/#event-swipe-macos) event only emitted when 3-finger swiping and we should use [scroll-touch-begin](http://electron.atom.io/docs/all/#event-scroll-touch-begin-macos), [scroll-touch-end](http://electron.atom.io/docs/all/#event-scroll-touch-end-macos) and [scroll-touch-edge](http://electron.atom.io/docs/all/#event-scroll-touch-edge-macos) to make sure 2-finger swiping working properly and also respect [swipe system preference](http://electron.atom.io/docs/all/#systempreferencesisswipetrackingfromscrolleventsenabled-macos) 
If we wanna fix #4383, we need to handle`swipe` event for sure but also need a granular system preferences indicated user chose 3-finger-swiping.
screen shot 2016-10-04 at 10 21 51

lucidNTR:
@darkdh i see i did not handle the swipeing with two fingers correctly but the merge  #4369 completely breaks my 3 finger swiping. i guess we need to check for system settings and either register swipe handlers or do the scroll touch begin and end handling. could you reopen this pr so i add the missing functionality?

darkdh:
#4369 just remove legacy code deprecated by https://github.com/brave/browser-laptop/pull/4083. And your PR will break current two finger swiping behavior. You don't need to remove the two finger swiping code to support three finger swiping. We should first implement the system preference detection in electron and based on the preference you can handle `swipe` event for three finger swiping.

bsclifton:
Since this has been open for a while, I'm going to close.  @lucidNTR if you would like to revise this per the comment by @darkdh, please open a new PR and we'll check it out :smile: Thanks!

PR #4396, opened 2016-09-30
ledger backup and recovery

### LINKED ISSUE #3350: implement ledger recovery keys screen
for advanced users who want to backup their ledger recovery keys

@bradleyrichter has mockups for this

jkup:

Fixes #3350

Auditors @bbondy @bradleyrichter 

Still remaining

- [x] Modals still need some custom CSS
- [x] Generate a new file with fs.writeStream for printing / saving
  - [x] Either ignore generated file or delete after printing or saving
- [x] Re-implement printing and saving with the new file
- [x] Send data from recovery form to ledger.js
  - [x] Send response back to UI and display message to user
- [x] Link through settings like minimum page time and minimum visits
- [x] At least write automated tests proving backup and recover flow


jkup:
Right now I'm working on the recovery flow. The thing I need help with is figuring out a good approach for printing / saving the keys!

bradleyrichter:
@jkup good catch. This should fix the UX issue:

image

ayumi:
@jkup yes, the appAction IPC -> ledger thing sounds legit

jkup:
This one is getting close. Could I get some eyes on the communication from @bbondy or @bridiver.

Right now preferences.js triggers and aboutAction to appStore to ledger.js (which is all working but maybe not the best approach.

Then (the last part I'm struggling with is getting the data back to preferences.js from ledger.js. I tried (as you can see in this branch) just emitting an ipc event but that doesn't seem to be working.

I can see that the recover task is working correctly, I just need to get that info back to the UI.

bridiver:
@jkup you communicate back to preferences by updating the state

js/stores/appStore.js
```diff
@@ -583,6 +586,27 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_BACKUP_KEYS:
+      const paymentId = appState.get('ledgerInfo').get('paymentId')
```
bridiver: I think you want to encapsulate this in a ledger.js method

js/stores/appStore.js
```diff
@@ -583,6 +586,27 @@ const handleAppAction = (action) => {
...
+          webContents.print({silent: false, printBackground: false})
+        } else {
+          webContents.downloadURL(fileUrl(filePath))
+        }
+      }).catch((err) => {
+        console.error(err)
+      })
+      break
+    case AppConstants.APP_RECOVER_WALLET:
+      ipcMain.emit(messages.LEDGER_RECOVER_WALLET, action.firstRecoveryKey, action.secondRecoveryKey)
```
bridiver: same as above. `ledger.js` runs in the same process as `appStore.js` so emitting an ipc event doesn't really make sense. Just call a method on ledger.js - see https://github.com/brave/browser-laptop/compare/tabs-api?expand=1#diff-23ca389e2bcb77191b5a9c10900eb3a3R717 for an example. The methods should always take the state and the action (reducer) as parameters and return the state (may or may not be modified). The flow should almost always be the same: call an action -> reducer functions update state in appStore -> ui updates to reflect the new state

js/about/aboutActions.js
```diff
@@ -88,6 +88,38 @@ const aboutActions = {
   },
 
   /**
+   * Generates a file with the users backup keys
+   */
+  generateKeyFile: function (backupAction) {
```
bridiver: it would probably be helpful to give these actions/constants names that include `ledger`. Eventually we'll probably break them out into ledger actions

js/about/preferences.js
```diff
@@ -722,7 +722,17 @@ class TabsTab extends ImmutableComponent {
 class PaymentsTab extends ImmutableComponent {
   constructor () {
     super()
+    this.state = {
```
diracdeltas: please put the state in `AboutPreferences` and pass it to `PaymentTab` as props.

js/stores/appStore.js
```diff
@@ -583,6 +586,27 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_BACKUP_KEYS:
+      const paymentId = appState.get('ledgerInfo').get('paymentId')
+      const passphrase = appState.get('ledgerInfo').get('passphrase')
+      const message = 'Your ledger keys are ' + paymentId + ' and ' + passphrase
```
diracdeltas: this should be localized

docs/state.md
```diff
@@ -478,8 +478,12 @@ WindowStore
       percentage: number,        // i.e., 0, 1, ... 100
       publisherURL: string,      // publisher site, e.g., "https://wikipedia.org/"
       faviconURL: string         // i.e., "data:image/...;base64,..."
+    } ],
+    options: {
```
diracdeltas: it would be useful to give this a more descriptive name, like `synopsisOptions`, because this property gets merged into ledgerData in js/components/frame.js

js/stores/appStore.js
```diff
@@ -583,6 +586,27 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_BACKUP_KEYS:
+      const paymentId = appState.get('ledgerInfo').get('paymentId')
+      const passphrase = appState.get('ledgerInfo').get('passphrase')
```
diracdeltas: use `getIn` here and in the line above to avoid chaining gets

diracdeltas:
please add a test or two for this in test/components/braveryPanelTest.js

js/about/preferences.js
jkup: Can you help me understand why? Will I also need to create and pass down actions for updating those keys as well? Thanks!

diracdeltas: yes. @bbondy can explain why

js/stores/appStore.js
```diff
@@ -583,6 +586,27 @@ const handleAppAction = (action) => {
...
+      const message = 'Your ledger keys are ' + paymentId + ' and ' + passphrase
+      const filePath = app.getPath('userData') + '/backup_keys.html'
+
+      fs.writeFile(filePath, message)
```
diracdeltas: @bridiver sorry, i meant either writeFileSync or put the rest of the block in the callback

js/stores/appStore.js
```diff
@@ -583,6 +586,27 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_BACKUP_KEYS:
+      const paymentId = appState.get('ledgerInfo').get('paymentId')
+      const passphrase = appState.get('ledgerInfo').get('passphrase')
+      const message = 'Your ledger keys are ' + paymentId + ' and ' + passphrase
+      const filePath = app.getPath('userData') + '/backup_keys.html'
```
bridiver: always use path.join for cross platform compat

js/stores/windowStore.js
```diff
@@ -431,6 +431,9 @@ const doAction = (action) => {
     case WindowConstants.WINDOW_NEW_FRAME:
       newFrame(action.frameOpts, action.openInForeground)
       break
+    case WindowConstants.WINDOW_VIEW_KEY:
+      newFrame(action.frameOpts, action.openInForeground)
```
bridiver: I think I like the idea of having a separate action for this, but there's not much point in having a separate action if all it does is call newFrame. I would suggest a few things:
1. This can be moved to the appStore now that we have a good way to open new tabs there
2. The handler for the action that opens the new window should know what the url and other options are

jkup:
@yan and @bridiver thank you so much for the feedback! I think I addressed all of your concerns except the localization one. I'm not sure how to localize a string while inside ledger.js. Do either of you know? Otherwise I can keep digging.

app/sessionStore.js
```diff
@@ -233,6 +233,8 @@ module.exports.cleanAppData = (data, isShutdown) => {
   data.temporarySiteSettings = {}
   // Delete Flash state since this is checked on startup
   delete data.flashInitialized
+  // Delete Recovery status on shut down
+  delete data.ui.about.preferences.recoverySucceeded
```
diracdeltas: plz document in docs/state.md. also isn't this going to throw an error for users who have never enabled the ledger?

bsclifton: @diracdeltas that is a great point- one thing we'll want to try is an upgrade from the previous version to make sure that loading or saving the state doesn't break.

@jkup Can these delete operations be wrapped in a "tryDelete" or "safeDelete" type of method that checks for existence before deleting?

jkup: Yeah, should I just wrap it in an if check or do other operations need this too?

diracdeltas: i think you just need to make sure this one particular line doesn't cause `Cannot convert undefined or null to object` errors, so a try-catch would be fine

js/stores/appStore.js
```diff
@@ -328,6 +328,9 @@ function handleChangeSettingAction (settingKey, settingValue) {
 }
 
 const handleAppAction = (action) => {
+  const ledger = require('../../app/ledger')
+  ledger.boot()
```
jkup: I'm not super familiar with the ledger but from looking at it I think I can remove the .boot() line. Because the "recover" and "backup" actions shouldn't be able to be fired unless ledger is already enabled -- right?

diracdeltas: deferring to @mrose17 on whether this line should be here

jkup: @diracdeltas I took it out and it seems to work fine but would love to hear @mrose17 weigh in!

app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -110,6 +117,20 @@ offerSearchSuggestions=Autocomplete search term as you type
 doNotTrackTitle=Do Not Track
 doNotTrack=Send a 'Do Not Track' header with browsing requests (requires browser restart)
 blockCanvasFingerprinting=Fingerprinting Protection (may break some websites)
+advancedSettings=Advanced Settings...
+advancedSettingsTitle=Advanced Settings for Brave Payments
+ledgerRecoveryTitle=Recover your Brave wallet
+ledgerRecoverySubtitle=Enter your recovery keys below
+ledgerRecoveryContent=The balance of the recovered wallet will be transferred to your new Brave wallet. The old wallet will still exist as an empty wallet.
+ledgerBackupTitle=Backup your Brave wallet
+ledgerBackupContent=Below, you will find the anonymized recovery keys that are required if you ever lose access to this computer. We recommend that you print or save these keys and store them in a safe place, like your local safe deposit box, or under your mattress. It's really up to you!
```
diracdeltas: nit: can we get rid of the `like your local safe deposit box, or under your mattress. It's really up to you!` part? if i were a hacker i would start looking under people's mattresses after this launched. :p

diracdeltas:
it has become unclear what the `show notifications` switch does. the placement makes it look like it is related to minimum visits and page time.

i would say move it to the far right of the modal and change text to `Show payment notifications`

diracdeltas:
the `Print keys` and `Save recovery file..` functionality isn't working. Just says LEDGERBACKUPTEXT right now.
screen shot 2016-10-07 at 4 17 49 pm

app/extensions/brave/locales/en-US/preferences.properties
diracdeltas: fwiw it is probably better in most cases to store this digitally in a password manager or such instead of printing a physical copy; but i'm okay with just leaving out storage suggestions for now.

app/sessionStore.js
bridiver: part of that change was removing the side-effects and returning the new state so we can make it immutable and use things like `deleteIn`

jkup:
@diracdeltas sorry about that, should be fixed now.

bbondy:
@jkup Please rebase this on master
@diracdeltas could you review this today?

diracdeltas:
sure i can review. @jkup please address the remaining comments and open an issue for adding tests for this PR, thx

app/extensions/brave/locales/en-US/preferences.properties
bradleyrichter: fine to remove on this one. Just remember that we do want to inject and retain some personality in places. (inspired by slack) But please suggest improvements in all areas when you encounter them.

jkup:
@diracdeltas cool thanks! I made #4657 to track automated tests and fixed those comments you made. Also we need to keep in mind this ticket is dependent on wip-tabs-api from @bridiver

diracdeltas:
thanks, lgtm! but why do my ledger recovery keys contain non-ascii characters?
the text file says
```
Your ledger keys are â¨[redacted]â© and â¨[redacted]â©
```

jkup:
Ah ha! @bridiver told me it was ghosts!

diracdeltas:
pushing a fix for unicode thing now

js/stores/appStore.js
```diff
@@ -585,6 +587,21 @@ const handleAppAction = (action) => {
     case AppConstants.APP_SET_DICTIONARY:
       appState = appState.setIn(['dictionary', 'locale'], action.locale)
       break
+    case AppConstants.APP_BACKUP_KEYS:
+      ledger.backupKeys(appState, action)
```
bridiver: this should be `appState = ledger.backupKeys(appState, action)` to capture any state updates that this method might make. I know it doesn't directly make any now, but the pattern should always be the same to avoid confusion

js/stores/appStore.js
```diff
@@ -585,6 +587,21 @@ const handleAppAction = (action) => {
...
+      break
+    case AppConstants.APP_RECOVER_WALLET:
+      ledger.recoverKeys(appState, action)
```
bridiver: same as above

app/ledger.js
```diff
@@ -216,6 +234,51 @@ var boot = () => {
...
+
+  return appState
+}
+
+/*
+ * Recover Ledger Keys
+ */
+
+var recoverKeys = (appState, action) => {
+  client.recoverWallet(action.firstRecoveryKey, action.secondRecoveryKey, (err, body) => {
```
bridiver: this callback may or may not happen asynchronously so the appActions should be called inside `setImmediate` to prevent any potential race conditions with state updates. The issue was discussed in slack, but this is how it happens:

```
For simplicity, all actions increment the current count by 1, but the same problem exists even if the actions change different parts of the state
1. action1 is called with the current appState = {count: 0}
2. action1 calls a method with a callback
3. callback returns synchronously (does not perform any io or other operations that would make it async)
4. action2 is called inside callback with the current appState = {count: 0}
5. action2 returns appState = {count: 1}
6. appState is emitted
7. action1 returns appState = {count: 1}
8 app state is emitted
```

bridiver:
++

PR #4458, opened 2016-10-03
Documentation nit

posix4e:
I did not think this required an issue. I just thought we should be a bit more explicit.


bbondy:
++

Issue #4416, opened 2016-10-01
Caption buttons inconsistency


luixxiul:
**Describe the issue you encountered:**

Open a window and F11 -> ![clipboard01](https://cloud.githubusercontent.com/assets/3362943/19011067/dcafd47c-87c6-11e6-96f0-ad9eefb16240.png)

Open a window, make it maximize, and F11 -> ![clipboard01](https://cloud.githubusercontent.com/assets/3362943/19011075/00050b0e-87c7-11e6-92c7-9c8b1e5a5eb7.png)

**Expected behavior:** The max button in the 2nd picture should change the browser from fullscreen mode to the maximized state. Currently the button does not work.

- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 10

- Brave Version: 0.12.3 RC5

- Steps to reproduce: noted above

- Any related issues: #4377

PR #4463, opened 2016-10-03
Polished up full screen detection logic (Windows only)

bsclifton:
Polished up full screen detection logic (Windows only)

- Full screen mode shows button as restore
- Clicking the restore button (when full screen) will:
  - Undo full screen
  - Restore window back to the non-maximized size

Fixes https://github.com/brave/browser-laptop/issues/4416

Extra notes captured (below) since this is late into 0.12.3

Auditors: @alexwykoff @bbondy 

## Test Plan
1. Launch Brave and make sure your window is not maximized
2. Go into full screen mode (F11)
3. Notice the caption button for maximize/restore. It should be set as restore button.
4. Hit the restore button
5. Notice status of window and caption buttons:
  - window should no longer be "full screen"
  - window should be back to original size
  - maximize button should match the status (single square for not-maximized) 
6. Use `maximize` caption button to maximize window
7. Once maximized, hit F11 to enter full screen mode
8. Push the `restore` button and confirm:
  - window should no longer be "full screen"
  - window should now be smaller
  - maximize button should match the status (single square for not-maximized) 

## Notes about existing behavior
- You can double click any draggable area to maximize the window (executes different code than clicking the caption buttons)
- If window is full screen (via F11):
  - you can NOT drag the window (this is expected)
  - you can NOT double click a draggable are to maximize (also expected)
- Full screen and maximize status are persisted to session

### DIFF
--- a/app/renderer/components/windowCaptionButtons.js
+++ b/app/renderer/components/windowCaptionButtons.js
@@ -43,6 +43,12 @@ class WindowCaptionButtons extends ImmutableComponent {
   }
 
   onMaximizeClick (e) {
+    if (currentWindow.isFullScreen()) {
+      // If full screen, toggle full screen status and restore window (make smaller)
+      currentWindow.setFullScreen(false)
+      if (currentWindow.isMaximized()) currentWindow.unmaximize()
+      return false
+    }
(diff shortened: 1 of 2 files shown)


bbondy:
++

PR #4469, opened 2016-10-03
Using temporarySiteSettings to store private tab preferences

### LINKED ISSUE #4468: Private tab preferences
Make bravery panel, site permissions, mixed content...(temporarySiteSettings) works in private tab.

darkdh:

fix #4468

Auditors: @diracdeltas , @bridiver

Test Plan:
Test bravery panel, site permissions, mixed content, flash in private tabs and the preferences should not write to disk and affect regular tabs

---

There is a better way to achieve this but require architectural change.
We plan to supersede siteSettings and temporarySiteSettings by reading and writing user prefs directly. Private sessions will keep in-memory overlaid user prefs, but still be able to read the normal session prefs that haven’t been changed in private tabs.

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -717,7 +717,7 @@ class Main extends ImmutableComponent {
   get allSiteSettings () {
     const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
     if (activeFrame && activeFrame.get('isPrivate')) {
-      return this.props.appState.get('siteSettings').merge(this.props.appState.get('temporarySiteSettings'))
+      return this.props.appState.get('siteSettings').mergeDeep(this.props.appState.get('temporarySiteSettings'))
     }
     return this.props.appState.get('siteSettings')
   }
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -715,6 +715,10 @@ class Main extends ImmutableComponent {
   }
 
   get allSiteSettings () {
+    const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
+    if (activeFrame && activeFrame.get('isPrivate')) {
+      return this.props.appState.get('temporarySiteSettings')
+    }
     return this.props.appState.get('siteSettings')
   }
 
@@ -741,11 +745,6 @@ class Main extends ImmutableComponent {
   }
 
   get braveShieldsDisabled () {
-    const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
-    if (activeFrame && activeFrame.get('isPrivate')) {
-      return true
-    }
-
     const activeRequestedLocation = this.activeRequestedLocation
     if (!activeRequestedLocation) {
       return true
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -717,7 +717,7 @@ class Main extends ImmutableComponent {
   get allSiteSettings () {
     const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
     if (activeFrame && activeFrame.get('isPrivate')) {
-      return this.props.appState.get('temporarySiteSettings')
+      return this.props.appState.get('siteSettings').merge(this.props.appState.get('temporarySiteSettings'))
     }
     return this.props.appState.get('siteSettings')
   }
--- a/js/state/contentSettings.js
+++ b/js/state/contentSettings.js
@@ -79,6 +79,53 @@ const getBlock3rdPartyStorage = (braveryDefaults) => {
   }
 }
 
+const hostSettingsToContentSettings = (hostSettings, contentSettings) => {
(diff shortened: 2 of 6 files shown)


js/state/contentSettings.js
```diff
@@ -126,51 +173,8 @@ const getContentSettingsFromSiteSettings = (appState) => {
...
-    if (hostSetting.shieldsUp === false) {
-      addContentSettings(contentSettings.cookies, hostPattern, '*', 'allow')
-      addContentSettings(contentSettings.canvasFingerprinting, hostPattern, '*', 'allow')
-      addContentSettings(contentSettings.adInsertion, hostPattern, '*', 'block')
-      addContentSettings(contentSettings.javascript, hostPattern, '*', 'allow')
-      addContentSettings(contentSettings.referer, hostPattern, '*', 'allow')
-    }
-  }
+  hostSettingsToContentSettings(appState.get('siteSettings').toJS(), contentSettings)
+  hostSettingsToContentSettings(appState.get('temporarySiteSettings').toJS(), contentSettings)
```
bridiver: this is going to send the private settings to the non-private tabs. We can't sync all the sessions, we have to send the temp + regular to the private session and just the regular to the regular session. Probably use `setUserPref` with the `incognito` flag instead of `registerUserPrefs`

diracdeltas: shouldn't `temporarySiteSettings` get added only if the content settings are being sent to a private frame?

diracdeltas: nvm @bridiver made the same comment above

js/components/main.js
```diff
@@ -715,6 +715,10 @@ class Main extends ImmutableComponent {
   }
 
   get allSiteSettings () {
+    const activeFrame = FrameStateUtil.getActiveFrame(this.props.windowState)
+    if (activeFrame && activeFrame.get('isPrivate')) {
+      return this.props.appState.get('temporarySiteSettings')
+    }
```
diracdeltas: need to be careful here that the `allSiteSettings` frame property is consistent with what is being sent in `js/state/contentSettings.js` and also `registerPermissionHandler` in `app/filtering.js`. the expected behavior is that private context inheirit settings from regular contexts but not vice-versa, so I think what you want here is a merge of `temporarySiteSettings` and `siteSettings` for private frames.

js/state/contentSettings.js
darkdh: Addressed in 7de0c05.

js/components/main.js
darkdh: Address in 7de0c05.

js/state/contentSettings.js
```diff
@@ -126,85 +175,51 @@ const getContentSettingsFromSiteSettings = (appState) => {
...
-      addContentSettings(contentSettings.cookies, hostPattern, '*', 'allow')
-      addContentSettings(contentSettings.canvasFingerprinting, hostPattern, '*', 'allow')
-      addContentSettings(contentSettings.adInsertion, hostPattern, '*', 'block')
-      addContentSettings(contentSettings.javascript, hostPattern, '*', 'allow')
-      addContentSettings(contentSettings.referer, hostPattern, '*', 'allow')
-    }
+  const regularSettings = hostSettingsToContentSettings(appState.get('siteSettings').toJS(), contentSettings)
+  if (isPrivate) {
+    const privateSettings =
+      hostSettingsToContentSettings(appState.get('siteSettings').merge(appState.get('temporarySiteSettings')).toJS(),
```
diracdeltas: s/merge/mergeDeep

bridiver: sorry I missed this earlier, but the deep merge is not required or desirable. The overlay pref store already handles this correctly if you set the regular settings for the regular profile and only the private settings for the private profile

bridiver:
```
bool OverlayUserPrefStore::GetValue(const std::string& key,
                                    const base::Value** result) const {
  // If the |key| shall NOT be stored in the overlay store, there must not
  // be an entry.
  DCHECK(ShallBeStoredInOverlay(key) || !overlay_.GetValue(key, NULL));

  if (overlay_.GetValue(key, result))
    return true;
  return underlay_->GetValue(GetUnderlayKey(key), result);
}
```

diracdeltas:
thanks a ton, this is mostly working! i did find one bug:
1. open http://homestarrunner.com/ in a regular tab and in a private tab
2. right-click and choose 'allow once' in the private tab
3. the regular tab also reloads and Flash is allowed in it

i am pretty sure this is because you need to edit `js/contextMenus.js` to keep track of whether the Flash right-click context menu is being opened in a private tab.

also, please add a private tab test to `braveryPanelTest.js`.

bbondy:
@darkdh going to merge this in so we can get some testing, but please do the last comments in a follow up bug.

darkdh:
will do

PR #4489, opened 2016-10-04
Regional adblock

bbondy:

Test Plan:
- Add gravatar.com to the custom adblock rules, and load brianbondy.com, the avatars should be gone.
- Test some regional sites, here are the lists https://github.com/bbondy/abp-filter-parser-cpp/blob/master/lib/regions.js you can load a listURL to see the resources which are blocked in sites.


bbondy:
@bsclifton will review and I'll do changes in a follow up.

app/adBlock.js
```diff
@@ -10,6 +10,18 @@ const ABPFilterParser = ABPFilterParserLib.ABPFilterParser
 const FilterOptions = ABPFilterParserLib.FilterOptions
 const DataFile = require('./dataFile')
 const Filtering = require('./filtering')
+const appConfig = require('../js/constants/appConfig')
+const debounce = require('../js/lib/debounce.js')
```
bsclifton: gotta drop the `.js` :smile:

app/filtering.js
```diff
@@ -31,6 +31,7 @@ const app = electron.app
 const uuid = require('node-uuid')
 const path = require('path')
 const getOrigin = require('../js/state/siteUtil').getOrigin
+const {adBlockResourceName} = require('./adBlock')
```
bsclifton: Small style question- our project has a mix of:
- destructuring assignments with and without spaces around the item(s) (see line 34)
- including using require and then referencing the method name (see line 33)

Is there a preferred syntax?

app/filtering.js
```diff
@@ -571,7 +581,10 @@ module.exports.isResourceEnabled = (resourceName, url) => {
   }
 
   if ((resourceName === appConfig.resourceNames.ADBLOCK ||
-       resourceName === appConfig.resourceNames.TRACKING_PROTECTION)) {
+      appConfig[resourceName] &&
```
bsclifton: It was there before, but has unnecessary double parens
if((condition === value)) { }

docs/state.md
```diff
@@ -94,33 +94,16 @@ AppStore
...
-  adblock: {
+  // resourceIdentifier is one of: `adblock`, `safeBrowsing`, `trackingProtection`,
+  //   `httpsEverywhere`, or another additional resource by name such as
+  //   adblock regional resource files.
+  [resourceIdentifier]: {
     etag: string, // last downloaded data file etag
     lastCheckVersion: string, // last checked data file version
     lastCheckDate: number, // last checked data file date.getTime()
-    enabled: boolean, // Enable adblocking
+    enabled: boolean, // Enable the resoruce
```
bsclifton: s/resoruce/resource :)

js/about/aboutActions.js
```diff
@@ -247,6 +247,32 @@ const AboutActions = {
...
+      actionType: appConstants.APP_UPDATE_ADBLOCK_DATAFILES,
+      uuid,
+      enable
+    })
+  },
+
+  /**
+   * Dispatches an event to the renderer process to register or deregister a datafile
+   *
+   * @param {rules} ABP filter syntax rule string
```
bsclifton: might be a copy/paste thing, but are these comments right? to register or deregister? (looks like it's updating existing rules that are in place)

bsclifton:
LGTM :smile: great job w/ the tests <3

app/adBlock.js
bbondy: doh that was from copy/paste, I'll get rid of all files at once when I fix this to avoid that

bbondy: db001147cba8ca30c853c753b98752d074337ac4

app/filtering.js
bbondy: Yep it was probably there because at one point they mattered and then after a bugfix they were left behind. Fixed as part of:
ecc67f28714ec923d59d639f79415364712ad214

docs/state.md
bbondy: Fixed as part of:
ecc67f28714ec923d59d639f79415364712ad214

js/about/aboutActions.js
bbondy: Fixed as part of:
ecc67f28714ec923d59d639f79415364712ad214

app/filtering.js
bbondy: 9b1aa09da5b8b213749c94da102ae92a2357f1dd

PR #4514, opened 2016-10-04
handle getBraveryProperties error

### LINKED ISSUE #4421: Brave Payments : First time user wallet creation : If a user attempts to change the value before the wallet is created, Brave will crash.
Yes

If a user is starting for the first time and attempt to change the monthly allowance while their wallet is being created, Brave will crash.

**Expected behavior:**
Brave should not crash and the selected value should be respected.

Tested on Win7 x64

- Brave Version:
0.12.3 RC5

- Steps to reproduce:
    1. With a clean install, go to the Payments preference pane, enable payments.
    2. While the wallet is being created, select another value from the 'Monthly budget' dropdown.

- Screenshot if needed:

- Any related issues:

mrose17:
Test Plan: enable 'brave payments', then IMMEDIATELY change the monthly budget. the browser should *not* crash

if a wallet is under creation, but not yet creating, `getBraveryProperties` should not be invoked.

fixes #4421

auditor: @ayumi

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -1245,9 +1245,15 @@ var getPaymentInfo = () => {
 }
 
 var setPaymentInfo = (amount) => {
+  var bravery
+
   if (!client) return
 
-  var bravery = client.getBraveryProperties()
+  try { bravery = client.getBraveryProperties() } catch (ex) {
+// wallet being created...
+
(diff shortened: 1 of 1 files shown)


ayumi:
tried out out / lgtm :gift:

PR #4342, opened 2016-09-28
Adds rounded borders to the top corners for Windows 7

bsclifton:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Adds rounded borders to the top corners for Windows 7
screen shot 2016-09-27 at 6 06 05 pm

- Introduces a platformUtil helper
- Top level element uses this helper to add classes for platform and release (Windows only)
- removed OS detection logic in WindowCaptionButtons component
- reworked styles in navigationBar.less

Manually tested on:
- Windows 7
- Windows 10
- Windows 7 at 150% DPI (corners didn't look exactly right, but might be a VM issue)

Fixes https://github.com/brave/browser-laptop/issues/4216

Auditors: @bbondy @bridiver

PR #4524, opened 2016-10-04
Remove border radius from Windows 7; it has issues on higher DPI setups.

bsclifton:
manually revert border radius change from https://github.com/brave/browser-laptop/pull/4342

Auditors: @bbondy

Test Plan:
1. Launch Brave on Windows 7
2. Ensure the top left and top right corners are not rounded with a black outline.
   They may be rounded with NO outline (and that's OK)
3. Crank DPI up to 150%
4. Ensure top left/right borders are square and look correct

### DIFF
--- a/less/navigationBar.less
+++ b/less/navigationBar.less
@@ -78,11 +78,6 @@
 
   // Windows 7 specific styles
   &.win7 {
-    > div#window.frameless {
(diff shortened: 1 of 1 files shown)


bbondy:
++

PR #4564, opened 2016-10-06
Do not re-create shortcuts on desktop after upgrading on Windows

aekeus:
Auditors: @bbondy

Test Plan:

  Located in https://github.com/brave/browser-laptop/issues/1181


bbondy:
thanks!

PR #4615, opened 2016-10-07
do not show coinbase for unsupported countries

### LINKED ISSUE #4292: UX: Strange label “Need to buy Bitcoin outside of the USA?”
“Need to buy Bitcoin outside the USA?” — Well, I don’t know. You tell me!

See the attached screenshot. The label on the first option makes sense to me because I know that it shows something else in the USA. To anybody else, however, the wording is very strange. It also doesn’t really explain what is going on in this dialog nor guide the user towards completing the required steps.

I suggest it be changed to something like “Start by purchasing some BitCoins from an exchange, and then transfer your BitCoins to the wallet address shown below. We recommend so-and-so.  Visit so-and-so.”

It should also mention an amount of BitCoins to purchase and how much that would cost in a local currency (or EUR?). These numbers are important to make the user feel more familiar with BitCoins and their worth in a currency they can relate to. I suggest recommending that the user start with 3 times the minimal monthly contribution so the user will be set up for the first three months. This helps reduce the transaction costs involved with currency exchanges and commits the user to three months. “We recommend you purchase 0,0009 BitCoins (15 EUR) at so-and-so, but you can get started with Brave Payments for as little as  0,0003 BitCoins (5 EUR) per month.” 

The headline is also off, but I’m not sure what can be done to clarify it. (The first option doesn’t transfer funds to Brave, and the second and third option is really the same option.) Maybe “Add BitCoins to your Brave Wallet to get started”

screenshot

jkup:

Fix #4292 

Test Plan:

Open Brave from a country where coinbase is not supported. I set my VPN to Israel. Then make sure the following changes occur in the add funds dialog.

+ Title goes from "Three ways to add.." to just "Add Bitcoin"
+ Coinbase logo disappears from the bottom
+ Click QR code and make sure coinbase logo is also gone from that footer.


js/about/preferences.js
```diff
@@ -300,9 +301,13 @@ class BitcoinDashboard extends ImmutableComponent {
...
-      <div id='coinbaseLogo' />
-      <div id='appstoreLogo' />
-      <div id='playstoreLogo' />
+      {
+        coinbaseCountries.indexOf(this.props.ledgerData.get('countryCode')) > -1
+        ? <div id='coinbaseLogo' />
+        : null
+      }
+      <a href='https://itunes.apple.com/us/app/coinbase-bitcoin-wallet/id886427730?mt=8' target='_blank' id='appstoreLogo' />
+      <a href='https://play.google.com/store/apps/details?id=com.coinbase.android' target='_blank' id='playstoreLogo' />
```
diracdeltas: why are we linking to the coinbase app for countries where coinbase is not supported?

js/about/preferences.js
```diff
@@ -376,11 +381,13 @@ class BitcoinDashboard extends ImmutableComponent {
...
-      <Button l10nId='done' className='pull-right whiteButton' onClick={this.props.hideParentOverlay} />
-    </div>
+    if (coinbaseCountries.indexOf(this.props.ledgerData.get('countryCode')) > -1) {
+      return <div className='panelFooter'>
+        <div id='coinbaseLogo' />
+        <span className='coinbaseMessage' data-l10n-id='coinbaseMessage' />
+        <Button l10nId='done' className='pull-right whiteButton' onClick={this.props.hideParentOverlay} />
+      </div>
+    }
   }
```
diracdeltas: should return null otherwise

js/about/preferences.js
jkup: sorry! fixed.

diracdeltas:
when my country code is FR, it tells me to buy btc through cubits but still shows the coinbase logo and footer.
screen shot 2016-10-10 at 1 55 23 pm

diracdeltas:
when my country code is `MD`, the coinbase buy button is there, but clicking it returns a 403 from buy.coinbase.com and does nothing. the same thing happens on master FWIW. i think the list of country codes in this PR and the list of supported countries returned by ledger is not the same.

@mrose17

jkup:
@diracdeltas hmm that's a good catch. Perhaps there is something wrong with our approach. @mrose17 found a list of "supported countries" here https://www.coinbase.com/global but they don't all seem to be supported.

diracdeltas:
2 options:
1. assume js/constants/coinbaseCountries is authoritative and change `BitcoinDashboard.exchangePanel` accordingly
2. assume ledgerInfo.exchangeInfo is authoritative and change the footer logo/links accordingly

mrose17:
@jkup - i agree with @diracdeltas

jkup:
@mrose17 @diracdeltas just to make sure I'm clear. We should do away with all of the checks for the user being in the US and only have whether or not they are in constants/coinbaseCountries?

diracdeltas:
did i understand correctly today that the widget only works in the US but the coinbase apps work elsewhere? if so, this at least should change: https://github.com/brave/browser-laptop/blob/f4418d64977dedb564c4be755e4de30f921ab71b/js/about/preferences.js#L353.

jkup:
@diracdeltas I had a misunderstanding with the logic, should all be working now!

diracdeltas:
lgtm

Issue #3888, opened 2016-09-11
Brave payments: Line should always appear when user scroll down


bitcoinuser:
Hi,

I think this line on brave payments should always appear when user scroll down.

![captura de tela de 2016-09-11 09-10-53](https://cloud.githubusercontent.com/assets/1178856/18417227/5946a924-7800-11e6-8f24-f6a5be035ea3.png)

mrose17:
@jkup - fyi: this probably should be dealt with at the same time as we deal with https://github.com/brave/browser-laptop/issues/3792

jkup:
@bsclifton I think you had some good feedback about this table in chat somewhere but I can't find it. Was it lacking keyboard navigation or something?

bsclifton:
Yes! the scrollbar was on the table itself, not on the window.  So it wasn't straightforward to scroll through the list (especially without a mouse)

jkup:
@bsclifton ah yes! ok great, I'll add that work to this ticket. Thanks.

bradleyrichter:
this solution doesn't seem to be working well enough

![image](https://cloud.githubusercontent.com/assets/13509546/19098617/8c100f6a-8a63-11e6-9f55-9caf537c5d2e.png)


@jkup  Can we instead float the entire top area including the list header row and then use the entire window for scrolling?

jkup:
@bradleyrichter yeah I mean this is working as expected with the CSS hack I added to make the headers sticky. I agree that it would look much nicer if we didn't set a max-height on the list and just let the window scroll. Should I make a new issue for this? Either way I can get it in this week.

bradleyrichter:
Sure. Close and make a new issue if you prefer.

bbondy:
Reclosing but pls do a new issue since the change would be in 0.12.5 and this is in 0.12.4.

PR #4616, opened 2016-10-07
Ledger list should not have maximum height

### LINKED ISSUE #4537: Ledger list should not have maximum height
This relates to #3888

Right now the list has a max height causing it to have a separate scroll container. It would be better if it went the entire page height so there was only a scrollbar on the window.

jkup:

Fix #4537

Test Plan:

Instead of a fixed-height box with it's own scroll. The entire Payments page should scroll if you have enough publishers.

screen shot 2016-10-07 at 3 11 06 pm


diracdeltas:
i added the fixed-height thing because it was in one of @bradleyrichter's original design specs. just making sure he is ok with this change

jkup:
@diracdeltas totally. People were complaining that they didn't know you could scroll so he suggested this change here: https://github.com/brave/browser-laptop/issues/3888#issuecomment-251562491

diracdeltas:
ah ok, cool

diracdeltas:
lgtm, thanks

PR #4668, opened 2016-10-11
Removes unnecessary --compilers arg

jonathansampson:
Test Plan:

In accordance with http://mochajs.org/#about-babel


bsclifton:
++

PR #4658, opened 2016-10-11
WIP: Bookmark manager style update

bsclifton:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [ ] Ran `git rebase -i` to squash commits (if needed).

Fixes brave#4562
Fixes brave#1994
Fixes brave#4523

## Completely updated styles
Almost done- needs some tweaks
screen shot 2016-10-11 at 2 02 04 am

## Test Plan:
TBD

PR #4669, opened 2016-10-11
Fix autofill regression caused by refactor

bsclifton:
Some of the refactoring done in #4361 unfortunately broke autofill. This commit reverts the parts that broke it back to the pre-refactor state and keeps the parts that were OK.

Properly fixes https://github.com/brave/browser-laptop/issues/1725

`js/store/windowStore.js` and `js/contextMenus.js` had a circular dependency with each other. Ultimately, `contextMenus.js` shouldn't have a dependency on windowStore (IMO).  I tried to keep the refactored code in place and while eliminating the windowStore dependency in `contextMenus.js` and realized this is not an easy task (we'll have to revisit in future)

Auditors: @darkdh @bridiver

Test Plan: follow steps from original PR. The original automated tests work great


js/stores/windowStore.js
```diff
@@ -841,7 +813,7 @@ const doAction = (action) => {
       if (getSetting(settings.AUTO_HIDE_MENU)) {
         doAction({actionType: WindowConstants.WINDOW_TOGGLE_MENUBAR_VISIBLE, isVisible: false})
       } else {
-        hideContextMenu(action)
+        doAction({actionType: WindowConstants.WINDOW_SET_CONTEXT_MENU_DETAIL})
       }
       doAction({actionType: WindowConstants.WINDOW_SET_SUBMENU_SELECTED_INDEX})
       doAction({actionType: WindowConstants.WINDOW_SET_BOOKMARKS_TOOLBAR_SELECTED_FOLDER_ID})
```
bsclifton: Glad this came up- this is something I missed while doing the refactor. Resetting the menu status is supposed to clear the selected folder ID. This action never got renamed to the new one

bsclifton:
This PR no longer needed- it was included as one of the commits in https://github.com/brave/browser-laptop/pull/4658 (I was still working on that one and was going to drop it via rebase)

bridiver:
agree on the contextMenus -> windowStore dependency. Any state it requires should be passed in like any other component

PR #4682, opened 2016-10-11
Tests require --global mocha

jonathansampson:
Tests assume `mocha` is globally recognized.

Auditors: @bsclifton


bsclifton:
++

PR #4719, opened 2016-10-12
Updated test documentation

bsclifton:
Updated test documentation after discussion in Slack about best practices

Auditors: @bridiver

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -6,7 +6,8 @@ Tests use the [mocha test framework](https://mochajs.org/). Start by installing 
 
     npm install --global mocha
 
-Most tests use [webdriver.io](http://webdriver.io/) framework (via [Spectron](https://github.com/electron/spectron) to bring up the browser and run through the tests.
+Most tests use [webdriver.io](http://webdriver.io/) framework (via [Spectron](https://github.com/electron/spectron)) to bring up the browser and run through the tests.
+Please [check out the API documentation](http://webdriver.io/api.html) for a comprehensive list of commands we can use in our tests.
 
 Tests are located in the top level `test` directory and their filenames have a suffix of `Test.js`.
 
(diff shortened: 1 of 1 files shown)


docs/tests.md
```diff
@@ -37,6 +38,78 @@ You can run a subset of tests which match a `description` or `it` with:
...
 
+## Best practices for writing tests
+
+- If you do anything that opens a new tab, you have to validate that the tab has opened before trying to switch to
+another tab/window. This is the most common cause of race conditions that I find in tests. webdriver automatically
+switches contexts to any new "window" that is created. chromedriver can’t differentiate between tabs and windows in
+brave which is why we have special helpers. If there is more than one tab with the same url, you can’t use waitForUrl:
+it will return on the tab/window that already exists and not wait for the new one. `getTabCount` is a good alternative
+because it doesn’t require a particular window context to run
+
```
bridiver: maybe link to the examples in #browser ?

docs/tests.md
```diff
@@ -37,6 +38,78 @@ You can run a subset of tests which match a `description` or `it` with:
...
+- getWindowState
+
+### Site management
+- addSite: add a bookmark/folder or history entry
+- addSiteList
+- removeSite
+
+### Preferences
+- changeSetting
+- changeSiteSetting
```
bridiver: This isn't necessarily relevant to this pr, but I think we need more helpers like this but with a slight change. Right now `changeSiteSetting` sets the setting, but doesn't wait for it to actually be updated in the store. This leaves the dev with the task of both recognizing that the update is async and also writing/duplicating code to check for the update completion cc @bbondy

bbondy: agreed, we should wait for the change on these so they don't need to be in every test.

bsclifton:
Updated! Merging... :smile:

Issue #3282, opened 2016-08-20
Reduced specificity of select class in css


rohanthacker:
**Did you search for similar issues before submitting this one?**
 Yes
**Describe the issue you encountered:**
The style of select element used is deeply nested under settings

**Expected behavior:**
Reduce code duplications by assigning classes, similar to the pattern used in other elements, For Eg :  button. 

- Platform (Win7, 8, 10? macOS? Linux distro?):
All 
- Brave Version:
11.3
- Steps to reproduce:
    1. #2894 , Required Dropdown for Bookmarks
    2.  Select Element used in Brave Menu 
<img width="476" alt="screen shot 2016-08-21 at 3 29 45 am" src="https://cloud.githubusercontent.com/assets/1190324/17834105/98eadde6-674f-11e6-83dd-30847d72523b.png">
    
3.Select Element Used in Settings, similar to bookmarks but nested under settings classes.
<img width="389" alt="screen shot 2016-08-21 at 3 29 56 am" src="https://cloud.githubusercontent.com/assets/1190324/17834106/98edc236-674f-11e6-8802-cc30a62ae336.png">

bradleyrichter:
the brave menu should ideally use the in-page style for popup menus.

PR #4726, opened 2016-10-13
Improve UI/UX of bookmark popup

### LINKED ISSUE #2894: Improve UI/UX of bookmark popup

jkup:

Fix #2894, #4854, #4864 and #3282 via @rohanthacker's changes!

Auditors @bsclifton @bradleyrichter 

Test Plan:

Make sure of a few things:

+ Bookmark is automatically created when you click the orange star
+ Bookmark doorhanger aligns with orange star
+ View all bookmarks page takes you to about:bookmarks


bsclifton:
Checking it out right now, already noticed a few things (doh)

Launching w/ an empty tab, the icon is off:
screen shot 2016-10-13 at 10 56 01 pm

After you visit a site though, it fixes itself and looks OK.  However, the star icon (when a site is favorited) moves up and down when you have the magic titlebar feature on and show/hide the URL bar.

When you click it to trigger a bookmark, I can tell it's trying to put an overlay there, but it's in the wrong area. I ran into a similar issue when doing the titlebar (but mine was in main.js, where your overlay appears to be in addEditBookmark). Here's a pic:
screen shot 2016-10-13 at 10 56 20 pm

bsclifton:
@jkup notes left :smile: hopefully, I didn't do something wrong when testing (I grabbed latest from your branch `bookmark-popup`).  Please let me know and I can retest 

@bradleyrichter when you click the star when it's already bookmarked, the title still says "Bookmark added". If the site is bookmarked already and you click that, should we even show a title at all?

jkup:
Wanted to run these fixes by you:

@bsclifton I don't think this technically needs to be a dialog anymore right? I think it can just be a div with some similar behaviors. I just removed the dialog class which fixes that shadow issue but I may be overlooking something.

@bbondy the positioning bug that @bsclifton found is because we are pushing an empty browserButton which has a width and a height but no content. Can I just return null here? https://github.com/brave/browser-laptop/pull/4726/files#diff-b9cf7ede0a812496eace3679483c2c40R129

bbondy:
yes null is fine

bradleyrichter:
wow. a breath of fresh UI air!

some initial problems I found:

1 - needs to Close on click-outside.
2 - pressing the star while open also closes (same as click-outside)
3 - with door hanger closed, clicking a marked star opens the hanger with "edit bookmark" as the title (not "bookmark added")
4 - ideally, the URL is stored in whatever state the URL is in when the hanger closes, not when it opens. This allows the user to actually edit the URL in the URL box to modify the bookmark info.

For styling issues, we can work on it in person.

bsclifton:
Looking really good! :smile: All the same things Brad said above... plus the empty button issue you already mentioned (on about pages, including new tab)

@bradleyrichter the "View all bookmarks" seems a little hard to read, gray on gray.

bradleyrichter:
no worries. We'll shine er' up on Thursday...

(spec image:)

image

bsclifton:
I think this should also fix:
https://github.com/brave/browser-laptop/issues/2347

bsclifton:
Tested on Mac and also Windows 10. Looks and works good- great job :smile:

PR #4752, opened 2016-10-13
Long bookmark folder names should be truncated

### LINKED ISSUE #4739: Long bookmark folder names should be truncated
Yes

I realize that @srirambv's comment in https://github.com/brave/browser-laptop/pull/4690 is contrary to my opinion, but the downside of having folder names which are not truncated is that the bookmark names become oddly truncated then eventually are stacked below the folders.

**Expected behavior:**
It would seem that proportional columns make the most sense in this case if we must maintain columns. @bradleyrichter it may be worth revisiting this interaction and allow for a different method of selecting folders and shuffling bookmarks between them.

Tested on Windows 7 and OS X.

- Brave Version:
0.12.5 Preview 2

- Steps to reproduce:
    1. Create a bookmark folder with an obnoxiously long name.
    2. Import a lot of bookmarks.

- Screenshot if needed:
screen shot 2016-10-13 at 7 33 54 pm
screen shot 2016-10-13 at 6 58 30 pm

- Any related issues:

jkup:

Fix #4739

Auditors: @bradleyrichter @alexwykoff 

Test Plan:

We can't just add max-width: 50% because then it'll only take up half the screen when the columns float underneath each other but wrapping it in a breakpoint works.

Add a bookmark folder with a really long name and make sure it looks good with both a huge and small screen.
screen shot 2016-10-13 at 2 44 34 pm
screen shot 2016-10-13 at 2 44 44 pm

### DIFF
--- a/less/about/bookmarks.less
+++ b/less/about/bookmarks.less
@@ -103,6 +103,12 @@
       }
     }
   }
+
+  @media (min-width: @breakpointNarrowViewport) {
+    .folderView {
+      max-width: 50%;
+    }
+  }
 }
 
 .bookmarkFolderList list >* {


alexwykoff:
lgtm, noticed a performance of tracking objects while resizing:
bookmark_squeeze

i'll open a separate issue for that

Issue #4751, opened 2016-10-13
Bookmark folder disappears after drag and drop in about:bookmarks


bsclifton:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
I was drag and dropping in the new about:bookmarks screen and one of my folder disappeared. No matter what I did, I could not find the folder. The bookmarks are still there; you can search and find them.  I opened the session file up and saw the the folder in question had it's parentFolderId set to the same number as the folderId (which should be impossible)

JSON from the session:
```javascript
{
  "lastAccessedTime": 1475044276843,
  "tags": [
    "bookmark-folder"
  ],
  "folderId": 1,
  "customTitle": "favorites",
  "parentFolderId": 1
},
```

**Expected behavior:**
You should never be able to move a folder onto itself

- Platform (Win7, 8, 10? macOS? Linux distro?):
Mac, but @alexwykoff had it occur on Windows

- Brave Version:
HEAD of master

- Steps to reproduce:
  I cannot reliably reproduce, but @alexwykoff had encountered the issue too

- Screenshot if needed:
![windows_folder_drag_poof](https://cloud.githubusercontent.com/assets/4733304/19367730/38337bc6-9151-11e6-93fe-6a7bdeb9cb4d.gif)

- Any related issues:

ajohnsen:
I just had this same bug happen to me in the bookmark toolbar.
I checked in the session-store1 file and when i look at my missing folder i can see it, but its referencing itself as the parentId. 

Could it be possible that this bug is back or maybe not fixed in the bookmark toolbar?

Is there any way to get the bookmark folder back? I tried editing the json file, but that didn't change anything.

Issue #4685, opened 2016-10-11
Moving a subfolder into Other Bookmarks does not work


luixxiul:
**Describe the issue you encountered:** Moving a subfolder into Other Bookmarks does not work properly.

**Expected behavior:**

- Platform (Win7, 8, 10? macOS? Linux distro?): OS X

- Brave Version: https://github.com/brave/browser-laptop/tree/43c0100c617b289da846d218a231d7a837f35aa7

- Steps to reproduce:
    1. Open about:bookmarks
    2. Add a temp folder (called "temp")
    3. Drag and drop the folder into "Other Bookmarks"

- Screenshot if needed:

<img width="780" alt="screenshot 2016-10-12 5 14 55" src="https://cloud.githubusercontent.com/assets/3362943/19286980/dab54e46-903a-11e6-9fc6-25e4b7d945fc.png">

You should be able to see "temp" folder under "Other Bookmarks". It looks like the folder is gone.

- Any related issues: #4684

PR #4756, opened 2016-10-13
Bookmark fixes for 0.12.5

bsclifton:

Auditors: @jkup @darkdh 

## Other Bookmarks
Other Bookmarks in about:bookmarks now displays sub-folders properly

Fixes https://github.com/brave/browser-laptop/issues/4685

Test Plan:
1. Open Brave and go to about:bookmarks
2. Create a bookmark if you don't already have one
3. Drag and drop this bookmark into the "Other Bookmarks" folder
4. Confirm when "Other Bookmarks" is clicked that it shows in the right pane
5. Create a new bookmark folder and put this bookmark into it
6. Drag and drop this new bookmark folder into "Other Bookmarks"
7. Confirm the UI shows a subfolder under "Other Bookmarks"
8. Click the subfolder and confirm bookmark shows in right pane

## Fix disappearing bookmark folders
Bookmark folders can no longer be moved into themselves (which causes them to disappear).
(This was a long-standing bug and this commit adds a test to ensure it's not possible)

Fixes https://github.com/brave/browser-laptop/issues/4751

Test Plan:
This was hard to reproduce... I would suggest trying to drag and drop a lot in the UI. @alexwykoff had captured the steps as dragging a bookmark folder into the right pane.  The bookmark folders should **never** disappear

### DIFF
--- a/js/about/bookmarks.js
+++ b/js/about/bookmarks.js
@@ -37,14 +37,17 @@ class BookmarkFolderItem extends ImmutableComponent {
 
   onDrop (e) {
     const bookmark = dndData.getDragData(e.dataTransfer, dragTypes.BOOKMARK)
+
     if (bookmark) {
+      // Don't allow a bookmark folder to be moved into itself
+      if (bookmark.get('folderId') === this.props.bookmarkFolder.get('folderId')) {
+        return
+      }
       aboutActions.moveSite(bookmark.toJS(), this.props.bookmarkFolder.toJS(), dndData.shouldPrependVerticalItem(e.target, e.clientY), true)
     }
   }
   render () {
-    const childBookmarkFolders = this.props.bookmarkFolder.get('folderId') === -1
-      ? []
-      : this.props.allBookmarkFolders
+    const childBookmarkFolders = this.props.allBookmarkFolders
           .filter((bookmarkFolder) => (bookmarkFolder.get('parentFolderId') || 0) === this.props.bookmarkFolder.get('folderId'))
     return <div>
       <div role='listitem'
@@ -106,7 +109,9 @@ class BookmarkFolderList extends ImmutableComponent {
       }
       {
         this.props.bookmarkFolders.map((bookmarkFolder) =>
-          <BookmarkFolderItem bookmarkFolder={bookmarkFolder}
+          this.props.isRoot && bookmarkFolder.get('parentFolderId') === -1
+          ? null
+          : <BookmarkFolderItem bookmarkFolder={bookmarkFolder}
             allBookmarkFolders={this.props.allBookmarkFolders}
             selected={!this.props.search && this.props.selectedFolderId === bookmarkFolder.get('folderId')}
             selectedFolderId={this.props.selectedFolderId}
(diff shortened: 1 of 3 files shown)


js/about/bookmarks.js
```diff
@@ -106,7 +109,9 @@ class BookmarkFolderList extends ImmutableComponent {
       }
       {
         this.props.bookmarkFolders.map((bookmarkFolder) =>
-          <BookmarkFolderItem bookmarkFolder={bookmarkFolder}
+          this.props.isRoot && bookmarkFolder.get('parentFolderId') === -1
+          ? null
+          : <BookmarkFolderItem bookmarkFolder={bookmarkFolder}
             allBookmarkFolders={this.props.allBookmarkFolders}
```
bsclifton: This check is so that folders that are children of "Other Bookmarks" don't show under "Bookmarks Toolbar".  Since the above check (line 50) was removed, it will now bind ALL bookmarks to this control.

js/about/bookmarks.js
```diff
@@ -37,14 +37,17 @@ class BookmarkFolderItem extends ImmutableComponent {
...
+        return
+      }
       aboutActions.moveSite(bookmark.toJS(), this.props.bookmarkFolder.toJS(), dndData.shouldPrependVerticalItem(e.target, e.clientY), true)
     }
   }
   render () {
-    const childBookmarkFolders = this.props.bookmarkFolder.get('folderId') === -1
-      ? []
-      : this.props.allBookmarkFolders
+    const childBookmarkFolders = this.props.allBookmarkFolders
```
jkup: Can this still return -1? If so won't a .filter() throw?

jkup:
Code LGTM! I posted that one concern but as long as that's not a problem ++ from me!

js/about/bookmarks.js
```diff
@@ -37,14 +37,17 @@ class BookmarkFolderItem extends ImmutableComponent {
 
   onDrop (e) {
     const bookmark = dndData.getDragData(e.dataTransfer, dragTypes.BOOKMARK)
+
     if (bookmark) {
+      // Don't allow a bookmark folder to be moved into itself
+      if (bookmark.get('folderId') === this.props.bookmarkFolder.get('folderId')) {
```
bbondy: we have another check somewhere that checks all decedents but I don't remember where off hand.   Is this a strong enough check? I.e. could someone do something like this?
```
A
--> B
     --- C
```
And then move A into C to create a circular ref?

bbondy:
merging, thanks

js/about/bookmarks.js
darkdh: I use -1 for "Other Bookmarks" in importer too.

darkdh:
++

js/about/bookmarks.js
bsclifton: we have something like this in the moveSite code; I'll make sure that code is rock solid by working on some unit tests :smile:

luixxiul:
follow-up: #4788 #4804

PR #4819, opened 2016-10-15
Disable View page source on chrome-extension pages

### LINKED ISSUE #4812: "View Page Source" on about:history opens nothing
**Describe the issue you encountered:** If you click "view page source" of the common context menu on about:history, nothing is shown

clipboard01

**Expected behavior:** it should be either removed or grayed out
- Brave Version: 0.12.5 RC1
- Any related issues: #4811

Sh1d0w:

Fixes #4812

Auditors:
@bbondy


luixxiul:
Rather I'm thinking that the common context menu itself on about pages should be disabled.

Sh1d0w:
@luixxiul Ok I am closing the PR. I read in the original issue you have opened you wanted the element eighter hidden or grayed out ... :)

luixxiul:
yes sorry @Sh1d0w still you don't have to close the issue as it was not decided to disable it. Also in this case other elements should be grayed out too like 1Password, reload page, etc. Find, print, bookmark page would be useful here too.

Sh1d0w:
@luixxiul I have removed Reload, Back, Forward, View Source and 1Password entries if you are on extension page. Can you please confirm the behaviour is correct like this? Thanks

bbondy:
We should only disable the things which do not work, can you confirm if that's the case in the PR?
Do any of these work?
> Also in this case other elements should be grayed out too like 1Password, reload page, etc, while Find, print, bookmark page would be useful here.

Sh1d0w:
@luixxiul Can you confirm the items I have removed are ones that do not work?

luixxiul:
These do not work
- 1Password
- Reload page

These work
- Back
- Forward
- Find 
- Print
- Bookmark page
- Inspect Element

Sh1d0w:
@bbondy The PR is updated. Let me know if I have to amend anything else.

bbondy:
What doesn't work about extension pages and reload? I'm wondering because our pdf viewing is done without reloading.  Are you sure it doesn't work?

bbondy:
It seems like it should

bsclifton:
Closing this PR in favor of https://github.com/brave/browser-laptop/pull/5559 which disables context menu items ONLY on the about pages.

Apologies that we didn't communicate more about this one :frowning_face: It probably should have only solved the about pages, unless there is an existing issue for the other work

PR #4820, opened 2016-10-15
When Modal Overlays are layered, only bottom layer has a dark background (fix #4808)

### LINKED ISSUE #4808: background color of modal window on ledger advanced settings gets darker by clicking backup wallet button
**Describe the issue you encountered:** the background color of modal window on the ledger advanced settings gets darker by clicking the backup wallet button.

**Expected behavior:** it should be the same.
- Brave Version: 0.12.5 RC1
- Steps to reproduce:
  1. Go to about:preferences#advanced
  2. Click Advanced Settings
  3. Click "backup your wallet"
- Any related issues:

willy-b:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist. (fixes #4808)

Test Plan:
1. Open Brave, navigate to Preferences->Payments tab
2. Enable Payments
3. Click "Advanced Settings"
4. Observe background (page outside modal) is darkened.
5. Click "Backup Wallet"
6. Observe background does **not** get darker.

### DIFF
--- a/js/components/modalOverlay.js
+++ b/js/components/modalOverlay.js
@@ -9,7 +9,39 @@ const ImmutableComponent = require('./immutableComponent')
  * Represents a modal overlay
  */
 
+var globalInstanceCounter = 0
+var mountedInstances = []
+
 class ModalOverlay extends ImmutableComponent {
+
+  componentWillMount () {
+    this.instanceId = globalInstanceCounter++
+
+    this.setState({last: true})
+
+    if (mountedInstances.length) {
+      let lastModal = mountedInstances[mountedInstances.length - 1]
+      lastModal.setState({last: false})
+      lastModal.forceUpdate()
+    }
+
+    mountedInstances.push(this)
+  }
+
+  componentWillUnmount () {
+    let instId = this.instanceId
+
+    mountedInstances = mountedInstances.filter(function (inst) {
+      return inst.instanceId !== instId
+    })
+
+    if (mountedInstances.length) {
+      let lastModal = mountedInstances[mountedInstances.length - 1]
+      lastModal.setState({last: true})
+      lastModal.forceUpdate()
+    }
+  }
+
   get dialogContent () {
     var close = null
     var button = null
@@ -35,7 +67,7 @@ class ModalOverlay extends ImmutableComponent {
   }
 
   render () {
-    return <div className='modal fade' role='alert'>
+    return <div className={'modal fade' + (this.state.last ? ' last' : '') + (this.props.transparentBackground ? ' transparentBackground' : '')} role='alert'>
       {this.dialogContent}
     </div>
   }
--- a/less/modalOverlay.less
+++ b/less/modalOverlay.less
@@ -4,19 +4,31 @@
 
 @import "variables.less";
 
+.modal.last + .modal.last {
+  background: transparent;
+}
+
 .modal {
   opacity: 1;
   overflow: auto;
   position: fixed;
-  background: @black25;
+  background: transparent;
   width: 100vw;
   height: 100vh;
   left: 0;
   top: 0;
   z-index: 8999;
 
-  &.fade {
-    transition: @transitionFast;
+  +.modal {
+    background: transparent;
+  }
+
+  &.last {
+    background: @black25;
+
+    &.transparentBackground {
+      background: transparent;
+    }
   }
 
   &.hidden {
@@ -25,8 +37,6 @@
   }
 
   .modal {
-    background: transparent;
-
     .dialog {
       left: 0;
       top: 0;


bbondy:
Could you provide before / after screenshots?

willy-b:
@bbondy, yep!

Before:
brave_issue_4808_before

After:
brave_issue_4808_after

luixxiul:
Viewing the code I think setting background:none to ``.modal + .modal`` is simpler.

willy-b:
@luixxiul agree. I just tested this and replaced my commit in this PR with your solution by force-push.

bsclifton:
One slight problem with this PR (sorry it hadn't been looked at before :frowning_face:)

If the top level modal is smaller than the previous one / ones, you can see that they aren't being grayed out.
screen shot 2016-11-26 at 12 27 44 am

In this case, I had to hack the DOM to get this showing... but if the child window is smaller, this would be a problem. I simulated this by cracking open the dev tools and removing the main content div from the "recover your wallet" page

willy-b:
good catch. looking into this now

willy-b:
hey @bsclifton, after trying a bunch of CSS-only solutions, I fixed this by adding some logic to keep track of all the ModalOverlay instances and ensure only the top-most one has a `last` class which gives it a dark background. If you can recommend a CSS-only way, please do.

It's a little trickier than it looks (`:last-child` , `:last-of-type` won't work here since modal/non-modal divs are mixed together as siblings). Still seems like there should be a simple opposite of `.modal + .modal { background: transparent }` (which successfully ensures only the first modal has a dark background) that would work for ensuring the last modal would have a dark background here, but I couldn't find it.

luixxiul:
If `div[class^="modal"]:last-of-type` could work, that would be the best solution, but it doesn't.

bsclifton:
@willy-b it does look good for `Advanced settings`... but unfortunately, if you go to the `Add funds` screen and then hit either the QR code or the add with debit/credit (assuming those show up for you), they'll either have a missing overlay or a super dark overlay :frowning_face: 

screen shot 2016-11-28 at 4 13 46 pm

screen shot 2016-11-28 at 4 14 02 pm

willy-b:
@bsclifton: there is now only one modal background in view for all cases.

this fixes the QR popup case, but the background for the Coinbase iframe modal is still a bit darker because the background is actually from Coinbase's site which uses a darker gray. not sure what the best approach there is.

bsclifton:
@willy-b sorry it's taken me a few days to get back to you; I tried this out...\
- overlay looks great between pages :smile: 
- unfortunately, there is a flicker when going from add funds to the buy widget. Can you update this to keep the overlay (which would make it even darker?)  That would prevent this flash

Regarding the iframe styles, the coloring is in the body tag for the buy widget.  You might be able to select it with document.querySelectAll() and then update the object returned's style attribute (removing background-color or replacing with transparent).

willy-b:
Hey @bsclifton, I made the change you asked for, but I just noticed @luixxiul has some changes on the way that may obsolete this (see https://github.com/brave/browser-laptop/pull/6009#issuecomment-264687241).

@luixxiul you're fixing #4808 in PR #5009 then?

willy-b:
Didn't mean to step on your toes @luixxiul, didn't see your changes. (It says they came in 11 days ago on commit but the commit reference on GitHub only popped up about 25 minutes ago -- fresh force push probably)

luixxiul:
@willy-b No, I cannot find a solution by myself. Please go ahead :-)

willy-b:
Sorry. I'm confused. Is this issue supposed to be fixed in commit https://github.com/luixxiul/browser-laptop/commit/2327e9eca53eb4a637dd4c3e3a2a1d1f613c2fbc#commitcomment-19937858?

willy-b:
This looks fixed in your screenshots, let's close this issue.

willy-b:
OK, I just tested your commit @luixxiul  https://github.com/luixxiul/browser-laptop/commit/2327e9eca53eb4a637dd4c3e3a2a1d1f613c2fbc#commitcomment-19937858 , and it looks perfect for the recovery dialog. The QR code popup still has a double overlay, which is not present in this PR, but I'm sure you an fix that too. The results are otherwise identical and your solution is MUCH cleaner (all CSS/LESS). You guys (@bsclifton, @luixxiul) make the call. Thanks!

bsclifton:
@willy-b how about let's just keep this on the back-burner until @luixxiul's PR is accepted and then we'll revisit?

luixxiul:
@willy-b I have been trying CSS solution too but couldn't find the solution. I thnk it doesn't solve the issue found by @bsclifton here above: https://github.com/brave/browser-laptop/pull/4820#issuecomment-263049318

bsclifton:
OK awesome; @luixxiul's PR has been accepted.  @willy-b, can you rebase?  

I think you're safe to own this change (the darkening of the modal issue) since you had already started this with a PR. Your last commit that I tested was working great, but we just needed to tune the behavior with regards to the buy widget... choosing either to:
- Keep the overlay and let it darken extra dark
- keeping the overlay but then putting a style over the iframe body style, making background transparent (for the buy widget)

willy-b:
Hey @bsclifton, thanks again for reviewing. I pushed a commit implementing the first option (keeping the overlay for the Coinbase iframe) 4 days ago and force-pushed a rebase of that here just now.

(in case anyone wants to check something, I kept the old/pre-rebase branch at https://github.com/willy-b/browser-laptop/tree/issue-4808-old)

bsclifton:
Great job! This looks and feels great :smile:

cezaraugusto:
++!! looks awesome, great work!

luixxiul:
@willy-b Hi Willy, is it possible to add `.last` _unless_ the modal contains `.coinbaseOverlay`? The iframe inside `.coinbaseOverlay` itself has the dark background.

bsclifton:
@luixxiul would you be able to open a separate issue for that? I think it's possible, but would be good to track separately :smile:

PR #4870, opened 2016-10-17
Fix alignment of home button between tab with content and about:newtab

bsclifton:

Fixes https://github.com/brave/browser-laptop/issues/4854

Auditors: @jkup @bbondy 

@jkup this is a proposed fix which has the same symptom as the favorites button you had adjusted with https://github.com/brave/browser-laptop/pull/4726.  This PR should fix it too- basically, it's putting a placeholder button in place (which has opacity 0) which causes the flexbox to adjust it's bounds properly. I'm curious if there's a better solution, but I would need to dig in more (I don't have enough CSS knowledge at the moment to know an alternative)

Test plan:
1. Launch Brave and open https://brave.com and also a new tab
2. Open preferences, enable home button
3. switch between brave.com and the new tab screen. home icon should not move

### DIFF
--- a/js/components/navigationBar.js
+++ b/js/components/navigationBar.js
@@ -115,7 +115,7 @@ class NavigationBar extends ImmutableComponent {
       <div className='startButtons'>
         {
           isSourceAboutUrl(this.props.location) || this.titleMode
-          ? <span className='browserButton' />
+          ? <span className='browserButton hidden-button fa fa-repeat' />
           : this.loading
             ? <Button iconClass='fa-times'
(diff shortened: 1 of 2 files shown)


jkup:
This is interesting, I just asked a question about that line over here https://github.com/brave/browser-laptop/pull/4726#issuecomment-254360593

Do we even need to render an empty browserButton?

bsclifton:
Closing in favor of solution provided w/ https://github.com/brave/browser-laptop/pull/4726

PR #4872, opened 2016-10-17
Add whether brave is default browser check

### LINKED ISSUE #2105: General Settings: "Brave is not your default browser" setting
Functionality:
Button: 
For each specific OS, the button may require a different action/script.

Mac OS - Utilize available API to toggle OS level setting, and show the standard dialog.
Dialog sample:
image

Windows: Either toggle the default browser setting and show the Windows standard confirmation dialog, or open the windows CP if the API is not available. (Win 7 vs 8 vs 10 for example)

Linux: Same as windows.

Switch:
"Always check on startup" is on by default, but may be switched off by the user.

Item label:
"Brave is not your default browser"

Button label:
"Set as default..."

Switch label:
"Always check on startup"

Settings window:
image

Startup Sheet Dialog: (slides down from below tabs toolbar)
image

And setting toggle state: (when Brave is your default browser)

The button hides and the "startup check" switch moves up.
image

darkdh:

fix #2105
requires https://github.com/brave/electron/pull/73

Auditors: @bridiver, @bbondy


app/index.js
```diff
@@ -421,7 +423,6 @@ app.on('ready', () => {
     basicAuth.init()
     contentSettings.init()
     privacy.init()
-    Autofill.init()
```
bbondy: Was Autofill.init() removed here intentionally?

darkdh: Ah, no it must be a rebase mistake

app/renderer/components/checkDefaultBrowserDialog.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+  onNotNow () {
+    appActions.defaultBrowserUpdated(false)
+    appActions.defaultBrowserCheckComplete()
+    appActions.changeSetting(settings.CHECK_DEFAULT_ON_STARTUP, this.props.checkDefaultOnStartup)
+  }
+  onUseBrave () {
+    appActions.defaultBrowserUpdated(true)
+    appActions.defaultBrowserCheckComplete()
+    appActions.changeSetting(settings.CHECK_DEFAULT_ON_STARTUP, this.props.checkDefaultOnStartup)
```
bbondy: I wonder if all these 3 things could be managed by a single appAction instead. I won't block on it but I think that's preferred for future ref.

bbondy:
looks great overall, I'll post followups for discovered problems.

PR #4880, opened 2016-10-18
Make bookmarks fill toolbar container for any window width. fix #4116

### LINKED ISSUE #4116: bookmarks toolbar still not fully filling up on some resolutions
I think it depends on the DPI setting, my wife's laptop doesn't fill up the whole toolbar though.

#9939

cezaraugusto:
Auditors: @bbondy 

Fix: #4116

Test Plan:
- Open Brave with Bookmarks toolbar enabled;
- Resize the window horizontally
- Check that our » (_view more_ bookmarks button) don't disappear;
- Check that bookmarks are added on blank spaces, as window gets wider;
- Check that there's shouldn't have any UI breaks/unaligned items;
- They should work flawlessly with our three options for bookmarks:
  - No favicons
  - With favicons
  - Only favicons

### DIFF
--- a/app/renderer/components/bookmarksToolbar.js
+++ b/app/renderer/components/bookmarksToolbar.js
@@ -297,7 +297,7 @@ class BookmarksToolbar extends ImmutableComponent {
     const noParentItems = this.bookmarks
       .filter((bookmark) => !bookmark.get('parentFolderId'))
     let widthAccountedFor = 0
-    const overflowButtonWidth = 25
+    const overflowButtonWidth = 15
 
     // Dynamically calculate how many bookmark items should appear on the toolbar
     // before it is actually rendered.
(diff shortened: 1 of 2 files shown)


bbondy:
I was able to checkout this branch and I just tried to resize my window smaller slowly and the overflow button was being cut.  I can also do that in master, but this isn't a full fix so I'm going to move it to 0.12.7.

bbondy:
no overflow button:
screenshot 2016-10-18 09 51 14

bbondy:
btw also better if you put "Fix: #4116" after the first commit line because then github makes the link clickable. Thx :)

bbondy:
This is not working in show only favicon mode:
screenshot 2016-10-24 16 52 06

So I'm going to move this issue and PR to 0.12.8.

bbondy:
pushing to 0.12.9 so focus can be on new tab

bsclifton:
Pushing to 1.0, so we can focus on new tab :smile:

bsclifton:
Closing this for now... but the issue is still assigned :smile: @cezaraugusto if you revisit this, please open a new PR.  Thanks!

Issue #4936, opened 2016-10-19
Bookmarks bar select element should not be disabled


bbondy:
Currently the switch control for always show the bookmarks bar controls the bookmarks bar enabled state. But they should be independent settings. 

<img width="338" alt="screenshot 2016-10-19 00 50 14" src="https://cloud.githubusercontent.com/assets/831718/19506064/10cdeba4-9596-11e6-9d6f-3d9ec5a668d6.png">

PR #4944, opened 2016-10-19
Split apart "always show bookmarks toolbar" and the bookmarks toolbar display mode

bsclifton:

Split apart "always show bookmarks toolbar" and the bookmarks toolbar display mode

Fixes https://github.com/brave/browser-laptop/issues/4936

Auditors: @bbondy

Test Plan:
1. Launch Brave and open Preferences
2. Verify you can toggle "always show bookmarks toolbar" and not have the drop down above be affected

### DIFF
--- a/js/about/preferences.js
+++ b/js/about/preferences.js
@@ -616,7 +616,6 @@ class GeneralTab extends ImmutableComponent {
     const homepageValue = getSetting(settings.HOMEPAGE, this.props.settings)
     const homepage = homepageValue && homepageValue.trim()
     const disableShowHomeButton = !homepage || !homepage.length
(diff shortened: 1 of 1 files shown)


bbondy:
I think it might actually make the most sense to have 4 options:
Do not show bookmarks toolbar
Show bookmarks as text only
Show bookmarks as text and icons
Show bookmarks as icons only

/cc @bradleyrichter I think this is an improvement so I'll merge.  Mainly it's really confusing seeing a disabled item with an option of "Text only" under bookmarks toolbar. And an option to control it underneath it.

But please advise if you'd like a followup to combine into one drop down which I think is the most clean.

PR #4959, opened 2016-10-19
default browser check should run only on release build

### LINKED ISSUE #4958: default browser check should run only on release build
we should skip
NODE_ENV=test 
NODE_ENV=development

darkdh:

fix #4958

Auditors: @bridiver, @bbondy

Test Plan:
1. Make sure there is no session data
2. npm run start/test should not popup the check default browser modal
and overrides brave as default browser.
3. BRAVE_IS_DEFAULT_BROWSER=true npm run start/test should not popup the check default browser modal
and overrides brave as default browser.
4. BRAVE_IS_DEFAULT_BROWSER=false npm run start/test should popup the check default browser modal
and overrides brave not be default browser.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -444,9 +444,18 @@ app.on('ready', () => {
     }
     process.emit(messages.APP_INITIALIZED)
 
-    // Default browser checking
-    let isDefaultBrowser = defaultProtocols.every(p => app.isDefaultProtocolClient(p))
-    appActions.changeSetting(settings.IS_DEFAULT_BROWSER, isDefaultBrowser)
+    if (process.env.BRAVE_IS_DEFAULT_BROWSER !== undefined) {
+      if (process.env.BRAVE_IS_DEFAULT_BROWSER === 'true') {
+        appActions.changeSetting(settings.IS_DEFAULT_BROWSER, true)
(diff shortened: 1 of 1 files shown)


app/common/state/defaultBrowserState.js
```diff
@@ -8,5 +8,7 @@ const settings = require('../../../js/constants/settings')
 module.exports.shouldDisplayDialog = (state) => {
   return !getSetting(settings.IS_DEFAULT_BROWSER) &&
     !state.get('defaultBrowserCheckComplete') &&
-    getSetting(settings.CHECK_DEFAULT_ON_STARTUP)
+    getSetting(settings.CHECK_DEFAULT_ON_STARTUP) &&
+    process.env.NODE_ENV !== 'development' &&
+    process.env.NODE_ENV !== 'test'
```
bridiver: this looks good, but I think you probably want to add an option to override this check so we can test this without triggering it on every test. Maybe something like `process.env.BRAVE_FORCE_CHECK_DEFAULT_ON_STARTUP` which would run it regardless of the NODE_ENV or other settings

bsclifton: +1 on @bridiver's recommendation :smile:

PR #4988, opened 2016-10-20
Fix default setting behavior for getSetting()

bsclifton:
I accidently introduced the bug here (merged into 0.12.6):
https://github.com/brave/browser-laptop/commit/889757e1c94433594112efac5a4baf6f2f3026df#commitcomment-19471078

The fix by @darkdh is correct, except password manager defaults to a null value (not undefined)
https://github.com/brave/browser-laptop/commit/10927580f0865cab2a7a6b7620d52417b185a071

This fix checks against undefined and null

Auditors: @darkdh @bbondy

Test Plan: run unit tests (most PW manager ones were failing)


darkdh:
++

PR #5001, opened 2016-10-20
Add new tab page state

cezaraugusto:
Add state for newTab and update docs. Split of work to avoid conflicts.

Auditors: @bbondy


bsclifton:
LGTM :smile:

PR #4967, opened 2016-10-19
Updated pluralization in bravery panel (shields panel) to handle 0, 1, or more than 1

bsclifton:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Updated pluralization in bravery panel (shields panel) to handle 0, 1, or more than 1

Also-
Hardened tests quite a bit for the Bravery Panel. This includes
- added new helpers for:
  - waitForTabCount
  - openBraveMenu
  - waitForDataFile
- added waitForTabCount in places where race condition was present
- increase timeout for the waitForDataFile portion, which sometimes took longer than 1 second
- added a logVerbose method to many helper methods. Can be toggled by setting variable "logVerboseEnabled" to true in `test/lib/brave.js`

**NOTE**: _tests are not included for these strings. I can either add them or create an issue to track this_

Fixes https://github.com/brave/browser-laptop/issues/4966

Auditors: @bbondy

Test Plan:
- Launch Brave, go to https://clifton.io, and open bravery panel
- Notice that:
  - "Ad or Tracker blocked" is used instead of the plural (since only 1 is present)
  - items with 0 use the plural tense
- Visit https://brianbondy.com and notice:
  - "Ads or Trackers blocked" is used since there are 3 present

PR #5039, opened 2016-10-21
Test documentation update w/ notes about verbose mode :)

bsclifton:
Follow up from https://github.com/brave/browser-laptop/pull/4967/files#r84397349

Auditors: @bbondy


bbondy:
++

Issue #5091, opened 2016-10-24
Fix auto-suggest sort order with different protocol prefixes


aekeus:
* Auto-complete history was not correctly ordering items with a mix of http:// and https:// prefix
* Too few history items where being shown

- Steps to reproduce:
    1. Visit a mixture of sites that have a similar name (twitter.com, twilio.com) but have a different protocol prefix (http, https)
    2. Locations with different prefixes will be sorted together regardless of visit count or last access 

Related: #5067 

cc @bbondy @bradleyrichter @Sh1d0w

PR #5092, opened 2016-10-24
Handle different protocol prefixes in auto-suggest

aekeus:
- [X] Added/updated tests for this change (for new code or code which already has tests).

  * Remove http and https from URLs when sorting
  * Increase history limit to 3, suggested to 3
  * Update unit tests

Auditors: @bbondy

Test Plan:

  1. Visit a mixture of sites that have a similar name (twitter.com, twilio.com) but have a different protocol prefix (http, https)
  2. Ensure that history sorting is entirely based on count and frequency
  3. Ensure that a maximum of 3 history items and 3 suggested sites are shown (so as not to overflow the space available and cause a scrollbar to appear)

Fixes: #5091

### DIFF
--- a/app/renderer/lib/suggestion.js
+++ b/app/renderer/lib/suggestion.js
@@ -107,5 +107,8 @@ module.exports.simpleDomainNameValue = (site) => {
  *
  */
 module.exports.normalizeLocation = (location) => {
-  return location.replace(/www\./, '')
+  location = location.replace(/www\./, '')
+  location = location.replace(/^http:\/\//, '')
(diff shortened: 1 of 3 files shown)


app/renderer/lib/suggestion.js
```diff
@@ -107,5 +107,8 @@ module.exports.simpleDomainNameValue = (site) => {
  *
  */
 module.exports.normalizeLocation = (location) => {
-  return location.replace(/www\./, '')
```
bbondy: ++ for extra logic in url utils but I'll merge this as is for now

Issue #5139, opened 2016-10-26
Dropdown menu in economist.com stops working


Quinn1415:
**Did you search for similar issues before submitting this one?**

Yes

**Describe the issue you encountered:**

Click on one of the dropdown menu on http://www.economist.com/ no longer works. I think this also happens to the Google Chrome when touch is enabled on a touch screen, disabling it will make it work, except here with Brave, no way for me to set it, plus I am running it on a non-touch screen Windows 10 laptop

**Expected behavior:**

Dropdown menu should show up
- Platform (Win7, 8, 10? macOS? Linux distro?):

Windows 10 professional, current Anniversary Update
- Brave Version:

Latest dev version
- Steps to reproduce:
  1.
  2.
  3.
- Screenshot if needed:
- Any related issues:

jonathansampson:
The drop-down menu (sub-menu revealed while hovering navigational items) is presently broken in Brave, Chrome, and Firefox. It does, however, work in Microsoft Edge.

![economist](https://cloud.githubusercontent.com/assets/815158/19792119/4fdc50f0-9c7c-11e6-94bd-b4a2594ce9d7.gif)

The `:hover` state is defined in http://cdn.static-economist.com/sites/default/files/css/css_6e9392aabd3c4f727a145049659314f6.css.

jonathansampson:
[Outreach has been made](https://twitter.com/BraveSampson/status/791858470182019072) to a Platform Engineer at The Economist.

JonasEconomist:
@jonathansampson thanks for reaching out. I'm unable to replicate this. Could you provide the steps to recreate this issue? Then I'll pass it along to the appropriate channels.

Quinn1415:
On the Brave version 0.12.6, and a PC desktop (Windows 10) with no touch
screen, go to http://www.economist.com/, and hover the mouse over the
dropdown menus (World Politics, Business & Finance, etc.), you will see
submenu popped up if it works properly.

jonathansampson:
@JonasEconomist @Quinn1415 I repro in Blink and Gecko browsers.

![economist-comparison](https://cloud.githubusercontent.com/assets/815158/19818861/9af23d5e-9d06-11e6-85c2-278f1f5a89ed.gif)

Quinn1415:
Brave 0.12.7 works now.
Many thanks.

bbondy:
@jonathansampson can this be closed?

Quinn1415:
Yes, please.

[...]

PR #5140, opened 2016-10-26
Added info about transifex and translations/localizations

bsclifton:
Added info about transifex and translations/localizations
- new translations.md under docs/
- link is now in README.md

Auditors: @aekeus, @luixxiul

### DIFF
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -1,0 +1,62 @@
+# How to contribute translations
+
+There are several things you can do to help us internationalize Brave and provide a great experience for everybody
+
+## Providing a translation
+
+We manage all of our translations using Transifex. Here's how you can get started:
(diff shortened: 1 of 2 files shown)


bsclifton:
quick heads up - I had accidentally had assigned you both to this https://github.com/brave/browser-laptop/issues/5139; if you got an alert or email just ignore it :smile:

docs/translations.md
```diff
@@ -0,0 +1,54 @@
+# How to contribute translations
+
+There are several things you can do to help us internationalize Brave and provide a great experience for everybody
+
+## Providing a translation
```
aekeus: We generally pull in all languages files at the time we cut a release. That allows us to keep everything up to date in a scalable way.

luixxiul:
We should localize this info itself later ;-)

docs/translations.md
bsclifton: Added this info in, thanks @aekeus :smile:

Issue #4973, opened 2016-10-20
[HackerOne] issue #175927	


diracdeltas:
https://hackerone.com/reports/175927

diracdeltas:
https://jsfiddle.net/n1L29m8w/ test

diracdeltas:
waiting again to test this after the chromium v54 upgrade

diracdeltas:
https://github.com/brave/browser-laptop/pull/5211 mitigated this somewhat, so bumping the milestone

diracdeltas:
appears to be fixed in the chromium54 branch

diracdeltas:
QA steps:
1. go to http://web.mit.edu/zyan/Public/googlespoof.html
2. click the Google link
3. when the new window stops loading https://www.google.com, the page should be blank

PR #5211, opened 2016-10-28
only set frame title when navigation is not renderer initiated

diracdeltas:

Test Plan:
1. go to http://web.mit.edu/zyan/Public/googlespoof.html
2. click on the link
3. when the new window stops loading, the tab title should be the tab URL instead of 'fake'

partial fix for #4973. this prevents the page-that-is-being-spoofed from displaying the title from the spoofing page.

requires https://github.com/brave/electron/pull/83

auditors: @bbondy


bbondy:
looks great but waiting for electron 1.4.21 to change this.

diracdeltas:
merging now that electron has been updated on master

PR #5251, opened 2016-10-29
Removing clear cache and clear site data options from the history menu

### LINKED ISSUE #3093: Replace Clear data menu items in History menu with new, single item
new: 

image

old:

image

maaz93:

Fix #3093 
There were 3 options to clear history, namely Clear History, Clear Cache, Clear All Cookies and Site Data. These seemed redundant. Removing the 3 and keeping only one option called "Clear Browsing Data". ~~The particular string seems to already exist in history.properties file. Added the same entry with the key "clearBrowsingData" in menu.properties as well.~~

Test Plan:

1. Make sure only one option exists for clearing history, from history menu
 - Open Brave
 - Select the History menu option
 - See that only "Clear Browsing Data" option is available

~~2. Test option name in different supported languages~~
~~- Verify that the option name is correct in different languages~~

PS2: Removed the translation file changes
PS3: Changed "Clear browsing data..." to "Clear Browsing Data...", as requested in the issue

Auditors: @bsclifton


bsclifton:
Change looks good- the modification for other language files is not needed ([more info here](https://github.com/brave/browser-laptop/blob/master/docs/translations.md)) :smile:

bbondy:
Good to merge if you remove the translation changes other than en-US, thanks!
`git reset head^ --soft`
Then git add just the files you need and `git checkout -- filename` the ones you don't should do it.
Then you can just `git push -f` sorry if you already know all of this :)

maaz93:
I thought since the translation of the string is already available, and I only need to append `...`, why not do that as well. 😄  But as per the link provided by @bsclifton I can see that you have a process regarding translations and I guess that'll take care of this. Will update the PR.

@bbondy Thanks for the git commands! I was just googling around how to update a PR 😄

bsclifton:
Here's the before:
screen shot 2016-11-04 at 2 01 42 pm

This change consolidates it down to one single menu: 
"Clear Browsing Data…"

PR #5296, opened 2016-10-31
Limit 24 hour payment notification to once per reconcile period

### LINKED ISSUE #4481: Aggressive notification
The info bar “Hello!Good news! Brave will pay your favorite publisher sites in less than 24 hours.” is way too aggressive. There is no way to dismiss the toolbar (no X). If you give in and click the one and only weirdly labeled button, the toolbar is dismissed temporarily only to reappear a short while later.

I must have dismissed the bugger ten times today.

It also appears inside tiny pop-up windows (like social logins and the like). Only site-specific toolbars should ever appear inside these small windows opened with JavaScript (see behavior in every other browser).

Once I’ve opened `about:preferences#payments` from the toolbar, don’t show me the notification again until next month.

ayumi:
Fix #4481

Auditors: @bsclifton

Test Plan:

1. Trigger the "Payment in 24 hours, please review" notification.
  - Update reconcileStamp to <24 hours from now
  - Have sufficient funds; OR disable the sufficient funds conditional by adding false to ledger.js L1485 (showEnabledNotifications())
  - Change startup notification delay in ledger.js L484 from 15m to 5s
2. Open Brave and observe 24h review notification
3. Close Brave and reopen. Notification should not reappear.
(Next 24h notification timestamp is set in Application Support/brave-development/session-store-1)

### DIFF
--- a/docs/state.md
+++ b/docs/state.md
@@ -445,6 +445,7 @@ WindowStore
   ledgerInfo: {
     creating: boolean,          // wallet is being created
     created: boolean,           // wallet is created
+    reconcileFrequency: number, // duration between each reconciliation in days
     reconcileStamp: number,     // timestamp for the next reconcilation
     transactions: [ {           // contributions reconciling/reconciled
       viewingId: string,        // UUIDv4 for this contribution
--- a/js/constants/appConfig.js
+++ b/js/constants/appConfig.js
@@ -119,6 +119,9 @@ module.exports = {
     'bookmarks.toolbar.showOnlyFavicon': false,
     'payments.enabled': false,
     'payments.notifications': false,
+    // "Out of money, pls add" / "In 24h we'll pay publishers [Review]"
+    // After shown, set timestamp to next reconcile time - 1 day.
+    'payments.notification-reconcile-soon-timestamp': null,
     'payments.notificationTryPaymentsDismissed': false, // True if you dismiss the message or enable Payments
     'payments.contribution-amount': 5, // USD
     'privacy.autofill-enabled': true,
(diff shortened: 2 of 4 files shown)


docs/state.md
```diff
@@ -445,6 +445,7 @@ WindowStore
   ledgerInfo: {
     creating: boolean,          // wallet is being created
     created: boolean,           // wallet is created
+    reconcilePeriod: number,    // duration of each reconciliation period in days
```
bsclifton: Totally a nitpick- would it be better to say frequency? like the reconcileFrequency would be every 30 days.  When I see period or duration, it's not immediately clear

ayumi: i like it

js/constants/appConfig.js
```diff
@@ -119,6 +119,9 @@ module.exports = {
     'bookmarks.toolbar.showOnlyFavicon': false,
     'payments.enabled': false,
     'payments.notifications': false,
+    // "Out of money, pls add" / "In 24h we'll pay publishers [Review]"
+    // After shown, set timestamp to next reconcile time - 1 day.
+    'payments.notification-reconcile-soon-timestamp': null,
     'payments.notificationTryPaymentsDismissed': false, // True if you dismiss the message or enable Payments
```
bsclifton: This might be more of a @bradleyrichter question... but I wonder if this should this be covering both cases? Let's say you're within 24 hours of reconciling and you only have the ~5 dollars in your account. You'd dismiss the nag window but then you're out of budget.  This is a case where you might want to get nagged again sometime before the next reconciliation period (especially since getting funds in might take > 24 hours)

mrose17:
can i get @ayumi or @bsclifton to merge this...

bsclifton:
Will merge in a sec... just manually testing it now...

bsclifton:
Looks great! :+1:

Issue #5334, opened 2016-11-01
regression: opening a new tab should not show 'about:blank'


diracdeltas:
as of a45ca4fafcc8090132a7de5cdda2aa204a998c74, `about:blank` is displayed in the urlbar to prevent phishing. therefore it should not be used as the default URL, which should be about:newtab

bbondy:
Changing milestone to 0.12.9, I think this is landing with newtab.

PR #5335, opened 2016-11-01
Removed "blank" option (about:blank !== about:newtab)

bsclifton:
Fixes https://github.com/brave/browser-laptop/issues/5334

Definitely my mistake folks- I had forgotten about the URL bar getting focus and didn't consider that the change broke tests also :frowning_face: 

This change:
- finalizes the name of the setting (the previous one had "TEST" in it)
- defaults unknown URLs to about:newtab (which makes sense, since those have URL bar focus)
- removes blank as an option (since it was going to be removed when we shipped)

If we need to disable the NEW about:newtab screen for 0.12.8, we can just return an empty div in `js/about/newtab.js`

Auditors: @diracdeltas, @bbondy

Test Plan:
1. Launch Brave and open a new tab
2. New tab should be about:newtab
3. Preferences > General should no longer show "blank"
4. Tests should work again

### DIFF
--- a/app/common/constants/settingsEnums.js
+++ b/app/common/constants/settingsEnums.js
@@ -9,7 +9,6 @@ const startsWithOption = {
 }
 
 const newTabMode = {
-  BLANK: 'blank',
   NEW_TAB_PAGE: 'newTabPage',
   HOMEPAGE: 'homePage',
   DEFAULT_SEARCH_ENGINE: 'defaultSearchEngine'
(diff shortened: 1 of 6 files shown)


diracdeltas:
already created issue at https://github.com/brave/browser-laptop/issues/5334

bsclifton:
awesome, rebasing...

diracdeltas:
lgtm, thanks

PR #5342, opened 2016-11-02
Add WebTorrent support

dcposch:
See #3256 

This is still pretty raw. It demonstrates torrent support in Brave.

### Visit a magnet URL

screen shot 2016-11-01 at 8 03 05 pm

### Bookmark a magnet link

screen shot 2016-11-01 at 8 04 02 pm

### Play files in a torrent

brave-webtorrent-1

### Here are some of the things we still need to do:
- [x] Design nicer UX
- [x] Implement nicer UX
- [x] Probably move WebTorrent to its own process. Currently it's running in the Brave extension content script, with some of the CSP restrictions disabled.
- [x] Add tests
- [x] Security review (sort of---cc @diracdeltas )

Feedback appreciated!

@BrendanEich @diracdeltas @ayumi @bbondy


bbondy:
Could you add some sample torrent and magnet links that would demonstrate the added things nicely? Thanks!

dcposch:
@bbondy sure, Feross collected some here: https://codepen.io/ferossity/full/qaezaB/

Right now you have to paste magnet links into the URL bar. *Clicking* a magnet link tries to find an external app handler. I'll try to fix that later today.

dcposch:
@bbondy you can click on magnet links now

do you think this is a reasonable way to implement that?

https://github.com/brave/browser-laptop/pull/5342/commits/0834c988bea07dc2030945b4e73f6cb53e7557aa

app/extensions.js
```diff
@@ -137,6 +137,12 @@ let generateBraveManifest = () => {
     cspDirectives['style-src'] = '\'self\' \'unsafe-inline\' http://' + devServer
   }
 
+  // TODO:
+  // * Move WebTorrent to its own renderer process, similar to the way it's done in
+  //   WebTorrent Desktop
+  // * Remove this CSP exception:
+  cspDirectives['connect-src'] = '*'
```
dcposch: @bbondy do these CSP rules apply only to the braveExtension, or to *all* extensions?

diracdeltas: @dcposch only to the braveExtension, iirc. for security reasons, i would prefer that WebTorrent be its own "extension" similar to how we handle PDF reader (PDF.JS)

app/extensions/brave/js/webtorrent.js
```diff
@@ -0,0 +1,19 @@
+(function () {
+  var queryString = window.location.search
```
diracdeltas: looks like queryString is not used?

js/lib/appUrlUtil.js
```diff
@@ -124,6 +124,44 @@ module.exports.isTargetAboutUrl = function (input) {
...
+
+/**
+ * Gets the magnet:// URL for a given local URL.
+ * Returns null if the input is not the local URL for a magnet link.
+ * Example: getSourceMagnetUrl('chrome-extension://<...>.html#magnet:?x=y') -> 'magnet:?x=y'
+ */
+module.exports.getSourceMagnetUrl = function (input) {
+  const url = module.exports.getAppUrl('webtorrent.html')
+  if (!input.startsWith(url)) return null
+  return input.substring(input.indexOf('#') + 1)
```
diracdeltas: shouldn't this check that the return value starts with 'magnet:?'

diracdeltas:
how should this work for users who already have a `magnet:` URL protocol handler set up in Brave? ex: i usually set up put.io to handle magnet links

feross:
It's also worth noting that as currently architected in this PR, WebTorrent can only connect to WebRTC peers, and not normal TCP/uTP ones. This is because we're running in a `chrome-extension:///` context that has no node integration. So we're using the browser version of WebTorrent (`webtorrent.min.js`) at the moment.

feross:
@diracdeltas As currently implemented, we ignore any protocol handlers registered by webpages through [`navigator.registerProtocolHandler()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler)

feross:
Does Brave expose an interface for configuring protocol handlers yet? 

Actually, here's an idea. What if we just add a new link to the Torrent Viewer page that says "Open in <website name here>" for users who have a "magnet" protocol handler registered. Here's the current mock @dcposch is working on:

screen shot 2016-11-03 at 3 24 55 pm

So imagine another link added to that list. @diracdeltas Would that satisfy your use case?

js/lib/appUrlUtil.js
feross: The target URL is internal-only and controlled by us, so I think that we can assume it will always be a magnet link or torrent file. This seems to be the assumption that is made by the analogous `getSourceAboutUrl` function. Correct me if I'm wrong?

app/extensions/brave/js/webtorrent.js
feross: Will remove

diracdeltas:
> What if we just add a new link to the Torrent Viewer page that says "Open in " for users who have a "magnet" protocol handler registered. Here's the current mock @dcposch is working on:

:+1:

dcposch:
### New screenshots
The file list now consists of regular hyperlinks.  You can link to individual files within a torrent now, using the `ix` magnet link parameter.

Right now, following one of those links is slow, because the new page has to connect to the swarm from scratch. That will be fixed once we move WebTorrent to its own process.

screen shot 2016-11-03 at 7 50 44 pm


screen shot 2016-11-03 at 7 50 48 pm

app/extensions.js
dcposch: done!

dcposch:
WebTorrent now runs in the browser process.

It communicates with the Torrent Viewer extension pages over IPC, and serves content via a streaming torrent-to-HTTP server.

image

dcposch:
Improved UI/UX coming soon, @feross has a PR! https://github.com/dcposch/browser-laptop/pull/7

dcposch:
Almost ready for review!

### Work left to do
- Merge https://github.com/dcposch/browser-laptop/pull/7
- Fix refresh: hitting F5 should not try to open an external torrent app
- Auto stop: once the last tab for a given infohash is closed, remove the torrent after a reasonable amount of time
- Auto start: if a torrent is already active, go directly to the file list, don't show the Start button
- (Maybe) No <iframe>s. Change the UX to avoid framing file contents. See https://github.com/dcposch/browser-laptop/pull/7#issuecomment-260073109
- (Maybe) Better Incognito support
- Manual testing. Open 20+ tabs full of magnet links. Hold down the Refresh button. Open the same video across multiple windows. Nothing should crash, glitch, or get slow.

feross:
Woo! Getting close.

feross:
- [ ] Fix the way magnet links show up in the History tab.
- [ ] Support right-click > "Open in New Tab" in File List.

js/components/frame.js
```diff
@@ -877,10 +887,27 @@ class Frame extends ImmutableComponent {
           method = (currentDetail, originalDetail) =>
             windowActions.setAutofillCreditCardDetail(currentDetail, originalDetail)
           break
+        case messages.TORRENT_MESSAGE:
+          // Relay torrent IPC from the webview to a browser process
```
diracdeltas: I don't think this relaying is necessary; the webview can just call `ipc.send` to send a message to the browser process. See `messages.CERT_ERROR_ACCEPTED` for instance.

js/components/frame.js
```diff
@@ -877,10 +887,27 @@ class Frame extends ImmutableComponent {
...
+            this.torrentChannelID = message.clientKey
+            ipc.send(messages.TORRENT_MESSAGE, message)
+          }
+          break
       }
       method.apply(this, e.args)
     })
 
+    // Relay torrent IPC from the browser process back to the webview
+    ipc.on(messages.TORRENT_MESSAGE, (e, message) => {
```
diracdeltas: this  may not be necessary either; i think the browser process can just do `event.sender.send` after it gets a message from the webview in `app/browser/webtorrent.js`

js/webtorrent/components/mediaViewer.js
```diff
@@ -0,0 +1,23 @@
+const React = require('react')
+
+class MediaViewer extends React.Component {
+  render () {
+    const torrent = this.props.torrent
+    const ix = this.props.ix
+
+    let content
+    if (torrent.serverURL != null) {
```
diracdeltas: should be `!==` ?

js/webtorrent/components/mediaViewer.js
```diff
@@ -0,0 +1,23 @@
...
+      content = <iframe src={torrent.serverURL + '/' + ix} sandbox='allow-same-origin' />
```
diracdeltas: why is allow-same-origin needed?

js/webtorrent/components/mediaViewer.js
```diff
@@ -0,0 +1,23 @@
...
+    } else {
+      content = <div>Loading...</div>
```
diracdeltas: needs localization

app/extensions.js
```diff
@@ -137,15 +151,56 @@ let generateBraveManifest = () => {
...
+// The torrentExtension handles magnet: URLs
+// Analagous to the PDFJS extension, it shows a special UI for that type of resource
+let generateTorrentManifest = () => {
+  let cspDirectives = {
+    'default-src': '\'self\'',
+    'frame-src': 'http://localhost:*',
+    'form-action': '\'none\'',
+    'referrer': 'no-referrer',
+    'style-src': '\'self\' \'unsafe-inline\'',
+    'img-src': '* data: blob:'
```
diracdeltas: is the img-src directive  needed?

app/extensions.js
```diff
@@ -137,15 +151,56 @@ let generateBraveManifest = () => {
...
+  }
+
+  // TODO:
+  // * Move WebTorrent to its own process, similar to the way it's done in WebTorrent Desktop
+  // * Remove this CSP exception:
+  cspDirectives['connect-src'] = '*'
```
diracdeltas: is this directive needed?

js/webtorrent/components/mediaViewer.js
feross: We need the iframe to actually have an origin of `http://localhost:xxxx` so that it can make http requests to the server at that same location. Otherwise, the existence of `sandbox` makes the iframe have it's own, unique origin.

js/webtorrent/components/mediaViewer.js
feross: Good catch, will fix

app/extensions.js
feross: Probably not, I'll test it without that

app/extensions.js
feross: This was only needed when `webtorrent` was running in a `<script>` tag in the chrome extension, so it could connect to http web seeds, and websocket servers. Not needed anymore

js/webtorrent/entry.js
```diff
@@ -0,0 +1,141 @@
...
+        console.error('Ignoring unknown dispatch type: ' + JSON.stringify(action))
+    }
+  }
+
+  render () {
+    const {torrent, torrentID, errorMessage, parsedTorrent} = state
+    const ix = parsedTorrent && parsedTorrent.ix // Selected file index
+    let name = parsedTorrent && parsedTorrent.name
+    if (!name) {
+      name = state.torrent
```
feross: Oh, nevermind. You mean the next lines. Gotcha

diracdeltas: yup

js/webtorrent/components/mediaViewer.js
diracdeltas: makes sense, thanks

js/components/frame.js
dcposch: hmm. that doesn't seem to work

dcposch: nvm, figured it out!

js/components/frame.js
dcposch: fixed

js/webtorrent/components/mediaViewer.js
dcposch: fixed

js/webtorrent/entry.js
dcposch: fixed

app/extensions.js
dcposch: removed

app/extensions.js
dcposch: removed

diracdeltas:
I got a crash when open magnet:?xt=urn:ed2k:31D6CFE0D16AE931B73C59D7E0C089C0&xl=0&dn=zero_len.fil&xt=urn:bitprint:3I42H3S6NNFQ2MSVX7XZKYAYSCX5QBYJ.LWPNACQDBZRYXW3VHJVCJ64QBZNGHOHHHZWCLNQ&xt=urn:md5:D41D8CD98F00B204E9800998ECF8427E and then click 'Start Download'.

error in terminal:
```

Error: Missing torrentKey: f1a477fb8020c49698c0b162cbb5ae56
    at getTorrentByKey (/Users/yan/repos/browser-laptop/node_modules/webtorrent-remote/server.js:269:23)
    at handleCreateServer (/Users/yan/repos/browser-laptop/node_modules/webtorrent-remote/server.js:127:19)
    at WebTorrentRemoteServer.receive (/Users/yan/repos/browser-laptop/node_modules/webtorrent-remote/server.js:48:16)
    at EventEmitter.<anonymous> (/Users/yan/repos/browser-laptop/app/browser/webtorrent.js:24:12)
    at emitTwo (events.js:106:13)
    at EventEmitter.emit (events.js:191:7)
    at WebContents.<anonymous> (/Users/yan/repos/browser-laptop/node_modules/electron-prebuilt/dist/Brave.app/Contents/Resources/electron.asar/browser/api/web-contents.js:210:13)
    at emitTwo (events.js:106:13)
    at WebContents.emit (events.js:191:7)
process exited with code 7
```

diracdeltas:
I notice that the `Do you want to allow X to handle external protocols` notification bar appears for a brief moment before WebTorrent redirects to the viewer. You can prevent this by editing `registerPermissionHandler` such that if the permission is `openExternal` and the URL scheme is `magnet:`, then `cb(false)` immediately instead of showing the message box.

diracdeltas:
it would be nice to make this feature optional, mainly for users who have set up protocol handlers already for magnet: files (ex: i use put.io). see app/ledger.js `messages.CHECK_BITCOIN_HANDLER` for an example of checking protocol handlers.

idea:
1. create a WEBTORRENT_ENABLED pref in js/constants/settings.js. set it to false for users who have a magnet protocol handler already, default to true otherwise.
2. the first time a magnet URL is clicked, if WEBTORRENT_ENABLED is false, show a prompt that says 'Would you like to use WebTorrent to handle magnet links?', which sets the pref. the pref can later be changed in about:preferences.

dcposch:
@diracdeltas looks like you have an old version of `webtorrent-remote`. try running `npm install` or `npm update`

dcposch:
@diracdeltas fixed the crash you described

it was a race condition in `webtorrent-remote`. that's a module i made over the last week for this project. glad i extracted it, otherwise the diff in this PR would be huge

docs: https://github.com/dcposch/webtorrent-remote

published to npm here: https://www.npmjs.com/package/webtorrent-remote

dcposch:
the magnet link you linked to still doesn't work, because it doesn't have an infohash, but it no longer causes a crash.

@feross the error ("Malformed .... missing infohash") is now being returned by `webtorrent-remote` correctly and ends up in `state.errorMessage`, but we aren't actually rendering that. want to update the UI to show the error message?

dcposch:
@diracdeltas good call on `registerPermissionHandler`! i think that also fixes a bug that @feross @bbondy and i were trying to figure out where, *some* of the time (but not always), following a magnet link will try to open it in an external application in addition to showing the Torrent Viewer page

i'll do that and make Torrent Viewer optional tomorrow.

question: do you want me to add a pref (as you described), or should this just work by enabling/disabling the Torrent Viewer extension? what happens when you disable the PDFJS extension and navigate to a PDF?

Liunkae:
With PDFJS disabled, Brave asks to download the PDF file.

diracdeltas:
> question: do you want me to add a pref (as you described), or should this just work by enabling/disabling the Torrent Viewer extension? what happens when you disable the PDFJS extension and navigate to a PDF?

not sure i understand the difference between adding a pref and disabling the extension. when you disable PDFJS, it downloads the file. i think when you disable webtorrent, it should do what it does now (open it in the registered protocol handler or ask if the site can open an external application).

diracdeltas:
> the magnet link you linked to still doesn't work, because it doesn't have an infohash, but it no longer causes a crash.

verified that it's no longer crashing

dcposch:
@diracdeltas i'm changing `registerPermissionHandler`/`openExternal` now. here's my plan:
* when the Torrent Viewer extension is enabled, there's no permission prompt and magnet links just open there
* when the Torrent Viewer is disabled, the normal permission prompt happens and they can open magnet links in an external app

does that look reasonable?

diracdeltas:
sgtm

dcposch:
@diracdeltas our PDFJS support has bugs :/

1. Open the Bitcoin paper. Works, but shows the wrong URL in the address bar:
screen shot 2016-11-17 at 2 40 02 pm

2. Disable PDFJS and restart

screen shot 2016-11-17 at 2 39 51 pm

3. The tab with the Bitcoin paper fails to load, shows a white screen.
screen shot 2016-11-17 at 2 39 31 pm

dcposch:
> I notice that the Do you want to allow X to handle external protocols notification bar appears for a brief moment before WebTorrent redirects to the viewer. It also reappears when you navigate back to the opener page. You can prevent this by editing registerPermissionHandler such that if the permission is openExternal and the URL scheme is magnet:, then cb(false) immediately instead of showing the message box

Dunno if that's possible. In `registerPermissionHandler`, you unfortunately only get the URL of the calling page, not the URL that it's trying to open.

However, I fixed an old TODO in `filtering.js` where it used to allow *all* extensions the `openExternal` permission automatically. It no longer does that, so now you can say Deny when it tries to open in an external app it will stop opening magnet links externally.

> it would be nice to make this feature optional

done!

https://github.com/brave/browser-laptop/pull/5342/commits/be12e123678544aa917d8bd53f5a3194b4baa335

dcposch:
### Torrent Viewer is now optional

You can turn it on or off in Prefs > Advanced

screen shot 2016-11-17 at 5 14 44 pm

In that case, the extension is disabled:
screen shot 2016-11-17 at 5 16 05 pm

Then, clicking magnet links goes back to its current behavior. It prompts to open an external app:
screen shot 2016-11-17 at 5 16 16 pm

dcposch:
@feross, you made two new checkboxes
* Magnet links don't show up in History at all. I think that's OK, though: the PDF Viewer works the same way. PDFs themselves don't go into History.
* I don't think we need special handling for .mkv files. When you click a .mkv link from a normal HTTP page, it goes into Downloads. I think it's fine for .mkvs in torrents to work the same way.

dcposch:
@diracdeltas what kind of security review do we need? thanks in advance for you help!

the last checkbox left after that is "add tests"... what do you think we need there?
it's interesting and a bit surprising that the CI seems to be down for weeks now. should i add unit or integration tests?

cc @bbondy

diracdeltas:
@dcposch nice catch on the PDFJS bug! i fixed at least the first part of it here: https://github.com/brave/browser-laptop/commit/552af86bb5cb0709585ae1a19398ae2c488028a9

dcposch:
added some unit tests: https://github.com/brave/browser-laptop/pull/5342/commits/894d08249df7852abcc84580f66923ef4a7d7070

dcposch:
@diracdeltas ^ lgtm

diracdeltas:
lgtm though it looks like you need to add the logo image to web_accessible_resources
screen shot 2016-11-17 at 8 35 35 pm

bbondy:
Took a look and it's looking great. I have some questions in terms of feature-set.
- Could we have  a fullscreen button on the video element?
- Is it still the plan to have a bookmark button for individual files in the torrent?
- Is it still the plan to have a download button for individual files in the torrent?

I think it'd be cool to do an initial release for 0.12.11 which freezes on end of day Monday and then do another one with follow ups.  Thoughts?

srirambv:
Would be nice to show progress (% done of all active downloads) on the tab when it is not the active tab
Opening a new magnet link in a new tab shows only that file in the list. It doesn't show other active/seeding/downloading files

dcposch:
> lgtm though it looks like you need to add the logo image to web_accessible_resources

@diracdeltas done

dcposch:
> Could we have a fullscreen button on the video element?

almost done. will add a commit for this in a few mins

> Is it still the plan to have a bookmark button for individual files in the torrent?

already possible. navigate to an individual file, then click the star button in the address bar to bookmark

> Is it still the plan to have a download button for individual files in the torrent?

coming soon, prob later today

> I think it'd be cool to do an initial release for 0.12.11 which freezes on end of day Monday and then do another one with follow ups. Thoughts?

sweet!

dcposch:
## I think we're ready

Please try it out & let me know if there's anything that needs improvement!

Code review also appreciated.

@bbondy @diracdeltas @bradleyrichter @feross @BrendanEich 

#### In this PR

* Support for magnet links. View a torrent, stream individual files, bookmark, save `.torrent`s, download files.

#### Not in this PR

Here are some additional things you could do, probably after Brave 1.0

* Support HTTP torrent URLs. Currently, navigating to a `.torrent` file will just download it. We could treat those the same way we treat magnet links. There are some UX challenges to consider. For example: how do you link to a single file within an HTTP `.torrent` URL? It's not as straightforward as appending `&ix=<n>` to a magnet URL. You might use a hash, eg `http[s]://<...>.torrent#ix=<n>`

* History. Currently, `magnet:` links don't appear in the History tab. Looks like this is the same behavior as PDFs in the PDF Viewer.

* Support torrents in the mobile versions of Brave.


## Latest screenshots

#### Regular webpage with magnet links

Navigate [https://codepen.io/ferossity/full/qaezaB/](here). Click one of the links.

screen shot 2016-11-18 at 5 14 22 pm

#### Magnet page

Brave doesn't connect to any peers or to the DHT until you give it permission. A short legal warning lets users know they're about to join a P2P network.

screen shot 2016-11-17 at 11 14 52 am

Click start.

#### File list

Right click on one of the named links, then click Open in New Tab.

screen shot 2016-11-18 at 5 17 05 pm

screen shot 2016-11-18 at 4 42 48 pm

#### Magnet page for an individual file

This is a magnet link with the same infohash as the one we were just on, but with an `ix` attribute specifying a single file inside the torrent.

Since that torrent is already active, there's not disclaimer and no start button, the media starts streaming immediately. WebTorrent downloads pieces from peers in the order that we need them so we can start playing right away.

screen shot 2016-11-18 at 6 27 58 pm

screen shot 2016-11-17 at 11 15 08 am

#### Download links

Back in the first tab, click Download. This creates a normal download. Under the hood, it uses the same streaming torrent-to-HTTP local server as the <video> tag in our second tab.

You can also click Download Torrent File to save a `.torrent`, eg for use with an external app.

screen shot 2016-11-18 at 4 42 41 pm

screen shot 2016-11-18 at 6 28 18 pm

#### Bookmarks

You can bookmark a whole torrent or an individual file.

#### Preferences

You can disable the Torrent Viewer. In that case, magnet files will prompt the user to open an external app.

screen shot 2016-11-18 at 6 30 40 pm

LMK if this works for you!

bbondy:
Rebased to get rid of the recent conflict and merged to master
Great work @dcposch @feross @diracdeltas ! 👍  💯 
eb004bd573c8644836c315493a3a9467f2d68f5c...05e1231287916329f33282bd5a43643245477b12

Issue #5357, opened 2016-11-02
delete folder cause stack overflow


darkdh:
version: master 9566bcd1e7c418e14cee1150917378ff4862623d

STR: 
1. Make sure you have clean session
2. new folder “123” under toolbar
3. new folder “456” under “123”
4. delete folder “123"

there will be a infinite recursive call in removeSite due to parentFolderId is its folderId

bsclifton:
this shouldn't be present in 0.12.8, since the about:newtab commits were not included

Root cause was found by @darkdh to be here:
https://github.com/brave/browser-laptop/commit/dbe041d0e7670aa5a563cb33e38c7d376279c6b4#diff-23ca389e2bcb77191b5a9c10900eb3a3R462

Fix in progress...

PR #5361, opened 2016-11-03
Created session helper for newTabState

bsclifton:

Created session helper for newTabState

- site will only be added if it's a bookmark or history item
- site will only be removed if it's a bookmark or history item
- site moves are now ignored
- favicon is now updated for the newTabState (which is a different list than AppState.sites)

Fixes https://github.com/brave/browser-laptop/issues/5357
Fixes https://github.com/brave/browser-laptop/issues/5310

Auditors:
@bridiver - check out the new session helper; hope I did this right :)

@darkdh - this fixes the folder delete issue by ignoring folders

@cezaraugusto - move is no longer tracked (which is fine; we don't care about the order of
the items with regard to their position in bookmarks toolbar, which is what moveSite is for).
Also, see the new session helper (which we can use for ordering by # visits)

Test Plan:
1. Launch Brave and visit brave.com
2. Open a new tab (about:newtab) and confirm you see brave.com there (favicon should show also)
3. Visit about:history and delete the visit you just made to brave.com
4. Confirm it's now gone in about:newtab

### DIFF
--- a/app/common/state/aboutNewTabState.js
+++ b/app/common/state/aboutNewTabState.js
@@ -18,8 +18,9 @@ const removeDuplicateSites = (sites) => {
   })
 }
 
-const newTabState = {
+const aboutNewTabState = {
   mergeDetails: (state, props) => {
+    state = makeImmutable(state)
     if (!props) {
       return state
     }
@@ -29,6 +30,7 @@ const newTabState = {
   },
 
   addSite: (state, props) => {
+    state = makeImmutable(state)
     if (!props) {
       return state
     }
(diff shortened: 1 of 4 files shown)


darkdh:
lgtm

cezaraugusto:
awesome! lgtm :ok_hand:

bridiver:
nice cleanup! Please add some tests and what do you think about naming it `aboutNewTabState`? When I first looked at it I thought it was code that applied to any new tab and took me a minute to realize that it was specifically for the `about:newtab` page.

bsclifton:
@bridiver added test cases :smile: Took a while to get it all sorted, but really glad I added them

While writing the tests, I realized that the remove wasn't working and ended up updating the code to only add the site info, (not the tags).  If you look at siteUtil.removeSite, it never actually deletes the entry (unless the tags are empty). By excluding the tags, the remove will always work (instead of keeping the entry for history)

bridiver:
++

PR #5421, opened 2016-11-05
refined string text for multiple items including change of Bravery to Shields

bradleyrichter:
cc @bbondy


bsclifton:
There was conflicts; I rebased and resolved them. So changes look like:
- remove global site settings from Bravery menu
- move import/check for update down a few positions in mac Brave menu
- Changing text to say "Shield settings" rather than "Bravery settings"
- Add "(will break many sites)" warning to no script

Changes LGTM :smile: Assigning to @bbondy per Brad's comment above

bradleyrichter:
thanks @bsclifton

js/contextMenus.js
```diff
@@ -683,7 +683,7 @@ function hamburgerTemplateInit (location, e) {
     }, {
       label: locale.translation('bravery'),
       submenu: [
-        CommonMenu.braveryGlobalMenuItem(),
+//        CommonMenu.braveryGlobalMenuItem(),
```
bbondy: nit: Comment should be aligned with the lines on the same indent level
nit2: You don't ever really need to comment out code since all code is recoverable with git history.

bbondy:
++ merging.

bradleyrichter:
Noted. Ty.

PR #5461, opened 2016-11-07
added release notes for 0.12.8

alexwykoff:
Auditors: @bbondy @luixxiul


bbondy:
nice! One tweak is just saying about:styles is where we will have the style guide in the future, didn't notice anything else though.  Just because it's a placeholder only for now.  Or you can just remove that one and announce it later.  Merging though.

PR #5469, opened 2016-11-07
"Add funds" notification: Reduce nag level and add button to "Turn off notifications"

### LINKED ISSUE #5299: In ledger notifications offer option to "Don't notify me anymore"
Yes

We have a preference for Ledger notifications, but it's unclear that exists when you actually view a notification.
For example, when getting the "Add more funds.." the default behavior is to wait a while then show the message again.

**Expected behavior:**
On ledger notifications, offer a clear way to disable ledger notifications (dismiss notifications forever).

- Brave Version:
0.12.7

- Any related issues:
#4481 / Aggressive notifications

@bradleyrichter @bsclifton

### LINKED ISSUE #5418: Add funds bar comes up too often and should have turn off button
When a user has 0 funds and Brave payments turned on, which will be a lot of people that just wanted to try the experience, shouldn't be prompted so often to add funds, and they should have an option to turn it off. 

I think I had about 20 prompts for it yesterday and have had 2 already today and have only been on my computer for an hour. 

I think just disabling brave payments would stop them, so maybe just adding a Turn Off button in the notification, and also make sure it doesn't come up more than once a day if they hit Later.

ayumi:
Fix #5418
Fix #5299

Auditors: @bsclifton

Test Plan / Quieter notification:
1. Trigger the "Add funds" notification.
  a. Update reconcileStamp to <24 hours from now (or a few days in the past)
  b. Ensure not enough balance
2. Change startup notification delay in ledger.js L484 from 15m to 5s
3. Open Brave and observe notification. Click Later.
4. Close Brave and reopen. Notification should not reappear.
(Next Add funds notification timestamp is set in Application Support/brave-development/session-store-1 `notification-add-funds-timestamp`)

Test Plan / Turn off button:
5. Trigger the Add funds notification again by closing Brave and editing session-store-1 `notification-add-funds-timestamp` to the past.
6. Open Brave, and when the notification appears choose Turn off.
7. Open Preferences > Payments and click Advanced Settings... and observe notifications are disabled.

Image / Before
screen shot 2016-11-07 at 12 22 25

Image / After
screen shot 2016-11-08 at 14 30 38

Image 2 / After
screen shot 2016-11-08 at 14 19 39

cc @bbondy @bradleyrichter

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -1494,29 +1494,43 @@ const showDisabledNotifications = () => {
 */
 const showEnabledNotifications = () => {
   const reconcileStamp = ledgerInfo.reconcileStamp
-  const balance = Number(ledgerInfo.balance || 0)
-  const unconfirmed = Number(ledgerInfo.unconfirmed || 0)
-
   if (reconcileStamp && reconcileStamp - underscore.now() < msecs.day) {
-    if (ledgerInfo.btc &&
-        balance + unconfirmed < 0.9 * Number(ledgerInfo.btc)) {
-      addFundsMessage = addFundsMessage || locale.translation('addFundsNotification')
-      appActions.showMessageBox({
-        greeting: locale.translation('updateHello'),
-        message: addFundsMessage,
-        buttons: [
-          {text: locale.translation('updateLater')},
-          {text: locale.translation('addFunds'), className: 'primary'}
-        ],
-        options: {
-          style: 'greetingStyle',
-          persist: false
-        }
-      })
+    if (shouldShowNotificationAddFunds()) {
+      showNotificationAddFunds()
     } else if (shouldShowNotificationReviewPublishers()) {
       showNotificationReviewPublishers()
     }
   }
+}
+
+const shouldShowNotificationAddFunds = () => {
+  const balance = Number(ledgerInfo.balance || 0)
+  const unconfirmed = Number(ledgerInfo.unconfirmed || 0)
+  const nextTime = getSetting(settings.PAYMENTS_NOTIFICATION_ADD_FUNDS_TIMESTAMP)
+  if (nextTime && (nextTime > underscore.now())) {
+    return false
+  }
+  return ledgerInfo.btc &&
+         (balance + unconfirmed < 0.9 * Number(ledgerInfo.btc))
+}
+
+const showNotificationAddFunds = () => {
+  const nextTime = underscore.now() + 3 * msecs.day
+  appActions.changeSetting(settings.PAYMENTS_NOTIFICATION_ADD_FUNDS_TIMESTAMP, nextTime)
+
+  addFundsMessage = addFundsMessage || locale.translation('addFundsNotification')
+  appActions.showMessageBox({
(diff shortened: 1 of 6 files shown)


bradleyrichter:
I am worried that this new button will be the first choice for most people and the result is a drop off in re-ups for month 2.


If we don’t want to do the work based on my suggestion, then let’s at least order the buttons so the CTA remains at the right side.


[TON]  [Later] [Add Funds]

ayumi:
Is it our convention to have primary buttons at the rightmost side? I put it on the left because I'd assumed it was reading order / left to right.

RE Drop off: I think nagging people every month currently doesn't lead to much re-upping, so removing notifications would lose just a small number of users. Also, nagging annoys users.

I feel auto monthly reload would fix these problems.

bbondy:
Agree if someone wants to turn off notifications, then let them.  They can go in preferences if they want it.   We don't want to nag people in this way.  Not every hour, not every 6 hours, not every day, not every month.

bbondy:
I think the convention is that default button is rightmost and I think we should be consistent with it.

bbondy:
Some examples of button order. 

screenshot 2016-11-08 08 40 24

screenshot 2016-11-08 08 38 31

screenshot 2016-11-08 08 38 27



screenshot 2016-11-08 08 40 14

bradleyrichter:
Nag them we should not. Not in a boat. Not with a goat.

And yes, button order should be flipped. 

I wish the notify-off button text was shorter but I don't have a better solution at the moment.

bsclifton:
Heads up (coming soon by @jkup !)  We will be updating the styles throughout the app to be consistent :smile: 

EX: these flat buttons
screen shot 2016-11-08 at 4 43 49 pm

Versus the fancy :tophat: 
screen shot 2016-11-08 at 4 45 05 pm

app/ledger.js
```diff
@@ -320,12 +320,13 @@ if (ipc) {
     const win = electron.BrowserWindow.getFocusedWindow()
     if (message === addFundsMessage) {
       appActions.hideMessageBox(message)
+      // See showNotificationAddFunds() for buttons.
+      // buttonIndex === 1 is "Later"; the timestamp until which to delay is set
+      // in showNotificationAddFunds() when triggering this notification.
       if (buttonIndex === 0) {
```
bsclifton: These indexes (indices?) would be great as constants, IMO. Instead of checking against 0, it could look like:
`if (buttonIndex === ledgerButton.Later) {`

ayumi: I like your comment, i think it'd make the code easier to follow.. maybe button indexes should be replaced by button values or something .. like if buttonValue === ledgerButton.later

The index shouldn't matter, just the button action. This would also handle cases where button sets don't contain Later or Turn off because they're not applicable.

bsclifton:
Dismissing (or choosing to turn off notifications) sets the timestamp as 3 days in the future.  So I can re-enable if I change my mind afterwards and I'll be prompted again 3 days from when I chose hide (effectively same as dismiss) 

Tested locally and it works great; merging :smile:

PR #5480, opened 2016-11-08
Make tab close button respect tab's container

### LINKED ISSUE #5431: Disable close button on tabs if tab size is too small
## Test plan

Under "QA steps" on https://github.com/brave/browser-laptop/pull/7799#issue-215652883

----

**Describe the issue you encountered:** if you click the close button of tab whose size is too small, sometimes you click the close button accidently.

**Expected behavior:** Disable close button on tabs if tab size is too small.

- Brave Version: 0.12.8 RC1 

- Steps to reproduce:
    1. Open many tabs
    2. Minimize the width of the window
    3. Select any tab carefully not to close it...

- Screenshot if needed:

clipboard01

- Any related issues:

cezaraugusto:

Fix #5431

Auditors: @bbondy, @bsclifton

Note:

This change fixed the issue, but affected tab icons on small windows (now being overflowed in a different way). Doesn't look weird imo but can be improved per team's request.

Test Plan:

* Reduce window's width checking for close button on tabs
* Close button should not overflow tab container on narrow views, avoiding UI mismatch

### DIFF
--- a/less/tabs.less
+++ b/less/tabs.less
@@ -42,8 +42,8 @@
 }
 
 .tab {
-  // background: rgba(222, 222, 222, 0.6);
-    background: linear-gradient(to bottom, rgba(255, 255, 255, 0.4), rgba(255, 255, 255, 0.1));
+  position: relative;
+  background: linear-gradient(to bottom, rgba(255, 255, 255, 0.4), rgba(255, 255, 255, 0.1));
   border-radius: @borderRadiusTabs @borderRadiusTabs 0px 0px;
   border-width: 1px 1px 0;
   box-sizing: border-box;
@@ -157,13 +157,24 @@
   }
 
   .closeTab {
+    transition: opacity 100ms linear;
     opacity: 0;
     text-align: center;
     width: 16px;
     height: 16px;
     margin-left: 4px;
     margin-right: 4px;
-
+    @media (max-width: @breakpointTabsTextTrim) {
+      position: absolute;
+      width: 100%;
+      margin: 0;
+      background: #fff;
+      border-radius: 0;
+      height: 100%;
+      display: flex;
+      justify-content: center;
+      align-items: center;
+    }
     &:hover {
       opacity: 1;
     }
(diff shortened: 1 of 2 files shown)


bbondy:
A bit worried this will break some tests at small window sizes

bbondy:
which maybe travis has?

bbondy:
hard to tell because our tests are so broken right now.

cezaraugusto:
better to just make "close" button to overflow tab's favicon instead of hiding it? We can avoid risks of breaking tests this way, if preferable.

bsclifton:
I think a better way to solve this would be to consider the size of the tab itself.

`.tabArea:not(.partOfFullPage)` could use a min width, so that it never gets smaller than the favicon (28px seems to work nicely)

You can then solve the actual issue w/ `.closeTabButton` by setting `overflow: hidden`. It'll auto-hide when the parent isn't big enough to accommodate the button (however, it'll need some playing around, as sometimes the button gets cut off- might need a margin-right)

Since I don't think this should be rushed and it's not critical for next release, I'm going to assign it to 0.12.10

cezaraugusto:
PR updated

bbondy:
Seems like an improvement but it doesn't get rid of the (x) and it makes the favicons get cut off when smaller.  @bradleyrichter is this the way you want it?
screenshot 2016-11-15 17 09 55

bradleyrichter:
@bbondy no, they should actually end up looking like fat pinned tabs, and eventually the same, but of course in the main tab area.

cezaraugusto:
PR updated

We now have close button centered filling tab's width on small (< 770px) screens:

close_btn

bradleyrichter:
Much better…

Is it possible to have the close button appear on the right?

cezaraugusto:
@bradleyrichter you mean something like this?

new_tab_close

If not, could you do a quick draft about how do you want it?

cezaraugusto:
Closing in favor of #5845

PR #5483, opened 2016-11-08
Remove row control for topSites on new tab page

### LINKED ISSUE #5482: Make new tab page to have only one topSite row with for first release
For new tab page's first release we should have only one row of topSites with no option to switch rows until we have better previews.

cezaraugusto:

Test Plan:

fix #5482

Auditors: @bbondy

Test Plan:

* there should be no row control for topSites


bbondy:
I was thinking we would just set display: none for now but it's ok since we can always get the code back.  Merging.

PR #5498, opened 2016-11-09
Don't show HTTP icon while site is loading (since it may be HTTPS)

bsclifton:

Don't show HTTP icon while site is loading (since it may be HTTPS)

Fixes https://github.com/brave/browser-laptop/issues/5490

Includes breaking the urlbar icon into it's own control (along with click/drag events)

Auditors: @diracdeltas, @jkup, @bbondy

Test Plan:
1. Launch Brave and open a new tab
2. Type in "https://twitter.com" and get really close to your screen
3. Hit enter and notice you do not see the yellow triangle at any time

### DIFF
--- a/app/renderer/components/urlBarIcon.js
+++ b/app/renderer/components/urlBarIcon.js
@@ -1,0 +1,104 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const React = require('react')
+const ImmutableComponent = require('../../../js/components/immutableComponent')
+const windowActions = require('../../../js/actions/windowActions')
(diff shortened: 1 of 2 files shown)


diracdeltas:
i still see the insecure icon flash briefly. possibly before `this.props.loading` is set to true.

how about, if the URL can be neither confirmed insecure nor secure based on the protocol, show no icon at all?

bsclifton:
@diracdeltas if you go to `http://twitter.com` for example, you will see the insecure icon flash quickly because it does a redirect.  We could work-around this by putting a rolling timer (ex: debounce) around the value (and storing in component state) to avoid this

diracdeltas:
@bsclifton i went to the https url directly but it still showed the insecure icon

diracdeltas:
I'm going to merge this for now but keep the issue open and create another PR on top of it using https://github.com/brave/electron/pull/90

PR #5500, opened 2016-11-09
[WIP] Move WebTorrent to its own process

dcposch:
For now, that's the main browser process. In the future, it'll be a separate utility process. The important parts:
* It's a single process with a single instance of WebTorrent, not one per tab
* It has access to the Node APIs, which allow WebTorrent to talk to the millions of standard BitTorrent clients out there

### Work left to do in this PR
- **Pull the WebTorrent remote API into its own module.** I'll make a module called `webtorrent-remote` soon. This means we can re-use work we already did in WebTorrent Desktop and get a nice clean separation of concerns.
- **Security review.**

### Future work, won't be in this PR
- **Support WebRTC in addition to standard BitTorrent.** Currently, the only good, non-flaky WebRTC implementation we have access to is Chromium's WebRTC stack, which is only available in renderer processes. Conversely, the Node APIs are only available in the browser process. (This is due to Brave's security features, different from vanilla Electron.) Ideally, WebTorrent would run in a utility process similar to a vanilla Electron renderer process: with access to both the Node APIs and WebRTC. For now, we support standard BitTorrent only, no WebRTC.

@bbondy  @feross @diracdeltas @BrendanEich


dcposch:
Actually, this should probably be a PR on top of this one: https://github.com/brave/browser-laptop/pull/5342

I'll close it here

dcposch:
I'll replace this PR with this one, which has a cleaner diff

https://github.com/dcposch/browser-laptop/pull/4

PR #5502, opened 2016-11-09
Setting location for a frame will now set loading=true

bsclifton:
Follow up for https://github.com/brave/browser-laptop/pull/5498

## Problem
Being on a new tab and then loading a URL would return loading: false, even after the location is updated.

## Resolution
setting location for a frame will now set loading=true

Is this OK? Usually, it gets set to true in WindowStore:: WINDOW_WEBVIEW_LOAD_START
The only places I see this used are:
- navigationBar (which binds to urlBar, which then binds to urlBarIcon (where the problem is))
- tabs (used to show the loading spinner)

Auditors: @diracdeltas, @bridiver

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -266,7 +266,8 @@ const doAction = (action) => {
             currentLocation.substring(0, 6).toLowerCase() !== 'about:') {
           windowState = windowState.mergeIn(frameStatePath(action.key), {
             activeShortcut: 'load-non-navigatable-url',
-            activeShortcutDetails: action.location
+            activeShortcutDetails: action.location,
+            loading: true
           })
         }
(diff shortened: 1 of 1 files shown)


bridiver:
this seems like it would potentially trigger the loading indication when a load hasn't really started. I think the loading state should only be triggered by the loading events or we will end up with odd behavior and load indicators that never stop

bsclifton:
@diracdeltas I'm going to go ahead and close this, per @bridiver's comment. I'm definitely scared of a regression and am not in a hurry to fix this. I'll keep an eye out for your electron change :smile:

bsclifton:
and here it is: :smile: 
https://github.com/brave/electron/commit/ca0aefe11f827d2b110561d5343e768804beac4d

diracdeltas:
sgtm

Issue #5529, opened 2016-11-10
Regression : autocomplete is deleting the first character typed in the url bar


alexwykoff:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When changing tabs and wanting to visit github.com, I clicked in the url bar and found the first character was deleted.

**Expected behavior:**
Whatever I type in the url bar should be respected.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Tested on OS X

- Brave Version:
0.12.9 RC1

- Steps to reproduce:
    1. Give focus to tab A
    2. Click on tab B
    3. Click on the url bar in tab B
    4. Quickly type a url like github.com

- Screenshot if needed:

- Any related issues:

diracdeltas:
this may be because the URLbar is getting selected after the first character is typed, so whatever is typed afterward overrides the initial character. just a thought.

bbondy:
I think this might be fixed by:
https://github.com/brave/browser-laptop/commit/fcde1241a8ae0201511921100da42f04448d6254

Please test and re-open if not.

PR #5573, opened 2016-11-12
fix syncing between react and DOM urlbar focus/select

diracdeltas:

previously it seems that the urlbar was getting selected
every time the react select state was updated; yet then the
react selected state would be updated to false.

this may fix https://github.com/brave/browser-laptop/issues/5529

Auditors: @bbondy

Test Plan:
follow Alex's test steps in https://github.com/brave/browser-laptop/issues/5529
and check that the first character is not deleted.

### DIFF
--- a/js/components/urlBar.js
+++ b/js/components/urlBar.js
@@ -69,8 +69,8 @@ class UrlBar extends ImmutableComponent {
     return this.props.urlbar.get('focused')
   }
 
-  updateDOMInputFocus (focused) {
-    if (this.urlInput && focused) {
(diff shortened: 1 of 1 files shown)


bbondy:
I think this is maybe the same thing that this fixed:
https://github.com/brave/browser-laptop/commit/fcde1241a8ae0201511921100da42f04448d6254
Thoughts?

bbondy:
I'll close becauseI think that's the case that it's fixed by that one, but please re-open if you don't think so, or if you think this should be merged too (it is bitrot though by mine)

PR #5608, opened 2016-11-14
Retrieve Transifex language resources using lower case

aekeus:
Test Plan:

`npm run download-languages` should run without displaying warning for all resources including those with mixed capitalization.


bbondy:
awesome idea, thanks.

Issue #5503, opened 2016-11-09
Account balance error cannot be fixed after disabling and enabling Internet connection


luixxiul:
**Describe the issue you encountered:** Account balance error cannot be fixed after disabling and enabling the Internet connection. You have to restart the browser to get the error fixed.

**Expected behavior:** The Internet connection should be detected automatically, and the error should be fixed as soon as possible.

- Platform (Win7, 8, 10? macOS? Linux distro?): macOS

- Brave Version: master branch

- Steps to reproduce:
    1. Close the browser
    2. Disable your Internet connection
    3. Open the browser
    4. Make sure the account balance error happens on ``about:preferences#payments``
    5. Enable your Internet connection
    6. The error on ``about:preferences#payments`` should be fixed as soon as possible

- Screenshot if needed:

- Any related issues:

@ayumi @mrose17

PR #5649, opened 2016-11-15
small ledger fixes

mrose17:
Auditor: @bbondy 
Test Plan:
1. Close the browser
2. Disable your Internet connection
3. Open the browser
4. Go to about:preferences#payments and see this:

Enable your Internet connection and a few seconds later see this:

Fixes https://github.com/brave/browser-laptop/issues/5606

Fixes https://github.com/brave/browser-laptop/issues/5503


bbondy:
@mrose17 just an FYI for the future, only en-US should be modified. The other files are auto generated and pulled in periodically, so any changes to them will be overwritten.

PR #5692, opened 2016-11-16
Fix about:brave; version information is always set (regardless of errors)

bsclifton:

Version information is always set (regardless of errors). 

Code itself is now fault tolerant as well. Includes tests.

Fixes: https://github.com/brave/browser-laptop/issues/5633

Auditors: @jonathansampson, @bbondy

Manual test plan:
1. Move your session out of the way (or do a fresh install)
2. Launch Brave
3. Go to about:brave
4. Notice version shows as expected

Automated test plan:
run `npm run test -- --grep="about:brave tests"`

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -349,6 +349,34 @@ module.exports.cleanSessionDataOnShutdown = () => {
 }
 
 /**
+ * version information (shown on about:brave)
+ */
+const setVersionInformation = (data) => {
+  try {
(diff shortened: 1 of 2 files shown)


jonathansampson:
LGTM

bsclifton:
BTW- meant to document this; The reason the version wasn't showing:
When the app loads for the first time, the session does not exist. When it tries to load, it fails because it's trying to process undefined.

PR #5729, opened 2016-11-18
Fix newtab's test "lets you pin a tile"

cezaraugusto:

Auditors: @bbondy

Test Plan:

* Test `lets you pin a tile (and shows the pinned icon afterwards)` should pass

### DIFF
--- a/test/about/newTabTest.js
+++ b/test/about/newTabTest.js
@@ -129,7 +129,7 @@ describe('about:newtab tests', function () {
         .waitForVisible('.topSitesElementFavicon')
         .moveToObject('.topSitesElement')
         .waitForVisible('.topSitesActionContainer')
-        .click('.topSitesActionBtn')
+        .click('.topSitesActionBtn:first-child')
         .moveToObject('.timeSaved')
         .waitForVisible('.pinnedTopSite')
     })


bsclifton:
This increases the timeout for this visible check to 3 seconds... Should we consider bumping up the default timeout? ([currently 1 second](https://github.com/brave/browser-laptop/blob/master/test/lib/brave.js#L566))

bbondy:
I haven't actually seen this failing again, but I think it was intermittent. @cezaraugusto did you reproduce it failing or was this a guess at a fix?

bbondy:
I bet lowering the waitForInterval would help speed up tests a lot too

cezaraugusto:
@bbondy I could reproduce only once. My guess is that since we don't define which selector we want to target it searches for all (even if it'll target the first) and the first found may not be visible yet, which is why you saw `RuntimeError: element not visible`. The fix would be setting `nth-child(1)`. 

Increasing timeout was just me making sure element could be found without errors. May not be required for it to pass. 

I removed timeout (PR updated) but feel free to close this if it's not necessary/currently working.

bbondy:
This has merge conflicts but I think it's a no-op anyway because of the documentation on webdriver's click function says this:

> element to click on. If it matches with more than one DOM-element it automatically clicks on the first element

So I think this PR, if merged, wouldn't help the issue from failing intermittent.

bbondy:
reference: http://webdriver.io/api/action/click.html

cezaraugusto:
Yes, my comment above was that even targeting the first element by default, error might happen while searching for targets. Explicitly referencing element was a way to avoid that. Not sure about it so ok to close.

WebDriver docs loses only for Immutable docs on my topSites 🙂

Issue #3433, opened 2016-08-25
consolidate state in app/ledger.js with appState


diracdeltas:
there is a lot of state that is duplicated between `appState` and `ledgerInfo`/`publisherInfo` in app/index.js, which becomes confusing to keep in sync. this should be deduplicated somehow.

bsclifton:
I'd like to propose taking this one step further: 
- creating a ledgerState
- refactoring code to be reducer friendly
- breaking up `app/ledger.js` into a utils and a state helper

I have an example that I spent maybe 20 minutes hacking on which shows some progress towards that goal.  Definitely curious what others think!
cc: @bbondy @bridiver @diracdeltas @mrose17 

https://github.com/bsclifton/browser-laptop/commit/bd9e1356a0e3c5819cad6dc88823d72a821420bc

Notice that with a reasonable amount of effort, we can pull most of the code out of `app/ledger.js` and into both a state helper (for methods manipulating state; eg: the reducers) and a utils class (business logic).  With the code broken out here, we can then begin the process of writing unit tests :smile:

bridiver:
a separate ledger state would be going against the plan to consolidate into a single appState. We also want to move away from the json files and move things into the chromium user prefs storage which is already setup for cross-device sync. We can also bring in the history component to keep it out of the general preferences

bsclifton:
@bridiver that makes sense- instead of ledgerState, these reducers could be manipulating appState 😄 Let me update the original post

PR #5824, opened 2016-11-23
minimum visits and minduration not persisting when changed

### LINKED ISSUE #4822: Set 2 visits on minimum visit for publisher relevancy and automatically back to 5
**Did you search for similar issues before submitting this one?** Yes

I set 2 visits on minimum visit for publisher relevancy and automatically back to 5

**Expected behavior:**

Keep on 2
- Platform (Win7, 8, 10? macOS? Linux distro?): Ubuntu 64
- Brave Version: Dev Channel Beta v0.12.5dev-RC1
- Steps to reproduce:
  1. Go to preferences/payment
     2.Open advanced settings
  2. Set 2 visits

Obs. See the video uploaded to understand the problem
[Brave.ogv.zip](https://github.com/brave/browser-laptop/files/531172/Brave.ogv.zip)

jkup:

Fix #4822

Auditors @mrose17 @bsclifton 

Two things here. One, we were multiplying by 1000 before storing the value but that causes a bug when the (now * 1000) value gets back to the UI. So instead, let's just keep the values in milliseconds.

The second thing is we initially store the minDuration and minimumVisits in ledgerData at the root but then we were saving it to ledgerData.synopsisOptions when it changed so the UI was grabbing the old value!

@bsclifton can you think of a good automated test for this? I know we chatted about it yesterday  but I was struggling to implement something that makes sense!

## Test Plan
1. Start Brave
2. Go to Preferences > Payments
3. Click on Advanced Settings
4. Change Minimum Duration to any new value
5. Close and re-open the Advanced Settings modal
6. Make sure the new value persists
7. Do 4-6 with minimum visits as well

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -145,7 +145,7 @@ const doAction = (action) => {
         case settings.MINIMUM_VISIT_TIME:
           if (action.value <= 0) break
 
-          synopsis.options.minDuration = action.value * 1000
+          synopsis.options.minDuration = action.value
           updatePublisherInfo()
           break
 
@@ -580,10 +580,11 @@ var enable = (paymentsEnabled) => {
       // cf., the `Synopsis` constructor, https://github.com/brave/ledger-publisher/blob/master/index.js#L167
       value = getSetting(settings.MINIMUM_VISIT_TIME)
       if (!value) {
-        value = 8
+        value = 8 * 1000
         appActions.changeSetting(settings.MINIMUM_VISIT_TIME, value)
       }
-      if (value > 0) synopsis.options.minDuration = value * 1000
+      // for earlier versions of the code...
+      if ((value > 0) && (value < 1000)) synopsis.options.minDuration = value * 1000
 
       value = getSetting(settings.MINIMUM_VISITS)
       if (!value) {
@@ -1225,9 +1226,6 @@ var getStateInfo = (state) => {
 
   ledgerInfo.paymentId = state.properties.wallet.paymentId
   ledgerInfo.passphrase = state.properties.wallet.keychains.passphrase
-
-  ledgerInfo.minDuration = synopsis.options.minDuration
-  ledgerInfo.minPublisherVisits = synopsis.options.minPublisherVisits
 
   ledgerInfo.created = !!state.properties.wallet
   ledgerInfo.creating = !ledgerInfo.created
--- a/js/about/preferences.js
+++ b/js/about/preferences.js
@@ -1014,8 +1014,8 @@ class PaymentsTab extends ImmutableComponent {
   }
 
   get advancedSettingsContent () {
-    const minDuration = this.props.ledgerData.get('minDuration')
-    const minPublisherVisits = this.props.ledgerData.get('minPublisherVisits')
+    const minDuration = this.props.ledgerData.getIn(['synopsisOptions', 'minDuration'])
+    const minPublisherVisits = this.props.ledgerData.getIn(['synopsisOptions', 'minPublisherVisits'])
 
     return <div className='board'>
       <div className='panel advancedSettings'>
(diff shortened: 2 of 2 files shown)


app/ledger.js
```diff
@@ -143,15 +143,14 @@ const doAction = (action) => {
 
         case settings.MINIMUM_VISIT_TIME:
           if (action.value <= 0) break
-
-          synopsis.options.minDuration = action.value * 1000
```
mrose17: in order for the ledger code to actually do the pruning, you'll need this line. just remove the "* 1000"

app/ledger.js
```diff
@@ -143,15 +143,14 @@ const doAction = (action) => {
...
+          ledgerInfo.minDuration = action.value
           updatePublisherInfo()
           break
 
         case settings.MINIMUM_VISITS:
           if (action.value <= 0) break
 
-          synopsis.options.minPublisherVisits = action.value
```
mrose17: in order for the ledger code to actually do the pruning, you'll also need this line.

bsclifton:
Will check this out tonite and give feedback :smile:

app/ledger.js
jkup: @mrose17 ok so the problem right now in master is that it's first stored in ledgerData and then here it's added to ledgerData.synopsisOptions.

If it needs to be in synopsisOptions is the fix moving the initial value from the root of ledgerData?

jkup:
@bsclifton thanks for checking it out!

@mrose17 how is this approach? What if we don't set the duration and the minimum visits on the root object? And instead only set it / look for it in synopsisOptions?

If this doesn't work we can keep it in both places, it just seems unnecessary :)

mrose17:
i agree it would be best to maintain the value in one place. in that case, i think that the place to keep the value is in `var synopsis` and to have the value copied from there in to/from `ledgerData/appData`. does that make sense?

jkup:
@mrose17 ok so take minDuration and minPublisher visits out of synopsis.options and move them into synopsis?

app/ledger.js
bsclifton: @mrose17 this is seconds now right? (previously was milliseconds?)

bsclifton:
The ledger code is distinctly different than the rest of the codebase, which can cause confusion (or at least it does for me).  I think with a little work, we can make this more Redux friendly. Give me a few mins to cook up a change...

bsclifton:
OK the ideas I had ended up being MUCH larger scope once I dug in. Details can be read here if anyone is interested:
https://github.com/brave/browser-laptop/issues/3433#issuecomment-263080542

For this PR, it sounds like we need to update `ledgerInfo` to also have this new value and then trigger an update to persist the values in appState

Would you guys be able to work this out on Slack? :smile: Or if it's easier, we can do a Google hangout tomorrow (let me know, I'll set it up)

jkup:
@mrose17 LGTM!

@bsclifton assuming you ++ I think this one is ready to merge!

mrose17:
agreed. @bsclifton - you get the "final say" !

app/ledger.js
```diff
@@ -580,10 +580,11 @@ var enable = (paymentsEnabled) => {
       // cf., the `Synopsis` constructor, https://github.com/brave/ledger-publisher/blob/master/index.js#L167
       value = getSetting(settings.MINIMUM_VISIT_TIME)
       if (!value) {
-        value = 8
+        value = 8 * 1000
         appActions.changeSetting(settings.MINIMUM_VISIT_TIME, value)
       }
-      if (value > 0) synopsis.options.minDuration = value * 1000
+      // for earlier versions of the code...
+      if ((value > 0) && (value < 1000)) synopsis.options.minDuration = value * 1000
```
bsclifton: :+1:

bsclifton:
@jkup @mrose17 LGTM! Thanks for working it out :smile: I'll check out the tests here and merge

It **could** use a unit test to help prevent any regressions or unintended altering of the behavior :blush: @jkup if you would be willing to check that out, you could address with a follow up :smile: (logic could be broken out into ledgerUtil for easy testing).  Shouldn't be too bad, just checking the value of `settings.MINIMUM_VISIT_TIME` in the assert

bsclifton:
[Tests finished running](https://travis-ci.org/brave/browser-laptop/builds/179555065); the ones passed when I ran them locally.  Merging! :smile:

jkup:
@bsclifton thanks for the merge! created this https://github.com/brave/browser-laptop/issues/5887

PR #5831, opened 2016-11-24
Open bookmarks in new foreground tabs

### LINKED ISSUE #3207: Middle click on a bookmark link opens tab in the background
Yes

When "Switch to new tabs immediately" setting has been selected, middle clicking on a bookmark link opens a new tab in the background.

**Expected behavior:**
The new tab should open in the foreground, like with right clicking and selecting "Open Link in New Tab".
- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 7, Ubuntu 16.04 LTS
- Brave Version: 0.11.5
- Steps to reproduce:
  1. Go to settings and turn on "Switch to new tabs immediately";
  2. Middle click on a link in the bookmarks toolbar;
  3. New tab opens in the background, not foreground.
- Screenshot if needed:
- Any related issues:

GreenRecycleBin:
These commits should fix #3207

There doesn't seem to be any middle-clicking bookmark item(s) test under https://github.com/brave/browser-laptop/blob/ca626503925c6f706a65896b1b679692374d6697/test/components/bookmarksToolbarTest.js.

In order to add some tests according to the test plans I've outlined in the commits, I'd need to know how to:

- Simulate middle clicking on bookmark item(s)
- Whether to do a URL match of a new foreground tab to that of the original bookmark item or check that such an event is fired

I can probably study other tests and figure out how to do these. But it'd be helpful if someone could give me a few pointers. Thank you.

### DIFF
--- a/js/actions/bookmarkActions.js
+++ b/js/actions/bookmarkActions.js
@@ -7,6 +7,8 @@
 const siteUtil = require('../state/siteUtil')
 const windowActions = require('./windowActions')
 const eventUtil = require('../lib/eventUtil')
+const {SWITCH_TO_NEW_TABS} = require('../constants/settings')
+const getSetting = require('../settings').getSetting
 
 const bookmarkActions = {
   openBookmarksInFolder: function (allBookmarkItems, folderDetail) {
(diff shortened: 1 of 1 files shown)


GreenRecycleBin:
A related question is whether we should implement `Shift` + middle clicking a bookmark folder to open in new foreground tabs regardless of the value of `SWITCH_TO_NEW_TABS` setting? This parallels the current behavior of `Shift` + middle clicking a bookmark item to always open it in a new foreground tab.

GreenRecycleBin:
@bbondy Could you help review this?

bsclifton:
Awesome, thanks for the fix, @GreenRecycleBin!

I manually tested and verified this works as I'd expect. If you wanted to add a test (which I'd highly encourage you to do and submit another PR for), you identified a good place to add the test :smile: 

Here's a [link to our test doc](https://github.com/brave/browser-laptop/blob/master/docs/tests.md), which includes a link to the [webdriver documentation](http://webdriver.io/api.html). To simulate a middle click, you'd want to use the middleClick action:
http://webdriver.io/api/action/middleClick.html

Issue #5896, opened 2016-11-29
Edit Bookmark dialog needs cursor indicators


jonathansampson:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
Some elements of the Edit Bookmark dialog do not change the cursor, thus giving little indication that they can/ought-to be clicked.

**Expected behavior:**
The cursor is always used to make context and behavior clear.

**Platform (Win7, 8, 10? macOS? Linux distro?):** Windows 10

**Brave Version:** 0.12.10

**Screenshot if needed:**
![edit-bookmark-hover-close](https://cloud.githubusercontent.com/assets/815158/20716759/e792919c-b618-11e6-873a-7595a7ab67d5.gif)

**Any related issues:** #5894

jonathansampson:
My mistake; neither element merits changing the cursor.

PR #5900, opened 2016-11-29
Added needed cursor indicators requested in #5896

getsetbro:

Test Plan:

This CSS change resolves issue #5896 requested that the cursor indicate that items are clickable. The two items were the close icon and a SELECT element.

### DIFF
--- a/less/forms.less
+++ b/less/forms.less
@@ -537,6 +537,7 @@
 
 .dialog > .bookmarkDialog .fa-close {
   color: #999;
+  cursor: pointer;
   font-size: 16px;
   float: right;
   margin-top: 7px;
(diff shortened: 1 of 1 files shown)


Liunkae:
#5896 was closed as invalid.

bsclifton:
@getsetbro apologies that I didn't comment here earlier :frowning_face: 

We want to keep the behavior of the default cursor, which gives a more app-like feel. However, we could update styles so that there is a hover effect.  For the X, it should be @buttonColor when idle and #000 (black) when hovered.  You can see a good example of this on the bookmarks manager, with the add folder or add bookmark icons

I'll close this PR... but if you were interested in adding the hover, you could create an issue and submit a new PR against it :smile:

Issue #5855, opened 2016-11-26
An uncaught exception occurred in the main process - Cannot find module 'speedometer'


msiyer:
**Did you search for similar issues before submitting this one?**
Yes. No prior issue.

**Describe the issue you encountered:**
The build was successful and I tried to launch brave from the dev package (**not** rpm install and launch). I see this:
```
An uncaught exception occurred in the main process Uncaught Exception:
Error: Cannot find module 'speedometer'
```

I did `npm install speedometer` and also, `dnf install nodejs-speedometer`, but no success.

**Expected behavior:**
brave should launch

- Platform (Win7, 8, 10? macOS? Linux distro?):
Fedora 25

- Brave Version:
brave@0.12.10

Detailed terminal output:
```
An uncaught exception occurred in the main process Uncaught Exception:
Error: Cannot find module 'speedometer'
    at Module._resolveFilename (module.js:455:15)
    at Function.Module._resolveFilename (/home/msiyer/workspace/browser-laptop/Brave-linux-x64/resources/electron.asar/common/reset-search-paths.js:35:12)
    at Function.Module._load (module.js:403:25)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/msiyer/workspace/browser-laptop/Brave-linux-x64/resources/app.asar/node_modules/webtorrent/index.js:19:19)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
Waiting 60 seconds for process to load
/home/msiyer/workspace/browser-laptop/Brave-linux-x64/resources/app.asar/node_modules/node-anonize2-relic-emscripten/anonize2.js:18
[...]

PR #5921, opened 2016-11-29
Fix webtorrent, don't exclude dependencies

### LINKED ISSUE #5849: LSOpenURLsWithRole() failed with error -10810 on the packaged build
From Terminal when I try to open the packaged build, I get this.
````
Brave-MacBook-Pro:browser-laptop Suguru$ open ./Brave-darwin-x64/Brave.app
LSOpenURLsWithRole() failed with error -10810 for the file /Users/Suguru/browser-laptop/Brave-darwin-x64/Brave.app.
````

- Brave Version: master branch

- Any related issues: #5818

CC @dcposch

dcposch:
Fixes #5849

Also lazy-load `webtorrent-remote` in the browser process.

* This should make startup a bit faster, since `require('webtorrent-remote/server')` will pull in `webtorrent`, which is a big module.
* It also makes exceptions like #5849 easier to debug. Instead of getting a cryptic error, you get a stack trace at runtime when you try to load your first torrent URL.

BTW, I recommend:
* Measuring startup time. This can be as simple as adding a `console.time` ... `console.timeEnd`. We do that in WebTorrent Desktop and it helps catch performance regressions.
* Add a bit more error handling on startup. If the packaged app crashes on launch because of a `require()` error, it would be better to print the stacktrace than the `LSOpenURLsWithRole` exception.


dcposch:
@feross @bbondy @luixxiul

bbondy:
++

luixxiul:
Should this fix #5855 too? @dcposch

dcposch:
@luixxiul yes

Issue #2468, opened 2016-07-13
Support 32-bit linux


luixxiul:
From support (tickets/4470)

Is it already possible to build the app on 32 bit Linux?

- Any related issues: #397 https://github.com/brave/browser-laptop/pull/10514#issuecomment-322642404

bbondy:
we're just waiting for some build automation for this.

bsclifton:
+1 from support
https://linkbubble.zendesk.com/agent/tickets/6655

Liunkae:
Related: #235 (make builds for 32-bit Linux)

posix4e:
yarr we can print these

posix4e:
I have uploaded a 32 bit preview appimage to my personal bucket
http://storage.googleapis.com/posis-brave-mirror/

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

ariSunz:
Any updates on this?

posix4e:
@ariSunz Sure did you see our preview?

ariSunz:
You mean your last comment? I only see an XML document and I have no idea of what to do with it. Sorry for my ignorance. Also, I'd rather avoid any services related to Google...

posix4e:
@ariSunz  you should be able to look at the document and tell the filenames. For instance
http://storage.googleapis.com/posis-brave-mirror/brave-0.20.0-i386.AppImage

bsclifton:
Closing as `wontfix`. If you think this should be implemented when we move to Brave Core, please feel free to open an issue there:
https://github.com/brave/brave-browser/issues

PR #4723, opened 2016-10-12
Change RPM builds architecture to 'x86_64'

Aeyoun:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Test Plan:

1. Build a release, install on Fedoara with `dnf install package.rpm`. Package name should end with `x86_64.rpm`. Package should install with no errors.
2. Build a new release (higher version number). Upgrade on Fedoara with `dnf install package.rpm`. Should update with no errors.
3. Debian/Ubuntu package names and installation should be unchanged.

Core issue:
Incorrect documentation from upstream. [Submitted correction](https://github.com/unindented/electron-installer-redhat/pull/38) to their documentation.

Commit message:
Red Hat/RPM universe expects 'x86_64' whereas Debian expects 'amd64'.

Also expands documentation in some areas.

This closes #3774.

PR #5977, opened 2016-12-01
Add support for 32 bit linux builds

posix4e:
This is will change the name of our deb packages to have x64 instead of amd64. Do we care? RPMs will also be x64 instead of x86_64. Happy to keep it as is if we prefer


Liunkae:
Fixes #2468.

bsclifton:
IIRC, someone had made a PR to specifically have RPMs have x86_64 (which I think is the arch for 64 bit on Red Hat OSes)... let me check

bsclifton:
Here we go :smile: 

@posix4e what do you think? (see notes)
https://github.com/brave/browser-laptop/pull/4723

posix4e:
I think this is pretty much all good on chrome54? Good to target muon?

luixxiul:
So did the 32 bit build become available already?

bsclifton:
cc @posix4e

posix4e:
Soon!

PR #5995, opened 2016-12-02
Fix unit tests

bsclifton:
~Fix urlutils tests after https://github.com/brave/browser-laptop/commit/dcfd67d86e7ddd4e6afff38af4b13f235a8b33a4#diff-7b5b041c814805dd8496488bc6a53d3c~

Fix platformUtil tests after https://github.com/brave/browser-laptop/commit/24a0493421cf0448ebda4dd035ce665b44a6adfc#diff-fd09a6dc151769bdeaaf61fb423090f5
Set a default value (false) for 'security.flash.installed'

Auditors: @darkdh @bridiver

### DIFF
--- a/js/constants/appConfig.js
+++ b/js/constants/appConfig.js
@@ -141,6 +141,7 @@ module.exports = {
     'security.passwords.one-password-enabled': false,
     'security.passwords.dashlane-enabled': false,
     'security.passwords.last-pass-enabled': false,
+    'security.flash.installed': false,
     'general.downloads.default-save-path': null,
     'general.disable-title-mode': process.platform === 'linux',
     'advanced.hardware-acceleration-enabled': true,
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -290,6 +290,34 @@ const UrlUtil = {
   },
 
   /**
+   * Checks whether a link is an Flash installer URL.
+   * @param {string} url
+   * @return {boolean}
+   */
+  isFlashInstallUrl: function (url) {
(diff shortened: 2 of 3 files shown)


test/unit/lib/urlutilTest.js
```diff
@@ -179,41 +179,6 @@ describe('urlutil', function () {
     })
   })
 
-  describe('isFlashInstallUrl', function () {
```
bridiver: I don't think these tests should go away. I didn't realize these tests existed when I made the changes because the tests weren't working, but those methods can move back to urlUtil to make them unit testable again

js/constants/appConfig.js
```diff
@@ -141,6 +141,7 @@ module.exports = {
     'security.passwords.one-password-enabled': false,
     'security.passwords.dashlane-enabled': false,
     'security.passwords.last-pass-enabled': false,
+    'security.flash.installed': false,
```
bridiver: nice catch

darkdh:
platformUtilTest looks good to me

bsclifton:
@bridiver I added the methods back in and also the tests :smile: Should be good to go

bridiver:
@bsclifton I don't see a change to use those methods in flash.js

bsclifton:
@bridiver I'll get those in a follow up 😄

luixxiul:
Please add a tracking issue for this, thanks.

bsclifton:
Marked with `automated-tests` and `QA/no qa needed`

PR #6032, opened 2016-12-06
these edits add 2 px to our tab height

bradleyrichter:
- this adds some needed space above and below the text string
- increases tab text to 12 on mac (already 12 on windows)
- tabs buttons adjusted to align properly

auditors:  @bsclifton

### DIFF
--- a/less/button.less
+++ b/less/button.less
@@ -22,8 +22,8 @@ span.buttonSeparator {
   background-color: @buttonColor;
   display: inline-block;
   width: 20px;
-  height: 12px;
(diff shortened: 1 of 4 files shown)


luixxiul:
screenshot 2016-12-06 15 51 41

I like this better:
````
background: linear-gradient(to bottom, rgba(255, 255, 255, 0.5), rgba(255, 255, 255, 0.4));
````

screenshot 2016-12-06 16 04 04

It creates some contrast between tabs and I think it's good in terms of accesibility. wdyt? @bradleyrichter

PR #6037, opened 2016-12-06
Add focus ring to about page form elements

### LINKED ISSUE #5971: Add focus ring to about page form elements
This came up in our most recent #design meeting but currently it is impossible for keyboard-only users to know which form element they are on because we don't display a visual change on focus state.

Some of this will be fixed by https://github.com/brave/browser-laptop/pull/5916 which switches us from span's to button tags (which have focus state!) but this ticket is for adding back the blue focus ring to select boxes, textarea's and input fields.

jkup:

Fix #5971

Auditor @bsclifton, @cezaraugusto 

## Test Plan

1. Open Brave
2. Go to Preferences
3. Start tabbing through the document. Make sure all buttons and inputs/selects have blue focus rings.
4. Then Add some bookmarks
5. Toggle on show bookmark bar
6. Click in the urlbar
7. Start tabbing through, make sure each bookmark also gets a blue focus ring

### DIFF
--- a/app/renderer/components/bookmarksToolbar.js
+++ b/app/renderer/components/bookmarksToolbar.js
@@ -172,7 +172,7 @@ class BookmarkToolbarButton extends ImmutableComponent {
         : siteDetailLocation
     }
 
-    return <span
+    return <button
       className={cx({
         bookmarkToolbarButton: true,
         draggingOverLeft: this.isDraggingOverLeft && !this.isExpanded,
@@ -217,7 +217,7 @@ class BookmarkToolbarButton extends ImmutableComponent {
         ? <span className='bookmarkFolderChevron fa fa-chevron-down' />
         : null
       }
-    </span>
+    </button>
   }
 }
 
--- a/less/bookmarksToolbar.less
+++ b/less/bookmarksToolbar.less
@@ -61,6 +61,12 @@
       box-shadow: 0px 1px 5px 0px rgba(0, 0, 0, 0.1);
     }
 
+    &:focus {
+      outline-color: rgb(59, 153, 252);
+      outline-style: auto;
+      outline-width: 5px;
+    }
+
     &.draggingOverLeft {
       &:not(.isDragging) {
         margin-left: 25px;
--- a/less/button.less
+++ b/less/button.less
@@ -6,7 +6,6 @@
 
 button[class*="Button"] {
   background: none;
-  outline: none;
   border: none;
   margin: 0;
   white-space: nowrap;
--- a/less/forms.less
+++ b/less/forms.less
@@ -15,7 +15,6 @@ select {
   color: @darkGray;
   font-size: 14.5px;
   height: 2.25em;
-  outline: none;
   padding: 0.4em;
   width: 100%;
 }
--- a/less/navigationBar.less
+++ b/less/navigationBar.less
@@ -545,6 +545,11 @@
   }
 }
 
+.navbutton,
+.navigationButton {
+  outline: none;
+}
+
 .navigationButton {
   background-color: @buttonColor;
   display: inline-block;


luixxiul:
Is this expected?

screenshot 2016-12-06 23 15 05


@bradleyrichter @jkup https://github.com/brave/browser-laptop/blob/f7162732d97639d18e3f03dacd07529da89a8d79/less/forms.less#L614 too?

luixxiul:
Also is this expected on about:preferences#payments?

screenshot 2016-12-06 23 35 03

bradleyrichter:
oh no! This may have arrived with a chromium update?

We need to prevent this selection from ALL mouse input outside of the page content box.

luixxiul:
If only elements on about pages should have the outline, I would add like this in preferences.less:

````
.prefBody {
  display: flex;
  height: 100%;
  margin-left: @sideBarWidth;

  button[class*="Button"],
  input,
  select {
    &:focus {
      outline: -webkit-focus-ring-color auto 5px;
    }
  }
}
````

luixxiul:
@bradleyrichter I thought only button, input and select element should have the outline. Please correct me if I'm wrong.

bradleyrichter:
To be more specific, this selection should only be shown when invoked for accessibility as an indicator for tabbing through elements when a mouse can not be physically used. If we can control this, it can be used in the toolbars as well. I just don't want it to appear outside of an accessibility mode.

luixxiul:
> I just don't want it to appear outside of an accessibility mode.

Can we detect it via JS?

bradleyrichter:
@luixxiul @jkup 

Reviewing Chrome and Firefox shows that no browser toolbar items are tab-advance selectable except for the URL field which should have its own style and not use the item-select border.

Thus, "-webkit-focus-ring-color " should only apply to content area elements and never the toolbar elements.

jkup:
We want not only prefs pages but also the bookmarks bar to show the outline like Chrome does. Also there is no way to use JS to detect accessibility mode. I'll update the selectors to be more specific.

bsclifton:
Sorry that I'm just now getting to this; the outline style looks great :smile: 

When I tried latest against our prefs page, it seems to only be highlighting:
- URL bar
- text fields
- select fields
- bookmarks toolbar (nice!! :sunglasses:)

I didn't see buttons get focus, even if I clicked them and then tried the tab :frowning_face:

jkup:
Fixed it for the navigation buttons like reload/back/forward but the links in preferences#payments should get a focus ring so users know they can interact with them.

What @luixxiul mentioned here is the correct behavior: https://github.com/brave/browser-laptop/pull/6037#issuecomment-265165002

jkup:
@bsclifton @luixxiul is this good to go?

jkup:
Hey @cezaraugusto I know @bsclifton is super busy lately.. any chance you can give this a review for me and add the ready-to-merge label if it looks good? 🙏

cezaraugusto:
++ Manually tested, LGTM

Not specifically related, but per @luixxiul comment [here](https://github.com/brave/browser-laptop/pull/6037#issuecomment-265165002), it looks weird for me as well. Simple solution would be add `display: flex` (or `block`) to link inside `.site`. But it's ok as-is, just polish.

In the future, as we talk about accessibility, would love to see switch buttons _tabbable_ as well.

jkup:
@cezaraugusto thanks! I created https://github.com/brave/browser-laptop/issues/6439 to track that UI polish!

bsclifton:
[Last travis-ci run](https://travis-ci.org/brave/browser-laptop/builds/183823284) showed several errors that I believe are not related; I re-ran each of those locally and they pass as expected :smile: Merging...

bsclifton:
Removed milestone after revert with https://github.com/brave/browser-laptop/commit/2aa39bb1073fe8bb40d9fc32d18fdd2ea9cee1d5

luixxiul:
@bsclifton I'm not sure why this needed to be reverted. Will you please let me know?

bsclifton:
@luixxiul the focus ring negatively affected too many elements on the page. For example, see these:
https://github.com/brave/browser-laptop/issues/6760
https://github.com/brave/browser-laptop/pull/6761

Also, having the bookmarks toolbar items outlines is undesirable. cc: @bbondy

bradleyrichter:
There must be a way to contain this to the DOM/page region so it never effects the toolbar elements...

luixxiul:
`#window > .mainContainer` could be that way

bsclifton:
I think at this point, it would be better to go back to the issue and re-open if we think it's important.  We can lay out requirements and set an appropriate milestone

PR #6013, opened 2016-12-04
cleanup about page handling and fix missing document.l10n

bridiver:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [ ] Ran `git rebase -i` to squash commits (if needed).

fixes #6012

Test Plan:
For each about page:
1. Load the page
2. Ensure that the page and favicon load correctly

PR #6062, opened 2016-12-07
cleanup about page handling and fix missing document.l10n

bridiver:
see https://github.com/brave/browser-laptop/pull/6013

moves the filter before anything that could ignore it and just check for the gen dir instead of specific entry points

@bbondy


bbondy:
thanks

Issue #5891, opened 2016-11-29
convert URL bar FA icons to sprite PNG (or SVG)


bradleyrichter:
Using a sprite grid for our URL icons will allow us to control the alignment of each icon more easily within the image file without the need for unreliable pixel bumping in CSS.

20 icon positions are available initially.

Each icon is centered on the grid.

![image](https://cloud.githubusercontent.com/assets/13509546/20694845/f798210a-b59c-11e6-9f60-7a5b701747e5.png)

A1 - unused (future verified certificate)
A2 - HTTPS (secure)
A3 - unused (future mixed HTTPS) 
A4 - HTTP (non-secure)
A5 - unused (normal file)
A6 - search (normal)
A7 - unused (info)
A8 - alert page (all error results pages)
A9 - brave payments = yes
A10 - all brave about:pages

B1 - verified tag (used in brave payment table)
B2 - brave hosted page
B3 - unused (brave alert)
B4 - HTTP for title mode (non-secure) 
B5 - unused (normal file - grey)
B6 - unused (search - grey)
B7 - unused (info - grey)
B8 - noscript active
B9 - brave payments = no
B10 - unused (all brave about:pages - grey)

![url_bar_icon_sprite 3x](https://cloud.githubusercontent.com/assets/13509546/20695209/0c8f25e8-b59f-11e6-8d5b-507d45ce162f.png)

![url_bar_icon_sprite 2x](https://cloud.githubusercontent.com/assets/13509546/20695450/e122a784-b5a0-11e6-96c4-ef7f6f0d2174.png)

![url_bar_icon_sprite 1x](https://cloud.githubusercontent.com/assets/13509546/20695453/e41a6238-b5a0-11e6-83fd-af8b4adc5975.png)

bradleyrichter:
CC @bbondy

bbondy:
nice work!

bsclifton:
Starting to implement these using Aphrodite! So far, so good :smile:

bsclifton:
@bradleyrichter I made good progress with this last nite (which you can see here: https://github.com/bsclifton/browser-laptop/commit/9091b3170d4906bc8a134108d26d2d73003656ae), but I was not able to finish it unfortunately.  Would you be OK if we bumped this to 0.13.1? None of the urlbar icons have changed or gotten worse between 0.12.10 and now

luixxiul:
I thought they should be SVG instead of PNG? @bradleyrichter

[...]

PR #6084, opened 2016-12-08
Restyled select and input to the spec image

### LINKED ISSUE #5972: Separate input and select styles using Aphrodite
The goal for this issue would be to separate input and select styles (which are currently coupled together).  [Per our style guidelines](https://github.com/brave/browser-laptop/blob/master/docs/style.md), solving this issue would involve rolling the input and select elements into an Aphrodite component.

- [x] Text entry field (editable)
- [x] Text entry field (focus outline)
- [x] Text entry field (locked/fixed value)
- [x] drop-down menu style
- [x] preferred drop down arrow style
- [x] flatten out the orange button
- [x] update about:styles

Here are some spec images:

## Text entry field (editable)
image

image

## Text entry field (focus outline)
image
 (use a 2 px radius, and "@statsBlue: #0796fa;")

## Text entry field (locked/fixed value)
image

image
(use variable color for fill: @veryLightGray: rgb(250, 250, 250); )

## drop-down menu style

image

image

## preferred drop down arrow style
image

We can switch to a different FA triangle
image

## and this will flatten out the orange button
image

image

luixxiul:

Closes #5972
The SVG caret down will be replaced with the sprite later (#5891)

Test Plan: in https://github.com/brave/browser-laptop/pull/6084/commits/fd0c10af0de5e76fb4bbe803a907f168ba4414d2

Auditors:

screenshot 2016-12-08 14 53 33

Into
screenshot 2016-12-11 19 01 19
screenshot 2016-12-08 15 36 51

It seems there is no property such as `outline-radius`.

screenshot 2016-12-08 15 44 19


luixxiul:
If there is (and should be) a better way to handle radius inside the input form, please edit the commit, thanks.

bradleyrichter:
sweet!

I do see that we need one more style difference between editable text entry fields and non-editable. (can be the same but light grey BG instead of white.)

luixxiul:
@bradleyrichter even `@veryLightGray` in variable.less is too dark on about:preferences, though I'm not sure if `#fefefe` makes difference.

luixxiul:
be2a185 fixes

screenshot 2016-12-10 16 51 33

into
screenshot 2016-12-10 16 51 40

luixxiul:
With 21a13ef

screenshot 2016-12-10 17 24 44

less/forms.less
```diff
@@ -22,7 +22,8 @@ select {
   width: 100%;
 }
 
-select.form-control {
+select.form-control,
+.bookmarkForm select {
```
luixxiul: I know this is not an elegant solution; general properties for `select` and `input` should be specified at first, followed by more specific properties for `.form-control`, etc.

luixxiul:
With f21ff68

screenshot 2016-12-11 18 55 55

bradleyrichter:
@luixxiul These look good. The one thing I see that is still wrong is that the form elements in the bookmark hanger should be taller to match the prefs pages.

luixxiul:
@bradleyrichter Should it be like this?

screenshot 2016-12-12 4 42 55

bradleyrichter:
yes. perfect.

luixxiul:
OK, updated.

bsclifton:
Will check this out soon (I promise!)

bradleyrichter:
@bsclifton Candidate for 13.0? Or wait?

bsclifton:
@bbondy since these are only style changes, if we test appropriately before accepting, could we pull this into 0.13.0?

luixxiul:
Does anyone know any hack to fix this?

screenshot 2016-12-08 15 36 51

because outline-radius does not exist.

bsclifton:
Let's keep this for 0.13.1 :smile:

luixxiul:
CI passed: https://travis-ci.org/brave/browser-laptop/builds/190889920

cezaraugusto:
look good but needs rebase

bsclifton:
I've been very hesitant to review this PR because of all of the time cost associated with bugs and regressions that style updates introduce.  Per our reviewer guidelines, I'm going to close this issue:
https://github.com/brave/browser-laptop/wiki/Reviewer-guideline

> 6. Front end: If styles are being added, make sure it is using Aphrodite.

I'll update the original issue to capture this requirement

luixxiul:
OK but please make sure that the guideline was published 8 days ago while the PR was initially opened 1 month ago

bsclifton:
@luixxiul I understand.  I definitely appreciate the work, but the cost of reviewing adhoc style changes is orders of magnitude higher than reviewing code which has tests.

Issue #5911, opened 2016-11-29
Clicking link from email will sometimes will do a search instead of directly following


bsclifton:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
With the following conditions:
- Brave being default browser
- clicking email link (email marketing) in a native client such as Microsoft Outlook

Brave will sometimes launch and then immediately do a search using your default search engine.  When I dug deeper, I noticed that a period at the end of the URL was triggering this.  You could repro by pasting the full URL into the omnibox and hitting enter.  Removing that single period fixed the issue 

@bradleyrichter also saw this behavior and can provide an example. Having a different default browser open works as you'd expect.

**Expected behavior:**
Links like this should be followed, not causing a search.

- Platform (Win7, 8, 10? macOS? Linux distro?):
macOS

- Brave Version:
0.12.10

- Steps to reproduce:
    1. Launch Brave and open a new tab
    2. Paste in `https://e.turbotax.intuit.com/pub/cc?_ri_=3vv-8-e.` and hit enter
    3. Notice a search is done using your default search engine
    4. Paste in `https://e.turbotax.intuit.com/pub/cc?_ri_=3vv-8-e` (notice: missing the period at the end) and hit enter
    5. Notice that Brave properly follows the link (in this case, the link doesn't go anywhere; I sanitized it so that folks can't re-sign me up for emails! :smile: )

- Screenshot if needed:

- Any related issues:

bsclifton:
cc: our URL bar experts :smile: 
@aekeus @bbondy @diracdeltas

bsclifton:
I created a PR which has a failing test:
https://github.com/brave/browser-laptop/pull/5912

If you grab this one, you can cherry-pick that commit in and run it to verify that issue is actually fixes :smile:

Sh1d0w:
@bsclifton Possible places to check where is goes wrong:

https://github.com/brave/browser-laptop/blob/master/js/lib/urlutil.js#L110

https://github.com/brave/browser-laptop/blob/master/js/lib/urlutil.js#L78

bsclifton:
Bumped up milestone / assignee to match https://github.com/brave/browser-laptop/issues/6058

Issue #4546, opened 2016-10-05
Search via URL bar doesn't work when ending with a dot


JamborJan:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When I'm using the URL bar to search for something and I or auto correct adds a dot at the end brave thinks it's a URL and wants to open this wrong URL

**Expected behavior:**
Brave should trigger the configured search engine to search for the term including the dot.

- Platform (Win7, 8, 10? macOS? Linux distro?):
tested with iOS 10.0.2 and OS X 10.11.6 with Safari 10.0

- Brave Version:
0.12.3

- Steps to reproduce:
    1. type "test." in the empty URL bar
    2. hit enter
    3. Brave tries to open "http://test./"

- Screenshot if needed:
![bildschirmfoto 2016-10-05 um 14 30 37](https://cloud.githubusercontent.com/assets/7508960/19113266/28609ace-8b09-11e6-9ae4-0a1c6b5408f4.png)

- Any related issues:

PR #6086, opened 2016-12-08
Hardened isNotURL / isURL code in urlutil.js

bsclifton:

Fixes https://github.com/brave/browser-laptop/issues/5911
(which was unintentionally introduced with https://github.com/brave/browser-laptop/issues/4546)

Also fixes these cases (no issues created yet):
- view-source/mail-to/etc were not checking from beginning of string
- ensures input is string (preventing failing call to trim())

And adds missing tests for:
- checking localhost toLowerCase()
- checking basic auth formatted URI (with and without protocol)
- is input a string
- each type of special page
- partial bad matches on special pages

Also updates tests for isURL to ensure value is always negated version of isNotURL

## Auditors
@bbondy, @darkdh, @diracdeltas

Since this code has a very large potential impact (everything from the URL bar calls this for example), please review and help make sure I covered everything safely :)

## Unit test
`npm run unittest -- --grep="urlutil"`

## Webdriver test
```
# window 1
npm run watch-test
#window 2
npm run uitest -- --grep="when following URLs"
```

### DIFF
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -99,29 +99,32 @@ const UrlUtil = {
     if (input === undefined || input === null) {
       return true
     }
-
+    if (typeof input !== 'string') {
+      return true
+    }
     // for cases where we have scheme and we dont want spaces in domain names
     const caseDomain = /^[\w]{2,5}:\/\/[^\s\/]+\//
     // for cases, quoted strings
     const case1Reg = /^".*"$/
-    // for cases, ?abc, "a? b", ".abc" and "abc." which should searching query
-    const case2Reg = /^(\?)|(\?.+\s)|^(\.)|(\.)$/
+    // for cases:
+    // - starts with "?" or "."
+    // - contains "? "
+    // - ends with "." (and was not preceded by a /)
+    const case2Reg = /(^\?)|(\?.+\s)|(^\.)|(^[^\/]*\.$)/
     // for cases, pure string
     const case3Reg = /[\?\.\/\s:]/
     // for cases, data:uri, view-source:uri and about
-    const case4Reg = /^data:|view-source:|mailto:|about:|chrome-extension:|magnet:.*/
+    const case4Reg = /^(data|view-source|mailto|about|chrome-extension|magnet):.*/
 
     let str = input.trim()
-    let scheme = this.getScheme(str)
+    const scheme = this.getScheme(str)
 
     if (str.toLowerCase() === 'localhost') {
       return false
     }
-
     if (case1Reg.test(str)) {
       return true
     }
-
     if (case2Reg.test(str) || !case3Reg.test(str) ||
         (scheme === undefined && /\s/g.test(str))) {
       return true
@@ -129,11 +132,9 @@ const UrlUtil = {
     if (case4Reg.test(str)) {
       return !this.canParseURL(str)
     }
-
     if (scheme && (scheme !== 'file://')) {
       return !caseDomain.test(str + '/')
     }
-
     str = this.prependScheme(str)
     return !this.canParseURL(str)
   },
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -109,8 +109,8 @@ const UrlUtil = {
     // for cases:
     // - starts with "?" or "."
     // - contains "? "
-    // - ends with "." (and was not preceded by a /)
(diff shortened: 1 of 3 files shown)


js/lib/urlutil.js
```diff
@@ -99,41 +99,42 @@ const UrlUtil = {
...
     // for cases where we have scheme and we dont want spaces in domain names
     const caseDomain = /^[\w]{2,5}:\/\/[^\s\/]+\//
     // for cases, quoted strings
     const case1Reg = /^".*"$/
-    // for cases, ?abc, "a? b", ".abc" and "abc." which should searching query
-    const case2Reg = /^(\?)|(\?.+\s)|^(\.)|(\.)$/
+    // for cases:
+    // - starts with "?" or "."
+    // - contains "? "
+    // - ends with "." (and was not preceded by a /)
```
darkdh: type `123/123.` will be search string on other browsers

darkdh: fixed in cd71e0e and test is also covered. Mainly changing logic of end with period, which is 
`brave.com/123.` will be url and `brave/com/123.` will not

test/unit/lib/urlutilTest.js
```diff
@@ -86,8 +86,11 @@ describe('urlutil', function () {
       it('is localhost (case-insensitive)', function () {
         assert.equal(UrlUtil.isNotURL('LoCaLhOsT'), false)
       })
-      it('ends with period (input contains a forward slash)', function () {
-        assert.equal(UrlUtil.isNotURL('https://brave.com/test/cc?_ri_=3vv-8-e.'), false)
+      it('ends with period (input contains a forward slash and domain)', function () {
+        assert.equal(UrlUtil.isNotURL('brave.com/test/cc?_ri_=3vv-8-e.'), false)
+      })
+      it('ends with period (input contains only a forward slash)', function () {
+        assert.equal(UrlUtil.isNotURL('brave/com/test/cc?_ri_=3vv-8-e.'), true)
```
bsclifton: Testing for an extra .might not be a good idea, because folks may use Brave to browse their intranet. It's not a best practice, but many folks will setup a host file entry and access hosts by computer name (instead of domain)

bsclifton: actually- I think this behavior might be OK!  I am almost positive you are correct about the other behavior (because it's happened to me so many times :stuck_out_tongue: ) When prefixed with protocol, it validates just fine: https://github.com/brave/browser-laptop/pull/6086/commits/cd88861cfa29f5e369001386b459ee0fe949cc57

diracdeltas:
i'm not sure why we are using regexes and `window.URL` in urlutil at all. seems like you should just be able to do `urlParse` and conclude the URL is invalid if `protocol` and `host` are null.

bbondy:
@diracdeltas 

> i'm not sure why we are using regexes and window.URL in urlutil at all. seems like you should just be able to do urlParse and conclude the URL is invalid if protocol and host are null.

I didn't write this original code but really it's more about heuristics about if a user meant to type a url or not.  For example `require('url').parse('a b')` will give output of no `protocol` but will give `pathname`, `path` and `href`.  The user means a search in that case. 

For example `require('url').parse('www.test.com')` will give the same output types.  And in that case it's meant to be a URL.  So this function is more about what the user intended for it to be, and not whether it can be considered a URL or not.

bbondy:
The second example has host not filled in by the way

js/lib/urlutil.js
```diff
@@ -99,41 +99,42 @@ const UrlUtil = {
...
+    // for cases:
+    // - starts with "?" or "."
+    // - contains "? "
+    // - ends with "." (and was not preceded by a domain or /)
+    const case2Reg = /(^\?)|(\?.+\s)|(^\.)|(^[^.+\..+]*[^\/]*\.$)/
     // for cases, pure string
     const case3Reg = /[\?\.\/\s:]/
     // for cases, data:uri, view-source:uri and about
-    const case4Reg = /^data:|view-source:|mailto:|about:|chrome-extension:|magnet:.*/
+    const case4Reg = /^(data|view-source|mailto|about|chrome-extension|magnet):.*/
```
bbondy: :|

bbondy: ouch

diracdeltas:
@bbondy i figured, but i was hoping we could just do urlParse after normalizing the input; i will open a separate issue

PR #6127, opened 2016-12-10
Simplified build instructions

posix4e:
This simplifies the build instruction and assumes a modern enough system to support lsb_release

PR #6135, opened 2016-12-11
Improve clarity of the Ubuntu install instructions

leighmcculloch:
What
===
Change the Ubuntu AMD64 installation instructions to show the commands required to append the sources.list entry and complete the full installation, rather than only show the entry that needs to be made.

Why
===
The instructions are not clear that the deb entry needs to be added to the sources.list and the way it is included makes it look like a command the user needs to execute. The instructions are also missing the `apt update` and `apt install brave` that are also necessary.


bradleyrichter:
@posix4e Can you approve and merge this one? Ty...

posix4e:
What about https://github.com/brave/browser-laptop/pull/6127

leighmcculloch:
@posix4e We need both PRs I guess. They are both essentially the same, but we should include the update and install commands too.

PR #6145, opened 2016-12-12
Deb Docs instructions simplied

posix4e:
Assume a new enough linux for lsb_release.
  Use Lsb_release to combine duplicate instructions


alexwykoff:
Tested for Ubuntu and Linux Mint, both worked without issue.

Issue #6209, opened 2016-12-14
Ubuntu update problem


Liunkae:
**Did you search for similar issues before submitting this one?**
Yes. None of the other Ubuntu installation issues seem the same.

**Describe the issue you encountered:**
From https://github.com/brave/browser-laptop/issues/6199#issuecomment-266956777:
>There seem to be some problems with the automatic update process. I'm getting:
>
>`W: Failed to fetch https://s3-us-west-2.amazonaws.com/brave-apt/dists/trusty/InRelease Unable to find expected entry 'main/binary-i386/Packages' in Release file (Wrong sources.list entry or malformed file)`
>
>even though I'm on amd64 architecture

**Expected behavior:**
Brave should install/update without error.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Ubuntu 14.04.LTS

- Brave Version:
Updating from 0.12.6 to 0.12.14

- Steps to reproduce:
    1. Have an old version of Brave installed on Ubuntu.
    2. Follow the [Linux installation instructions](https://github.com/brave/browser-laptop/blob/0403102d51903bb71db3ad0a76c88bb769bd0684/docs/linuxInstall.md).

posix4e:
We haven't release i386 yet

Liunkae:
Has this been fixed with 24b58cd?

Liunkae:
I think this was fixed with 24b58cd. Closing.

PR #6212, opened 2016-12-14
Fixed APT Unable to find expected entry

hmrodrigues:
When adding the repository to Debian/Ubuntu, you must add [arch=amd64] so apt doesn't complain about not finding the i386 binaries.

**Issue**: #6209 

**Steps to fix**: [arch=amd64] between _deb_ and the repository URL

### DIFF
--- a/docs/linuxInstall.md
+++ b/docs/linuxInstall.md
@@ -9,7 +9,7 @@ want to try installing `git` using the package manager for your distro._
 To install brave using apt:
 ``` 
 curl https://s3-us-west-2.amazonaws.com/brave-apt/keys.asc | sudo apt-key add -
-deb https://s3-us-west-2.amazonaws.com/brave-apt jesse main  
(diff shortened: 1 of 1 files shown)


jimfenton:
With the added `[arch=64]`, it's even less obvious that one should replace `[xenial/trusty]` with either xenial or trusty, especially for someone who isn't very familiar with apt.

It might be clearer to give separate examples for the releases, or to add some text telling the user to make the substitution.

bbondy:
@posix4e could you review?
@hmrodrigues in the meantime I think the PR needs to be rebased if you don't mind.

posix4e:
Good idea although in a few weeks we will have 32 build.

posix4e:
-1 please rebase

hmrodrigues:
Hey hmmm, to do the rebase, I need to pick my last commit and squash to where?
Sorry first time doing this

posix4e:
It might be easier just to do it manually

PR #6249, opened 2016-12-16
Add currency symbols where appropriate

### LINKED ISSUE #3533: Add currency symbols where appropriate
Using a library like https://www.npmjs.com/package/currency-symbol-map it would be nice to add currency symbols on the ledger settings pages for the appropriate currency type.

It looks like "currency" is stored in state as a string (https://github.com/brave/browser-laptop/blob/master/docs/state.md#L403), so we should just be able to pipe it through a library like the one above and then prepend it to the currency number.

jkup:

Fix #3533

Auditors @bsclifton @cezaraugusto 

Test Plan:

1. Go to Preferences > Payments
2. Check the payment amount, the current balance, and your payment history and make sure they all have $ for currency.
screen shot 2016-12-16 at 1 25 07 am

### DIFF
--- a/js/about/preferences.js
+++ b/js/about/preferences.js
@@ -23,6 +23,7 @@ const preferenceTabs = require('../constants/preferenceTabs')
 const messages = require('../constants/messages')
 const settings = require('../constants/settings')
 const coinbaseCountries = require('../constants/coinbaseCountries')
+const getSymbolFromCurrency = require('currency-symbol-map').getSymbolFromCurrency
 const {passwordManagers, extensionIds} = require('../constants/passwordManagers')
 const {startsWithOption, newTabMode, bookmarksToolbarMode} = require('../../app/common/constants/settingsEnums')
 const {l10nErrorText} = require('../../app/common/lib/httpUtil')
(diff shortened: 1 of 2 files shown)


bsclifton:
How is `this.props.ledgerData.get('currency')` determined? I'm guessing it's based on the IP?

@jkup can you screenshot the places where currency is updated? I'm also curious how the formatting looks for other currency

jkup:
@bsclifton yep it's based on the IP. I added a screenshot of what this looks like for USD but I'm a little hesitant to ship this ticket because after talking with @mrose17 it turns out faking the currency is difficult to do.

I suppose my only real worry is that we support some currency that doesn't exist in this package's currency map (https://github.com/bengourley/currency-symbol-map/blob/master/map.js) so it will default to USD instead. Thoughts?

mrose17:
the list is fairly comprehensive, so i'm no worried about it.

we do not have plans to turn on additional currencies right now, so i don't think this is a priority.

i would do the PR and then mark the issue closed.

cezaraugusto:
++ lgtm

bsclifton:
Thanks for the update! Looks good :smile:

luixxiul:
Note: `npm i` is required to fix the error

````
Uncaught TypeError: getSymbolFromCurrency is not a function{TypeError: getSymbolFromCurrency is not a function
````

jkup:
@luixxiul yep this is how adding npm packages works

bradleyrichter:
@cezaraugusto This PR is useful but needs to be corrected ASAP to show the ISO 4217 currency code (3 letters) instead of the symbol.

bsclifton:
This was reverted; removing milestone

PR #6186, opened 2016-12-13
Redo: Add copy to clipboard for about brave page

luixxiul:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [ ] Ran `git rebase -i` to squash commits (if needed).

Closes #6183

Test Plan:
1. Visit about:about

Issue #6183, opened 2016-12-13
Regression : Alignment off for about:about


alexwykoff:
Only reproducible with the PR from:
https://github.com/brave/browser-laptop/issues/5790#issuecomment-265839135

**Did you search for similar issues before submitting this one?**
No

**Describe the issue you encountered:**
Visiting about:about 'List of about pages' is flush left with no margin.

**Expected behavior:**
'List of about pages' should left-align under 'About pages' header.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Tested on os x

- Brave Version:
0.12.15 RC2

- Steps to reproduce:
    1. Visit about:about

- Screenshot if needed:
<img width="367" alt="screen shot 2016-12-13 at 11 24 29 am" src="https://cloud.githubusercontent.com/assets/1434040/21151177/23707062-c127-11e6-83cd-266bf2ec74a5.png">

- Any related issues:

luixxiul:
https://github.com/brave/browser-laptop/commit/15976325924431b5fdbac96df72af5ef9ad16027#diff-1fcd5f7d89a2a38953d711aeb2e1dfeaL84 causes a regression.

Because `siteDetails.less` is used globally, actually the changes in the file with the commit should be reverted to set up a special file which is to be loaded only in about:brave.

https://github.com/brave/browser-laptop/commit/15976325924431b5fdbac96df72af5ef9ad16027#diff-1fcd5f7d89a2a38953d711aeb2e1dfeaR81 would therefore cause a regression too.

bbondy:
resetting milestone since the PR that introduced it was reverted:
https://github.com/brave/browser-laptop/pull/6107/files

luixxiul:
Also make sure every change in `history.less` would cause a regression because about:brave is based on that file.

luixxiul:
Test plan:

Open about:about

PR #6298, opened 2016-12-19
Redo: Add copy to clipboard for about brave page (fixes #5790)

### LINKED ISSUE #5790: about:brave page should have "copy to clipboard" button
I like the recent change to version info. It looks cleaner than presenting a dialog. One downside though is that selecting and copying the info gives me this:

Brave	0.12.10
Muon	1.4.26
libchromiumcontent	53.0.2785.143
V8	5.3.332.47
Node.js	6.5.0
Update Channel	dev
os.platform	linux
os.release	4.8.0-27-generic
os.arch	x64

It might be better to add a "copy to clipboard" button which provides something slightly easier to scan when pasted into issue tickets:

Brave: 0.12.10
Muon: 1.4.26
libchromiumcontent: 53.0.2785.143
V8: 5.3.332.47
Node.js: 6.5.0
Update Channel: dev
os.platform: linux
os.release: 4.8.0-27-generic
os.arch: x64

That's basically the text content format of the old dialog.

bsclifton:
Original PR was closed https://github.com/brave/browser-laptop/pull/6186

This PR fixes https://github.com/brave/browser-laptop/issues/6183

Does not include changes to about:brave, because that was re-committed with https://github.com/brave/browser-laptop/commit/5324f3c64449b2043ffce4a7d9ba86ece2f88591

Auditors: @luixxiul

Test Plan:
1. Visit about:about

### DIFF
--- a/js/about/brave.js
+++ b/js/about/brave.js
@@ -11,6 +11,7 @@ const aboutActions = require('./aboutActions')
 const ipc = window.chrome.ipcRenderer
 
 require('../../less/about/history.less')
+require('../../less/about/brave.less')
 require('../../node_modules/font-awesome/css/font-awesome.css')
 
 const tranformVersionInfoToString = (versionInformation) =>
--- a/less/about/brave.less
+++ b/less/about/brave.less
@@ -1,0 +1,16 @@
+@import "./common.less";
+
+.siteDetailsPage {
+  .siteDetailsPageContent {
+    .sectionTitle {
+      padding-left: 0;
+    }
+
+    .title {
+      width: 400px;
+      display: flex;
+      justify-content: space-between;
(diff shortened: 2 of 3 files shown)


luixxiul:
I was actually not sure to set property `.siteDetailsPage` in brave.less, as it might cause a regression. If it's fine, it's ok for me :)

bsclifton:
@luixxiul I can see the new less file possibly causing a regression in the future, but for now it's perfect

Long term, we'll want to move toward Aphrodite components and defining the styles inline, in the component. If you have time, you may check out this new style here :smile: 
https://github.com/brave/browser-laptop/blob/master/app/renderer/components/tabIcon.js

PR #6292, opened 2016-12-18
Fixed .urlbarIcon size change

luixxiul:
- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Closes #6270

Auditors: @srirambv @bsclifton /cc: @bradleyrichter

Test Plan:
1. Open about:preferences
2. Open http site in a new tab
3. Open https site in a new tab
4. Open a new tab and hit ":yt" in the url bar to display Youtube icon
5. Open a new tab and hit ":g" to display Google icon
6. Make sure .urlbarIcon does not change its size on each step

urlbaricon

PR #6319, opened 2016-12-20
Properly handle URL bar icon when searching and focus changes

### LINKED ISSUE #6273: lock icon is sticky in URL bar when it should display search icon
image

gyandeeps:

Fixes https://github.com/brave/browser-laptop/issues/6273

## Test Plan: 
- enable "always show URL bar", to make testing easier
- go to google.com and notice the lock in the omnibox
- start typing a search into a the omnibox and notice icon changes to magnifying glass
- click inside the webpage. icon should stay as magnifying glass and search terms should still be shown.  This confirms the fix (the bug before was that it would revert to showing the lock icon next to the search terms)
- click in the omnibox and press escape. This should reset icon back to lock and reset input back to google.com

### DIFF
--- a/app/renderer/components/urlBarIcon.js
+++ b/app/renderer/components/urlBarIcon.js
@@ -11,6 +11,19 @@ const dndData = require('../../../js/dndData')
 const {isSourceAboutUrl} = require('../../../js/lib/appUrlUtil')
 const searchIconSize = 16
 
+const getIconCssClass = (ctx) => {
+  if (ctx.isSearch) {
+    return 'fa-search'
+  } else if (ctx.isAboutPage && !ctx.props.titleMode) {
+    return 'fa-list'
+  } else if (ctx.isSecure) {
+    // NOTE: EV style not approved yet; see discussion at https://github.com/brave/browser-laptop/issues/791
+    return 'fa-lock'
+  } else if (ctx.inInsecure) {
+    return 'fa-unlock'
+  }
+}
+
 class UrlBarIcon extends ImmutableComponent {
   constructor () {
     super()
@@ -40,7 +53,7 @@ class UrlBarIcon extends ImmutableComponent {
    * - is a catch-all for: about pages, files, etc
    */
   get isSearch () {
-    const showSearch = this.props.active
+    const showSearch = this.props.isSearching && !this.props.titleMode
 
     const defaultToSearch = (!this.isSecure && !this.isInsecure && !showSearch) &&
                             !this.props.titleMode &&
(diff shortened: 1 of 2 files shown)


bradleyrichter:
@bsclifton Let's check and pull this one. Nice fix...thanks to @gyandeeps !

bsclifton:
awesome, thanks @gyandeeps :smile: I'll make sure we review this for our next release (we're a little backed up on PRs- hang in there with us :smile: )

app/renderer/components/urlBarIcon.js
```diff
@@ -11,6 +11,19 @@ const dndData = require('../../../js/dndData')
 const {isSourceAboutUrl} = require('../../../js/lib/appUrlUtil')
 const searchIconSize = 16
 
+const getIconCssClass = (ctx) => {
+  if (ctx.isSearch) {
```
luixxiul: I'd add `ctx.isAboutPage` to stop returining `fa-search fa-list`.

gyandeeps: I am not able to understand the use case here. Can you please explain the issue u r seeing?

luixxiul: ah sorry, I have pushed a PR to place the icon at the center of the input bar, and adding !ctx.isAboutPage avoids the CSS hack for `fa-search` to push 1px downward being applied to the url bar icon `fa-list` on about:preferences. https://github.com/brave/browser-laptop/pull/6292/files#diff-f5d8e85b88bbb4fdd8ebaf50901ebb3fR69

luixxiul: still because it's not the issue which this PR tries to resolve, you might skip it. I'll take care of it then

gyandeeps: k. But i dont think you need to make this change since now its if-else so only one class gets returned by this function. So you can never have a `fa-search fa-list` together.

luixxiul: I'm not quite sure how it does work but do you mean `ctx.isSearch && !ctx.isAboutPage` is not necessary? I'm thinking the case where you try to search something on the URL bar on about:preferences.

luixxiul: CC @bradleyrichter for this change

luixxiul: @gyandeeps I tested on the latest master and it looks good on about:prefenreces :-D

bsclifton:
Rebasing this one right now and then I'll check out the failing test :smile:

PR #6342, opened 2016-12-20
Update: Keep tabs same size when closing (fixes #6088)

### LINKED ISSUE #6088: Keep tabs same size when closing
https://community.brave.com/t/keep-tabs-same-size-when-closing/74?u=suguru

> When I have a lot of tabs open in Chrome and go to close them, the tabs stay the same size. That way I can click and close a lot of tabs one after the other without moving my mouse. But in Brave, the tabs resize to fit the window, which means, I have to move my mouse to close another tab.

gyandeeps:
#6088

Same behavior as chrome

Test Plan:

1. open 7-8 tabs (donest matter what the content are)
1. move your cursor to the close button of the second tab from left
1. Now as u keep clicking it will maintain the size of the tabs
1. The size will reset as soon as you move your cursor outside the tab area.

After the change:

tabs


js/components/tabs.js
```diff
@@ -27,6 +27,25 @@ class Tabs extends ImmutableComponent {
...
+    this.state = {
+      fixTabWidth: null
+    }
+  }
+
+  onMouseLeave () {
+    this.setState({
+      fixTabWidth: null
+    })
+    this.forceUpdate()
```
cezaraugusto: We should avoid using `forceUpdate()` to force re-render, it's ill advised by React itself. 

Ref: https://facebook.github.io/react/docs/react-component.html#forceupdate

> Normally you should try to avoid all uses of forceUpdate() and only read from this.props and this.state in render().

Could you do it in a way to avoid that? thanks!

gyandeeps: I have to use forceUpdate here because we inherit from `ImmutableComponent` and not `ReactComponent`. `ImmutableComponent` checks for props only to make the render update work.
forcedUpdate has been used in other places also (possibly because of the above reason).

bsclifton: @gyandeeps the ideal approach would be flux-like and to do the following:
- store the state in our AppStore / AppState. You can [view the current layout of the appState here](https://github.com/brave/browser-laptop/blob/master/docs/state.md)
- Inside `onFrameClose`, create and call a new [appAction](https://github.com/brave/browser-laptop/blob/master/js/actions/appActions.js) which has should also have an event based name.  `onFrameClose` seems OK
- The action then (like other actions) makes a call to dispatcher. This gets handled in [appStore.js](https://github.com/brave/browser-laptop/blob/master/js/stores/appStore.js#L376)
- AppStore then updates the application state
- `main.js` should be binding the appState to this control. [Here's an example with a different UI control](https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L1169)

Using the immutable control approach is ideal because components themself don't have state; everything goes through the app store.  Even if we insisted on keeping state, it would be better to just change the control to inherit from `React.Component`

gyandeeps: @bsclifton I do understand your approach. I dint do that since this tab width change s such a small state change. It has no impact on the other areas of the application.

Will make the change as you suggested.

gyandeeps: @bsclifton Here we need to change the width based on 2 different events. One is when the frame is closed and one onMouseLeave. Any suggestions on what kind of event name you would prefer for latter.

bsclifton: @gyandeeps I think onMouseLeave is appropriate :smile:

gyandeeps: Why I asked was because `onMouseLeave` event is very generic looking from a appState perspective. 
I can make it specific to tab like `onMouseLeaveTab`, etc

bsclifton: @gyandeeps ahh- ok gotcha... so yes- I would say an event name which is specific is great.  So perhaps `onTabClosed`?

gyandeeps: based on our slack chat, will use `onTabClose` and `onTabMouseLeave` as event names.

gyandeeps:
@cezaraugusto @bsclifton I have made changes as suggested. Please review the changes and let me know what u guys think.

bsclifton:
For some weird reason, I'm not seeing a CI build for this? I'm going to try to submit one manually

bsclifton:
Closing/re-opening triggered CI build :smile: If tests look good, I'll merge!

bsclifton:
Re-ran the [tests which failed with CI](https://travis-ci.org/brave/browser-laptop/builds/194636747) and all passed except search suggestions (has been failing for a while) and preferences component (the SVG I checked in with https://github.com/brave/browser-laptop/pull/6689 broke this; I'm looking at it now)

Thanks again, @gyandeeps! Merging...

Issue #6306, opened 2016-12-20
navigator.plugins is not handled correctly for Flash


bridiver:
navigator.plugins should be suppressed by default
exceptions:
navigator.plugins will always be available to the adobe flash installer page so it can correctly detect the current installation state
some sites detect flash through navigator.plugins and replace the embed with a link to the adobe installer page. Those sites will have a special permission that provides access to navigator.plugins if the user has allowed it

If flash is installed and enabled, links to the adobe flash installer page should trigger a permission dialog to allow flash for the site. If the user allows access the page will reload and navigator.plugins will contain the plugin info for Flash
If flash is installed and not enabled a request to enable flash should be displayed with an option to proceed to the flash installer page instead

bridiver:
fixed https://github.com/brave/browser-laptop/commit/0726ae0644624e5222dbb166351b2b7dd093844e

bridiver:
Test Plan:
1. Install and enable flash
2. Go to Pandora.com
3. allow flash
page should reload and enable flash

PR #6354, opened 2016-12-21
handle sites that check navigator.plugins

bridiver:
allow adobe to auto-detect install
todo - add flash enable message box to adobe install page when flash is installed
todo - reload page after enabling flash for pages that detect navigation.plugins
partial fix for #6306

Test Plan:
1. Go to Pandora.com
2. Click on adobe flash link
3. Install flash
4. Enable flash
5. Go to Pandora.com
6. Click on adobe flash link
7. Message box should open
8. Deny
9. Reload Pandora.com
10. Click on adobe flash link
11. Allow
12. Reload Pandora.com 

Flash should be enabled

cc @diracdeltas


diracdeltas:
it should auto-reload when clicking 'allow' if possible. however, when i manually reload, Flash doesn't seem to load.

bridiver:
@diracdeltas for the navigator.plugins sites? That's listed in the todo

bridiver:
but this PR isn't meant to cover that case so I think we can still merge unless there are other problems and take care of the rest in a follow up

diracdeltas:
@bridiver i'm trying this out on pandora. when i click 'remember' and reload manually it still doesn't seem to load flash

diracdeltas:
i do see the flash allowed site setting in about:preferences#security though

bridiver:
I think both issues are from using prebuilt instead of electron master. Let me know if you see the same thing after updating

diracdeltas:
merging for now since code lgtm

PR #6368, opened 2016-12-21
Fix broken array concat behavior during app relaunch

bsclifton:
The + turns it from an array into a string on Windows (new node behavior?). Definitely wrong.

This fix is to help with https://github.com/brave/browser-laptop/issues/5938.  Tested on Mac and Windows.  Please note: it doesn't fix the issue, but it does correct the wrong behavior

Fixes https://github.com/brave/browser-laptop/issues/6369

Auditors: @bbondy, @bridiver

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -359,7 +359,6 @@ const setVersionInformation = (data) => {
       {name: 'os.platform', version: os.platform()},
       {name: 'os.release', version: os.release()},
       {name: 'os.arch', version: os.arch()}
-      // TODO(bsclifton): read the latest commit hash from a file, etc.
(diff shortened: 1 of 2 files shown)


bbondy:
++

app/sessionStore.js
```diff
@@ -359,7 +359,6 @@ const setVersionInformation = (data) => {
       {name: 'os.platform', version: os.platform()},
       {name: 'os.release', version: os.release()},
       {name: 'os.arch', version: os.arch()}
-      // TODO(bsclifton): read the latest commit hash from a file, etc.
```
bbondy: 💃

PR #6377, opened 2016-12-21
Order tabs correctly when they have openers

bbondy:


test/unit/state/frameStateUtilTest.js
```diff
@@ -17,92 +17,6 @@ describe('frameStateUtil', function () {
     this.windowState = Immutable.fromJS(Object.assign({}, defaultWindowStore.toJS()))
   })
 
-  describe('cloneFrame', function () {
-    before(function () {
-      this.frameOpts = {
-        key: 1,
-        tabId: 2,
-        parentFrameKey: 3,
-        audioMuted: true,
```
bridiver: I think this is the only one we need to handle and we can do that in electron -  https://github.com/brave/electron/issues/126

test/unit/state/frameStateUtilTest.js
```diff
@@ -17,92 +17,6 @@ describe('frameStateUtil', function () {
...
-        canGoBack: true,
-        canGoForward: true,
-        icon: 'icon',
-        title: 'title',
-        isPrivate: true,
-        partitionNumber: 4,
-        themeColor: 'ffffff',
-        computedThemeColor: 'aaaaaa',
-        history: ['http://brave.com'],
```
bridiver: do we still need this now that we have history methods on the webview/webcontents? Are we still using it anywhere?

bridiver:
++

test/unit/state/frameStateUtilTest.js
bbondy: Possibly we're using it in about: pages.

bridiver: can we replace that with tab/webview history now?

Issue #6371, opened 2016-12-21
Copy version information tooltip is incorrect/not localized


Liunkae:
**Did you search for similar issues before submitting this one?**
Yes.

**Describe the issue you encountered:**
On about:brave, the button to copy the version information has an incorrect tooltip. The tooltip currently says "Copy password to clipboard".

**Expected behavior:**
The tooltip should say something like "Copy version information to clipboard".

- Platform (Win7, 8, 10? macOS? Linux distro?):
All (found on macOS 10.12.x)

- Brave Version (revision SHA):
Master (3c39b44)

- Steps to reproduce:
    1. Open a new Brave tab and navigate to about:brave
    2. Mouse over the copy to clipboard button.

- Screenshot if needed:
![image](https://cloud.githubusercontent.com/assets/19424103/21402010/34107c90-c77c-11e6-9067-3c54954b2cba.png)

- Any related issues:
None

Liunkae:
https://github.com/brave/browser-laptop/blob/d9824b34f0dc0d190dacf1faac88e133be7d056f/js/about/brave.js#L46

Also, it is not localized.

Liunkae:
I can't get `copyToClipboard` to work. And `copyToClipboard` is listed in `app/locale.js`. Many other keys on that list do work, though. 😕 
````js
const locale = require('../l10n')
...
<span className='fa fa-clipboard' title={locale.translation('copyToClipboard')} onClick={this.onCopy} />
````

luixxiul:
@bbondy do you think this is the same as #3946?

bbondy:
yes I think so, I think we need to implement the l10n extensions API to get the localization here

Liunkae:
I submitted PR #6392 for a temporary fix that changes the wording to "Copy to clipboard", until we get localization to work.

bsclifton:
@Liunkae I put a note about how you can fix the issue here:
https://github.com/brave/browser-laptop/pull/6392/files#r93702690

You can see js/about/history.js for examples (except that one uses it's own file, versus using app.properties)

Liunkae:
PR: #6414

luixxiul:
Test plan: https://github.com/brave/browser-laptop/pull/6414#issue-197425848

PR #6392, opened 2016-12-22
Fix copy version info tooltip

Liunkae:
Temporary fix until we get l10n to work (see #6371)

### DIFF
--- a/js/about/brave.js
+++ b/js/about/brave.js
@@ -43,7 +43,7 @@ class AboutBrave extends React.Component {
       <div className='siteDetailsPageContent aboutAbout'>
         <div className='title'>
           <span className='sectionTitle' data-l10n-id='versionInformation' />
-          <span className='fa fa-clipboard' title='Copy password to clipboard' onClick={this.onCopy} />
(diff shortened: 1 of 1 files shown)


js/about/brave.js
```diff
@@ -43,7 +43,7 @@ class AboutBrave extends React.Component {
       <div className='siteDetailsPageContent aboutAbout'>
         <div className='title'>
           <span className='sectionTitle' data-l10n-id='versionInformation' />
-          <span className='fa fa-clipboard' title='Copy password to clipboard' onClick={this.onCopy} />
```
bbondy: Something like this should work, but maybe another fix is needed for passing attributes to about pages, I'll merge this in the meantime either way though.
copyToClipboard.title=Copy to clipboard

bsclifton: Two things ( cc: @Liunkae ):
1. We're missing `data-l10n-id` for this item. We need it to look like this:
`<span data-l10n-id='copyToClipboard' className='fa fa-clipboard' onClick={this.onCopy} />`
2. When you look at about-brave.html, you'll see that for localization it uses `app.properties`.  Inside this file, you'll need to be having something like this: `copyToClipboard.title=Copy to clipboard`

PR #6410, opened 2016-12-23
Get each version in a more safe way; if one field fails, it won't break the whole thing

bsclifton:

Get each version in a more safe way; if one field fails, it won't break the whole thing

Fixes https://github.com/brave/browser-laptop/issues/6406

Auditors: @bbondy

Test Plan:
1. Remove existing Brave and deleting existing brave-development folder from %appdata%
2. Build from master
3. Open about:brave, verify info is shown; rev may or may not be blank

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -342,30 +342,48 @@ module.exports.cleanSessionDataOnShutdown = () => {
   return p
 }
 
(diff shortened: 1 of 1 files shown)


bbondy:
++ but maybe it's something relating to not getting ipc events or something like that. I guess we will see in the future.

PR #6452, opened 2016-12-29
Added arrow navigation to the autofill

### LINKED ISSUE #1302: Autofill: can't select popup item using arrow keys
STR: begin typing a login that will be autofilled, immediately a popup will appear below the login form field with the suggested autofill. One can click on this suggested autofill, but I can't down-arrow and press enter.

Expected: that down-arrow will select the suggested autofill, and enter key will enter it in the associated form field

NejcZdovc:

### Description
Resolves #1302

### Auditors 
@bsclifton

### Test Plan
- go to website with a form
- submit form so that autofill will be triggered
- go to the same website and double click on the input to trigger autofill
- use down arrow to select value and press enter
- value should be field into the input

## Automated tests
- `npm run watch-test`
- `npm run uitest -- --grep="ContextMenu"`

### DIFF
--- a/app/renderer/components/menubar.js
+++ b/app/renderer/components/menubar.js
@@ -53,22 +53,19 @@ class MenubarItem extends ImmutableComponent {
     }
     // If clicking on an already selected item, deselect it
     const selected = this.props.menubar.props.selectedIndex
-      ? this.props.menubar.props.selectedIndex[0]
-      : null
     if (selected && selected === this.props.index) {
       windowActions.setContextMenuDetail()
-      windowActions.setSubmenuSelectedIndex()
+      windowActions.setMenuBarSelectedIndex()
       return
     }
     // Otherwise, mark item as selected and show its context menu
-    windowActions.setSubmenuSelectedIndex([this.props.index])
+    windowActions.setMenuBarSelectedIndex(this.props.index)
+    windowActions.setContextMenuSelectedIndex([0])
     const rect = e.target.getBoundingClientRect()
     showContextMenu(rect, this.props.submenu, this.props.lastFocusedSelector)
   }
   onMouseOver (e) {
     const selected = this.props.menubar.props.selectedIndex
-      ? this.props.menubar.props.selectedIndex[0]
-      : null
     if (typeof selected === 'number' && selected !== this.props.index) {
       this.onClick(e)
     }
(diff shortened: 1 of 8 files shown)


NejcZdovc:
@bsclifton Me again :smiley: This PR is still WIP, but if you can take a look I would appreciate it.

1. What do you think about my approach for this issue? Should I use action and dispatch it when you press keydown and then modify selectedIndex in the store?
2. I have a problem with an enter keydown event. Is it possible to completely disable event propagation, so that pages will not receive this enter event. I was testing this on a form and when I pressed enter, menu item is selected, but you also triggered form validator. Current implementation does not stop form validation:
```
e.preventDefault()
e.stopPropagation()
```

NejcZdovc:
@bsclifton More a look at the code, more I think that I should be using store here :smile:

NejcZdovc:
Number 2 is solved.

bsclifton:
@NejcZdovc awesome, thanks for sticking with it :smile: This change will make a huge impact.  Apologies I haven't been of any use- I've been working through issues labelled with 0.13.0 milestone

bsclifton:
@NejcZdovc regarding number 2, e.preventDefault() is what you need and you did find it :smile: 

Regarding the approach, I really like it! Great job finding some of the keyboard code I had introduced when doing the Windows titlebar project (on Windows, you can hit alt to show/hide the menu and navigate everything using keyboard).  This re-uses that code which is perfect

As far as what to do when ENTER is hit (what approach to take there), you can look at the `Menubar` control that I wrote.  If you hit enter and the menu is open, menubar.js is expecting to process it (something we'll have to make sure doesn't get broken on Windows).
https://github.com/brave/browser-laptop/blob/master/app/renderer/components/menubar.js#L196

NejcZdovc:
@bsclifton thank you for your help, but maybe I wasn't clear enough what my problem was in point 1. When you press `enter` I reused code that was already implemented if you clicked on the item in menu. So this is already working. Basically everything is working. What I was wondering is if I would, rather then having local state `this.currentSelectedIndex` (this is the current implementation), use store and dispatch `selectedIndex` when you use arrows up/down that is used [here](https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L828).

bsclifton:
@NejcZdovc everything works great! I tested earlier today, awesome job :smile: 

Your suggestion of using the store rather than local state is actually our ideal solution :smile: Please do implement it this way (using appStore) and then the arrow up/down can be an action.  This may be the only situation using the index (for now), but we always want state coming from the store.

Please be sure to update the [`state.md`](https://github.com/brave/browser-laptop/blob/master/docs/state.md) to show your new session variable.  We should be able to sneak this into our 0.13.0 release if you are able to rework the session like this :smile:

NejcZdovc:
@bsclifton Ok, now I moved all logic to the store and added tests. Please review. After review I will squash all commits into one.

bsclifton:
@NejcZdovc excellent job with the tests! I ran locally and they work perfectly

There are some problems with this though, which may not be obvious unless you are on Windows.  Like I mentioned above, this code is also used by the menubar.  There is a very easy way for you to test this for yourself :smile: 

## intro to menubar component

Start by opening up [`js/components/main.js` around line 825](https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L825).  You can change:
`const customTitlebarEnabled = isWindows`
to
`const customTitlebarEnabled = true`

Now, when you launch the app, it'll behave like it does on Windows.  You can press ALT and the menu will show up:
screen shot 2017-01-04 at 11 43 35 pm

The menu should be mirroring the system menu 100% (assuming you're on Mac or Ubuntu).  If you're on another variant of Linux, you might see two menus! The system menu will also pop up when you press ALT and it'll look confusing.

## how is the selection tracked?

Anyways- the changes you made modify the index.  This index is multi-dimensional.  Let's say that you pick an option like so:
screen shot 2017-01-04 at 11 47 01 pm
Here, I pick `Bookmarks > Car > EPC-net online`.  This index is represented as:
[5, 5, 0]

5 = index of Bookmarks menu
5 = index within bookmarks menu of the item selected ("Car" bookmark folder)
0 = index within "Car" bookmark folder

## deeper dive into menubar component
The menubar component also has a keydown/keyup handler which I believe takes priority.  Prior to your change, the handler in ContextMenu was only being used in this component to capture the ESC key being pressed (which would close the context menu).

With the introduction of this new change, there is now a conflict where both handlers can fire.  Additionally, the primary index for the selection is ALWAYS being set to 0, which breaks the menu component.
ex: `windowActions.setSubmenuSelectedIndex([*0*, currentIndex + 1])`

## recommended solution
I believe the solution should be fairly easy actually (you're really close!).  I would recommend creating a new variable in the AppStore (which you can document in the state.md).  This would ONLY be used to store the index of the context menu **when a menu is not present**.

The next step you can take: bind the appropriate variable (`customTitlebar.menubarVisible`) in main.js to the ContextMenu control, so that it propagates down to the ContextMenu component.  Then, in the ContextMenu component you can check if this menu is visible.  If so, do not process keydown events (with the exception of ESC).

If menu is NOT visible, then you can process up/down arrow keys and use a state which is not related to the menubar

## conclusion
This was a lot of text and I am going to make an effort to transfer the knowledge to our wiki :smile: 

If you're up for the challenge, please go for it :smile: If the remaining tasks aren't clear or if you'd prefer, I can make some of the changes.  Let me know what you think :smile:

NejcZdovc:
@bsclifton will fix it, helpful as always :) I noticed that there is another dimension there, but I just couldn't find it. I am working on mac yes

NejcZdovc:
Will add top menu to this test's as well, so that we can catch this problems in the future

bsclifton:
@NejcZdovc I can't explain how big the smile on my face is right now :smile: I'm using the above as a stub and taking this time to fully document the Windows titlebar / menu.

One significant advantage that the custom menus offer is that they are testable with webdriver (the system menus are not).  You may need to expose functionality to enable the menu programmatically on a non-Windows system- I would love to help you with this!

luixxiul:
@NejcZdovc Is this ready for review?

bsclifton:
@luixxiul yup- I'm in process of re-reviewing (he had hit me up over Slack)

app/renderer/components/menubar.js
```diff
@@ -102,193 +100,69 @@ class Menubar extends ImmutableComponent {
...
-          windowActions.setSubmenuSelectedIndex(newIndices)
-
-          let openedSubmenuDetails = this.props.contextMenuDetail.get('openedSubmenuDetails')
-            ? this.props.contextMenuDetail.get('openedSubmenuDetails')
-            : new Immutable.List()
-          openedSubmenuDetails = openedSubmenuDetails.pop()
-
-          windowActions.setContextMenuDetail(this.props.contextMenuDetail.set('openedSubmenuDetails', openedSubmenuDetails))
-          break
+        const currentContextMenuItem = windowStore.getState().getIn(['ui', 'contextMenu', 'selectedIndex'])
```
bsclifton: Instead of getting like this, could you remove the reference to windowStore and then bind currentContextMenuItem as a property in `main.js`?
https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L938

app/renderer/components/menubar.js
```diff
@@ -102,193 +100,69 @@ class Menubar extends ImmutableComponent {
...
-            const newIndices = selectedIndex.slice()
-            if (this.hasMenuSelection) {
-              newIndices[selectedIndex.length - 1] = nextIndex
-            } else {
-              newIndices.push(0)
-            }
-            windowActions.setSubmenuSelectedIndex(newIndices)
-          }
+      case keyCodes.ENTER:
+        const contextMenuIndex = windowStore.getState().getIn(['ui', 'contextMenu', 'selectedIndex'])
```
bsclifton: same as comment above; would be good to remove reference to windowStore and then bind it using a property in `main.js`

docs/windowActions.md
```diff
@@ -855,10 +855,23 @@ Used by `main.js` when click happens on content area (not on a link or react con
...
+**index**: `number`, zero based index of the item.
+  Index excludes menu separators and hidden items.
+
+
+
+### setContextMenuSelectedIndex(index) 
+
+(Windows only)
+Used to track selected index of a menu bar
 Needed because arrow keys can be used to navigate the custom menu
```
bsclifton: Thanks for updating this :+1:

js/components/main.js
```diff
@@ -822,19 +822,17 @@ class Main extends ImmutableComponent {
   }
 
   get customTitlebar () {
-    const customTitlebarEnabled = isWindows
+    const customTitlebarEnabled = true // isWindows TODO revert before merge
     const captionButtonsVisible = customTitlebarEnabled
     const menubarVisible = customTitlebarEnabled && (!getSetting(settings.AUTO_HIDE_MENU) || this.props.windowState.getIn(['ui', 'menubar', 'isVisible']))
-    const selectedIndex = this.props.windowState.getIn(['ui', 'menubar', 'selectedIndex'])
+    const tempContextMenuIndex = this.props.windowState.getIn(['ui', 'contextMenu', 'selectedIndex'])
```
bsclifton: Regarding the previous comments, it looks like you already had it in here :smile: Any reason you chose to use the version from WindowStore over this one?  It's important to rely on this version, so that the control re-renders when the store changes

js/components/main.js
```diff
@@ -822,19 +822,17 @@ class Main extends ImmutableComponent {
...
       enabled: customTitlebarEnabled,
       captionButtonsVisible: captionButtonsVisible,
       menubarVisible: menubarVisible,
       menubarTemplate: menubarVisible ? this.props.appState.getIn(['menu', 'template']) : null,
-      menubarSelectedIndex: selectedIndex,
-      contextMenuSelectedIndex: typeof selectedIndex === 'object' && Array.isArray(selectedIndex) && selectedIndex.length > 1
-        ? selectedIndex.slice(1)
-        : null,
+      menubarSelectedIndex: this.props.windowState.getIn(['ui', 'menubar', 'selectedIndex']),
+      contextMenuSelectedIndex: tempContextMenuIndex !== null ? tempContextMenuIndex : [0],
```
bsclifton: I believe it would be useful to keep the previous logic here for the check (because the index should always be an array with length > 0).  The new default to `[0]` make perfect sense, because this is what we'd like for when autofill pops up :smile:

js/stores/windowStore.js
```diff
@@ -898,14 +898,18 @@ const doAction = (action) => {
...
-        ? action.index
-        : null)
+    case windowConstants.WINDOW_SET_MENUBAR_SELECTED_INDEX:
+      windowState = windowState.setIn(['ui', 'menubar', 'selectedIndex'], action.index)
+      break
+    case windowConstants.WINDOW_SET_CONTEXT_MENU_SELECTED_INDEX:
+      windowState = windowState.setIn(['ui', 'contextMenu', 'selectedIndex'],
+          Array.isArray(action.index)
+          ? action.index
+          : null)
```
bsclifton: ~I believe you have the logic for these two items mixed up; since `WINDOW_SET_SUBMENU_SELECTED_INDEX` was changed to `WINDOW_SET_MENUBAR_SELECTED_INDEX`, the array check should be there (which then also has a default to null).  You may also keep the check in for `WINDOW_SET_CONTEXT_MENU_SELECTED_INDEX`, but instead defaulting to `[0]` like you did above in `main.js`~

Disregard my comment above (striked out).  I now see that MenubarSelectedIndex is always having an integer value; If you'd like to keep type checking in place, you can instead maybe compare it against typeof 'number' and make sure it's greater than 0

Regarding the check for `WINDOW_SET_CONTEXT_MENU_SELECTED_INDEX`, you may want to default to `[0]` like you do above in `main.js` (instead of null)

test/components/contextMenuTest.js
```diff
@@ -0,0 +1,106 @@
...
+        .pause(10)
+        .keys('\uE015')
+        .pause(10)
+        .keys('\uE015')
+        .pause(10)
+        .keys('\uE015')
+        .pause(10)
+        .keys('\uE013')
+        .pause(10)
+        .keys('\uE007')
```
bsclifton: Love these tests :smile:

js/components/main.js
```diff
@@ -822,19 +822,17 @@ class Main extends ImmutableComponent {
   }
 
   get customTitlebar () {
-    const customTitlebarEnabled = isWindows
+    const customTitlebarEnabled = true // isWindows TODO revert before merge
```
bsclifton: don't forget to remove this :stuck_out_tongue:

js/components/contextMenu.js
```diff
@@ -252,12 +237,165 @@ class ContextMenuSingle extends ImmutableComponent {
...
+        windowActions.resetMenuState()
+        break
+
+      case KeyCodes.ESC:
+      case KeyCodes.TAB:
+        windowActions.resetMenuState()
+        break
+
+      case keyCodes.LEFT:
+        if (e.which === keyCodes.LEFT && this.hasSubmenuSelection) {
```
bsclifton: because the code is complex, if you could keep the original comments in there, it would be much appreciated :smile: For example, this one had:

```
// Left arrow inside a submenu
// <= go back one level
```

bsclifton:
I tested this out on Windows and it works wonderfully :smile: Besides leaving feedback, I also added @darkdh to the review, since he implemented most of the autofill functionality :smile: 

@darkdh- lots of the code for menu navigation (for Windows) was able to be re-used. I'm curious what you think

app/renderer/components/menubar.js
NejcZdovc: Done

app/renderer/components/menubar.js
NejcZdovc: Done

js/components/main.js
NejcZdovc: :smile: that's why I added TODO :stuck_out_tongue:

js/components/main.js
NejcZdovc: Honestly I forgot about it, it's fixed now

js/components/main.js
NejcZdovc: Done

js/components/contextMenu.js
NejcZdovc: Added

js/stores/windowStore.js
NejcZdovc: We need null here, so that other component's (menu bar) know what status we have. I moved default `[0]` to contextMenu.

test/components/contextMenuTest.js
NejcZdovc: Currently some test's are failing, still investigating why this is happening. If I do test manually (added longer pause) with my mouse and keyboard, test is successful.

NejcZdovc: Ok, the only thing that is now not working, and that's why test are failing, is enter. Enter is not prevented and that's why form is refreshed and test fails. Although it's strange, because event is prevented in contextMenu. I added test that check's if enter is prevented or not.

js/components/contextMenu.js
```diff
@@ -257,54 +253,149 @@ class ContextMenu extends ImmutableComponent {
...
+          openedSubmenuDetails = openedSubmenuDetails.pop()
+
+          windowActions.setContextMenuDetail(this.props.contextMenuDetail.set('openedSubmenuDetails', openedSubmenuDetails))
+        }
+        break
+
+      case keyCodes.RIGHT:
+        // Right arrow on a menu item which has a sub menu
+        // => go up one level (default next menu to item 0)
+        if (e.which === keyCodes.RIGHT && selectedMenuItem.has('submenu')) {
```
darkdh: We should check `!!selectedMenuItem.get('submenu')` here. You can press right at the end of submenu to see what happened. It will be `undefined` on popup menu and `null` on menubar

NejcZdovc: Added

NejcZdovc:
@darkdh Funny thing is that before rework (in my first commit in this PR) enter was prevented, but the code for this is basically the same. Regarding the test  `right press at the end of submenu` should I add it in this PR?

NejcZdovc:
@bsclifton & @darkdh I fixed event prevent for enter :tada: Can you please review again? I don't have any more pending things to do

NejcZdovc:
@bsclifton Commits are now squashed. I also removed true for isWindows

js/components/contextMenu.js
```diff
@@ -252,13 +237,165 @@ class ContextMenuSingle extends ImmutableComponent {
...
+      return true
+    }).size - 1
+  }
+
   get openedSubmenuDetails () {
     return this.props.contextMenuDetail.get('openedSubmenuDetails') || new Immutable.List()
   }
+
+  get hasSubmenuSelection () {
+    return this.props.selectedIndex.length > 1
```
darkdh: I think this line should be `(this.props.selectedIndex === null) ? false : this.props.selectedIndex.length > 1`
There is a chance `this.props.selectedIndex` will be `null` regarding https://github.com/brave/browser-laptop/pull/6452/files#diff-195859572336a7940350e808382f7fe8R398
STR:
9co43ds6ku

NejcZdovc:
@darkdh I can add test, but for this specific PR (autofill) we don't have submenu menu. I can add test, that will check if everything is ok when you press right/left on non submenu item.

js/components/contextMenu.js
NejcZdovc: Fixed

NejcZdovc:
@bsclifton @darkdh Test was added. Can you please check it again?

bsclifton:
Because 0.13.0 is already large and has been delayed, I'm choosing to merge this into 0.13.1 :smile: 

Great work, @NejcZdovc! :smile: [CI has passed](https://travis-ci.org/brave/browser-laptop/builds/190185876), merging!

bsclifton:
Pulled into 0.13.0 😄

Issue #6524, opened 2017-01-04
some certificate fingerprints cause decoding error


diracdeltas:
reported by eric lawrence:
![image uploaded from ios](https://cloud.githubusercontent.com/assets/549654/21656481/31fcded6-d2b6-11e6-8b1a-a741861a3c36.jpg)

bsclifton:
@diracdeltas is this fixed by https://github.com/brave/browser-laptop/pull/6525?

diracdeltas:
i'm not 100% sure due to lack of the test certificate, but we can close this and reopen if not.

bsclifton:
Closing; if not fixed by https://github.com/brave/browser-laptop/pull/6525, let's re-open :smile:

bsclifton:
was gonna add the new `QA/required` tag but you beat me to it, @luixxiul :smile:

bbondy:
Test failed on macOS, I posed a follow up here:
https://github.com/brave/browser-laptop/issues/7042

@diracdeltas could you take a look?

PR #6525, opened 2017-01-04
Handle cert fingerprints that contain '/'

diracdeltas:
The cert fingerprint emitted by Muon/Electron is base64 encoded [1], so it may contain '/'. I think we were truncating the certificate fingerprint when this was the case, which may have caused the decoding error in #6524.

[1] https://cs.chromium.org/chromium/src/net/base/hash_value.h?q=net::HashValue&sq=package:chromium&l=58

Auditors: @darkdh 

Test Plan:
I don't have an example of a cert with a fingerprint that contains `/`, unfortunately, and Eric Lawrence did not provide his bug-triggering cert. At least to do a regression test:
1. go to a site like https://wrong.host.badssl.com/ with an invalid ssl cert
2. click 'view certificate'
3. the hex-encoded certificate fingerprint should appear

### DIFF
--- a/js/about/certerror.js
+++ b/js/about/certerror.js
@@ -57,6 +57,8 @@ class CertErrorPage extends React.Component {
       var validExpiry = new Date()
       validStart.setTime(detail.validStart * 1000)
       validExpiry.setTime(detail.validExpiry * 1000)
+      var fingerprint = detail.fingerprint.split('/')
+      var algorithm = fingerprint.shift()
       this.setState({
         certDetail: true,
         certIssuerName: detail.issuerName,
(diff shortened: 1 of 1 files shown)


diracdeltas:
^ pinging @ericlaw1979

ericlaw1979:
> Can we ask Eric Lawrence to provide the url of the site whose certificate contains "/" fingerprint in base64? Just wanna record it in this PR.

This wasn't a live site, but instead a self-generated certificate (like one a bad guy could generate for his own server). 

My repro used Fiddler. Enable HTTPS decryption, then click 
Rules > Customize Rules > OnBeforeRequest. Inside OnBeforeRequest, set the desired CN value, e.g.

`static function OnBeforeRequest(oSession: Session) {		
		oSession["X-overridecertCN"]= "<b>alert(1)";
`	

Save the file, then visit any HTTPS site to have the self-generated certificate with the bad CN value returned to the client.

bsclifton:
[Last travis-ci run](https://travis-ci.org/brave/browser-laptop/builds/188973528) had an error; I re-ran this test locally and it's working as expected :smile: Merging...

PR #6548, opened 2017-01-05
make travis run tests

diracdeltas:
and add some test coverage for siteUtil when sync is enabled

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 script:
-  - npm run lint && npm test
+  - npm run lint && npm run download-sync-client && npm test
 notifications:
   slack:
     secure: bDwO2uce5JAZvjrvWj4+/+yEXJAIK4O0RcgUWvZ2IMbi7Q9I89Mw40JmkLWL6x2gWZwxr8+FoLtErJA7RVrsfImjrX+NmMyAB7AydLdrBJtkLozNnuacnhcnBRyp1gGCa1ymxCEXGbgC6onAD3kiJJhggr70T+2lu3IuJYXENhc=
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 script:
-  - npm run lint && npm test
+  - npm run lint && npm run download-sync-client && npm test
```
posix4e: Why are these connected via && instead of multiple - lines?

tools/downloadSyncClient.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
posix4e: I actually think the new normal is to name things tools/downloadSyncClient not tools/downloadSyncClient.sh unless you really are a .sh script (not bash)

tools/downloadSyncClient.sh
```diff
@@ -0,0 +1,9 @@
...
+
+# stop on errors (nonzero exit codes), uninitialized vars
+set -eu
+
+SYNC_PATH=app/extensions/brave/content/scripts/sync.js
+SYNC_URL=https://raw.githubusercontent.com/brave/sync-bundles/master/bundle.js
+
+curl -o $SYNC_PATH $SYNC_URL
```
ayumi: I added bundle.js and testbundle.js from brave/sync-bundles as the first release of brave/sync.

Download latest prod (non pre-release) version:
```sh
curl -O $(curl -s https://api.github.com/repos/brave/sync/releases/latest | grep \/bundle.js | head -n 1 | cut -d '"' -f 4)
```

.travis.yml
diracdeltas: dunno

diracdeltas:
made requested changes

diracdeltas:
travis still seems broken for reasons unrelated to this change, looking at that now

diracdeltas:
fixing the other travis error now

diracdeltas:
travis is now actually running tests

diracdeltas:
travis then stalled on the tests and gave up, but merging for now since it's some progress

PR #6570, opened 2017-01-07
Close tab page option

### LINKED ISSUE #5489: Add context-menu option to close tabs in tab page
I often find I don't need to use a tab page at all anymore and want to close the whole group, I end up right clicking on the tab page indicator, but then I don't see a close tab page option.

NejcZdovc:

## Description
Resolves #5489

## Auditors
@bbondy @bsclifton

## Test Plan
- open two tab pages
- right click on one and click close tab page
- all tabs in tab pages should be closed


js/actions/windowActions.js
```diff
@@ -697,6 +697,19 @@ const windowActions = {
   },
 
   /**
+   * Dispatches a close call to all frames in tab page
+   * The provided frame will be closed.
+   *
+   * @param {Object} framesList - List of frames
+   * @param {Object} framePropsList - List of frame properties to consider
+   */
+  closeTabPage: function (framesList, framePropsList) {
```
bbondy: Could you name this tabPageClosed?

NejcZdovc: Done

app/extensions/brave/locales/en-US/menu.properties
```diff
@@ -13,6 +13,7 @@ openLocation=Open Location…
 openSearch=Search for "{{selectedVariable}}"
 importFrom=Import from…
 closeWindow=Close Window
+closeTabPage=Close tab page
```
luixxiul: how about renaming this into `tabPageClosed` as well?

app/locale.js
```diff
@@ -76,6 +76,7 @@ var rendererIdentifiers = function () {
     'closeOtherTabs',
     'closeTabsToRight',
     'closeTabsToLeft',
+    'closeTabPage',
```
luixxiul: ditto

js/contextMenus.js
```diff
@@ -88,6 +88,11 @@ function tabPageTemplateInit (framePropsList) {
     click: (item, focusedWindow) => {
       windowActions.muteAllAudio(framePropsList, true)
     }
+  }, {
+    label: locale.translation('closeTabPage'),
```
luixxiul: ditto

app/extensions/brave/locales/en-US/menu.properties
NejcZdovc: I was thinking about it, but I am not sure.

bsclifton: I think it's OK as-is; it matches the English and also consistent format with `closeWindow`

luixxiul: I agree!

js/actions/windowActions.js
```diff
@@ -729,6 +729,19 @@ const windowActions = {
...
 
   /**
+   * Dispatches a close call to all frames in tab page
+   * The provided frame will be closed.
+   *
+   * @param {Object} framesList - List of frames
+   * @param {Object} framePropsList - List of frame properties to consider
+   */
+  tabPageClosed: function (framesList, framePropsList) {
+    framePropsList.forEach((frameProps) => {
```
bbondy: this is ok for now but we'll want to change this later to something that dispatches to the store and does a single action.  This would be pretty slow as is sometimes with a full tab page set.

bbondy:
Could you add a test? Optionally I left a comment for doing it in one operation instead of N.

bsclifton:
+1 for adding a test cc: @NejcZdovc

NejcZdovc:
Will add a test and will also try to solve problem with N operations.

NejcZdovc:
@bsclifton added tests. AFAIK we don't have mechanism for testing this type of menu with web driver?

@bbondy I removed N calls into the store, now we have only 1 call

@bradleyrichter When we close the whole tab page, I was thinking to have two possible scenarios for which active tab to select:
- last tab (we select the last tab from the remaining tabs) - implemented
- last active tab (we select last active tab from the remaining tabs) - not implemented
I wouldn't add parent tab option here, because we could have multiple tabs with multiple parents and even though we could focus only on active tab, but what happens when you close tab page which doesn't have an active tab. What happens when active tab have parent tab in the closing tab page. We could add the whole logic, but I think it's not worth it.

bsclifton:
@NejcZdovc you are 100% right- I neglected to notice this uses a popup (native menu) and not our ContextMenuItem. I don't believe this can fully be tested w/ webdriver

NejcZdovc:
Implemented 'last active tab' option

js/actions/windowActions.js
```diff
@@ -333,6 +334,33 @@ const windowActions = {
   },
 
   /**
+   * Dispatches a message to close multiple frames
+   * @param {Object[]} framePropsList - The properties of all frames to close
+   */
+  closeFrames: function (framePropsList) {
+    const activeFrameKey = windowStore.getState().get('activeFrameKey')
```
bbondy: ideally there should be no logic inside the window action, I think you should just make it an easy api and have it declarative, so like windowAction.tabPageClosed(i) for the index to close.  The store then does the work to figure out which frames it needs to close and closes them all in one swoop.

bsclifton: +1 on that

bsclifton:
@benmosher is this ready for review? if so, let's rebase 😄

NejcZdovc:
This PR is done, tested and it's working correctly. The only thing is that there is some logic in window action, so it should be refactored according to @bbondy comment

bbondy:
yes sorry, we can't accept new PRs with logic in the window action.

NejcZdovc:
@bbondy totally understandable, will refactor it, not a problem

NejcZdovc:
@bbondy refactored, ready for a review

test/unit/state/frameStateUtilTest.js
```diff
@@ -283,4 +283,75 @@ describe('frameStateUtil', function () {
       })
     })
   })
+
+  describe('removeFrames', function () {
```
bsclifton: ❤️

app/renderer/reducers/frameReducer.js
```diff
@@ -0,0 +1,147 @@
...
+
+    case windowConstants.WINDOW_SET_FULL_SCREEN:
+      state = setFullScreen(state, action)
+      break
+  }
+
+  return state
+}
+
+module.exports = frameReducer
```
cezaraugusto: awesome++++

PR #6572, opened 2017-01-07
punycode.decode only accept ASCII-only input

### LINKED ISSUE #6540: Opening a link causes a crash
Open http://xn--jp%29-5y3b.f1-live.com/f1/jp/headlines/news/detail/060912193351.shtml and the browser crashes

**Expected behavior:**
It should return `ERR_NAME_NOT_RESOLVED` like Chrome does

- Brave Version (revision SHA): 0.12.15

- Steps to reproduce:
    1. Open http://xn--jp%29-5y3b.f1-live.com/f1/jp/headlines/news/detail/060912193351.shtml

- Any related issues:

Before verifying the bug, please make sure that the session data has been safely saved.

darkdh:

fix #6540

Auditors: @diracdeltas

Test Plan:
1. Go to http://xn--jp%29-5y3b.f1-live.com/f1/jp/headlines/news/detail/060912193351.shtml
2. There should be a DNS error page

### DIFF
--- a/js/lib/baseDomain.js
+++ b/js/lib/baseDomain.js
@@ -9,6 +9,18 @@ const LRUCache = require('lru_cache/core').LRUCache
 
 let cachedBaseDomain = new LRUCache(50)
 
+const checkASCII = function (str) {
+  if (typeof str !== 'string') {
+    return false
+  }
+  for (let i = 0; i < str.length; i++) {
+    if (str.charCodeAt(i) > 127) {
+      return false
+    }
+  }
+  return true
+}
+
 /**
  * Returns base domain for specified host based on Public Suffix List.
  * @param {string} hostname The name of the host to get the base domain for
@@ -16,8 +28,13 @@ let cachedBaseDomain = new LRUCache(50)
 
 module.exports.getBaseDomain = function (hostname) {
   // decode punycode if exists
-  if (hostname.indexOf('xn--') >= 0) {
-    hostname = punycode.toUnicode(hostname)
+  if (hostname.indexOf('xn--') >= 0 &&
+    checkASCII(hostname)) {
+    try {
+      hostname = punycode.toUnicode(hostname)
+    } catch (e) {
+      console.error('punnycode.toUnicode() failure:', e)
+    }
   }
 
   let baseDomain = cachedBaseDomain.get(hostname)


diracdeltas:
I still get a crash when I copy and paste the URL and hit enter. Stack trace:
```
RangeError: Invalid input
    at error (punycode.js:42:8)
    at decode (punycode.js:235:5)
    at punycode.js:389:6
    at map (punycode.js:57:20)
    at mapDomain (punycode.js:84:18)
    at Object.toUnicode (punycode.js:387:9)
    at module.exports.getBaseDomain (/Users/yan/repos/browser-laptop/js/lib/baseDomain.$s:33:25)
    at Object.module.exports.isThirdPartyHost (/Users/yan/repos/browser-laptop/app/filt$ring.js:448:26)
    at Function.<anonymous> (/Users/yan/repos/browser-laptop/app/filtering.js:254:26)
process exited with code 7
```

darkdh:
`checkASCII` in the PR will catch percent-encoded string

darkdh:
This is what I saw with this PR
screen shot 2017-01-09 at 23 02 12
 and without it I will also see the crash call stack

diracdeltas:
`checkASCII(hostname)` returns true for me. what do you get for `url.parse('http://xn--jp%29-5y3b.f1-live.com/f1/jp/headlines/news/detail/060912193351.shtml').hostname` ?

js/lib/baseDomain.js
```diff
@@ -9,14 +9,27 @@ const LRUCache = require('lru_cache/core').LRUCache
 
 let cachedBaseDomain = new LRUCache(50)
 
+const checkASCII = function (str) {
+  if (typeof str !== 'string') {
+    return false
+  }
+  for (let i = 0; i < str.length; i++) {
+    if (str.charCodeAt(i) > 127) {
```
bbondy: I think charCodeAt is implemented to UCS-2 even know the strings are UTF-16.  So for example a single unicode symbol could be for charCodeAt(0) and charCodeAt(1).  Are there any code units that have their second byte < 127?  Maybe there is a utility function we can use instead for this since these types of checks can easily go wrong? On a side note I think our TP and adblock libs probably need fixes like this.

bbondy: I think the impl is probably right though in regards to what I brought up because you will just return false before it gets to the second code unit, so you can probably ignore my comment above.

js/lib/baseDomain.js
```diff
@@ -9,14 +9,27 @@ const LRUCache = require('lru_cache/core').LRUCache
...
 /**
  * Returns base domain for specified host based on Public Suffix List.
  * @param {string} hostname The name of the host to get the base domain for
  */
 
 module.exports.getBaseDomain = function (hostname) {
   // decode punycode if exists
-  if (hostname.indexOf('xn--') >= 0) {
+  if (hostname.indexOf('xn--') >= 0 &&
+    checkASCII(hostname)) {
```
diracdeltas: i figured out why this was crashing. the `hostname` that gets checked here when i go to http://xn--jp%29-5y3b.f1-live.com/f1/jp/headlines/news/detail/060912193351.shtml is `xn--jp`, which is valid ASCII but invalid punycode. so punycode.js throws a (different) error, and the crash happens. I verified that wrapping `hostname.toUnicode` in a try-catch fixes it for me.

js/lib/baseDomain.js
```diff
@@ -9,15 +9,32 @@ const LRUCache = require('lru_cache/core').LRUCache
...
  * @param {string} hostname The name of the host to get the base domain for
  */
 
 module.exports.getBaseDomain = function (hostname) {
   // decode punycode if exists
-  if (hostname.indexOf('xn--') >= 0) {
-    hostname = punycode.toUnicode(hostname)
+  if (hostname.indexOf('xn--') >= 0 &&
+    checkASCII(hostname)) {
+    try {
```
bsclifton: ++ on the try/catch! You'd be surprised that some domains look legit punycoded but actually fail the decoding process.  Not having try/catch has bitten me multiple times in the past :smile:

PR #6574, opened 2017-01-07
Rename React root div from windowContainer to appContainer

### LINKED ISSUE #6568: windowContainer is used as React root div id and Window component div id
React root div id is ["windowContainer"](https://github.com/jide/browser-laptop/blob/master/app/extensions/brave/index.html#L22), and is where the `<Window/>` [root component is mounted](https://github.com/jide/browser-laptop/blob/master/js/entry.js#L82) :

```js
  ReactDOM.render(
    <Window includePinnedSites={disposition !== 'new-popup'} frames={frames} initWindowState={initWindowState} />,
    document.getElementById('windowContainer'))
```

But the `<Window/>` component itself renders a `<div/>` with a ["windowContainer"](https://github.com/jide/browser-laptop/blob/master/js/components/window.js#L68) id :

```js
    return <div id='windowContainer' className={cx(classes)} >
      <Main windowState={this.state.immutableData.windowState}
        appState={this.state.immutableData.appState} />
    </div>
```

So we end up with 2 nested divs with a "windowContainer" id:

capture d ecran 2017-01-07 a 02 15 36

The root div could simply be renamed "root".

jide:
Rename windowContainer to appContainer to avoid duplicate IDs in the DOM.

Fixes #6568

Auditors: @bridiver @darkdh

### DIFF
--- a/app/extensions/brave/index-dev.html
+++ b/app/extensions/brave/index-dev.html
@@ -19,7 +19,7 @@
     <link rel="localization" href="locales/{locale}/downloads.properties">
   </head>
   <body>
-    <div id="windowContainer">
+    <div id="appContainer">
       <div id="webpackLoading" style="display:none">
         <p>
           <span>Loading...</span>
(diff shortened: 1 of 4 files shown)


bbondy:
++ thanks

PR #6575, opened 2017-01-08
Build: Make babel tranpile smarter by using chrome version (fixes #6367)

### LINKED ISSUE #6367: Build: Only transpile JS which is not supported based on chrome version
n/a

Currently, we transpile all the JavaScript to ES5 code. But I think we can be smart by using https://github.com/babel/babel-preset-env to only tranpile code which is not supported by a particular version of chrome (I guess we can say 54 now). 

**Expected behavior:**

This will help us to run most of the code we write as it is. Chrome as better perfs on certain ES6 features i guess.

all
master

gyandeeps:

Test Plan: N/A

### Before

```sh
    aboutPages.entry.js    2.62 MB       0  [emitted]  aboutPages
           app.entry.js    2.32 MB       1  [emitted]  app
```

### After

```sh
    aboutPages.entry.js    2.52 MB       0  [emitted]  aboutPages
           app.entry.js    2.16 MB       1  [emitted]  app
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,12 @@
 {
-  "presets": ["react", "es2015"],
+  "presets": [
+    "react",
+    ["env", {
+      "targets": {
+        "chrome": 54
+      }
+    }]
+  ],
   "plugins": [
     "transform-react-inline-elements",
     "transform-react-constant-elements"
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "babel-plugin-transform-react-constant-elements": "^6.4.0",
     "babel-plugin-transform-react-inline-elements": "^6.4.0",
     "babel-polyfill": "^6.3.14",
-    "babel-preset-es2015": "^6.1.18",
+    "babel-preset-env": "^1.1.6",
     "babel-preset-react": "^6.1.18",
     "babel-register": "^6.3.13",
     "babel-runtime": "^6.3.13",


bbondy:
Thanks!

bsclifton:
@gyandeeps great find! :smile:

hzoo:
Awesome to see this! Wanted to leave some comments (as a Babel maintainer):

Instead of chrome 54 (since brave is on 56 now), you can do something like

```js
"targets": {
  "browsers": ["last 2 Chrome versions"]
}
```

Although now that I think about it, we should add one for brave 😝? A package was made for electron https://github.com/babel/babel-preset-env/pull/144

I'd recommend using the `useBuiltIns` option to reduce the number of polyfills loaded as well. https://github.com/babel/babel-preset-env#usebuiltins if you use `babel-polyfill` (although not sure if that is loaded or not in the source?

bbondy:
thanks for the tips @hzoo 
@gyandeeps would you mind trying to use useBuiltIns and always use latest for target

hzoo:
Might need some help with determining where the polyfill is loaded in (if it is at all?)

The way it works currently is that it literally tries to find `import "babel-polyfill";` in your code (or require) and then replace that code with the individual imports so it needs to be imported in the entry point file rather than in a script tag or the webpack entry point etc

gyandeeps:
Thanks @hzoo for feedback.

* I dont think `babel-polyfill` is used in source code.
* Since we know what version of chrome we use then i guess i can update it to `56`.

So just change the chrome version. Any thing i am missing?

PR #6591, opened 2017-01-10
sync site settings, add SYNC_CLEAR HISTORY

diracdeltas:
also refactors reusable sync methods from app/sync.js to js/state/syncUtil.js

Auditors: @ayumi

Test Plan:
1. start Brave
2. go to bing.com
3. disable shields
4. you should see the console log a setting with 'shieldsUp: false', indicating the sync-client sent the record.
5. enable shields
6. same as 4 but with 'shieldsUp: true'

### DIFF
--- a/app/sync.js
+++ b/app/sync.js
@@ -6,7 +6,6 @@
 
 const Immutable = require('immutable')
 const electron = require('electron')
-const crypto = require('crypto')
 const ipcMain = electron.ipcMain
 const messages = require('../js/constants/sync/messages')
 const categories = require('../js/constants/sync/proto').categories
@@ -16,6 +15,7 @@ const appActions = require('../js/actions/appActions')
 const syncConstants = require('../js/constants/syncConstants')
 const appDispatcher = require('../js/dispatcher/appDispatcher')
 const AppStore = require('../js/stores/appStore')
+const syncUtil = require('../js/state/syncUtil')
 
 const categoryNames = Object.keys(categories)
 const categoryMap = {
@@ -27,13 +27,6 @@ const categoryMap = {
 
 let deviceId = null /** @type {Array|null} */
 let pollIntervalId = null
-
-/**
- * Gets current time in seconds
- */
-const now = () => {
-  return Math.floor(Date.now() / 1000)
-}
 
 /**
  * Sends sync records of the same category to the sync server.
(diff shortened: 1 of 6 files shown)


app/sync.js
```diff
@@ -187,13 +66,40 @@ const doAction = (sender, action) => {
...
+      }
+      break
+    case syncConstants.SYNC_REMOVE_SITE_SETTING:
+      sendSyncRecords(sender, writeActions.DELETE,
+        [syncUtil.createSiteSettingsData(action.hostPattern, action.item.toJS())])
+      break
+    case syncConstants.SYNC_CLEAR_SITE_SETTINGS:
+      sender.send(messages.DELETE_SYNC_PREFERENCE_TYPE, 'siteSetting')
+      // TODO: sync-client should listen for this message and delete
+      // all existing preferences with the specified objectData type
```
ayumi: :+1:

`DELETE_SYNC_PREFERENCE_TYPE`, `'device'` sounds odd. i wonder if it's more intuitive to split it into DELETE_SITE_SETTINGS and DELETE_DEVICE.

js/state/syncUtil.js
```diff
@@ -20,3 +34,124 @@ module.exports.setObjectId = (item) => {
...
+  const adControlEnum = {
+    showBraveAds: 0,
+    blockAds: 1,
+    allowAdsAndTracking: 2
+  }
+  const cookieControlEnum = {
+    block3rdPartyCookie: 0,
+    allowAllCookies: 1
+  }
+  const value = Object.assign({hostPattern}, siteSettingDefaults)
```
ayumi: should it be `Object.assign(siteSettingDefaults, {hostPattern})` ? as is i think the custom hostpattern is always overwritten as empty string

js/stores/appStore.js
```diff
@@ -589,15 +592,25 @@ const handleAppAction = (action) => {
     case appConstants.APP_CHANGE_SITE_SETTING:
       {
         let propertyName = action.temporary ? 'temporarySiteSettings' : 'siteSettings'
-        appState = appState.set(propertyName,
-          siteSettings.mergeSiteSetting(appState.get(propertyName), action.hostPattern, action.key, action.value))
+        let newSiteSettings = siteSettings.mergeSiteSetting(appState.get(propertyName), action.hostPattern, action.key, action.value)
+        if (!action.temporary) {
+          let s = syncUtil.setObjectId(newSiteSettings.get(action.hostPattern))
```
diracdeltas: noted

js/state/syncUtil.js
diracdeltas: oops, good catch!

app/sync.js
diracdeltas: wfm

js/state/syncUtil.js
diracdeltas: i guess it should actually be `Object.assign({}, siteSettingDefaults, {hostPattern})`

diracdeltas:
addressed comments

PR #6625, opened 2017-01-12
Fix SYNC_CLEAR_HISTORY

ayumi:
Clearing history wasn't clearing SyncRecord historySites.
This happened because `app/sync.js` `doAction()` always checks if `action.item` is a proper Sync item.

This PR resolves it by checking for action.item only when the action requires it.

Test Plan:

0. Sync should be enabled.
1. Clear browsing history.
2. In the console observe `Deleting category: HISTORY_SITES`

Auditors: @diracdeltas

### DIFF
--- a/app/sync.js
+++ b/app/sync.js
@@ -21,6 +21,7 @@ const settings = require('../js/constants/settings')
 
 const CATEGORY_MAP = syncUtil.CATEGORY_MAP
 const CATEGORY_NAMES = Object.keys(categories)
+const SYNC_ACTIONS = Object.values(syncConstants)
 
 const log = (message) => {
   if (!config.debug) { return }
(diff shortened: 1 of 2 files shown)


app/sync.js
```diff
@@ -22,6 +22,16 @@ const settings = require('../js/constants/settings')
...
 
+const SYNC_ACTIONS = Object.values(syncConstants)
+const SYNC_ACTIONS_WITH_ITEMS = [
+  syncConstants.SYNC_ADD_SITE,
+  syncConstants.SYNC_UPDATE_SITE,
+  syncConstants.SYNC_REMOVE_SITE,
+  syncConstants.SYNC_ADD_SITE_SETTING,
+  syncConstants.SYNC_UPDATE_SITE_SETTING,
+  syncConstants.SYNC_REMOVE_SITE_SETTING
+]
```
diracdeltas: it would be slightly easier for this to be SYNC_ACTIONS_WITHOUT_ITEMS since most actions in the future will probably have items

diracdeltas: i think these constants should go inside validateAction since they are only used in there

ayumi:
No new test breakages.

Issue #3151, opened 2016-08-12
Refactor: URLBarSuggestions should not generate suggestions in the component


bbondy:
This is still legacy code from the gecko browser days and probably some of the earliest code we have but URLBarSuggestions data generation shouldn't be done in the component. This is both for generating the suggestions and for the search suggestion xhr. It should move to the store level.

This is currently the cause of some hacks we have and code would be much more clear and performant with this fixed, as it would also fix wasted re-renders that are currently needed.

bbondy:
For a test plan please re-test urlbar and urlbar suggestion functionality

PR #6634, opened 2017-01-14
Suggestions refactor

bbondy:
Auditors: @bsclifton


bbondy:
This is partially reviewed already but I'm going to merge. I'll do any review comments as follow ups. I'd like to get some testing for these changes with the next preview.

bbondy:
This removes various hacks and workarounds we had in place, including getting rid of urlBar calling into the urlBarSuggestions component. It detangles a lot of complex logic.
There are still further refactors that can happen here, such as getting rid of various window actions that are setters.  Also I'd like to get component unit tests now for urlBar and urlBarSuggestions, but I'm not blocking on those since we have good UI coverage for this already.   I did add a bunch of reducer unit tests though.

luixxiul:
If no qa is required please add the label, thanks!

bbondy:
there was a basic note for testing but hard to see from this commit in the PR
https://github.com/brave/browser-laptop/pull/6634/commits/b1bbeffc316ff6a2faa4271972e3015f72c2b2c9

In this issue:
https://github.com/brave/browser-laptop/issues/3151

I think QA works from issues and not from PRs for QA labels for qa needed / not needed, let me know if that's not the case though. Thanks!

luixxiul:
> QA works from issues and not from PRs

Actually I've been doing it both from the issues and the PRs for less possibility of leaving commits untested.

bbondy:
I think it should only be done by issues and then make sure every PR has an associated issue.

Issue #6641, opened 2017-01-14
Brave considers https and wss as different origins when blocking cookies


STRML:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
This is intended to be a general ticket encompassing the underlying issue behind #4694 and #4918. There may be others.

The issue is that in the default mode, Brave considers a websocket connection on the same origin as "3rd party" and blocks sending cookies. This is unexpected behavior and breaks auth on many websockets.

**Expected behavior:**

For the purpose of blocking cookies, `http://mysite.com` and `ws://mysite.com` should be considered the same origin, as should `https://mysite.com` and `wss://mysite.com`.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Any

- Brave Version (revision SHA):
0.12.15

- Steps to reproduce:
    1. Go to any site that does websocket authentication with cookies (for example, bitmex.com)
    2. Notice the websocket does not receive the `Cookie` header and auth fails.

- Any related issues:

Strangely, this appears to be triggered by a bug of some sort; on a fresh install I'm not able to replicate the Cookie header being blocked on `wss://` until a few cycles of logging in and out.

luixxiul:
CC @diracdeltas for a thought

STRML:
Working on a fix; this appears to be related to the devtools in some way, can replicate behavior on localhost but only with devtools open.

STRML:
I was able to replicate this in `v0.12.15` but not in `master`.

In `master`, when the websocket upgrade request goes through [filtering](https://github.com/brave/browser-laptop/blob/1d3384a9186e451e64f72b5d8f6d9106b82cb82e/app/filtering.js#L254), `details` contains the following:

```json
{
  firstPartyUrl: 'http://localhost:2001/app/wallet',
  id: 79,
  method: 'GET',
  referrer: '',
  requestHeaders: {
    'Accept-Encoding': 'gzip, deflate',
    'Accept-Language': 'en-US',
    Connection: 'Upgrade',
    Cookie: 'connect.sid=s%3AkEYCYf7e_-k4-jTasWZiFFshUGv6i_FA.G%2F0FDMFLrfH46a7Wdj7tLJYAUPravxe6xIAw7d5QSQ0',
    Origin: 'http://localhost:2001',
    'Sec-WebSocket-Version': '13',
    Upgrade: 'websocket',
    'User-Agent': '...'
  },
  resourceType: 'other',
  tabId: -1,
[...]

PR #6642, opened 2017-01-14
Avoid an unnecessary url.parse() in filtering.

STRML:
If cookie blocking is disabled, or the resource being requested
is not to be filtered (such as unfilterable protocols like ws:)
or requests from the internal extension, the parse can be skipped.

This is just a tiny optimization I saw while working through https://github.com/brave/browser-laptop/issues/6641.

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -222,7 +222,6 @@ function registerForBeforeSendHeaders (session, partition) {
     }
 
     let requestHeaders = details.requestHeaders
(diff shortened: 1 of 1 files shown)


bsclifton:
@darkdh you were just in this code recently; would you be able to check this out?

luixxiul:
Let's merge this if nobody disagrees.

bbondy:
++

PR #6721, opened 2017-01-19
Sync siteSettings with browser data

ayumi:
Auditors: @diracdeltas

Test Plan:

Prep:

0. Update sync lib to `brave/sync #fix/resolve-delete-nonexistant-props`.
1. Prepare 2 instances (pyramids) of Brave with Sync enabled.
  a. Enable Sync and close Brave.
  b. Copy `{userData}/brave-development` to `{userData}/brave-development-2`.
  c. Edit `brave-development-2/session-store-1` `deviceId` to `1`.
  d. To `browser-laptop` `package.json` add `"start2": "node ./tools/start.js --user-data-dir=brave-development-2 --debug=5859 --enable-logging --v=0 --enable-extension-activity-logging --enable-sandbox-logging --enable-dcheck",`
2. In `appConfig.js` `sync.fetchInterval` reduce to 5 seconds.

Play:

3. Open both pyramid 1 and pyramid 2.
4. In pyramid 1 visit a webpage and open up the bravery panel.
5. In pyramid 2 visit the same page and open up the bravery panel.
6. In pyramid 2 toggle each available siteSetting. Observe it appears in pyramid 1 after 1–5s.
7. Try toggling multiple settings at once, and toggling different settings simulatenously on both pyramids.
8. In both go to Preferences #Shields. Clear siteSettings with the Clear links and the red X's. Observe they sync over.

### DIFF
--- a/js/state/syncUtil.js
+++ b/js/state/syncUtil.js
@@ -64,6 +64,64 @@ const applySiteRecord = (record, siteDetail, tag) => {
   }
 }
 
+const applySiteSettingRecord = (record) => {
+  // TODO: In Sync lib syncRecordAsJS() convert Enums into strings
+  const adControlEnum = {
+    0: 'showBraveAds',
+    1: 'blockAds',
+    2: 'allowAdsAndTracking'
+  }
+  const cookieControlEnum = {
+    0: 'block3rdPartyCookie',
+    1: 'allowAllCookies'
+  }
+  const getValue = (key, value) => {
+    if (key === 'adControl') {
+      return adControlEnum[value]
+    } else if (key === 'cookieControl') {
+      return cookieControlEnum[value]
+    } else {
+      return value
+    }
+  }
+  const appActions = require('../actions/appActions')
+  const objectId = new Immutable.List(record.objectId)
+  const category = CATEGORY_MAP[record.objectData].categoryName
+  const hostPattern = record.siteSetting.hostPattern
+  if (!hostPattern) {
+    throw new Error('siteSetting.hostPattern is required.')
+  }
+
+  let applySetting = null
+  switch (record.action) {
+    case writeActions.CREATE:
+    case writeActions.UPDATE:
+      // Set the objectId if needed so we can access the existing object
+      let existingObject = this.getObjectById(objectId, category)
+      if (!existingObject) {
+        appActions.changeSiteSetting(hostPattern, 'objectId', objectId, false, true)
+        existingObject = this.getObjectById(objectId, category)
+      }
+      const existingObjectData = existingObject[1]
+      applySetting = (key, value) => {
+        const applyValue = getValue(key, value)
+        if (existingObjectData.get(key) === applyValue) { return }
+        appActions.changeSiteSetting(hostPattern, key, applyValue, false, true)
+      }
+      break
+    case writeActions.DELETE:
+      applySetting = (key, _value) => {
+        appActions.removeSiteSetting(hostPattern, key, false, true)
+      }
+      break
+  }
+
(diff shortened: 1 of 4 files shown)


js/state/syncUtil.js
```diff
@@ -64,6 +64,59 @@ const applySiteRecord = (record, siteDetail, tag) => {
...
+      applySetting = (key, _value) => {
+        appActions.removeSiteSetting(hostPattern, key, false, true)
+      }
+      break
+  }
+  // First set the objectId if needed so we can access the existing objectj
+  let existingObject = this.getObjectById(objectId, category)
+  if (!existingObject) {
+    applySetting('objectId', objectId)
+    existingObject = this.getObjectById(objectId, category)
```
diracdeltas: does this fail when `record.action` is DELETE?

js/state/syncUtil.js
```diff
@@ -64,6 +64,59 @@ const applySiteRecord = (record, siteDetail, tag) => {
...
+  const existingObjectData = existingObject[1]
+  for (let key in record.siteSetting) {
+    if (key === 'hostPattern') { continue }
+    let value = record.siteSetting[key]
+    if (key === 'adControl') {
+      value = adControlEnum[value]
+    } else if (key === 'cookieControl') {
+      value = cookieControlEnum[value]
+    }
+    if (existingObjectData.get(key) === value) { continue }
```
diracdeltas: i think this needs to call `applySetting` anyway (instead of `continue`) when the action is DELETE. `null` may be a valid siteSetting

js/stores/appStore.js
```diff
@@ -620,11 +628,15 @@ const handleAppAction = (action) => {
         let propertyName = action.temporary ? 'temporarySiteSettings' : 'siteSettings'
         let newSiteSettings = new Immutable.Map()
         appState.get(propertyName).map((entry, hostPattern) => {
+          // const value = entry.get(action.key)
```
diracdeltas: minor: unnecessary comment

js/state/syncUtil.js
ayumi: I think Deletes are only sent when objectId is set, however it's probably good to check and skip it if it's a Delete.

diracdeltas:
i tried the test plan and it was mostly glorious. however there is an issue where some site settings are not being removed.

repro
1. goto https://docs.npmjs.com/cli/init in 2 pyramids
2. toggle every site setting in the lion menu on one of the pyramids
3. go to about:preferences#shields in both pyramids
4. in one of the pyramids, click 'clear all' in each category
5. in the other pyramid, all of the site settings should be cleared. however 'Ad Control', 'Block Phishing' and 'HTTPS Everywhere' are not cleared.

screen shot 2017-01-19 at 3 05 50 am

js/state/syncUtil.js
diracdeltas: ++

diracdeltas:
when i go to `'History' > 'Clear browsing data' > 'Clear site settings'` it doesn't clear the site settings in the other pyramid

ayumi:
@diracdeltas i made the changes!

For the last thing: Currently clearing Synced data categories clears it locally and only prevents future records from being synced; it doesn't clear categories on devices which have already synced the data.
I think that feature/fix should be a separate PR.

diracdeltas:
> For the last thing: Currently clearing Synced data categories clears it locally and only prevents future records from being synced; it doesn't clear categories on devices which have already synced the data.
> I think that feature/fix should be a separate PR.

sgtm, can you open an issue in brave/sync for tracking?

ayumi:
https://github.com/brave/sync/issues/40

PR #6755, opened 2017-01-20
On Sync init only send siteSettings without objectId

ayumi:
Auditors: @diracdeltas

Test Plan:
1. Have Sync disabled.
2. Browse a site and toggle a Bravery Panel setting.
3. Enable Sync and restart your pyramid. Note that siteSettings are sent.
4. Restart your pyramid again. Note this time siteSettings are *not* sent.


ayumi:
Merging because this improves start up performance when you have lots of siteSettings.

PR #6773, opened 2017-01-20
Don't add brave to the system apt sources.list file for ubuntu/debian installs

dflock:
Just drop a brave `.list` file into the `/etc/apt/sources.list.d/` folder instead.


bsclifton:
@posix4e would you be able to review this?

bbondy:
lgtm and works.

PR #6832, opened 2017-01-24
Sync bookmarks in order, without conflict

ayumi:
Fix https://github.com/brave/sync/issues/48

Auditors: @diracdeltas

Test plan:
1. Start Pyramid 0. It should have bookmarks and bookmark folders.
2. Enable Sync on Pyramid 0, and restart to complete setup.
3. Start Pyramid 1. Add it to Pyramid 0's Sync profile and restart.
4. Bookmarks should sync in the same order.

### DIFF
--- a/js/state/syncUtil.js
+++ b/js/state/syncUtil.js
@@ -39,17 +39,43 @@ const siteSettingDefaults = {
   ledgerPaymentsShown: true
 }
 
+// Whitelist of valid browser-laptop site fields. In browser-laptop, site
+// is used for both bookmarks and history sites.
+const SITE_FIELDS = ['objectId', 'location', 'title', 'customTitle', 'tags', 'favicon', 'themeColor', 'lastAccessedTime', 'creationTime', 'partitionNumber', 'parentFolderId']
+
+const pickFields = (object, fields) => {
+  return fields.reduce((a, x) => {
+    if (object.hasOwnProperty(x)) { a[x] = object[x] }
+    return a
+  }, {})
+}
+
 /**
  * Apply a bookmark or historySite SyncRecord to the browser data store.
  * @param {Object} record
- * @param {Array.<string>} tags
+ * @param {Immutable.Map} siteDetail
  */
-const applySiteRecord = (record, siteDetail, tag) => {
+const applySiteRecord = (record) => {
   const appActions = require('../actions/appActions')
+  const siteTags = require('../constants/siteTags')
   const objectId = new Immutable.List(record.objectId)
   const category = CATEGORY_MAP[record.objectData].categoryName
   const existingObject = this.getObjectById(objectId, category)
   const existingObjectData = existingObject && existingObject[1]
+
+  let siteProps = Object.assign({}, record.historySite, record.bookmark, record.bookmark && record.bookmark.site, {objectId})
+  const isFolder = (typeof siteProps.isFolder === 'boolean')
+    ? siteProps.isFolder
+    : (existingObjectData && existingObjectData.get('folderId'))
+  const tag = isFolder
+    ? siteTags.BOOKMARK_FOLDER
+    : siteTags.BOOKMARK
+  const parentFolderObjectId = siteProps.parentFolderObjectId
+  if (parentFolderObjectId && parentFolderObjectId.length > 0) {
+    siteProps.parentFolderId =
+      getFolderIdByObjectId(new Immutable.List(parentFolderObjectId))
+  }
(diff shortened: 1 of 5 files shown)


diracdeltas:
it does not appear related to this change but travis builds are erroring :crying_cat_face:

js/state/syncUtil.js
```diff
@@ -280,12 +300,37 @@ module.exports.newObjectId = (objectPath) => {
...
+    return module.exports.newObjectId(['sites', folderIndex])
+  }
+}
+
+/**
  * Converts a site object into input for sendSyncRecords
  * @param {Object} site
  * @param {number=} siteIndex
+ * @param {boolean} generateObjectIds objectIds are generated via AppActions IPC
+ *   which is async. You may prefer to generate it for immediate use.
```
diracdeltas: not sure i understand this comment / why this is necessary. `module.exports.newObjectId` returns the objectID immediately.

ayumi: so i had this issue on initial bookmarks sync because although `newObjectId()` returns the objectId, `appState.getIn(['sites', 123])` doesn't immediately reflect the objectId.
I decided to move the generation to [`sync.js`](https://github.com/brave/browser-laptop/pull/6832/files#diff-b7794baa4e0cd1fb8cab9449660662baR183) and memoize them as they're generated.

diracdeltas: you can get the value of the objectId immediately in the caller via `record.objectId` though (where `record` is the object returned by `createSiteData`?

diracdeltas:
looks like this test is failing: https://travis-ci.org/brave/browser-laptop/builds/194972665#L4738

js/state/syncUtil.js
ayumi: thx, i feel it now and removed the param.

the problem i had was in going from `parentFolderId` to `parentFolderObjectId`.

example:
- 2 bookmarks in a folder
- sync bookmark 1. it looks up its folder by parentFolderId, and sees the folder has no objectId so sets it.
- sync bookmark 2. it also looks up folder by parentFolderId, and again sees the folder has no objectId so sets it. this happens faster than the setting of the objectId.

i've retained the memoing in sync.js, and updated createSiteData() to use `site.parentFolderObjectId` if already set.

ayumi:
@diracdeltas I've updated the code. Could you take another look?

js/state/syncUtil.js
```diff
@@ -39,18 +39,43 @@ const siteSettingDefaults = {
   ledgerPaymentsShown: true
 }
 
+// Shared between bookmarks and history sites.
+const SITE_FIELDS = ['objectId', 'location', 'title', 'customTitle', 'tags', 'favicon', 'themeColor', 'lastAccessedTime', 'creationTime', 'partitionNumber', 'parentFolderId']
```
diracdeltas: [minor] partitionNumber is only useful on laptop since AFAIK the other clients don't have partitions

ayumi: thx, will clarify.

this is the browser-laptop site field whitelist.

diracdeltas: ^ NVM

js/state/syncUtil.js
```diff
@@ -127,22 +152,11 @@ const applySiteSettingRecord = (record) => {
  * @param {Object} record
  */
 module.exports.applySyncRecord = (record) => {
-  const objectId = new Immutable.List(record.objectId)
+  console.log(`applying: ${JSON.stringify(record)}`)
```
diracdeltas: unneeded logging ?

ayumi: good catch

diracdeltas:
i tried the test plan and it seemed to sync my bookmarks without folders

pyramid 0 bookmark structure:
screen shot 2017-01-27 at 6 15 36 pm

pyramid 1 bookmark structure:
screen shot 2017-01-27 at 6 18 38 pm

ayumi:
did you get the latest sync bundle with `yarn download-sync-client`

diracdeltas:
@ayumi i ran `yarn dist` from sync/staging. i'll try again

diracdeltas:
same result using  yarn download-sync-client

diracdeltas:
i  narrowed down the problem somewhat. syncing folders works for folders that are created after sync is enabled. it fails for folders that were created before sync was enabled.

diracdeltas:
test plan now works for me!!

diracdeltas:
siteUtil unittest is failing but i'm fixing that now

PR #6865, opened 2017-01-26
Harden / improve the tests for autofill / contextMenu

bsclifton:
- Adds a data-item attribute to each context menu item
- Test waits for selection before moving forward (instead of pause 10)
- Updated test to use constants

Auditors: @bbondy, @NejcZdovc

## Test Plan
Window 1: `npm run watch`
Window 2: `npm run uitest -- --grep="ContextMenu"`

### DIFF
--- a/test/components/contextMenuTest.js
+++ b/test/components/contextMenuTest.js
@@ -11,10 +11,14 @@ describe('ContextMenu', function () {
       .waitForEnabled(urlInput)
   }
 
+  const contextMenuIndexOf = (index) => {
+    return `.contextMenu div.selectedByKeyboard[data-index="${index}"]`
+  }
+
   describe('navigation', function () {
     Brave.beforeAll(this)
 
-    before(function *() {
+    before(function * () {
       yield setup(this.app.client)
       this.formfill = Brave.server.url('formfill.html')
       this.input = '[name="01___title"]'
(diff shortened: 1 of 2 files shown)


bsclifton:
@bbondy this is what I mentioned fixes the intermittent failures with autofill arrow keys

NejcZdovc:
Let's wait until #6888 is merged, so that we can check if everything is in order then

test/components/contextMenuTest.js
```diff
@@ -90,37 +93,35 @@ describe('ContextMenu', function () {
...
+    it('select item via keys only', function * () {
       yield this.app.client
         .tabByIndex(0)
         .loadUrl(this.formfill)
         .waitForVisible('<form>')
         .click(this.input)
         .windowByUrl(Brave.browserWindowUrl)
-        .keys('\uE015')
+        .keys(Brave.keys.DOWN)
         .waitForVisible('.contextMenu')
```
bbondy: I was just commenting on these pause's, thanks!

bbondy:
++

luixxiul:
Do we have an associated issue for this?

bbondy:
If it's just related to automated tests it's ok to exclude, in which case there is no QA on it.

PR #6751, opened 2017-01-19
Add sync

diracdeltas:
this adds basic support for browser sync, using https://github.com/brave/sync

Fix #1854

# Test Plan:

## setup
- make sure automated tests related to sync are passing
- `npm install` to add dependencies
- `npm run download-sync-client` to download the sync client to the right subdirectory
- add the following to package.json to create a separate browser 'device' to sync to:
```
 "start2": "node ./tools/start.js --user-data-dir=brave-development-2 --debug=5859 --enable-logging --v=0 --enable-extension-activity-logging --enable-sandbox-logging --enable-dcheck"
```

## test
- npm start
- go to about:preferences#sync
- enable sync
- restart, then click 'add device...'
- npm run start2
- in device 2, go to about:preferences#sync > i have an existing code > enter the code from device 1
- go to any website in either of the browsers, bookmark it and make some changes to the siteSettings. after a minute, it should show up in the other browser.

# todos before public release (not necessarily before merge)
- https://github.com/brave/sync/milestone/2

PR #6882, opened 2017-01-27
Add sync

### LINKED ISSUE #1854: Implement sync
Does Brave support sync ?       linux  Brave: 0.10.0 

I see sync mentioned in your code base yet sync does not appear in Preferences

diracdeltas:
replaces https://github.com/brave/browser-laptop/pull/6751

this adds basic support for browser sync, using https://github.com/brave/sync

Fix #1854

# Test Plan:

## setup
- make sure automated tests related to sync are passing
- `npm install` to add dependencies
- `npm run download-sync-client` to download the sync client to the right subdirectory
- add the following to package.json to create a separate browser 'device' to sync to:
```
 "start2": "node ./tools/start.js --user-data-dir=brave-development-2 --debug=5859 --enable-logging --v=0 --enable-extension-activity-logging --enable-sandbox-logging --enable-dcheck"
```

## test
- npm start
- go to about:preferences#sync
- enable sync
- restart, then click 'add device...'
- npm run start2
- in device 2, go to about:preferences#sync > i have an existing code > enter the code from device 1
- go to any website in either of the browsers, bookmark it and make some changes to the siteSettings. after a minute, it should show up in the other browser.

# todos before public release (not necessarily before merge)
- https://github.com/brave/sync/milestone/2


diracdeltas:
The test plan previously worked against 0.13.1 but seems buggy with master (bookmark favicon is synced but not the actual bookmark location). Site settings sync seems to be working.

luixxiul:
If this will be included 0.13.1, would you please merge `.properties` files first so contributors on Transifex can start localization?

diracdeltas:
syncing bookmarks now working again

bsclifton:
Moving to 0.13.3; I know there are some challenges with rebase (and this is contingent on https://github.com/brave/browser-laptop/pull/6938 )

diracdeltas:
@bsclifton actually i resolved all the merge conflicts yesterday and this is ready for review. sorry, i should have mentioned that.

bsclifton:
@diracdeltas awesome! will try to start reviewing this soon (maybe tomorrow?)

bsclifton:
@diracdeltas @ayumi when I try to run this locally, I'm getting a crash ☹️   I made sure to remove `node_modules` and redo the `npm install`.  I've also tried deleting my session (so it would make a fresh new one) and this didn't work

diracdeltas:
@bsclifton did you do `npm run download-sync-client`?

bsclifton:
@diracdeltas no I did not (I am not a smart person!).  Thanks 😄

bsclifton:
## Feedback (current WIP)
Just going to type up as I use the branch and find things 😄   this post will be updated as I move there here 

### Enabling Sync
screen shot 2017-02-01 at 2 06 32 pm
- For the textbox on this screen, [can we use one the new textbox controls](https://github.com/brave/browser-laptop/blob/master/app/renderer/components/textbox.js#L25)? (ex: `FormTextbox`).  This will make the text larger and easier to read
- Should we choose a better default name (than `browser-laptop`? 
  - Since we have node.js available, we should be able to detect the hostname of the computer.
  - We could also make our own "hostname" based on:
    - the OS (mac / win / linux)
    - the version (mac-16.4.0)
    - partial guid (mac-16.4.0-3782a409)

diracdeltas:
i am working on making sync restartless so those bugs above related to sync restart should go away soon

diracdeltas:
sync is now restartless so i checked off the todo items that had to do with restarting. WRT:

> How can I break the relationship with my "sync group"? ex: I can turn off syncing, but ever after I quit... if I re-enable syncing, it seems that I'm still tied to that group (i'll get updates). This can be verified by checking the secret words

@bradleyrichter believes we should not expose the concept of a 'sync group' in the UI, so this is not a supported action right now. however, i think this will be solved with https://github.com/brave/sync/issues/41 because then you can delete the current device.

>  In the case that the code words (or QR code) are compromised... maybe someone shares a screenshot on reddit accidentally... what options do users have?
> can you request to delete your sync data?
> can you rotate your keywords / QR code?

https://github.com/brave/sync/issues/51 would delete both the userID/keys and the synced data

diracdeltas:
updated the sync wiki page to be more friendly

> I think we should consider allowing to sync preferences. For example, I have bookmarks toolbar shown (with show favicon and text). I didn't even realize the sync already happened because the toolbar didn't show :stuck_out_tongue: This is a down the road thing; we could always just make a follow up issue

Agree; the bookmarks toolbar thing may be a separate issue though. I noticed that the toolbar on a fresh profile automatically shows up when the first bookmark is added manually but not when it is added through sync. seems like a UI trigger issue.

> Should there be any visual indicator for when the last sync was? Like a growl/popup which says "Sync last run: 25 seconds ago...", stays for a few seconds, then fades away. This would be a good use-case for actually using the built in JavaScript notification API

It syncs every minute right now so it would be kind of spammy

ayumi:
- I think these issues are resolved, and bookmarks and history sync ok:
> on pyramid 2, I added a folder called "folder1". Before (or during?) the sync, I tried to delete the folder. As a user, nothing happened but I saw the following get logged:
https://gist.github.com/bsclifton/6ce897bc81a292addef13f4fdbce7ee6
> I ran into a weird situation adding a folder (may be timing related? like I added it during the sync?) where the two devices remain out of sync for one item.
> In this pic, the "33" folder is only on one of them. I don't know if the logs above included error details? I quit and restarted and syncing worked fine after that

- Added sync integration tests:
  - Bookmark – Create, Update (Rename), Delete, Move into folder, Order
  - Bookmark folder – Create, Delete
  - History – Works, Order, Clearing prevents future sync
  - Site settings – Works, Clearing prevents future sync

- When a sync fetch contains new folders which have bookmarks, the bookmarks sometimes sync not in their folders but directly on the bookmarks bar (https://github.com/brave/sync/issues/49).

bsclifton:
Great job resolving the issues 😄  (and awesome job w/ the tests too)!  I'll give this a go again today

ayumi:
the-brave-pyramids-orange-medium

diracdeltas:
os.hostname() appears to be the Brave extension ID in the context of about:preferences

diracdeltas:
most or all of the review issues have been addressed, but the merge seems to have caused some new issues

test/unit/state/siteUtilTest.js
```diff
@@ -425,9 +433,47 @@ describe('siteUtil', function () {
         const processedSites = siteUtil.addSite(Immutable.fromJS(sites), newSiteDetail, siteTags.BOOKMARK, oldSiteDetail)
         const expectedSiteKey = siteUtil.getSiteKey(expectedSiteDetail)
         let expectedSites = {}
-        expectedSites[expectedSiteKey] = expectedSiteDetail.toJS()
+        expectedSites[expectedSiteKey] = expectedSiteDetail.set('objectId', undefined).toJS()
         assert.deepEqual(processedSites.toJS(), expectedSites)
       })
+      it('sets an objectId when syncCallback is provided', function () {
```
alexwykoff: This appears to be describing and update/edit case if I am not mistaken. 

What I did not notice is a deletion test case.

test/components/syncTest.js
```diff
@@ -0,0 +1,788 @@
...
+    // For Create: Bookmark page 1
+    yield bookmarkUrl(this.page1Url, this.page1Title)
+
+    // For Update: Bookmark page 2 and rename it
+    yield Brave.app.client
+      .tabByIndex(0)
+      .waitForUrl(this.page1Url)
+    yield bookmarkUrl(this.page2Url, this.page2Title)
+    yield Brave.app.client
+      .pause(1000) // XXX: Helps to correctly order Create and Update
```
alexwykoff: Rather than using explicit pauses, is there no promise (or set of nested promises) we can create?

test/components/syncTest.js
```diff
@@ -0,0 +1,788 @@
...
+          .then((title) => title === updatedTitle)
+      })
+      .waitUntil(function () {
+        return this.getText('.bookmarkToolbarButton:nth-child(3) .bookmarkText')
+          .then((title) => title === folder1Title)
+      })
+  })
+})
+
+describe('Syncing history', function () {
```
alexwykoff: What happens if the separate machines have overlapping timestamps for history? Do we have tiebreaker logic?

test/components/syncTest.js
```diff
@@ -0,0 +1,788 @@
...
+  })
+
+  it('private browsing does not sync', function * () {
+    yield Brave.app.client
+      .waitForElementCount('table.sortableTable tbody tr', 2)
+      .waitForVisible(`table.sortableTable td.title[data-sort="${this.page3Title}"]`, 1000, true)
+  })
+})
+
+describe('Syncing site settings', function () {
```
alexwykoff: What happens for conflicting site preferences? How do we tiebreak?

test/components/syncTest.js
diracdeltas: @ayumi put some effort into this but it seemed like not easily

diracdeltas: in general i agree on removing explicit pauses; if i understand correctly, it is tricky here because this is waiting on the server to receive the records and process them, which is not something that can be detected client-side (ex: using a waitUntil for some appState change). since the server can't push notifications to the client, we would have to have the client poll the server periodically to detect when a sync operation has finished.

test/components/syncTest.js
diracdeltas: not sure what you mean by a 'tie'. if the same history entry (same location) has two sync records with the same timestamp, the one which is received later by the sync server wins.

test/components/syncTest.js
diracdeltas: i believe the record which is received later always wins. https://github.com/brave/sync/blob/staging/client/recordUtil.js#L103

ayumi:
Discussed with @bsclifton and @bbondy in slack and decided this is good to merge for MVP sync!

@alexwykoff we will follow up with polish right away.

luixxiul:
@diracdeltas @ayumi Please update the test plan in the first post, if it is outdated. Also please add anything which should be tested during QA tests, including edge cases. Thanks in advance!

ayumi:
@luixxiul thanks, I updated the Test Plan above!

PR #6910, opened 2017-01-28
Split sites, removed customTitle

### LINKED ISSUE #6585: Remove customTitle from siteDetail object
Discussion starts here https://github.com/brave/browser-laptop/issues/6108#issuecomment-265934841

> **To keep things simple, I'd like to propose that we eliminate custom title**.
> - When a bookmark is added for a page, it will default to the title of the page
> - Once set, the title for that bookmark will never be updated again

Removing this should resolve the following issues:
https://github.com/brave/browser-laptop/issues/6104
https://github.com/brave/browser-laptop/issues/6108
https://github.com/brave/browser-laptop/issues/3694

There are likely more which have this as the root cause (we can search for issues where titles change)

NejcZdovc:

Resolves #6585

Auditors:
@bbondy, @ayumi, @diracdeltas, @bsclifton

Test Plan:
- everything should work the same as before

### DIFF
--- a/docs/state.md
+++ b/docs/state.md
@@ -61,19 +61,40 @@ AppStore
       }
     }
   },
-  sites: {
-    [siteKey]: { // folder: folderId; bookmark/history: location + partitionNumber + parentFolderId
+  bookmarks: {
+    [siteKey]: {
       location: string,
       title: string,
-      customTitle: string, // User provided title for bookmark; overrides title
-      tags: [string], // empty, 'bookmark', 'bookmark-folder', 'pinned', or 'reader'
+      tags: [string], // 'bookmark' or 'bookmark-folder'
       favicon: string, // URL of the favicon
       themeColor: string, // css compatible color string
       lastAccessedTime: number, // datetime.getTime()
       creationTime: number, //creation time of bookmark
       partitionNumber: number, // Optionally specifies a specific session
-      folderId: number, // Set for bookmark folders only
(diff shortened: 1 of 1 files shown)


bbondy:
The documentation update looks good to me, thanks for starting with that.

NejcZdovc:
@bbondy Correct me if I am wrong, but I think `pinnedSites` is not needed,  as far as I can see from this [code](https://github.com/brave/browser-laptop/blob/master/js/about/newtab.js#L71).

Also I think [this parameter](https://github.com/brave/browser-laptop/blob/master/js/components/main.js#L1187) is redundant, it's not used in component tree afterwords.

Can you please double check?

bsclifton:
One issue- although we don't have to solve this right now though 😄 

With regards to `visitedSites`, the ideal solution would **not** be key based- since that would be the history.  Our current solution is unfortunately key based and if you visit the same site twice, only  the last instance is recorded.  This is important to keep in mind since the session files need to be upgradeable. If we choose to change the format down the road (instead of planning it out now), we'll need to write a migration (something to convert the old format to the new format and detect if it's needed).

NejcZdovc:
@bbondy already warn me about upgrade, so I am implementing upgrade path in this  [function](https://github.com/brave/browser-laptop/blob/master/app/sessionStore.js#L408).

NejcZdovc:
Will continue with work on this after #6929 is done.

bsclifton:
Closing this one for now- we can re-open once it's unblocked 😄

PR #6939, opened 2017-01-31
Notification Bar Changes

### LINKED ISSUE #6935: Move page-level alerts to below the tabs bar, and add an overlap nub
Moving the alert bar (for tab-specific alerts) under the tabs bar will give the correct hierarchical appearance to the alert. (global alerts can remain in their top position.)

Adding an overlapping nub reduces the chance for exact spoofing. 

similar example compliments of Chrome on OS X:
image

jonathansampson:

Auditors: @bradleyrichter 

Resolves #6935

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -23,7 +23,7 @@ const TabPages = require('../../app/renderer/components/tabs/tabPages')
 const TabsToolbar = require('../../app/renderer/components/tabs/tabsToolbar')
 const FindBar = require('./findbar')
 const UpdateBar = require('./updateBar')
-const NotificationBar = require('./notificationBar')
+const {NotificationBar} = require('./notificationBar')
 const DownloadsBar = require('../../app/renderer/components/downloadsBar')
 const SiteInfo = require('./siteInfo')
 const BraveryPanel = require('./braveryPanel')
@@ -732,6 +732,10 @@ class Main extends ImmutableComponent {
 
     const appStateSites = this.props.appState.get('sites')
 
+    const notifications = this.props.appState.get('notifications')
+    const hasNotifications = notifications && notifications.size
+    const notificationBarOrigin = notifications.map(bar => bar.get('frameOrigin'))
+
     return <div id='window'
       className={cx({
         isFullScreen: activeFrame && activeFrame.get('isFullScreen'),
@@ -861,12 +865,6 @@ class Main extends ImmutableComponent {
 
         <UpdateBar updates={this.props.appState.get('updates')} />
         {
-          this.props.appState.get('notifications') && this.props.appState.get('notifications').size && activeFrame
-          ? <NotificationBar notifications={this.props.appState.get('notifications')}
-            activeFrame={activeFrame} />
-          : null
-        }
-        {
           showBookmarksToolbar
           ? <BookmarksToolbar
             draggingOverData={this.props.appState.getIn(['dragData', 'dragOverData', 'draggingOverType']) === dragTypes.BOOKMARK && this.props.appState.getIn(['dragData', 'dragOverData'])}
@@ -909,12 +907,18 @@ class Main extends ImmutableComponent {
           key='tab-bar'
           activeFrameKey={(activeFrame && activeFrame.get('key')) || undefined}
           onMenu={this.onHamburgerMenu}
+          notificationBarActive={notificationBarOrigin}
           hasTabInFullScreen={
             sortedFrames
               .map((frame) => frame.get('isFullScreen'))
               .some(fullScreenMode => fullScreenMode === true)
           }
         />
+        {
+          hasNotifications && activeFrame
+          ? <NotificationBar notifications={notifications} activeFrame={activeFrame} />
+          : null
+        }
 
         {
           activeFrame && activeFrame.get('findbarShown') && !activeFrame.get('isFullScreen')
--- a/js/components/notificationBar.js
+++ b/js/components/notificationBar.js
@@ -15,6 +15,7 @@ const Button = require('./button')
 
 const {StyleSheet, css} = require('aphrodite/no-important')
 const commonStyles = require('../../app/renderer/components/styles/commonStyles')
+const globalStyles = require('../../app/renderer/components/styles/global')
 
 class NotificationItem extends ImmutableComponent {
   clickHandler (buttonIndex, e) {
@@ -115,6 +116,14 @@ class NotificationBar extends ImmutableComponent {
   }
 }
 
+class NotificationBarCaret extends ImmutableComponent {
+  render () {
+    return <div className={css(styles.caretWrapper)}>
+      <div className={css(styles.caret)} />
+    </div>
+  }
+}
+
 const styles = StyleSheet.create({
   flexJustifyBetween: {
     display: 'flex',
@@ -142,7 +151,32 @@ const styles = StyleSheet.create({
     textDecoration: 'underline',
     fontSize: '13px',
     margin: '5px'
+  },
+  caretWrapper: {
+    position: 'absolute',
+    bottom: 0,
+    left: 0,
+    right: 0,
+    zIndex: globalStyles.zindex.zindexTabs,
+    filter: 'drop-shadow(rgba(0,0,0,0.25) 0px 0px 1px)'
+  },
+  caret: {
+    position: 'relative',
+    margin: 'auto',
+    width: '16px',
+    ':before': {
+      content: '""',
+      position: 'absolute',
+      bottom: 0,
+      left: 0,
+      borderBottom: `6px solid ${globalStyles.color.notificationItemColor}`,
+      borderLeft: '8px solid transparent',
+      borderRight: '8px solid transparent'
+    }
   }
 })
 
-module.exports = NotificationBar
+module.exports = {
+  NotificationBar,
+  NotificationBarCaret
+}
--- a/app/renderer/components/styles/commonStyles.js
+++ b/app/renderer/components/styles/commonStyles.js
@@ -174,13 +174,20 @@ const styles = StyleSheet.create({
     width: '100%',
     cursor: 'default',
     userSelect: 'none',
-    marginTop: globalStyles.spacing.navbarMenubarMargin
+    // if there's more than one notification per site,
+    // ensure border is on the last only
+    ':last-child': {
+      borderBottom: `5px solid ${globalStyles.color.notificationBottomBorderColor}`
+    },
+    // last-child will always be orange, but others can be gray
+    ':not(:last-child)': {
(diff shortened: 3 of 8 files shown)


less/notificationBar.less
```diff
@@ -7,7 +7,19 @@
 
 .notificationBar {
   display: inline-block;
+  position: relative;
+  z-index: 1000;
```
luixxiul: I'd change this into `@zindexNavigationBar` according to https://github.com/brave/browser-laptop/blob/ce25254d5bb0a391ffa5ebbc0d5f70ef2348155a/less/variables.less#L129

bsclifton:
@jonathansampson is this one ready for review? If not, can we move it to 0.13.3? thanks 😄

jonathansampson:
@bsclifton I'm not positive; @bradleyrichter, thoughts?

less/notificationBar.less
jonathansampson: @luixxiul That appears to cause the type of problems we faced in the design discussion last night. Namely, interacting with some of the document elements will cause the shadow to vanish; often only regions of it will vanish.

odd-shadow

bradleyrichter:
I thought we were stuck at the lame-triangle problem? Did you find a solution @jonathansampson ?

jonathansampson:
@bradleyrichter I'm looking into some of the tab-change, window-resize events now to see if we can just subscribe to those and pick up the needed information to position the triangle. Positioning it will require us to use a true element, rather than a pseudo element.

less/notificationBar.less
luixxiul: How about changing @zindexNavigationBar on variables.less into `1000` then??

cezaraugusto:
I have a fix for triangle issue and going to self-assign this and make a follow-up of @jonathansampson changes.

cezaraugusto:
@jonathansampson I edited PR with caret feel free to edit anything. Didn't test all use cases for app-wise notifications, tried only fullscreen

jonathansampson:
@cezaraugusto Wonder if we should be concerned with the following scenario: A user has multiple tabs opened, and may confuse the notification on one for being a notification on another. Note below how the notification bars for github.io appear to be coming from google.com.

image

diracdeltas:
@jonathansampson the problem with having the notification bars fully in the webview area is that a page can spoof them; this is why initially i implemented them to be above the tab bar. this is solved by having the triangle poke into the "trusted UI" area, but yeah i agree it's confusing to have it point to a different tab. IMO it should always point to either the URL bar or the tab where the notification is coming from.

bradleyrichter:
This should never happen anyway because we have switched the alerts to be tab specific.

diracdeltas:
@bradleyrichter i don't know what you mean. the notifications right now appear in the ("global") trusted UI area, but they only appear when the tab they are for is active. 

@cezaraugusto as long as the caret isn't going to an unrelated tab, sgtm

bsclifton:
@cezaraugusto if the notifications are per-tab, you'll also be able to close this issue out 😄  https://github.com/brave/browser-laptop/issues/7737

jonathansampson:
@bradleyrichter @diracdeltas In the above screenshot, Google.com has a notification bar. I left it untouched, and moved over to webrtc.github.io, where two additional notification bars were opened. Each bar points only to its related tab. But triangles are visible even when associated notification bars are not.

@cezaraugusto I would suggest placing the triangle in the center of the tab rather than on the left or right. Should we also hide the triangle if the tab isn't focused? Or, perhaps mark tabs in such a manner so as to show that they have a pending notification? Something other than the triangle for blurred tabs?

diracdeltas:
@jonathansampson i see, i didn't realize there was a notification for the goog tab. in that case seems like the triangle for google should just disappear (or turn grey)

cezaraugusto:
edited, now caret only appears if tab is active and has a notificationBar. Also moved caret to the center.

js/components/main.js
```diff
@@ -1221,12 +1216,19 @@ class Main extends ImmutableComponent {
...
           onMenu={this.onHamburgerMenu}
+          notificationBarActive={notificationBarOrigin}
           hasTabInFullScreen={
             sortedFrames
               .map((frame) => frame.get('isFullScreen'))
               .some(fullScreenMode => fullScreenMode === true)
           }
         />
+        {
+          this.props.appState.get('notifications') && this.props.appState.get('notifications').size && activeFrame
```
jonathansampson: We call out to `this.props.appState.get('notifications')` 3-4 times. Can we do this once, and then store the result?

cezaraugusto: you mean adding to a variable? If so, we can:

```
const notifications = this.props.appState.get('notifications')
const hasNotifications = notifications && notifications.size
...
hasNotifications && activeFrame
  ? <NotificationBar notifications={notifications} activeFrame={activeFrame} />
  : null
```

jonathansampson: @cezaraugusto Yes. This way we only have to go searching for the object once.

bradleyrichter:
And I confused a pair of tab alerts for separate tab alerts. Sorry! 

So, I confirmed we are good with tab specific alerts. Next we need to stack them in z order and not y order. 

And hide the non-active nub as Yan mentioned.

cezaraugusto:
updated code, here's the actual result:

notificationbargif

It's likely to have some UI polish from here but will left up to you guys @bradleyrichter @jonathansampson. I think Brad has a new UI/colors for notificationBar and IMO it's a good opportunity to do it.

bradleyrichter:
looks awesome....

this was the last mockup I did. I don't think we need the yellow color to signify an alert but I could be wrong. 

image

jonathansampson:
@bradleyrichter @cezaraugusto Presently, a tab could have a notification bar opened, and the user would have no idea. I'm a little uneasy about this. Some type of indicator would be helpful. Edge (IIRC) would gently pulsate the background color of the tab, informing you that there is a notification bar that needs to be addressed. I feel we should do something as well, rather than leaving the user completely unaware.

js/components/main.js
```diff
@@ -936,6 +936,9 @@ class Main extends ImmutableComponent {
 
     const appStateSites = this.props.appState.get('sites')
     const activeTabShowingMessageBox = !!(activeTab && activeTab.get('messageBoxDetail'))
+    const notifications = this.props.appState.get('notifications')
+    const hasNotifications = notifications && notifications.size
+    const notificationBarOrigin = this.props.appState.get('notifications').map(bar => bar.get('frameOrigin'))
```
jonathansampson: We can call `notifications.map` instead, to avoid the redundant lookup.

js/components/main.js
```diff
@@ -1221,12 +1218,19 @@ class Main extends ImmutableComponent {
...
+          notificationBarActive={notificationBarOrigin}
           hasTabInFullScreen={
             sortedFrames
               .map((frame) => frame.get('isFullScreen'))
               .some(fullScreenMode => fullScreenMode === true)
           }
         />
+        {
+          hasNotifications && activeFrame
+          ? <NotificationBar notifications={this.props.appState.get('notifications')}
```
jonathansampson: `notifications={notifications}` to avoid extra lookup.

bradleyrichter:
Subtle alerts sound fine. Originally we were on the path of annoying the user to death, leaving them hunting for the button to disable all alerts. 

We need to keep balancing protection and annoyance. :  )

luixxiul:
I'm just wondering, if we really need the caret over the tab title. I can easily imagine a bunch of users would say, "please add an option to hide it or just remove it because I want to read the full title".

For me the aim of the caret is unclear, because, if I select a tab and there appears a notification bar under the tab, it is obvious that the notification bar appears there exactly for the tab which I selected just now.

Also, in this case, if the tab preview is enabled, the notification bar should be hidden if another tab is previewed which do not have notification.

diracdeltas:
The caret is needed because otherwise a notification bar can be spoofed by a malicious page.

diracdeltas:
The way chrome/ff handle this is by having a notification box pop up from the URL bar when the tab it is for is in focus. That would be better.

cezaraugusto:
I'm ok either way change it to be nested to urlbar should be minimal

cezaraugusto:
@bradleyrichter ok to use it the same way as Chrome/FF?

bradleyrichter:
I am not sure this would work because it would cover all of the other tabs, forcing the user to interact with the message before using other tabs?

bradleyrichter:
Let's try adding a border and making the nub more defined and obvious.

bsclifton:
How's this one looking? @jonathansampson, @cezaraugusto, @luixxiul

jonathansampson:
@bsclifton I'm assuming it's _ready to go_ since @cezaraugusto added [🏷PR/ready-for-review](https://github.com/brave/browser-laptop/labels/PR/ready-for-review) a day ago. @bradleyrichter, how do you feel about the present presentation?

js/components/notificationBar.js
```diff
@@ -115,6 +116,14 @@ class NotificationBar extends ImmutableComponent {
   }
 }
 
+class NotificationBarCaret extends ImmutableComponent {
```
NejcZdovc: Why do we need this component? I don't see real benefit of having it.

jonathansampson: @NejcZdovc What do you suggest as an alternative?

NejcZdovc: I am not sure if we need component for two divs. What I would suggest is to create divs where we now use this component and move styles into notificationStyle file and use it from there.
cc @cezaraugusto

cezaraugusto: I prefer this way because then I can leave styles on the component itself. Otherwise, I'll have to create a new style for that and put it on the tab itself or inside a third file, and if I do it then I'm not taking advantage of Aphrodite.

Style and component should live together so I don't have to run several files to find a simple thing. Things we do such as `commonStyles.js` is just a hack we found to keep refactoring important stuff while common components aren't ready yet/too nested. Having both together makes regressions less likely. 

It also simplifies the code, so if for some reason Brad or other less code-experienced member/collaborator wants to change something for a given component, they'll need to go to into just one file.

bradleyrichter:
I'm happy now. Let's merge and refine the styles separately as @NejcZdovc suggested.

cezaraugusto:
Based on latest Brad's spec:

screen shot 2017-04-20 at 1 04 20 am

jonathansampson:
@cezaraugusto Holy smokes, that's gorgeous! :)

srirambv:
Nub not shown for pinned tabs 
image

Also notification for pinned tabs is retained accross normal tabs as well. This is on packaged 0.15.2 packaged preview 1 build

bradleyrichter:
it should also work the same for pinned tabs.

PR #6964, opened 2017-02-01
Sync initial bookmarks in order

ayumi:
Re-introduce SyncRecord.Bookmark.index after Sites Map refactor.

Fix https://github.com/brave/sync/issues/45

Auditors: @diracdeltas

Test plan:
1. Start Pyramid 0. It should have bookmarks and bookmark folders.
2. Enable Sync on Pyramid 0, and restart to complete setup.
3. Start Pyramid 1. Add it to Pyramid 0's Sync profile and restart.
4. Bookmarks should sync onto Pyramid 1 in the same order as Pyd 0.

### DIFF
--- a/app/sync.js
+++ b/app/sync.js
@@ -165,38 +165,44 @@ module.exports.onSyncReady = (isFirstRun, e) => {
 
   /**
    * Sync a bookmark that has not been synced yet, first syncing the parent
-   * folder if needed.
+   * folder if needed. For folders, set and memoize folderId to ensure
+   * consistent parentFolderObjectIds.
+   * Otherwise siteUtil.createSiteData() will generate new objectIds every
+   * call; there's not enough time to dispatch id updates to appStore.sites.
    * @param {Immutable.Map} site
    */
-  // Setting objectIds happens with AppActions IPC but we need it immediately.
-  const folderObjectIds = {}
-  // TODO (@ayumi): siteIndex is now ignored since switching to siteKey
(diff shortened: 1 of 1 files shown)


app/sync.js
```diff
@@ -165,38 +165,48 @@ module.exports.onSyncReady = (isFirstRun, e) => {
...
-   * folder if needed.
+   * folder if needed. For folders, set and memoize folderId to ensure
+   * consistent parentFolderObjectIds.
+   * Otherwise siteUtil.createSiteData() will generate new objectIds every
+   * call; there's not enough time to dispatch id updates to appStore.sites.
    * @param {Immutable.Map} site
+   * @param {number} index Sets the index for SyncRecord.Bookmark.index
+   *   so other clients apply records in a batch in the correct order.
+   * @returns {number} Next index to use. Note this may not be index +1
+   *   because for foldered bookmarks we first recurse and sync parent folders.
```
diracdeltas: looks like this jsdoc is no longer correct?

ayumi: good 🔼

diracdeltas:
i noticed that not all of my bookmarks were synced in the correct folders :(. but this is also happening on feature/syncing so not related to this PR.
top: pyramid 0
bottom: pyramid 1
screen shot 2017-02-01 at 1 28 58 am

luixxiul:
I added the label 0.13.3 in case that the sync would be implemented with that version. Let's reset it later if it is not.

PR #6995, opened 2017-02-02
remove getURL

bridiver:
cc @bbondy


luixxiul:
Don't we need to have an associated issue for this?

bridiver:
@bbondy did you create an umbrella issue for removing the sync webview methods?

bbondy:
@bridiver  https://github.com/brave/browser-laptop/issues/7002

bbondy:
++

luixxiul:
@bbondy will this be included in 0.13.3?

bbondy:
no issue is ok on this one, it will bypass QA but this would break from general regression pass if it was a problem.

PR #7023, opened 2017-02-03
Added default download path

### LINKED ISSUE #2110: General Settings: "Save my downloads here" setting
Functionality:
Providing a user-definable default downloads destination allows for a more efficient user work flow when downloading files. The default path is their downloads folder. This path can be changed by clicking on the pencil (edit) button. This opens a download location chooser dialog which is OS specific.

A toggle switch (off by default) allows the user to change the setting to ask them for the desired location upon each file download.

item label:
"Save my downloads here"

toggle switch label:
"Always ask me where to save files"

Download location sheet-dialog in Mac:
image

Spec image: (this issue in blue)
image

NejcZdovc:

Resolves #2110

## Auditors
@bradleyrichter @bsclifton

## Test Plan
- go to the preferences, general tab
- define your default download path
- toggle show dialog
- based on your settings, download dialog should be opened or not
- if you set to hide download dialog, files should be downloaded automatically to your default download path


NejcZdovc:
@bradleyrichter in #2110 you added pencil at the end of the input. Can you please provide an icon for it, if we will use it. For now I set input to read only and when you click on it, dialog is shown automatically.

NejcZdovc:
@bsclifton Sorry I added you as a review too soon. My mistake. This PR is not ready for a final review as you notice there is no test for this new feature. I added label `ready-for-review` so that @bradleyrichter can look through it and prepare icon for browse/edit. And yes I agree we need to have a browse/edit icon.

bradleyrichter:
@NejcZdovc SVG icon for the edit (pencil) button attached:


[browser_icon_pencil.svg.zip](https://github.com/brave/browser-laptop/files/765198/browser_icon_pencil.svg.zip)

bsclifton:
@NejcZdovc was this one done?

NejcZdovc:
@bsclifton I added test's as well, so we are ready for a review

app/browser/reducers/downloadsReducer.js
```diff
@@ -89,6 +92,18 @@ const downloadsReducer = (state, action) => {
...
       break
+    case appConstants.APP_DOWNLOAD_DEFAULT_PATH:
+      const focusedWindow = BrowserWindow.getFocusedWindow()
+
+      dialog.showOpenDialog(focusedWindow, {
+        defaultPath: app.getPath('downloads'),
+        properties: ['openDirectory']
+      }, (folder) => {
+        if (Array.isArray(folder) && fs.lstatSync(folder[0]).isDirectory()) {
+          appActions.changeSetting(settings.DOWNLOAD_DEFAULT_PATH, folder[0])
```
bsclifton: I think this is fine as-is... but it could be better as it's own discreet action (in the event that we'd like other reducers to be listening).  Perhaps an action like `defaultDownloadPathChanged`

NejcZdovc:
@bsclifton First of all sorry, I somehow over looked your second review.

I added hover color effect. Regarding the pointer I remember talking with @bradleyrichter about it for bookmark manager and he said that we should use only color hover and not a pointer. Bookmark manager for example don't have single pointer hover, only color hover. I agree that we should be consistent, so we need a guide line for it. 

For user select I suggest to add it after #7481 is merged, because I added aphrodite param for component `SettingItem`.

NejcZdovc:
Just for a reference, @bridiver added flag to handle "Save as" calls https://github.com/brave/muon/commit/de90c69d2a1432bedf8f43f4fa5498bcb56a4655

js/about/aboutActions.js
```diff
@@ -407,6 +407,15 @@ const aboutActions = {
       url: url,
       savePath: savePath
     })
+  },
+
+  /**
+   * Open dialog for default download path setting
+   */
+  defaultDownloadPath: function () {
```
bridiver: these no longer have to go in `aboutActions` and this file should be deprecated. Use `appActions` for now instead

bridiver:
this will need changes to correspond with my muon updates to correctly handle the prompt

js/about/aboutActions.js
NejcZdovc: Moved

NejcZdovc:
@bsclifton added save prompt for save as. Ready for a review

bsclifton:
An area we can improve (cc: @bradleyrichter): 

If I have a long path, it gets clipped
screen shot 2017-03-16 at 2 56 24 pm

It looks the exact same in Chrome though, so not a deal-breaker or anything... just an area we could do better 😄 
screen shot 2017-03-16 at 2 57 51 pm

PR #7039, opened 2017-02-03
Focus webview when ESC pressed on UrlBar

### LINKED ISSUE #3248: Make pressing Esc in address-bar return focus to viewport (Enhancement)
Yes

If the address-bar has focus, there is no way to switch focus to the viewport by using the keyboard.

**Expected behavior:**
In Firefox, pressing Esc eventually returns focus to the viewport.

This is also the case in Opera.

In Chrome, there is a workaround using Cmd+P to open the Print dialog and then pressing Esc to close it, but that doesn't work in Brave.
- Brave Version: 0.11.5
- Steps to reproduce:
  1. Open a page
  2. Switch focus to the address bar by pressing Cmd+L
  3. Try to return focus to the viewport by pressing Esc several times
  4. Try to return focus to the viewport by pressing Tab

Expected: Focus is returned to the viewport after 1-3 presses of Esc or Tab

Actual: Focus remains in address-bar

**Possibly related issues:**
#2920, #2914, #3057

michalbe:
Fixes #3248

Test Plan:
  - Focus the URL bar
  - Press `ESC`
  - notice the focus returned to the webview

### DIFF
--- a/app/renderer/components/urlBar.js
+++ b/app/renderer/components/urlBar.js
@@ -7,6 +7,7 @@ const urlParse = require('../../common/urlParse')
 
 const ImmutableComponent = require('../../../js/components/immutableComponent')
 const windowActions = require('../../../js/actions/windowActions')
+const webviewActions = require('../../../js/actions/webviewActions')
 const appActions = require('../../../js/actions/appActions')
 const KeyCodes = require('../../common/constants/keyCodes')
 const cx = require('../../../js/lib/classSet')
@@ -335,9 +336,11 @@ class UrlBar extends ImmutableComponent {
 
   onKeyUp (e) {
     switch (e.keyCode) {
+      case KeyCodes.ESC:
+        webviewActions.setWebviewFocused()
+        return
       case KeyCodes.UP:
       case KeyCodes.DOWN:
-      case KeyCodes.ESC:
         return
     }
     if (this.isSelected()) {


bbondy:
Sorry this is not the correct behaviour and would diverge us from how all other browsers work. We're all about being different for our differentiators, but for standard UX things that people expect to work one way, and have trained themselves to think they should work that way, we shouldn't diverge.  In particular browser users often press esc to clear what they are typing and revert the URL text to the previous one.

michalbe:
@bbondy I understand your point and it makes perfect sense to me, but I think things like this should be pointed out in the original issue spec (in #3248) before it was marked as `good-first-bug` and `feature request` half a year ago. There's 1.8k+ opened issues and it's hard for non-employees to follow your vision of the project when those tasks are not triaged properly. Anyway - I'm impressed with how Brave is growing, I love what you guys are doing and I hope sooner or later I will find an issue that's valid and properly scoped :). Thanks, and have a nice weekend.

bbondy:
@michalbe definitely, and I'm sorry you had to spend time on it and we didn't realize it earlier. I don't have full visibility over all issues but we try to do our best to catch things at the issue level.  Most issues are fine that are open, but feel free to hit me up to make 100% sure in the issue if you do another.  Thanks for taking the time to contribute and sorry we dropped the ball on this one.

PR #7072, opened 2017-02-04
Download Item Removal Confirmation

### LINKED ISSUE #2604: Add delete confirmation to downloads bar
## Test Plan
- Download a file
- try to delete it from the `downloadsBar`
- notice new confirmation bar
  - click `ok` to remove the file
  - move the mouse outside the `downloadItem` panel to hide the confirmation bar.

## Original issue description
"Delete Download" on download bar should have a warning prompt.
- Any related issues: #1076 #1202

michalbe:
## Test Plan
- Download a file
- try to delete it from the `downloadsBar`
- notice new confirmation bar
  - click `ok` to remove the file
  - move the mouse outside the `downloadItem` panel to hide the confirmation bar.

## Automated tests
`npm run test -- --grep="downloadItem"`

## Description
Fixes #2604

How it works:
download-remove-confirmation


luixxiul:
I like it :-D

michalbe:
@luixxiul glad to hear that, thanks :)

bsclifton:
@michalbe well done! 😄 

Reviewing the code... would you be able to leave this component as an ImmutableComponent and instead have our AppStore be updated? Ideally, our controls would have state bound to them as a property (with binding happening in `components/main.js` or a child component).  To trigger this state changing we would want to use an action (see `appAction.js`)

michalbe:
@bsclifton Thanks, I implemented changes you've requested. 
I have a problem with tests - I'm probably doing something wrong with mocking the appState in the test suite - even fires as expected, app action fires as well (so I assume the event is dispatched), but it never gets to the reducer in appStore. Could you help me with this please?

siemiatj:
@michalbe I haven't yet read through the browser code but from my experience for instance some other bar/popup/modal may potentially be hidden by this ?

app/renderer/components/downloadItem.js
```diff
@@ -74,10 +82,17 @@ class DownloadItem extends ImmutableComponent {
...
       onContextMenu={contextMenus.onDownloadsToolbarContextMenu.bind(null, this.props.downloadId, this.props.download)}
       onDoubleClick={this.onOpenDownload}
+      onMouseLeave={this.onHideDeleteConfirmation}
       className={cx({
         downloadItem: true,
+        deleteConfirmationVisible: this.props.downloadDeleteConfirmationVisible,
         [this.props.download.get('state')]: true
       })}>
+      {
+        this.props.downloadDeleteConfirmationVisible
```
siemiatj: You could also do `this.props.downloadDeleteConfirmationVisible && <Component ... />`

michalbe: I just followed the pattern from rest of the codebase, but thanks for the effort.

test/unit/app/renderer/downloadItemTest.js
```diff
@@ -53,13 +50,19 @@ describe('downloadItem component', function () {
       before(function () {
         this.downloadId = uuid.v4()
         this.download = newDownload(state)
-        this.result = mount(<DownloadItem downloadId={this.downloadId} download={this.download} downloadDeleteConfirmationVisible={this.download.downloadDeleteConfirmationVisible} />)
+        this.result = mount(
+          <DownloadItem
+            downloadId={this.downloadId}
+            download={this.download}
+            deleteConfirmationVisible={this.download.get('deleteConfirmationVisible')} />
```
bsclifton: NOTE: this was wrong before.  Before, it was `this.download.downloadDeleteConfirmationVisible`. Accessing child objects from an Immutable object will return `undefined`.

I updated this to use the Immutable.js `get` method to fetch the variable

michalbe: Thanks! I never worked with ImmutableJS before, so most of my code in there was just a blind guess.

bsclifton: @michalbe no worries- it's something that bites me all the time (because it fails silently). Frankly, I wish an error would be thrown when it's accessed that way. I can't count the # times I've debugged an issue relating to that 😄

test/unit/app/renderer/downloadItemTest.js
```diff
@@ -119,16 +122,39 @@ describe('downloadItem component', function () {
...
+          assert(spy.called)
+        } finally {
+          appActions.showDownloadDeleteConfirmation.restore()
+        }
       })
+
+      if ([downloadStates.CANCELLED, downloadStates.INTERRUPTED, downloadStates.COMPLETED].includes(state)) {
+        describe('when delete button has been clicked', function () {
+          beforeEach(function () {
+            this.result.setProps({
```
michalbe: THIS is the part I missed! Thanks!

test/unit/app/renderer/downloadItemTest.js
```diff
@@ -119,16 +122,39 @@ describe('downloadItem component', function () {
...
+              deleteConfirmationVisible: true
+            })
+          })
+
+          testButton('.confirmDeleteButton', [downloadStates.CANCELLED, downloadStates.INTERRUPTED, downloadStates.COMPLETED], function (button) {
+            const spy = sinon.spy(appActions, 'downloadDeleted')
+            try {
```
michalbe: Why there's a `try` block in here?

bsclifton: In case something fails.  For example, with your commit, it failed because 0 items were found.   This raised an exception which then did not properly restore the sinon spy.  When the next test case ran, it would try to put a spy on the same method but fail (because spy has already been done).  The try is required only because I wanted to use a finally to restore the method which had been spied

michalbe:
@bsclifton fixed
download-confirmation

michalbe:
@bsclifton fixed and conflict resolved

PR #7077, opened 2017-02-04
String change for overdue contributions

da2x:
The word “next” is redundant as we’re talking about the present.


bradleyrichter:
I approve. And agree!

Issue #6901, opened 2017-01-28
alert() popups should appear below tabs


ayumi:
- Did you search for similar issues before submitting this one?
Yes
- Describe the issue you encountered:
JS `alert()` messages appear above the [Line of Death](https://textslashplain.com/2017/01/14/the-line-of-death/), suggesting that alert messages are triggered by Brave the browser and not the webpage.
- Platform (Win7, 8, 10? macOS? Linux distro?):
MacOS 10.11.6
- Brave Version (revision SHA):
0.13.0 @ 8ca0127
- Steps to reproduce:
    1. triggger a JS alert. you could open the dev console with Cmd+Alt+J then run `alert("i love pyramids")`
    
- Actual result:
<img width="785" alt="screen shot 2017-01-27 at 21 10 52" src="https://cloud.githubusercontent.com/assets/521861/22394397/245f4ca6-e4d5-11e6-9fac-971d6fe76553.png">

- Expected result:
Alert should appear entirely below browser tabs, within the webpage.
Also probably shouldn't have a Lion icon.

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
Yes
- Is this an issue in the currently released version?
Yes
- Can this issue be consistently reproduced?
Yes

- Appendix A: Chrome
<img width="663" alt="screen shot 2017-01-27 at 20 57 47" src="https://cloud.githubusercontent.com/assets/521861/22394345/5228d3b6-e4d3-11e6-8099-918f3628129b.png">

- Appendix B: Firefox
<img width="708" alt="screen shot 2017-01-27 at 21 05 11" src="https://cloud.githubusercontent.com/assets/521861/22394377/5abe4c1c-e4d4-11e6-9794-ab74eba3f7d7.png">

bradleyrichter:
+ 1 for inter-mockup humor

Liunkae:
Related issue: #2755 (prevent js alert spoofing attacks)

bsclifton:
Moving to 0.13.4

PR #7107, opened 2017-02-06
Rework alert/confirm to be tab-modal (instead of application-modal)

bsclifton:
## Test Plan
Testing alert
1. Visit http://jsbin.com/fiyojusahu/edit?html,output
2. In the `output` area, click the `Click me to test an alert` button
3. Confirm the alert works as expected

Testing confirm 
1. Visit http://jsbin.com/sadunogefu/edit?html,output
2. In the `output` area, click the `Click me to test a confirm` button
3. Confirm the confirm works as expected (try clicking OK and also cancel)

See issues for issue specific tests:
#3794 - Hard to exit/close Brave when site spams you with message box / alerts
#2755 - prevent js alert spoofing attacks
#6901 - alert() popups should appear below tabs
#7213 - Brave crashed on macOS 10.12.3 via.window.open()
#7280 - Invalid sync code alert needs proper heading

## Automated tests
```bash
# unit tests
npm run unittest -- --grep="MessageBox component unit tests"
npm run unittest -- --grep="tabMessageBoxState unit tests"
npm run unittest -- --grep="tabMessageBox unit tests"
npm run unittest -- --grep="tabState unit tests"
npm run unittest -- --grep="Main component unit tests"
npm run unittest -- --grep="NavigationBar component unit tests"
npm run unittest -- --grep="UrlBar component unit tests"
npm run unittest -- --grep="UrlBarIcon component unit tests"

# webdriver tests
npm run uitest -- --grep="MessageBox component tests"
```

Rework alert/confirm to be tab-modal (instead of application-modal)

Message box:
- shows an on/off switch for "Prevent this page from creating additional dialogs". Fixes https://github.com/brave/browser-laptop/issues/3794
- does not include Brave logo; shows origin; does not need to show "switch to tab" because it's already bound to tab. Fixes https://github.com/brave/browser-laptop/issues/2755
- is shown BELOW tabs now. Fixes https://github.com/brave/browser-laptop/issues/6901

Other changes include:
- Removed "An embedded page at " from source message (shown on every message)
- Rename showMessageBox/hideMessageBox/clearMessageBoxes actions to showNotification/hideNotification/clearNotification
- Converted most styles to Aphrodite

Auditors: @bridiver, @diracdeltas
Folks who may also be interested: @bbondy, @darkdh

## Special note
Electron [recently received a PR to implement an optional checkbox](https://github.com/electron/electron/pull/8590) in electron.dialog.showMessageBox... but this is a better choice for us because it's tab-modal (you can switch tabs while the alert is up, etc).

## Screenshots

screen shot 2017-02-06 at 10 32 02 am

screen shot 2017-02-06 at 10 30 04 am

screen shot 2017-02-06 at 10 27 50 am

screen shot 2017-02-06 at 11 48 20 am

cc: @ayumi


luixxiul:
Nice work @bsclifton ! Does this PR fix the spoofing dialog like #4992? (I know it's labelled duplicated, I just wanted to make it clear)

app/browser/messageBox.js
```diff
@@ -0,0 +1,47 @@
...
+  show: (tabId, detail, cb) => {
+    messageBoxCallbacks[tabId] = cb
+    setImmediate(() => {
+      appActions.showMessageBox(tabId, detail)
+    })
+  },
+
+  close: (state, action) => {
+    state = makeImmutable(state)
+    action = makeImmutable(action)
```
diracdeltas: [minor] setResponse calls setImmutable on the args, so it's not necessary here

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,138 @@
...
+  get messageBoxButtons () {
+    const buttons = []
+
+    for (let index = (this.buttons.size - 1); index > -1; index--) {
+      buttons.push(<Button l10nId={this.buttons.get(index)}
+        className={index === 0 ? 'primaryButton' : 'whiteButton'}
+        onClick={this.onDismiss.bind(this, this.tabId, index)} />)
+    }
+
+    return buttons
```
diracdeltas: [minor] could return `this.buttons.map(...)` instead of iterating through `this.buttons`

bsclifton:
@luixxiul yes- it does.  By removing the logo and also showing the origin, I believe the concern raised in https://github.com/brave/browser-laptop/issues/4992 should be fixed. @diracdeltas can also confirm 😄

app/renderer/components/messageBox.js
bsclifton: I did originally implement it like that- but the buttons have to be shown in reverse order (in order to keep the same "interface" as the electron `dialog.showMessageBox`).

ex: buttons would be ['OK', 'Cancel'] with cancel being index 1 (and the caller specifies cancelId === 1)

app/common/state/messageBoxState.js
```diff
@@ -0,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const tabState = require('./tabState')
+const {makeImmutable} = require('./immutableUtil')
+
+const messageBoxDetail = 'messageBoxDetail'
```
diracdeltas: does this need to be added to `docs/state.md`? does it need to be cleaned  from tabs on shutdown?

app/index.js
```diff
@@ -356,39 +356,32 @@ app.on('ready', () => {
...
         message,
         title,
-        buttons: buttons
-      })
+        buttons: ['OK'],
+        suppress: false,
+        showSuppress: shouldDisplaySuppressCheckbox
+      }
+
+      MessageBox.show(tabId, detail, cb)
```
diracdeltas: [minor] the `enter` key should be bound to `OK`

diracdeltas:
the test plan worked for me. :fireworks: 

however, now that the modal is below the line-of-trust (AKA fully in the webview area), it can be fully spoofed by the page. this is not the case in Chrome/etc. because the modal is its own window, which can be moved outside the webview area; in our case it is immovable.

i see two potential downsides:
1. this can be abused by the page to make it appear that Brave keeps showing dialogs even though the user has checked the 'suppress' checkbox. 
2. if the alert box is covering something that the user wants to read, they have to close the alert to read it.

app/renderer/components/messageBox.js
diracdeltas: i think you can do `maxIndex - index` to get the index reversed

bridiver:
accidentally made the comments in your repo @bsclifton, but looks like they still appear above

bsclifton:
Thanks for the feedback, folks! I'm working on tests and am/will be addressing all of the above items 😄

app/renderer/components/styles/commonStyles.js
```diff
@@ -19,6 +19,17 @@ const styles = StyleSheet.create({
     outline: 'none',
     padding: '0.4em',
     width: '100%'
+  },
+  'flyoutDialog': {
+    backgroundColor: `${globalStyles.color.toolbarBackground}`,
+    borderRadius: `${globalStyles.radius.borderRadius}`,
```
luixxiul: Do we need the blackets here? I think this should work: `backgroundColor: globalStyles.color.toolbarBackground`. cf: https://github.com/brave/browser-laptop/pull/7107/files#diff-fa59d1e0221a5bee94fd2abd9b96dcfdR12

app/renderer/components/styles/commonStyles.js
```diff
@@ -19,6 +19,17 @@ const styles = StyleSheet.create({
...
+    boxShadow: '2px 2px 8px #3b3b3b',
+    color: '#000',
+    fontSize: '13px',
+    padding: '10px 30px',
+    position: 'absolute',
+    textAlign: 'left',
+    top: `${globalStyles.spacing.dialogTopOffset}`
```
luixxiul: ditto

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,138 @@
...
+    marginBottom: '1em',
+    marginTop: '0.5em'
+  },
+  'body': {
+    marginTop: '1.5em',
+    minWidth: '425px',
+    marginBottom: '1.5em'
+  },
+  'buttons': {
+    float: 'right',
```
luixxiul: This works enough but I would not set with float to avoid the possibility of the case like this: https://github.com/brave/browser-laptop/issues/6749#issuecomment-274096915. I would instead use display: flex and justify-content: flex-end

app/renderer/components/styles/commonStyles.js
```diff
@@ -19,6 +19,17 @@ const styles = StyleSheet.create({
     outline: 'none',
     padding: '0.4em',
     width: '100%'
+  },
+  'flyoutDialog': {
```
luixxiul: I'm going to create a tracking issue to delete https://github.com/brave/browser-laptop/blob/4fb5d5ae8202b5afffff4e3ce106a64d81fbfdb4/less/forms.less#L8-L18 in favor of these lines  :-)

bridiver:
@diracdeltas I'm not sure if we need to address those issues right now. A page can spoof an alert dialog and even if it doesn't resemble what Brave does, a lot of users don't know the difference anyway. The big issue I think is preventing people from navigating away when that happens and this fixes that issue. For the second issue, the old dialog could potentially cover content as well. Maybe file separate issues for them as improvements?

app/renderer/components/styles/commonStyles.js
luixxiul: posted here: https://github.com/brave/browser-laptop/issues/7114

bsclifton:
Round 1 of fixes done w/ https://github.com/brave/browser-laptop/pull/7107/commits/5f2d239e2e5e0c0b335f8ea77ac12d2aef757256

Coming up next:
- binding enter/escape keys
- checking tab order on controls
- webdriver tests enforcing the behaviors called out above in feedback (renderer crash, etc)

@luixxiul regarding the `float: right` comment, did you want to push a fix? (my flexbox skills are not the best) I'd love to have a proper style update (and you're welcome to commit one 😄 )

bsclifton:
Also one open I had for you all ( @bridiver, @diracdeltas ):

While writing my tests, I noticed that (because the UI isn't blocked), you can type in the URL bar (including hitting enter) when an alert is up. Nothing happens of course... at least, until you close the alert/confirm.

How should we address this behavior? This could be a chance to address the feedback about pages being able to fake the alert. We can update our alert to disable/gray out the URL bar or any other action we'd like.  I do think it's worth keeping the ability to switch tabs though

cc: @bradleyrichter

luixxiul:
> regarding the float: right comment, did you want to push a fix?

I will do with a follow-up task!

bradleyrichter:
@bsclifton It would be better to make the URL bar non-editable and perhaps disabled in appearance for this case if you can not actually execute a new search or page change in this state.

diracdeltas:
> Maybe file separate issues for them as improvements?

sgtm

bridiver:
@bradleyrichter we can go either way with it. We can make it editable and dismiss the dialog automatically on navigation or we can prevent navigation and make it non-editable. Personally I think we should make it editable/navigable, but right now I think whatever is easiest

bradleyrichter:
There are several use cases where you can manipulate the URL without leaving, or changing anything which leaves the page intact, and thus - the dialog. So the dialog would need to close upon loading of a different URL. 

@bridiver This would be the easiest for the user. (vs disabling anything)

bsclifton:
Moving to 0.13.4

bsclifton:
Working through webdriver tests; I'll be having to make modifications to chromedriver to get this working.  It's currently timing out because it thinks the alert is shown:
```
Error retreiving window handle CDwindow-e342606f-d6ec-4a63-a899-6d242dfa8f28 { Error
  message: 'unexpected alert open: {Alert text : this is an alert}',
  type: 'RuntimeError',
  seleniumStack:
   { status: 26,
     type: 'UnexpectedAlertOpen',
     message: 'A modal dialog was open, blocking this operation',
     orgStatusMessage: 'unexpected alert open: {Alert text : this is an alert}\n  (Session info: content shell=)\n  (Driver info: chromedriver=2.25 (314da7cc1e56fc9fa9271bac2b029922feb4b6f2),platform=Mac OS X 10.12.3 x86_64)' } }
    1) shows when alert() is called
```

app/index.js
```diff
@@ -358,39 +358,32 @@ app.on('ready', () => {
   process.on('window-alert',
     (webContents, extraData, title, message, defaultPromptText,
```
diracdeltas: i think if you replaced braveExtensionId in `title` with 'Brave', that would resolve https://github.com/brave/browser-laptop/issues/7280

bsclifton:
OK this is getting close 😄  (finally!)

@bridiver I need your help with the remaining tests.  I need help understanding a good approach for testing that the callback is called for these scenarios
- when tab is closed without dismissing alert
- when window is closed without dismissing alert
- when tab crashes before dismissing alert
- when tab navigates away before dismissing alert

Here's an example for the first one (tab with alert is closed). I can easily verify that the tabState is gone, but need help w/ a good approach for detecting callback called or not.
https://github.com/bsclifton/browser-laptop/blob/dbeada5da3f6e66b75930307f40d767514f2d284/test/app/renderer/components/messageBoxTest.js#L285

bsclifton:
Really good progress made (I checked off the cases which are done).  @bridiver please review 😄 

Besides the above item I need to implement (`when tab navigates away before dismissing alert`), I need help testing the remaining functionality (ex: test which verifies that callback was actually called).  That is something not covered yet

bsclifton:
Here's a screencap of the only issue which I know remains.  I think we can ship without fixing (behavior seems OK)- but am curious what others think
alert-issue

app/common/state/tabState.js
```diff
@@ -183,6 +183,7 @@ const api = {
   getPersistentState: (state) => {
     // TODO(bridiver) - handle restoring tabs
     state = makeImmutable(state)
+    // TODO(bsclifton) - if above TODO is implemented, we may not want to save/restore all tab data (ex: alerts, etc)
```
bridiver: we definitely don't want to restore alerts because they won't be connected to the original method call anyway

bsclifton:
Ready for re-review! cc: @bridiver, @diracdeltas, @bbondy 

This PR is mostly tests- it adds 93 tests (81 unit tests, 12 webdriver).  Be sure to check all the scenarios tested... they should be covered 😄  If you want to run the tests, you'll find the `npm run` commands in the top post here ^^

I've also manually tested each scenario. Should be good to go 😄

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+
+const tabMessageBox = {
+  init: (state, action) => {
+    process.on('window-alert', (webContents, extraData, title, message, defaultPromptText,
+      shouldDisplaySuppressCheckbox, isBeforeUnloadDialog, isReload, cb) => {
+      const tabId = webContents.getId()
+      const detail = {
+        message,
+        title,
+        buttons: ['OK'],
```
NejcZdovc: I think we should add translated string here

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+      tabMessageBox.show(tabId, detail, cb)
+    })
+
+    process.on('window-confirm', (webContents, extraData, title, message, defaultPromptText,
+        shouldDisplaySuppressCheckbox, isBeforeUnloadDialog, isReload, cb) => {
+      const tabId = webContents.getId()
+      const detail = {
+        message,
+        title,
+        buttons: ['OK', 'Cancel'],
```
NejcZdovc: I think we should add translated string here

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+    const msgBoxTitle = (this.props.detail && this.props.detail.get('title')) || ''
+    return msgBoxTitle.replace(config.braveExtensionId, 'Brave')
+  }
+
+  get message () {
+    return (this.props.detail && this.props.detail.get('message')) || ''
+  }
+
+  get buttons () {
+    return (this.props.detail && this.props.detail.get('buttons')) || makeImmutable(['OK'])
```
NejcZdovc: I think we should add translated string here

docs/state.md
```diff
@@ -268,18 +268,27 @@ AppStore
...
-    audible: boolean,  // is audio playing (muted or not)
+    audible: boolean, // is audio playing (muted or not)
     canGoBack: boolean, // the tab can be navigated back
-    canGoForward: boolean. // the tab can be navigated forward
-    muted: boolean,  // is the tab muted
-    windowId: number  // the windowId that contains the tab
+    canGoForward: boolean, // the tab can be navigated forward
+    muted: boolean, // is the tab muted
+    windowId: number, // the windowId that contains the tab
+    messageBoxDetail: { // fields used if showing a message box for a tab
```
NejcZdovc: Please order these and child properties alphabetically

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+        showSuppress: shouldDisplaySuppressCheckbox
+      }
+
+      tabMessageBox.show(tabId, detail, cb)
+    })
+
+    process.on('window-prompt', (webContents, extraData, title, message, defaultPromptText,
+          shouldDisplaySuppressCheckbox, isBeforeUnloadDialog, isReload, cb) => {
+      console.warn('window.prompt is not supported yet')
+      let suppress = false
```
NejcZdovc: use const

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+  show: (tabId, detail, cb) => {
+    messageBoxCallbacks[tabId] = cb
+    setImmediate(() => {
+      appActions.tabMessageBoxShown(tabId, detail)
+    })
+  },
+
+  close: (state, action) => {
+    action = makeImmutable(action)
+    let tabId = action.get('tabId')
```
NejcZdovc: use const

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+    let detail = action.get('detail')
```
NejcZdovc: use const

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+    let cb = messageBoxCallbacks[tabId]
```
NejcZdovc: use const

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+    return state
+  },
+
+  onTabClosed: (state, action) => {
+    action = makeImmutable(action)
+    const tabId = action.getIn(['tabValue', 'tabId'])
+    if (tabId) {
+      // remove callback; call w/ defaults
+      let cb = messageBoxCallbacks[tabId]
```
NejcZdovc: use const

app/browser/tabMessageBox.js
```diff
@@ -0,0 +1,130 @@
...
+    const detail = tabMessageBoxState.getDetail(state, tabId)
+    if (detail && detail.get('opener')) {
+      const url = action.getIn(['tabValue', 'url'])
+      // check if user has navigated away from site which opened the alert
+      if (url && url !== detail.get('opener')) {
+        const removeAction = makeImmutable({tabId: tabId})
+        // remove detail from state
+        state = tabMessageBoxState.removeDetail(state, removeAction)
+        // remove callback; call w/ defaults
+        let cb = messageBoxCallbacks[tabId]
```
NejcZdovc: use const

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+const KeyCodes = require('../../common/constants/keyCodes')
+const config = require('../../../js/constants/config')
+const {makeImmutable} = require('../../common/state/immutableUtil')
+const {StyleSheet, css} = require('aphrodite')
+const commonStyles = require('./styles/commonStyles')
+
+class MessageBox extends ImmutableComponent {
+  constructor () {
+    super()
+    this.onClick = this.onClick.bind(this)
```
NejcZdovc: I think this is not necessary, because you are just stopping propagation in this function

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+      buttons.push(<Button l10nId={this.buttons.get(index)}
+        className={index === 0 ? 'primaryButton' : 'whiteButton'}
+        onClick={this.onDismiss.bind(this, this.tabId, index)} />)
+    }
+
+    return buttons
+  }
+
+  render () {
+    return <Dialog className='noOutline'>
```
NejcZdovc: We should use Aphrodite here as well

less/dialogs.less
```diff
@@ -16,6 +16,10 @@
   align-items: center;
   background: rgba(0, 0, 0, 0.15);
 
+  &.noOutline {
```
NejcZdovc: We should use Aphrodite

test/unit/app/renderer/components/messageBoxTest.js
```diff
@@ -0,0 +1,338 @@
...
+        const detail2 = Object.assign({}, detail1)
+        detail2.title = config.braveExtensionId + ' says:'
+        const wrapper = shallow(
+          <MessageBox
+            tabId={tabId}
+            detail={Immutable.fromJS(detail2)}
+          />
+        )
+        const instance = wrapper.instance()
+        assert.equal(instance.title, 'Brave says:')
```
diracdeltas: ++

diracdeltas:
this seems to make the notificationBar login form tests fail: https://travis-ci.org/brave/browser-laptop/builds/205211560#L5533.

they pass on master with `npm run test -- --grep='notificationBar'` but the same command also fails locally in this branch

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+}
+
+const styles = StyleSheet.create({
+  'container': {
+    outline: 'none',
+    display: 'block'
+  },
+  'title': {
+    fontWeight: 'bold',
+    fontSize: '12pt',
```
cezaraugusto: what's the reason of using `pt` instead of `px`?

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+        <div className={css(styles.buttons)} data-test-id='msgBoxButtons'>
+          {this.messageBoxButtons}
+        </div>
+      </div>
+    </Dialog>
+  }
+}
+
+const styles = StyleSheet.create({
+  'container': {
```
cezaraugusto: no need to use quotes on keys for Aphrodite

app/renderer/components/messageBox.js
cezaraugusto: what's the reason of using `pt` instead of `px`?

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+    outline: 'none',
+    display: 'block'
+  },
+  'title': {
+    fontWeight: 'bold',
+    fontSize: '12pt',
+    marginBottom: '1em',
+    marginTop: '0.5em',
+    '-webkit-user-select': 'text'
```
cezaraugusto: would be better using `WebkitUserSelect` instead of stringifying

app/renderer/components/messageBox.js
```diff
@@ -0,0 +1,165 @@
...
+  },
+  'body': {
+    marginTop: '1.5em',
+    minWidth: '425px',
+    marginBottom: '1.5em',
+    '-webkit-user-select': 'text'
```
cezaraugusto: ditto

app/renderer/components/urlBarIcon.js
```diff
@@ -99,10 +99,15 @@ class UrlBarIcon extends ImmutableComponent {
...
+    if (!this.props.activeTabShowingMessageBox) {
+      props.draggable = true
+      props.onClick = this.onClick
+      props.onDragStart = this.onDragStart
+    }
     return <span
-      onDragStart={this.onDragStart}
-      draggable
-      onClick={this.onClick}
+      {...props}
```
cezaraugusto: ++ spread operator ftw

app/renderer/components/messageBox.js
bsclifton: Good question- I was thinking it would be nice to use official point sizes and then have everything else relative using em.  @NejcZdovc suggested using rem for font size.  We should work together and propose a standard cc: @luixxiul

app/renderer/components/messageBox.js
bsclifton: @cezaraugusto does this work even for proprietary extensions? (starting with a -)?

app/renderer/components/messageBox.js
bsclifton: good catch! It already is being translated (these strings get passed to l10n-Id in render)... however, it's expected (per the app.properties file) to be lowercase. I fixed and verified it works by changing the string to nonsense 😄

app/renderer/components/messageBox.js
bsclifton: Also good catch- I wasn't sure about it at first, because Dialog is not an Aphrodite component... but adding a className using Aphrodite works just fine 👍

app/renderer/components/messageBox.js
NejcZdovc: Yeah I notice that button component has I10n-Id, but when I changed to the another language there was still the same string.

app/renderer/components/messageBox.js
bsclifton: Good call on this- I copied from another modal and it was unnecessary. It may have originally been put there to prevent the modal from closing on click... but that doesn't happen here 👍

docs/state.md
```diff
@@ -279,16 +279,16 @@ AppStore
     audible: boolean, // is audio playing (muted or not)
     canGoBack: boolean, // the tab can be navigated back
     canGoForward: boolean, // the tab can be navigated forward
-    muted: boolean, // is the tab muted
-    windowId: number, // the windowId that contains the tab
     messageBoxDetail: { // fields used if showing a message box for a tab
       message: string,
```
NejcZdovc: These properties should be alphabetize as well

bsclifton: There is a comment showing some which are session and some which are persistent. We can alphabetize but then remove that label, but I don't see value with that

bsclifton:
```js
tabs: [{
    // persistent properties
    active: boolean,  // whether the tab is selected
    favIconUrl: string,
    id: number,
    index: number,  // the position of the tab in the window
    title: string,
    url: string,
    windowUUID: string,  // the permanent identifier for the window
    // session properties
    audible: boolean, // is audio playing (muted or not)
    canGoBack: boolean, // the tab can be navigated back
    canGoForward: boolean, // the tab can be navigated forward
    messageBoxDetail: { // fields used if showing a message box for a tab
      message: string,
      title: string, // title is the source; ex: "brave.com says:"
      buttons: [string], // array of buttons as string; code only handles 1 or 2
      suppress: boolean, // if true, show a suppress checkbox (defaulted to not checked)
      showSuppress: boolean, // final result of the suppress checkbox
      cancelId: number // optional: used for a confirm message box
    },
    muted: boolean, // is the tab muted
    windowId: number // the windowId that contains the tab
  }],
```

NejcZdovc: I am talking about these properties: 

```
messageBoxDetail: { // fields used if showing a message box for a tab
      message: string,
      title: string, // title is the source; ex: "brave.com says:"
      buttons: [string], // array of buttons as string; code only handles 1 or 2
      suppress: boolean, // if true, show a suppress checkbox (defaulted to not checked)
      showSuppress: boolean, // final result of the suppress checkbox
      cancelId: number // optional: used for a confirm message box
    }
```

bsclifton: doh my bad- fixed 😄

NejcZdovc:
Should we change this two as well:

- https://github.com/brave/browser-laptop/blob/master/app/importer.js#L236
- https://github.com/brave/browser-laptop/blob/master/app/importer.js#L250

bsclifton:
@NejcZdovc yes- great catch 😄  I went ahead and updated the import code to also use this alert.

bbondy:
I assume there will be some follow ups but this is looking very good, great work on this.  I'll merge now.

app/renderer/components/messageBox.js
cezaraugusto: yes, you can double-check looking at devTools for resulted code. I guess Aphrodite uses React built-in styles vendor mechanism. For reference:  

https://facebook.github.io/react/docs/dom-elements.html#style

PR #7121, opened 2017-02-08
Fix websockets bypass in Brave filtering

### LINKED ISSUE #6997: investigate webRequest bypasses
related: #6831 #6843 

currently ads may be able to bypass Brave's builtin blocking by being excluded from webRequest interception. i suspect websockets is the major one, based on https://bugs.chromium.org/p/chromium/issues/detail?id=129353. 

we should also ignore https://cs.chromium.org/chromium/src/extensions/browser/api/web_request/web_request_permissions.cc?sq=package:chromium&dr=C&l=46 for the Brave extension

diracdeltas:
Fix #6997.
Requires https://github.com/brave/muon/pull/160

Auditors: @bridiver

Test Plan: covered by automated test

### DIFF
--- a/test/components/braveryPanelTest.js
+++ b/test/components/braveryPanelTest.js
@@ -268,6 +268,24 @@ describe('Bravery Panel', function () {
         .click(blockAdsOption)
         .waitForTextValue(adsBlockedStat, '2')
     })
+    it('blocks websocket tracking', function * () {
+      const url = Brave.server.url('websockets.html')
+      yield this.app.client
+        .waitForDataFile('adblock')
+        .tabByIndex(0)
+        .loadUrl(url)
+        .waitForTextValue('#result', 'success')
+        .waitForTextValue('#error', 'error')
+        .openBraveMenu(braveMenu, braveryPanel)
+        .waitForTextValue(adsBlockedStat, '1')
+        .click(adsBlockedStat)
+        .waitUntil(function () {
+          return this.getText('.braveryPanelBody li')
+            .then((body) => {
+              return body[0] === 'ws://ag.innovid.com/dv/sync?tid=2'
+            })
+        })
+    })
     // TODO(bridiver) using slashdot won't provide reliable results so we should
     // create our own iframe page with urls we expect to be blocked
     it('detects blocked elements in iframe in private tab', function * () {
--- a/test/fixtures/websockets.html
+++ b/test/fixtures/websockets.html
@@ -1,0 +1,14 @@
+<div id='result'></div>
+<div id='error'></div>
+<script>
+  // This should get blocked by adblock if it is on
+  var exampleSocket = new WebSocket("ws://ag.innovid.com/dv/sync?tid=2");
+  exampleSocket.onerror = (e) => {
+    document.getElementById('error').innerText = 'error'
+  }
+  // This should be a valid connection
+  var ws = new WebSocket("wss://wss.websocketstest.com:443/service")
+  ws.onopen = function() {
+    document.getElementById('result').innerText = 'success'
+  }
+</script>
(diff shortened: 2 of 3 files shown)


test/components/braveryPanelTest.js
```diff
@@ -268,6 +269,56 @@ describe('Bravery Panel', function () {
...
+              if (time > 0) {
+                time2 = time
+                return true
+              }
+              return false
+            })
+        })
+      // Assume blocking websockets is working if the time difference between
+      // onerror firing with adblock on vs off is sufficiently high.
+      assert(time2 - time1 > 100)
```
diracdeltas: fwiw this test would be nicer if there were a way in webdriver to inspect the page console output. if the connection is blocked in app/filtering.js, it will show a CONNECTION_BLOCKED message. it shows a different error message when the connection is attempted but fails.

test/fixtures/websockets.html
```diff
@@ -0,0 +1,14 @@
+<div id='result'></div>
+<div id='errorTime'></div>
+<script>
+  // This should get blocked by adblock if it is on
+  var exampleSocket = new WebSocket("ws://ag.innovid.com/dv/sync?tid=2");
```
bridiver: can we use webpack for the blocked ws to prevent a false positive for the error?

test/components/braveryPanelTest.js
```diff
@@ -268,6 +269,56 @@ describe('Bravery Panel', function () {
         .click(blockAdsOption)
         .waitForTextValue(adsBlockedStat, '2')
     })
+    it('blocks websocket tracking', function * () {
+      const url = Brave.server.url('websockets.html')
+      let time1 = 0
+      let time2 = 0
+      yield this.app.client
+        .waitForDataFile('adblock')
```
bridiver: we should really add a way to mock the adblock data so we don't have to wait for the download or worry about connection errors and changes in the adblock file contents

test/components/braveryPanelTest.js
bridiver: shouldn't it succeed if it isn't blocked?

test/fixtures/websockets.html
diracdeltas: not sure i understand. the webpack ws URL is not in adblock, so it shouldn't get blocked?

test/components/braveryPanelTest.js
diracdeltas: @bridiver no because "ws://ag.innovid.com/dv/sync?tid=2" doesn't support websockets. it fails either way: either because it's blocked by adblock (the fast case) or because the server response is incorrect (the slow case)

test/fixtures/websockets.html
bridiver: see comment above about mocking, but are there any patterns for ad block that have a wildcard domain? Another options would be to have a separate adblock file for testing or add a localhost match with a uuid path (wildcard port)

test/components/braveryPanelTest.js
bridiver: that seems really fragile to me. I think we should mock up or add something that should be blocked. For now maybe we could even do it through site hacks

diracdeltas:
talked to brianj on slack. the underlying issue causing this PR to be fragile is that the tabId for websockets is -1 because `GetRenderFrameForRequest` is false for ws and wss URLs. however this may be fixed with the Chromium 56 upgrade [1] so i'll redo this PR after that is out.

[1] https://chromium.googlesource.com/chromium/src/+/4a593833d44a457f177f99b2c907bd0f6ae397f7

diracdeltas:
@bridiver i fixed the websockets tabid issue in https://github.com/brave/muon/pull/160 so the test is no longer fragile

diracdeltas:
muon changes are now in master

bridiver:
++

PR #7129, opened 2017-02-08
Improved dnd for bookmarks manager (dnd refactor)

### LINKED ISSUE #6246: Improve Drag and Drop for Bookmarks Manager
For the same reason as #6242 and #6245.

NejcZdovc:

Resolves #6246 

Part of this PR is also a refactoring, so that we can use dnd in other components.

@bsclifton Can you please check out my first commit `Removed dnd in url bar`. In 5837fc5 you added dnd for URLIcon. I don't see why this should be here (couldn't find anything about it in parent issue). Is it ok to remove it? If you dnd search icon in the current release you get error like this:
image

### DIFF
--- a/app/renderer/components/urlBarIcon.js
+++ b/app/renderer/components/urlBarIcon.js
@@ -6,8 +6,6 @@ const React = require('react')
 const ImmutableComponent = require('../../../js/components/immutableComponent')
 const windowActions = require('../../../js/actions/windowActions')
 const cx = require('../../../js/lib/classSet')
-const dragTypes = require('../../../js/constants/dragTypes')
-const dndData = require('../../../js/dndData')
 const {isSourceAboutUrl} = require('../../../js/lib/appUrlUtil')
 const searchIconSize = 16
 
@@ -28,7 +26,6 @@ class UrlBarIcon extends ImmutableComponent {
   constructor () {
     super()
     this.onClick = this.onClick.bind(this)
-    this.onDragStart = this.onDragStart.bind(this)
   }
   get isSecure () {
     return this.props.isHTTPPage &&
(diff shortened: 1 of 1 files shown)


bbondy:
all browsers allow you to drag that to bookmarks bar to create a  bookmark, I don't think removing it is something we should do.

NejcZdovc:
@bbondy I am talking about URL bar, not bookmarks bar. This code: https://github.com/brave/browser-laptop/blob/af368d3caaf5038bcd33252b72a3ad023dd72280/app/renderer/components/urlBarIcon.js#L103-L104.

Try dnd search icon and you will see an error.

image

bbondy:
Yes I'm talking about the same, but you can take the space to the left of the URL and drag to the bookmarks bar to create a new bookmark from the site you're on.

bsclifton:
like @bbondy said, it's expected to be able to drag that are (which works great for me in master and our latest release too).  That is actually how I store all of my bookmarks 😄

NejcZdovc:
@bbondy @bsclifton If I am honest I was not aware this was possible. I am always creating bookmarks from tabs. I tried it and it's working, so yes we need to leave it as it is, only fix this error.

bbondy:
ya I always create my bookmarks using it too, previously browsers used to put a site's favicon in the urlbar, but that led to sites using lock icons as their favicon which is a security problem, so browsers stopped displaying the lock there, but still allow dragging from that spot.

bbondy:
just a heads up that DND is going through some changes due to tab tear off work.

NejcZdovc:
Yeah I saw your PR and that's why I put a breaks on this PR and split `sites` PR.

bsclifton:
Closing this PR for the moment, since it's blocked.  Please do reopen when you have time to revisit 😄  Thanks!

PR #7136, opened 2017-02-08
Test syncing history and site settings

ayumi:
Also test that clearing history and siteSettings prevents syncing.

Test Plan:
`yarn test -- --grep="'^Syncing'"`


test/components/syncTest.js
```diff
@@ -26,6 +31,30 @@ function toHex (byteArray) {
...
+    .click(syncTab)
+    .waitForVisible(syncSwitch)
+    .click(syncSwitch)
+    .windowByUrl(Brave.browserWindowUrl)
+    .waitUntil(function () {
+      return this.getAppState().then((val) => {
+        return val.value.settings['sync.enabled'] === true
+      })
+    })
+    .pause(1000) // XXX: Wait for sync init (request AWS credentials from server)
```
diracdeltas: does it work to replace this with waitUntil sync.lastFetchTimestamp is > 0?

ayumi: it didn't work, but maybe i can try a little more

PR #7155, opened 2017-02-10
Split settings toggle component to its own files

### LINKED ISSUE #7153: Migrate preferences toggle component/helpers to its own files
## Test Plan:
* All toggle switches inside about:preferences should not be affected
  * Should toggle ON/OFF
  * Should store its last state upon a browser restart

## Original issue text
As we gradually move preferences panels to its own components (such as paymentsTab and upcoming syncTab, pluginsTab and extensionsTab), we could also migrate helpers such as settings checkboxes and alike to its own components/helper files as well.

We already have a [settings file](https://github.com/brave/browser-laptop/blob/master/app/renderer/components/settings.js) and for this issue we can include there our [settingCheckbox](https://github.com/brave/browser-laptop/blob/master/js/about/preferences.js#L119) and [siteSettingCheckbox](https://github.com/brave/browser-laptop/blob/master/js/about/preferences.js#L152) components. Thus making a better separation of concerns.

As a side effect, it will make it easier to refactor our UI to styled-components (Aphrodite).

cezaraugusto:

This is a component split and code was just migrated. The only page affected was about:preferences, and migration happened for both toggle switches.

Fix #7153

Auditors: @bsclifton

## Test Plan:
* All toggle switches inside about:preferences should not be affected
  * Should toggle ON/OFF
  * Should store its last state upon a browser restart

### DIFF
--- a/app/renderer/components/preferences/paymentsTab.js
+++ b/app/renderer/components/preferences/paymentsTab.js
@@ -11,6 +11,7 @@ const getSetting = require('../../../../js/settings').getSetting
 const settings = require('../../../../js/constants/settings')
 const ModalOverlay = require('../../../../js/components/modalOverlay')
 const coinbaseCountries = require('../../../../js/constants/coinbaseCountries')
+const {changeSetting} = require('../../lib/settingsUtil')
 const moment = require('moment')
 moment.locale(navigator.language)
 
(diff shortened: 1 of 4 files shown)


cezaraugusto:
components are already on `app/renderer/components/settings.js`, only `changeSetting` (which is only a helper) that was split to settingsUtil. 

I kept it there because of that but lmk if preferable to be on the same file as components.

Issue #6965, opened 2017-02-01
Re-add newtab background images (as an option)


bbondy:
## Test plan
1. Launch Brave and visit about:preferences#tabs
2. Default setting should be true
3. Open new tab; verify backgrounds are shown
4. Close new tab and change setting to false
5. Open new tab; verify gradients are shown

## Automated tests
```
npm run unittest -- --grep="NewTab component unit tests"
```

## Original issue text

Because of high memory usage bug in chromium, background images on newtab page were removed here:
https://github.com/brave/browser-laptop/commit/94a620a0af13a8656ff15e9c16bdfe0673dfe951

This task is to track re-adding it, but this time as an option, and we can enable it by default once the memory problem is fixed.

dschwank:
+1 would be great to have this as an option!

dschwank:
Yeah - the switch is implemented and works fine! Thanks!

PR #7160, opened 2017-02-10
Background for new tab is now toggleable via prefs

bsclifton:
## Test plan
1. Launch Brave and visit about:preferences#tabs
2. Default setting should be true
3. Open new tab; verify backgrounds are shown
4. Close new tab and change setting to false
5. Open new tab; verify gradients are shown

## Automated tests
```
npm run unittest -- --grep="NewTab component unit tests"
```

Fixes https://github.com/brave/browser-laptop/issues/6965

Noteable items include:
- New setting is called `Show images` and defaults to `true`
- Much of new tab is now unit tested. Redundant (and also slower) webdriver tests removed

Auditors: @cezaraugusto, @bbondy 

screen shot 2017-02-10 at 1 49 58 am

### DIFF
--- a/app/common/lib/randomUtil.js
+++ b/app/common/lib/randomUtil.js
@@ -1,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports.random = () => Math.random()
--- a/less/about/newtab.less
+++ b/less/about/newtab.less
@@ -20,14 +20,13 @@ body {
   background: #fff;
 }
 
-body, .dynamicBackground {
+body, .dynamicBackground, bgGradient {
   opacity: 0;
   animation: fadeIn 200ms;
   animation-fill-mode: forwards;
 }
 
 .copyrightNotice {
-  visibility: hidden !important;
   -webkit-user-select: none;
   cursor: default;
 }
@@ -43,10 +42,6 @@ ul {
     vertical-align: middle;
     list-style-type: none;
   }
-}
-
-#appContainer > div {
-  height: auto; // overrides default #appContainer > div: 100%
 }
 
 .dynamicBackground {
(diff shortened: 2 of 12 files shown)


app/common/lib/randomUtil.js
```diff
@@ -0,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports.random = () => Math.random()
```
bsclifton: This was added specifically so we can stub the method with sinon (and verify it gets called)

less/about/newtab.less
```diff
@@ -45,10 +44,6 @@ ul {
   }
 }
 
-#appContainer > div {
-  height: auto; // overrides default #appContainer > div: 100%
-}
-
```
bsclifton: This has to be removed; It looks like starting with Chromium 56, changes were made to `background-image` which causes the image to render itself at 100% DPI (rather than scaling to the size of the window)

less/about/newtab.less
```diff
@@ -20,14 +20,13 @@ body {
...
 
-body, .dynamicBackground {
+body, .dynamicBackground, bgGradient {
   opacity: 0;
   animation: fadeIn 200ms;
   animation-fill-mode: forwards;
 }
 
 .copyrightNotice {
-  visibility: hidden !important;
```
bsclifton: This is no longer needed because the FooterInfo component checks for (and hides) copyright info if no background is shown

app/common/lib/randomUtil.js
bbondy: you should still be able to stub Math.random even without this with sinon

bsclifton: I tried, but was unsuccessful ☹️  I [verified other folks have the same issue](https://stackoverflow.com/questions/17011765/stubbing-date-now-and-math-random) and then implemented it this way

bbondy: k np

bbondy:
++ nice tests

PR #7152, opened 2017-02-10
Syncing styling

ayumi:
also show device name

Auditors: @diracdeltas

screen shot 2017-02-09 at 18 35 13
screen shot 2017-02-09 at 18 35 35
screen shot 2017-02-09 at 18 35 41
screen shot 2017-02-09 at 18 35 52

PR #7164, opened 2017-02-10
Use <SettingTextbox> for the textbox on the dialog of enabling Sync

luixxiul:
For #7152

Based on https://github.com/brave/browser-laptop/pull/6882#issuecomment-276785903

Auditors: @ayumi @bsclifton

Test Plan: n/a

### DIFF
--- a/app/renderer/components/preferences/syncTab.js
+++ b/app/renderer/components/preferences/syncTab.js
@@ -12,6 +12,7 @@ const niceware = require('niceware')
 const ModalOverlay = require('../../../../js/components/modalOverlay')
 const Button = require('../../../../js/components/button')
 const {SettingsList, SettingItem, SettingCheckbox} = require('../settings')
+const {SettingTextbox} = require('../../../../app/renderer/components/textbox')
 
 const aboutActions = require('../../../../js/about/aboutActions')
 const getSetting = require('../../../../js/settings').getSetting
@@ -135,7 +136,9 @@ class SyncTab extends ImmutableComponent {
   get deviceNameInputContent () {
     return <SettingItem>
       <span data-l10n-id='syncDeviceNameInput' />
-      <input className='deviceNameInput formControl' spellCheck='false'
+      <SettingTextbox
+        data-test-id='deviceNameInput'
+        spellCheck='false'
         ref={(node) => { this.deviceNameInput = node }}
         placeholder={this.defaultDeviceName} />
     </SettingItem>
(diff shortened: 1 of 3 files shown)


luixxiul:
The change will make the input form into:

screenshot 2017-02-10 21 14 39

less/about/preferences.less
```diff
@@ -1589,9 +1589,7 @@ table.sortableTable {
   .setupContent {
     margin: 1em 0;
   }
-  .deviceNameInput {
-    width: 50%;
```
luixxiul: Please let me know if this should not be removed.

app/renderer/components/preferences/syncTab.js
```diff
@@ -117,7 +118,9 @@ class SyncTab extends ImmutableComponent {
       : getSetting(settings.SYNC_DEVICE_NAME, this.props.settings)
     return <SettingItem>
       <span data-l10n-id='syncDeviceNameInput' />
-      <input className='deviceNameInput formControl' spellCheck='false'
+      <SettingTextbox
```
ayumi: Earlier I tried to use a Textbox however I couldn't get the prop `ref={}` to work so the input didn't work[0]. I gave up and used a standard input.
This branch doesn't seem to set the device name either.
[0] Maybe this: https://facebook.github.io/react/docs/refs-and-the-dom.html

Test Plan:
- Clear profile
- Enable sync and name the device
- Observe terminal for the device name– should not be "browser-laptop"

Also tested by: `yarn test -- --grep='"^Sync Panel"'`

less/forms.less
```diff
@@ -766,6 +766,7 @@ select {
 }
 
 // From commonStyles.js
+// TODO: remove this after input element in addOverlayContent on syncTab.js is replaced with the new textbox control in Aphrodite
```
ayumi: :100:

app/renderer/components/preferences/syncTab.js
luixxiul: @bsclifton @cezaraugusto do you have any idea here?

ayumi:
Hmm, I think as per earlier https://github.com/brave/browser-laptop/pull/7164#discussion_r100586892 this patch breaks naming devices.

The relevant failing test is:
```
npm run test -- --grep='"^Sync Panel sync setup sync profile"'
```

ayumi:
I've reverted this and created https://github.com/brave/browser-laptop/issues/7360 to track the issue.

PR #7236, opened 2017-02-14
Sync bookmarks in order during initial sync

ayumi:
Test Plan:
1. `npm run test -- --grep='"^Syncing bookmarks from an"'`


diracdeltas:
++

PR #7276, opened 2017-02-16
convert download-sync-client to node script

diracdeltas:
Auditors: @bbondy

Test Plan: npm run download-sync client, then start browser and sync should work


bbondy:
works great on Windows, thanks.

Issue #7308, opened 2017-02-17
apply sync records in batches


diracdeltas:
testers found that a huge sync update (over 6000 bookmarks) locks up the UI and causes huge memory usage. i noticed that this is not a problem at all for bookmarks import, which calls `appActions.addSite(Immutable.fromJS(sites))` (for sites.length > 6000). this suggests that the issue is that we are calling `appActions.addSite` for one site at a time instead of in a batch.

we should sort sync records in `applySyncRecords` and apply them in batches. naively, i would sort things by objectData type, then apply all the CREATEs for a given type, then all the UPDATES, then all the DELETEs. cc @ayumi

PR #7309, opened 2017-02-18
batch synced site creation

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/7308

Auditors: @ayumi @alexwykoff

Test Plan:
1. in about:preferences#sync, choose the 'i have an existing sync code' option. enter in the code words that Alex posted in #testers (starts with 'forsaken')
2. open about:bookmarks and wait for 6000+ bookmarks to be synced. the browser should not freeze or use excessive memory during the sync.

### DIFF
--- a/js/state/syncUtil.js
+++ b/js/state/syncUtil.js
@@ -5,6 +5,7 @@
 
 const Immutable = require('immutable')
 const writeActions = require('../constants/sync/proto').actions
+const siteTags = require('../constants/siteTags')
 const siteUtil = require('./siteUtil')
 
 const CATEGORY_MAP = {
@@ -51,20 +52,27 @@ const pickFields = (object, fields) => {
   }, {})
 }
 
+// Cache of bookmark folder object IDs mapped to folder IDs
+let folderIdMap = new Immutable.Map()
+
 /**
- * Apply a bookmark or historySite SyncRecord to the browser data store.
+ * Converts sync records into a form that can be consumed by AppStore.
  * @param {Object} record
- * @param {Immutable.Map} siteDetail
+ * @param {Immutable.Map} appState
  */
-const applySiteRecord = (record) => {
-  const appActions = require('../actions/appActions')
-  const siteTags = require('../constants/siteTags')
+module.exports.getSiteDataFromRecord = (record, appState) => {
   const objectId = new Immutable.List(record.objectId)
   const category = CATEGORY_MAP[record.objectData].categoryName
-  const existingObject = module.exports.getObjectById(objectId, category)
-  const existingObjectData = existingObject && existingObject[1]
-  let tag
+  let existingObjectData
 
+  if (record.action !== writeActions.CREATE) {
+    // XXX: Is this necessary for CREATEs also?
+    const existingObject = module.exports.getObjectById(objectId, category,
+      appState)
+    existingObjectData = existingObject && existingObject[1]
+  }
+
+  let tag
   let siteProps = Object.assign(
     {},
     existingObjectData && existingObjectData.toJS(),
@@ -87,22 +95,11 @@ const applySiteRecord = (record) => {
     const parentFolderObjectId = siteProps.parentFolderObjectId
     if (parentFolderObjectId && parentFolderObjectId.length > 0) {
       siteProps.parentFolderId =
-        getFolderIdByObjectId(new Immutable.List(parentFolderObjectId))
+        getFolderIdByObjectId(new Immutable.List(parentFolderObjectId), appState)
     }
   }
   const siteDetail = new Immutable.Map(pickFields(siteProps, SITE_FIELDS))
-
-  switch (record.action) {
-    case writeActions.CREATE:
-      appActions.addSite(siteDetail, tag, undefined, undefined, true)
-      break
-    case writeActions.UPDATE:
-      appActions.addSite(siteDetail, tag, existingObjectData, null, true)
-      break
-    case writeActions.DELETE:
-      appActions.removeSite(siteDetail, tag, true)
-      break
-  }
+  return {siteDetail, tag, existingObjectData}
 }
 
 const applySiteSettingRecord = (record) => {
@@ -163,11 +160,20 @@ const applySiteSettingRecord = (record) => {
   }
 }
 
+const applyNonBatchedRecords = (records) => {
+  if (!records.length) { return }
+  setImmediate(() => {
+    const record = records.shift()
+    applySyncRecord(record)
+    applyNonBatchedRecords(records)
+  })
+}
+
 /**
  * Given a SyncRecord, apply it to the browser data store.
  * @param {Object} record
  */
-module.exports.applySyncRecord = (record) => {
+const applySyncRecord = (record) => {
   if (!record || !record.objectData) {
     console.log(`Warning: Can't apply empty record: ${record}`)
     return
@@ -175,7 +181,8 @@ module.exports.applySyncRecord = (record) => {
   switch (record.objectData) {
     case 'bookmark':
     case 'historySite':
-      applySiteRecord(record)
+      // these are handled in batches now
+      console.log(`Warning: Skipping unexpected site record: ${record}`)
       break
     case 'siteSetting':
       applySiteSettingRecord(record)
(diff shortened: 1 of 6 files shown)


diracdeltas:
looks like syncing bookmarks tests are failing so i need to fix those

bridiver:
I think this is part of a more general issue with appstate performance and we want to batch all updates in the dispatcher to solve some other perf issues. Is there any reason this code can't run in the background page?

diracdeltas:
> Is there any reason this code can't run in the background page?

it would be some effort to replace all the calls to `AppStore.getState()` and `appActions` with IPC between background page and browser. which part of this is most helpful to move to the background page?

bridiver:
@diracdeltas the loop that calls the appActions. That part seems like it should be pretty straightforward to move and then I can handle the batching with the appDispatcher stuff I'm working on

js/state/syncUtil.js
```diff
@@ -52,57 +52,95 @@ const pickFields = (object, fields) => {
...
+      const isFolder = (typeof siteProps.isFolder === 'boolean')
+        ? siteProps.isFolder
+        : !!existingFolderId
+      tag = isFolder
+        ? siteTags.BOOKMARK_FOLDER
+        : siteTags.BOOKMARK
+      const parentFolderObjectId = siteProps.parentFolderObjectId
+      if (parentFolderObjectId && parentFolderObjectId.length > 0) {
+        siteProps.parentFolderId =
+          getFolderIdByObjectId(new Immutable.List(parentFolderObjectId))
```
ayumi: for initial sync, if we apply all bookmarks and folders at once, this will always return null; so folders don't sync.

we could fix this by first applying folders and folderless bookmarks and folders (to ensure correct ordering); then applying foldered bookmarks. see below for more info

js/state/syncUtil.js
```diff
@@ -194,10 +232,22 @@ module.exports.applySyncRecord = (record) => {
  */
 module.exports.applySyncRecords = (records) => {
   if (!records || records.length === 0) { return }
+  const siteRecords = []
+  const otherRecords = []
+  records.forEach((record) => {
```
ayumi: bookmark folders:

instead of applying all bookmarks, i think we should first apply all folderless bookmarks and folders, then apply foldered bookmarks. otherwise we won't be able to resolve bookmark.parentFolderObjectId -> parentFolderId, so bookmarks will end up folderless.

bonus points to recursively apply (top level folders/items, then second level, etc) to better ensure folders ids are available when we need them.

js/state/syncUtil.js
diracdeltas: @ayumi i think there is a general race condition here even if we apply stuff in the right hierarchy order. see https://github.com/brave/browser-laptop/issues/7293#issuecomment-280883336

diracdeltas:
@bridiver i tried moving this code to the background script; it didn't beach-ball the browser but it did make the UI non-responsive indefinitely. i think the batching is necessary to solve the perf issue.

bridiver:
Batching is almost ready

js/state/syncUtil.js
```diff
@@ -87,22 +83,64 @@ const applySiteRecord = (record) => {
...
+      const siteDetail = siteData.siteDetail
+      const tag = siteData.tag
+      const existingObjectData = siteData.existingObjectData
+      if (record.action === writeActions.UPDATE) {
+        appActions.addSite(siteDetail, tag, existingObjectData, null, true)
+      } else if (record.action === writeActions.DELETE) {
+        appActions.removeSite(siteDetail, tag, true)
+      }
+    })
+  }, 2000)
```
bridiver: if this has to happen after creation a timeout doesn't seem very safe, or is the timeout just an extra buffer? If so I don't think it's really necessary with the dispatcher changes

js/state/syncUtil.js
```diff
@@ -51,20 +52,27 @@ const pickFields = (object, fields) => {
...
+module.exports.getSiteDataFromRecord = (record, appState) => {
   const objectId = new Immutable.List(record.objectId)
   const category = CATEGORY_MAP[record.objectData].categoryName
-  const existingObject = module.exports.getObjectById(objectId, category)
-  const existingObjectData = existingObject && existingObject[1]
-  let tag
+  let existingObjectData
+
+  if (record.action !== writeActions.CREATE) {
+    // XXX: Is this necessary for CREATEs also?
```
diracdeltas: omitting the existingObject search for CREATES reduces the sync lag significantly, from a minute or so to a couple seconds.

diracdeltas:
updated PR so now it fixes the bookmark folder issue too

ayumi:
it worked really fast for the `forsaken ...` profile!

then i tested with my special user data which is mildly corrupted (for example some bookmarks include nonexistent parentFolderIds), and it crashed.

you can replicate this by editing session-store-1 to change a bookmark's parentFolderId to something bogus, then starting Brave and doing the initial sync upload. it doesn't crash, but it prevents records from being uploaded.

diracdeltas:
@ayumi does that issue exist without this PR?

ayumi:
@diracdeltas no, it works on master as of d9234b59065113071d45e03d323f973bb242b511

diracdeltas:
@ayumi i think its fixed now

ayumi:
@diracdeltas sweet it works now!

js/stores/appStore.js
```diff
@@ -485,6 +487,32 @@ const handleAppAction = (action) => {
     case appConstants.APP_DATA_URL_COPIED:
       nativeImage.copyDataURL(action.dataURL, action.html, action.text)
       break
+    case appConstants.APP_APPLY_SITE_RECORDS:
```
bridiver: can we leave a comment here about refactoring this into a single action with APP_ADD_SITE? When we do that we should probably pull it out of appStore anyway and create a `siteReducer`

bridiver: or maybe the comment should go on the action. Either way is fine, just want to capture it in the code

diracdeltas: will do

alexwykoff:
I ran bookmark tests on 7a18689fb67c6a85849f1453b6df45cd805c8063 and found 4 failures 

  71 passing (3m)
  4 failing

  1) about:bookmarks page content "before all" hook:
     WaitUntilTimeoutError: Promise was rejected with the following reason: Error: unknown error: Cannot read property 'appStoreRenderer' of undefined
      at execute(<Function>) - brave.js:407:19

  2) Syncing bookmarks create bookmark in folder:
  Error

  3) Syncing bookmarks update bookmark, moving it into the folder:
     Promise was rejected with the following reason: Error: An element could not be located on the page using the given search parameters.
  Error

  4) Syncing bookmarks from an existing profile "before all" hook:
     WaitUntilTimeoutError: Promise was rejected with the following reason: timeout
      at windowHandles() - brave.js:208:32

diracdeltas:
@alexwykoff could you try on 9a2b535 ?

alexwykoff:
Sure!

Here's the latest:
3 failing

  1) about:bookmarks page content "before all" hook:
     WaitUntilTimeoutError: Promise was rejected with the following reason: Error: unknown error: Cannot read property 'appStoreRenderer' of undefined
      at execute(<Function>) - brave.js:407:19

  2) bookmark tests bookmarks bookmark pdf check location:
     WaitUntilTimeoutError: Promise was rejected with the following reason: timeout
      at execute(<Function>) - brave.js:413:19

  3) navigationBar tests auto open bookmarks toolbar for the first bookmark should remain hidden if user has bookmarks but has toolbar hidden:
     WaitUntilTimeoutError: element (#navigator .removeBookmarkButton) still not existing after 10000ms
      at isExisting("#navigator .removeBookmarkButton") - waitForExist.js:37:22

I don't think these are on you, they look like flakey timeouts.

bridiver:
@alexwykoff I think some of those are fixed in https://github.com/brave/browser-laptop/pull/7353 
the number of intermittent test failures is way down with those changes

diracdeltas:
syncing tests all fixed now!

srirambv:
@diracdeltas  There is a momentary spike in the CPU usage when bookmarks are sync'd. Brave CPU usage goes upto 30% and drops down. It took ~2mins to sync the bookmarks on my machine with a clean profile.

image

bridiver:
@srirambv when you say it took ~2 mins, what exactly are you referring to? Did the spike last 2 min? Was it 2 min until they appeared on another device?

srirambv:
@bridiver  The total time to sync was around 2 mins for the bookmarks to shown up on the second device. The CPU spike was momentary around 10 secs (could hear the fan loud and clear), happened initially when the sync code was added and then right before the bookmarks showed up on  the second device.

diracdeltas:
sync polls every 1 minute, so on average i would expect ~30s for a sync to occur

bridiver:
If the poll time is 1 minute the average should be closer to 1 minute to sync across devices. 2 minutes would be a theoretical worst case assuming the timers are accurate (which they aren't) and the sync process is instantaneous (which it isn't)

bridiver:
Future versions of this will improve latency with something like SQS

diracdeltas:
> If the poll time is 1 minute the average should be closer to 1 minute to sync across devices.

it sends ~immediately after the state change but the receiver polls every 1min

PR #7366, opened 2017-02-23
Torrent Viewer CSP rule blocks <iframe> content

feross:
Torrent content is rendered into <iframe> when it's not video or audio
content. For example, a .jpg or a .pdf file. This is because we're
using a viewer page that includes the content; we're not returning the
content directly.

This is because the torrent may not be active and in that case, we show
the "Start Download?" page.

When this occurs, CSP prevents the iframe from loading content from the
webtorrent server at http://localhost:port

This is because we only make a CSP exception for media elements, not
iframe elements. This is an easy fix.

Fixes: https://github.com/brave/browser-laptop/issues/7243

Test Plan:

1. Load the WIRED CD torrent from https://codepen.io/ferossity/full/qaezaB/
2. Start the torrent
3. Click on a non-media file like `poster.jpg` or `README.md` to view it in Brave.
4. It should load and display correctly.

Since this touches CSP, I'd appreciate if @diracdeltas could take a look 🔐 ✅

### DIFF
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -166,22 +166,23 @@ let generateTorrentManifest = () => {
   let cspDirectives = {
     'default-src': '\'self\'',
     // TODO(bridiver) - remove example.com when webtorrent no longer requires it
+    //                  (i.e. once Brave uses webpack v2)
     'connect-src': '\'self\' https://example.com',
     'media-src': '\'self\' http://localhost:*',
     'form-action': '\'none\'',
     'referrer': 'no-referrer',
     'style-src': '\'self\' \'unsafe-inline\'',
-    'frame-src': '\'self\''
+    'frame-src': '\'self\' http://localhost:*'
   }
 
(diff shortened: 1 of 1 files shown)


app/extensions.js
```diff
@@ -166,22 +166,23 @@ let generateTorrentManifest = () => {
...
     'form-action': '\'none\'',
     'referrer': 'no-referrer',
     'style-src': '\'self\' \'unsafe-inline\'',
-    'frame-src': '\'self\''
+    'frame-src': '\'self\' http://localhost:*'
   }
 
   if (process.env.NODE_ENV === 'development') {
     // allow access to webpack dev server resources
     let devServer = 'localhost:' + process.env.npm_package_config_port
```
feross: None of the changes below this line should be substantive. They just remove repetition that might allow the dev/prod versions of CSP to accidentally diverge.

feross:
This has been here for a while, going to go ahead and land it.

PR #7396, opened 2017-02-27
Prevent windows from being opened off-screen

### LINKED ISSUE #3558: Window off screen when monitors change
Yes

Note: I've experienced this issue before.
-  Using Windows
-  From a laptop
-  Plugged into an external monitor
-  Using "Extended" desktop rather than "mirrored"
-  Has also happened in Chrome in the past

When opening Brave, I could not locate the `ui` window.

**Expected behavior:**
When Opening brave, I should be able to see the `ui` window.
  Win 10
- Brave Version: 0.11.5
- Steps to reproduce:
  1. Use a Laptop with Windows
  2. Plug in external monitors
  3. Leave Brave Open
  4. Reboot Laptop
  5. ???
  6. Attempt to open Brave
  7. Cannot see `ui` window

**Related Symptoms for bug squashing:**
To fix this issue temporarily, I did the following.
    1.went to `C:/Users/{username}/AppdData/Roaming/brave`
    2.opened `session-store-1` in a text editor
    3.saw this json:

``` javascript
...
        "ui": {
                "downloadsToolbar": {
                    "isVisible": false
                },
                "isMaximized": false,
                "mouseInTitlebar": true,
                "position": [
                    -15244,
                    7
                ],
                "tabs": {
                    "tabPageIndex": 0
                }
            }
        }
    ]
}
```

   4.I changed the `ui.position` from [_-15244_, 7] to [**0**, 7]
   5,Now when opening brave, I could see the `ui` window
- Screenshot if needed:
- Any related issues:
  I have experienced something similar on Chrome(desktop on windows) in the past.
- Suggestion
  Maybe doing a check on the `ui.position` and making sure it isn't out of some sort of pre-defined range.  If it is out of range, then set it to 0 on start.

-Al

jonathansampson:

Test Plan:
Close Brave when the browser is positioned at a negative x or y offset. For example, move the left half of the browser off the left of the screen, leaving only the right-side of the browser visible. Close the browser. Re-open the browser. The window should not be opened to a negative x or y coordinate.

Resolves #3558

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -511,6 +511,10 @@ class Main extends ImmutableComponent {
 
     const onWindowMove = debounce(function (event) {
       const position = event.sender.getPosition()
+      // Windows may be placed off-screen
+      // See https://github.com/brave/browser-laptop/issues/3558
+      position[0] = Math.max(0, position[0])
+      position[1] = Math.max(0, position[1])
       // NOTE: the default window position is whatever the last window move was
       appActions.defaultWindowParamsChanged(undefined, position)
       windowActions.savePosition(position)
--- a/js/stores/appStore.js
+++ b/js/stores/appStore.js
@@ -155,8 +155,10 @@ const createWindow = (browserOpts, defaults, frameOpts, windowState) => {
     const maxX = Math.max(...displays.map((display) => { return display.bounds.x + display.bounds.width }))
     const maxY = Math.max(...displays.map((display) => { return display.bounds.y + display.bounds.height }))
 
-    browserOpts.x = Math.min(browserOpts.x, maxX - defaults.windowOffset)
-    browserOpts.y = Math.min(browserOpts.y, maxY - defaults.windowOffset)
+    // Windows may be placed off-screen
+    // See https://github.com/brave/browser-laptop/issues/3558
+    browserOpts.x = Math.min(Math.max(0, browserOpts.x), maxX - defaults.windowOffset)
+    browserOpts.y = Math.min(Math.max(0, browserOpts.y), maxY - defaults.windowOffset)
   }
 
   const minWidth = isModal(browserOpts) ? defaults.minModalWidth : defaults.minWidth


jonathansampson:
The change to `main.js` isn't really needed, since the commit corrects during _window creation_. Happy to make further modifications to this suggested merge. It may be the case that neither of these fixes should be made, and instead that `event.sender.getPosition` should clamp the values before returning an array.

js/components/main.js
```diff
@@ -511,6 +511,10 @@ class Main extends ImmutableComponent {
 
     const onWindowMove = debounce(function (event) {
       const position = event.sender.getPosition()
+      // Windows may be placed off-screen
+      // See https://github.com/brave/browser-laptop/issues/3558
+      position[0] = Math.max(0, position[0])
```
bridiver: should we be preventing partial off-screen placement? Seems like this could cause weird issues with multi-monitor placement or intentionally placing windows a little out of the way. I believe you can use the screen api to determine if a particular point is on any screen and I think we should probably use that to make sure that some portion of the draggable area is on the screen

bridiver: if the api doesn't provide what you need we can add the functionality to it

jonathansampson: @bridiver I feel a little strange by clamping to 0 indiscriminately for all negative values too. That being said, if 0,0 is the top-left most position of the top-left most display, I can't think of a reason why a user would have their browser hanging off that end. _I can think of reasons why they may want their browser to over-hang bottom-right._

I'll check out the screen API, to see if there's a more elegant approach. The primary issue I'd like to resolve is the case where the browser window is further to the negative than it is wide, resulting in an unreachable UI.

bridiver: oh right, it's for all windows, but depending on the configuration you could still have gaps that fall outside any window without being negative. I think testing for onscreen points would still be safer, but for now we could use 0 in the Y axis and use window width to ensure a fixed number of > 0 pixels for the X axis

bridiver: it could also go offscreen with position > max width/height

bsclifton:
Closing this PR as stale for the moment; let's discuss on Slack and please re-open if necessary.  Thanks! 😄

jonathansampson:
@bsclifton Fine to close, but we will definitely need to discuss. I suspect the issue only impacts those of us who frequently use a secondary monitor as their primary display. I've noticed similar issues with other applications, but I'm certain there's something we can do to alleviate some of the pains :)

PR #7467, opened 2017-03-02
fix "Cannot read property 'size' of undefined" error when getting persistent state

### LINKED ISSUE #7466: updating from 0.13.4 to 0.13.5 RC(1, 2) loses session-store-1 data
platform: OS X

when i update from 0.13.4 to 0.13.5 RC 1 or 2, most of my browser data (history, bookmarks, preferences, tabs, etc.) disappear. session-store-1 goes from 900+ KB to 27 KB.

this does not happen when upgrading to Preview 3, so the culprit is somewhere between 7373a01 and 0e68342

STR:
1. open 0.13.4, add some preferences and bookmarks
2. upgrade to 0.13.5 RC
4. open browser, observe that preferences and bookmarks are gone

diracdeltas:
When upgrading from 0.13.4 to 0.13.5, this was causing getting persistent state
to fail on startup, thereby losing session state. Fix #7466

Test Plan:
1. copy your session-store-1 from Brave 0.13.4 into the brave-development profile folder
2. npm start
3. Brave should show bookmarks, settings, pinned tabs, etc from the 0.13.4 profile

### DIFF
--- a/test/unit/app/common/state/tabStateTest.js
+++ b/test/unit/app/common/state/tabStateTest.js
@@ -483,6 +483,11 @@ describe('tabState unit tests', function () {
       const expectedAppState = defaultAppState.set('tabs', tabsWithoutField)
       assert.deepEqual(newAppState, expectedAppState)
     })
+    it('returns the state (unchanged) if tabs is falsey', function () {
+      const emptyTabState = defaultAppState.delete('tabs')
+      const newAppState = tabState.removeTabField(emptyTabState, 'loginRequiredDetail')
+      assert.equal(newAppState, emptyTabState)
+    })
   })
 
   describe('getPersistentState', function () {
--- a/app/common/state/tabState.js
+++ b/app/common/state/tabState.js
@@ -184,6 +184,9 @@ const api = {
     state = makeImmutable(state)
 
     let tabs = state.get('tabs')
+    if (!tabs) {
+      return state
+    }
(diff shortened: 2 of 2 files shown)


test/unit/app/common/state/tabStateTest.js
```diff
@@ -483,6 +483,10 @@ describe('tabState unit tests', function () {
       const expectedAppState = defaultAppState.set('tabs', tabsWithoutField)
       assert.deepEqual(newAppState, expectedAppState)
     })
+    it('returns the state (unchanged) if tabs is falsey', function () {
+      const newAppState = tabState.removeTabField(defaultAppState, 'loginRequiredDetail')
```
diracdeltas: seems that you would want defaultAppState without the `tabs` field for this test?

diracdeltas: otherwise lgtm

bsclifton: yup! it gets passed as input- the `tabState` handle is just the reference to what got exported in `app/common/state/tabState.js`

bsclifton: Fixing!

bsclifton: Fixed- that variable exists in other files (with other states- whoops!). Thanks for catching 😄

PR #7515, opened 2017-03-06
new window loads are sometimes dropped in chromium for unknown reason…

bridiver:
…s so periodically check and retry if necessary

cc @bbondy


bsclifton:
@bridiver is this PR still applicable?

bridiver:
it is, but I think it needs to be updated

PR #7524, opened 2017-03-07
Need new strings to handle overdue payments

### LINKED ISSUE #7078: Need new strings to handle overdue payments
* “Your first contribution will be 24 days ago.”
* “Your next contribution will be 18 days ago.”

So these need to be modified to handle the correct tense when payment is overdue. I’m uncertain about any localization implications this may have or how this should be best handled. Pull request #7077 handles a related string, but not this issue.

mrose17:
Fixes #7078

Test Plan:

- Create a new wallet
- Stop browser
- Set `reconcileStamp` `ledger-state.json` to the current timestamp or more than a day in the past.
- Verify that the text indicates "is due" an "is coverdue" instead of suggesting a future event.


app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -198,7 +200,7 @@ ledgerRecoveryFailedTitle=Recovery Failed
 ledgerRecoveryFailedMessage=Please re-enter keys or try different keys.
 ledgerRecoveryNetworkFailedTitle=Network Error
 ledgerRecoveryNetworkFailedMessage=Please check your internet connection and try again.
-ledgerRecoveryContent=The balance of the recovered wallet will be transferred to your new Brave wallet. The old wallet will still exist as an empty wallet.
+ledgerRecoveryContent=Your previous wallet will now be used. Your new wallet will be discarded.
```
bsclifton: Just to confirm:
@mrose17 when you do a restore, it'll use the previous wallet (what you just restored), right? So the old text here (about transferring the balance to your new wallet) was wrong?

NejcZdovc:
Regarding payment history button I think this is not relevant any more, because we removed payment history button and added only the icon, more in #7532

app/extensions/brave/locales/en-US/preferences.properties
mrose17: correct. wallet recovery is now actually "recovery of old wallet" rather than "transfer the balance of the old wallet to the new wallet"

bsclifton:
@NejcZdovc I believe the strings are also shown when you click the icon (and it shows the statements)
screen shot 2017-03-10 at 1 31 05 pm

Issue #7549, opened 2017-03-08
if flash is enabled, 'shields down' should allow the site to see that flash is installed


diracdeltas:
https://github.com/brave/browser-laptop/issues/4789#issuecomment-285112448

PR #7563, opened 2017-03-08
allow sites to detect flash when shields are down

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/7549

Auditors: @bridiver

Test Plan:
0. enable flash in preferences
1. go to http://tv-trwam.pl/na-zywo
2. turn off shields. the flash placeholder should appear.


luixxiul:
Note: http://tv-trwam.pl/na-zywo is redirected to https://tv-trwam.pl/na-zywo, which prompts the cert error. You manually need to change https to http.

PR #7642, opened 2017-03-09
Payments features 1.0

### LINKED ISSUE #7347: Update ledger list control to support new features for 1.0
image

Item 1: Update column headers to match this config. The rank number column has been removed, and an actions column has been added.

Item 2: Create a new grouped column for action buttons. The delete button moves here, and "Pin site" will be added. Eventually, a tweet button may be added.

Item 3: The percent column needs to support an edit box with active and inactive states.

Item 4: Add ability to show permanent positioned items by adding sort hierarchy. This will allow sorting by "pinned state" and then column.

NejcZdovc:

Resolves #7347

### Auditors
@bsclifton @bradleyrichter @mrose17 @alexwykoff 

### Test plan
- `npm run test -- --grep="LedgerTable"`
- `npm run test -- --grep="Ledger table"`
- features specified in #7347


test/unit/app/renderer/components/ledgerTableTest.js
```diff
@@ -0,0 +1,544 @@
...
+    assert.equal(wrapper.find('[data-tbody-index="0"] [data-test-id="siteName"]').length, 2, '2 pinned')
+    assert.equal(wrapper.find('[data-tbody-index="1"] [data-test-id="siteName"]').length, 2, '2 unpinned, 1 hidden')
+    assert.equal(wrapper.find('[data-l10n-id="showAll"]').length, 1, 'show all button visible')
+  })
+
+  it('two pinned tabs, 3 unpinned (hide less button, all visible)', function () {
+    const wrapper = mount(
+      <LedgerTable
+        ledgerData={Immutable.Map({synopsis: fivePublishers.synopsis})}
+        settings={Immutable.Map({
```
NejcZdovc: @bsclifton This test is failing, because settings is not found in the `getSettings` in `ledgerTable.js` file https://github.com/brave/browser-laptop/pull/7642/files#diff-e168dd994d1e4de3f340af83899a20f5R247.

Problem is that this if is failing https://github.com/brave/browser-laptop/blob/master/js/settings.js#L58. More specific this is `false` `settingsCollection.constructor === Immutable.Map`. Which is strange, becuase as you can see I am passing `settings` prop as an `Immutable.Map`.

Any idea what is the problem?

mrose17:
many thanks!

bradleyrichter:
@cezaraugusto yes. Although I don't know which is easier to understand for usability. We could leave the switch enabled and allow a pinned site to be "excluded" which would be a macro for "un pin and set to zero %.

PR #7685, opened 2017-03-13
Adjust min threshold for play button

### LINKED ISSUE #7665: Sound icon + close button takes up too much space when tabs are small
When both sound bar icon and tab close is available, it can be hard to figure out where to click.

screenshot 2017-03-12 15 16 27

I think in this case we should use the blue bar variant of the sound indicator.

My wife hits this all the time and often mutes or closes the tab when she just wants to switch to it.

CC @bradleyrichter @cezaraugusto

cezaraugusto:
Fix #7665

Auditors: @bbondy

Preview:

new_bpoint_audio_tab

**Note:** I had to remove a test case since for all `medimLarge` sizes we no longer show the default play indicator

Test Plan:

Tabs playing audio with less than `120px` should have alternative play indicator (top blue border)

### DIFF
--- a/app/renderer/components/styles/global.js
+++ b/app/renderer/components/styles/global.js
@@ -6,6 +6,7 @@ const globalStyles = {
     breakpointSmallWin32: '650px',
     breakpointTinyWin32: '500px',
     tab: {
+      large: '120px',
       largeMedium: '83px',
       medium: '66px',
       mediumSmall: '53px',
--- a/app/renderer/components/styles/tab.js
+++ b/app/renderer/components/styles/tab.js
@@ -43,11 +43,7 @@ const styles = StyleSheet.create({
   narrowViewPlayIndicator: {
     borderWidth: '2px 1px 0',
     borderStyle: 'solid',
-    borderColor: 'lightskyblue transparent transparent'
-  },
-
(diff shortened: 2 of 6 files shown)


cezaraugusto:
We can also benefit by reducing play button size. /cc @bradleyrichter

bbondy:
Could you make this a bit more, more? Maybe 120px

bradleyrichter:
We may want to consider a swap with favicon --> mute button?

The percentage of mutable tabs is low, but recognizing quickly which tab is making noise is probably more important than recognizing the favicon.

cezaraugusto:
how would that work with alternative audio indicator (top blue border)?

bradleyrichter:
we may not need the audio indicator. (which BTW needs testing with new styling branch since the shadow you were using is now gone.  :   )

cezaraugusto:
audio is working fine. I'm ok with favicon replacement, but we should consider favicon-only cases for small sizes, where I think top blue border would apply. Otherwise we'll not be able to change tabs without muting. Thoughts?

bradleyrichter:
never mind. Let’s try the top border as the only indicator below 48 width, and rt-click to mute of course.

bbondy:
much better, thank you.

Issue #5512, opened 2016-11-09
Brave should restore session gracefully after a crash


bsclifton:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
Brave is not aware during re-launch that it had a crash during the last run. 

**Expected behavior:**
Brave should save a "clean shutdown" flag of some sort when saving.  Doing so would mean that this session file would need to be saved without that flag on startup, so that we can detect the condition.

In those cases, regardless of what the "start brave with" setting has, we should show an alert letting the user know:
- Brave crashed and it was unexpected
- We can try to restore the sites that were open, if they'd like us to

- Steps to reproduce:
    1. Launch Brave and open some tabs, each to a different site
    2. On Windows, open up task manager. On Linux/Mac, use `ps -ef | grep brave`
    3. Kill the main Brave process (end task or w/ kill command)

- Screenshot if needed:

- Any related issues:

bsclifton:
+1 from https://community.brave.com/t/bookmarks-missing/99

bsclifton:
Manual work-around I've given to folks for trying to restore their session:

------------------

First, quit Brave and make sure all instances of it are closed (if you know how, you can check task manager to make sure no Brave.exe are running)

Then push Windows key + R to bring up run prompt and then run:
%appdata%\brave

You should then see your local files. The session files have the format of "session-store-". Look for others with a similar name which are bigger in size. Backups are saved every so often, including when quitting. You should be able to rename the file you want to restore as "session-store-1"

Please let me know how this works for you. I would advise not to delete files... but if needed, rename them or put into a folder

ktd99:
bsclifton, I saw only 2 session files...one from today and a tmp file from Aug 30.  I checked several folders...did not see other session files.

bsclifton:
@ktd99 apologies- maybe I responded to your other issue too quickly. Did you lose your data? or did you only lose the open tabs for one session?

ktd99:
I lost all but 5 tabs...quite a lot.  And some had been open for months, so the History option is far too tedious....not even sure it would go back far enough.

bsclifton:
@ktd99 I updated the original post above (here in this issue) to make sure when this is fixed that we restore tabs and windows that were open before the crash happened.

It doesn't help you now, but I am glad to hear that you have all your other data though :smile:

lukemulks:
Experienced a similar variation of this issue about an hour ago. 

[...]

PR #7690, opened 2017-03-14
Prevent session from being destroyed when errors are raised during cleanAppData

bsclifton:
Taking a step towards https://github.com/brave/browser-laptop/issues/5512

## unit test coverage
Before
% Stmts: 52.65
% Branch: 45.77
% Funcs: 54.05
% Lines: 53.21

After
% Stmts: 55.33
% Branch: 48.59
% Funcs: 56.41
% Lines: 56.41

## overview
Losing your session sucks- every report by a user means they lost all of their bookmarks and likely didn't have a backup.  This PR takes a small step to help prevent session from being destroyed.  Here's how it works

1. App starts up; loadAppState() is called
2. Data loaded is passed into cleanAppData()
3. If an error occurs (and isn't caught), loadAppState will default to a fresh profile (meaning the user lost their work)
4. This PR will catch specific errors and continue on
  - errors cleaning window state are logged and window state is deleted
  - errors cleaning tab state are logged and tab state is deleted
  - errors clearing history are logged (nothing is deleted)
  - errors while clearing autofill are logged, nothing deleted
5. If something else happens which isn't caught by these handlers, the catch handler will rename the current session file (instead of doing nothing and overwriting it)

Auditors: @lukemulks, @diracdeltas, @bridiver, @bbondy

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -12,7 +12,7 @@
 // - NODE_ENV of ‘test’ bypassing session state or else they all fail.
 
 const Immutable = require('immutable')
-const fs = require('fs')
+const fs = require('fs-extra')
 const path = require('path')
 const electron = require('electron')
 const app = electron.app
@@ -48,6 +48,14 @@ const getTopSiteMap = () => {
     return siteMap
   }
   return {}
+}
+
+const getTempStoragePath = (filename) => {
+  const epochTimestamp = (new Date()).getTime().toString()
+  filename = filename || 'tmp'
+  return process.env.NODE_ENV !== 'test'
+    ? path.join(app.getPath('userData'), 'session-store-' + filename + '-' + epochTimestamp)
+    : path.join(process.env.HOME, '.brave-test-session-store-' + filename + '-' + epochTimestamp)
 }
 
 const getStoragePath = () => {
@@ -91,10 +99,7 @@ module.exports.saveAppState = (payload, isShutdown) => {
     }
     payload.lastAppVersion = app.getVersion()
 
-    const epochTimestamp = (new Date()).getTime().toString()
-    const tmpStoragePath = process.env.NODE_ENV !== 'test'
-      ? path.join(app.getPath('userData'), 'session-store-tmp-' + epochTimestamp)
-      : path.join(process.env.HOME, '.brave-test-session-store-tmp-' + epochTimestamp)
+    const tmpStoragePath = getTempStoragePath()
 
     let p = promisify(fs.writeFile, tmpStoragePath, JSON.stringify(payload))
       .then(() => promisify(fs.rename, tmpStoragePath, getStoragePath()))
@@ -266,7 +271,11 @@ module.exports.cleanAppData = (data, isShutdown) => {
   }
   const clearAutocompleteData = isShutdown && getSetting(settings.SHUTDOWN_CLEAR_AUTOCOMPLETE_DATA) === true
   if (clearAutocompleteData) {
-    autofill.clearAutocompleteData()
+    try {
+      autofill.clearAutocompleteData()
+    } catch (e) {
+      console.log('cleanAppData: error calling autofill.clearAutocompleteData: ', e)
+    }
   }
   const clearAutofillData = isShutdown && getSetting(settings.SHUTDOWN_CLEAR_AUTOFILL_DATA) === true
   if (clearAutofillData) {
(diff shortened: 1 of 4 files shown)


bsclifton:
@bridiver something is broken if an exception occurs, and I agree that just logging masks the problem... would we be able to hook into the error reporting for these instances?

I committed a part 2 which we can consider also (with or without the first commit) which renames the session file in a worst case scenario (so that the user's session doesn't get overwritten with a default state when they close the browser)

bridiver:
we talked about keeping a versioned history so you can rollback, but really we just need to write this to UserPrefs so it updates individual keys instead of writing the entire file every time which would be pretty easy to do

bsclifton:
@bridiver the UserPrefs approach would be nice... I'm not aware of how little or how much work is needed to take advantage of that. It seems like it would have a large impact though (lots of edge cases; upgrading from VERY old versions to new version, migrations from newer to older versions, older to newer, etc).

The changes in this PR are pretty low-tech and can start helping immediately.  I can definitely understand how the try/catch can mask the error (which is not good and wasn't my intention)... but I'd at least like to advocate for accepting the session file backup when an error is caught (while loading)

diracdeltas:
Actually why is `cleanAppData` always called after a version update? Seems like it should only be called if `cleanedOnShutdown` is false.

bsclifton:
Good question @diracdeltas; it was added in https://github.com/brave/browser-laptop/commit/120177939c772e2add298267ed773d0965e452b8 by @bbondy

bsclifton:
@lukemulks just lost his session file and I believe the fixes here could prevent it from happening again

@bridiver instead of eating the exceptions, what would you think about me reworking this PR to only keep the rename/preserve the session file part (when exceptions are caught)?

Whatever the root cause is (corruption, etc), it causes the session to fail parsing. This causes the app to have a default state, which overwrites the existing session file (permanently erasing the user's session)

bsclifton:
@aekeus would we be able to plug into our error reporting services to anonymously report the errors which happen (instead of printing them out to the console?).  An alternative I guess could be that we store persistent log files which we could ask users for when debugging

bsclifton:
Before reworking this, I want to get some tests in place. I'll be submitting another PR later with tests for our existing session code.  If you'd like a preview, here's the first commit:
https://github.com/bsclifton/browser-laptop/commit/82082caa127537dd3b84b2d9147fd96fe58d438b

app/sessionStore.js
```diff
@@ -524,6 +567,20 @@ module.exports.loadAppState = () => {
...
 /**
+ * Called when session is suspected for corruption; this will move it out of the way
+ */
+module.exports.backupSession = () => {
+  const src = getStoragePath()
+  const dest = getTempStoragePath('backup')
+
+  console.log('An error occurred. For support purposes, file "' + src + '" has been renamed to "' + dest + '".')
+
+  fs.rename(src, dest, (err) => {
```
bridiver: this introduces a potential race condition between the rename and the next save. It's unlikely to happen, but I don't think we want to add any more potential race conditions so I think this should be a sync operation and we should use copy instead of rename in case the bug is somehow related to rename

bsclifton:
Per discussion w/ @bridiver, we may also want to implement flushing when writing session to disk

app/sessionStore.js
```diff
@@ -421,64 +514,30 @@ module.exports.loadAppState = () => {
...
-          data.settings[settings.DEFAULT_SEARCH_ENGINE] = 'Google'
-        }
-        if (data.settings[settings.DEFAULT_SEARCH_ENGINE] === 'content/search/duckduckgo.xml') {
-          data.settings[settings.DEFAULT_SEARCH_ENGINE] = 'DuckDuckGo'
-        }
+      loaded = true
+    } catch (e) {
+      // Session state might be corrupted; let's backup this
+      // corrupted value for people to report into support.
+      module.exports.backupSession()
```
bbondy: There's more to do here as per discussions but doesn't need to all be done in this bug.
But I think this will try to do a sync copy on disk on first startup which would slow down startup. Could you verify if that's the case?

bridiver:
to me the lack of fsync seems like the most likely culprit given how randomly these things occur

luixxiul:
The test plan would be to force a BSOD?

http://www.wikihow.com/Force-a-Blue-Screen-in-Windows

Issue #7655, opened 2017-03-11
Fresh Install of Brave Advertises Brave Properties


jonathansampson:
## Test plan

https://github.com/brave/browser-laptop/pull/7753#issue-214852343

----

- Did you search for similar issues before submitting this one?
Yes

- Describe the issue you encountered:
A fresh install of Brave yields advertisements for Brave. This took at least one user by surprise when they tried the browser. Their feedback is available [on Twitter](https://twitter.com/JulianNorton/status/840614856919252994):

![image](https://cloud.githubusercontent.com/assets/815158/23825495/6a380f8c-0650-11e7-8783-88215b669214.png)

- Platform (Win7, 8, 10? macOS? Linux distro?):
All

- Brave Version (revision SHA):
0.13.5

- Steps to reproduce:
    1. Install Brave
    2. Type 'facebook' into address bar
    3. Brave auto-completes to 'facebook.com/brave'
    
- Actual result:
Brave _advertises_ its own properties on a fresh install.

- Expected result:
User expects no advertisements or promotion

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
Yes

- Is this an issue in the currently released version?
Yes

- Can this issue be consistently reproduced?
Yes

- Extra QA steps:
    1. Steps repro for 'twitter'
    2. Steps repro for 'youtube'
    3. Steps repro for 'itunes'

#5594

bsclifton:
Duplicate of https://github.com/brave/browser-laptop/issues/5892

cc: @bradleyrichter, @bbondy, @alexwykoff (in case we wanted to take action other than closing as a dupe)

bradleyrichter:
This is the worst (or best) example of this "bug" and we should block on it. 

Of course this is a side effect of preloading top sites in the dashboard. 

In this example, the search result suggestion should not be pre-filled but rather presented in the list and require a down arrow to access. 

We may want to also create a new issue to mask out the preloaded top sites from the search suggestions so that this unintentional advertising is not perceived as intentional.

diracdeltas:
Seems like the user is asking for the preloaded Brave sites to not be in the urlbar suggestions at all

bsclifton:
@diracdeltas yup- we should be able to edit the site details to include a "leave me out of suggestions" flag. If the user then chooses to bookmark that site, we'd just want to clear that 😄

bsclifton:
[...]

PR #7534, opened 2017-03-08
Add short domains to url suggestions when long url is present

aekeus:
Auditors: @bbondy

Test Plan:

  1. Clear history
  2. Visit http://www.github.com/brave
  3. Verify one history item in about:history
  4. Open new tab
  5. type 'github'
  6. Verify that http://www.github.com suggestion is present

Fixes: #7533

- [X] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [X] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [X] Added/updated tests for this change (for new code or code which already has tests).
- [X] Ran `git rebase -i` to squash commits (if needed).

PR #7692, opened 2017-03-14
Remove Brave software links from urlbar suggestions

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/7655

Auditors: @bsclifton

Test Plan:
1. open a clean instance of brave
2. type 'face' into the urlbar
3. it should autocomplete to facebook instead of Brave's facebook page

Test Plan:

### DIFF
--- a/js/data/newTabData.js
+++ b/js/data/newTabData.js
@@ -80,3 +80,5 @@ module.exports.topSites = [
       "title": "Brave Browser: Fast AdBlock – Apps para Android no Google Play"
   }
 ]
+
+module.exports.topSiteLocations = module.exports.topSites.map((site) => site.location)
--- a/js/data/newTabData.js
+++ b/js/data/newTabData.js
@@ -10,18 +10,7 @@ const iconPath = getBraveExtUrl('img/newtab/defaultTopSitesIcon')
  */
 const now = Date.now()
 
-module.exports.pinnedTopSites = [
-    {
-      "count": 1,
-      "favicon": `${iconPath}/twitter.png`,
-      "lastAccessedTime": now,
-      "location": "https://twitter.com/brave",
-      "partitionNumber": 0,
-      "tags": [],
-      "themeColor": "rgb(255, 255, 255)",
-      "title": "Brave Software (@brave) | Twitter"
-  }
-]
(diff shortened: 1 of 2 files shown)


bsclifton:
Actually, I may have found a deal breaker (even if this behavior is OK)

@diracdeltas you are allowed to pin the items... if you choose to pin any of the default bookmarks, it doesn't stick.  In the above GIF, you can see that the Brave Twitter site is pinned and it gets erased after visiting the Facebook page

diracdeltas:
@bsclifton great catch! i fixed the pinned-site-going-away-after-a-site-visit bug. 

personally i am okay with the new tab default sites going away as soon as you have a real site for the new tab page, but that is probably not the desired UX.

diracdeltas:
going to merge this for now. @bradleyrichter if you have thoughts on persisting Brave sites on the newtab page, please open a follow-up issue.

bradleyrichter:
@bsclifton @diracdeltas doesn't this PR solve this issue and let us keep the default tiles as they were?

https://github.com/brave/browser-laptop/pull/7534

diracdeltas:
@bradleyrichter as mentioned in the issue, i don't think so, because the Brave sites are still in the autosuggest bar. i think the OP wanted them to not be in there at all.

bradleyrichter:
We need full tiles there on first run please. The row-selector is coming back which allows the user to select 1/2/3 rows as well.

diracdeltas:
they are in there on first run

bradleyrichter:
and vanish on the first visit to a new page? I think this will just look like a new bug.  :  )

diracdeltas:
i'm a fan of minimizing how much we self-advertise, but feel free to open an issue

bsclifton:
OK- I think we jumped the gun on this one.  @bradleyrichter is going to work out the desired behavior.  For now, we can leave as-is, but we'll want to change (and revisit) before release. I'll create an issue to track that (so that we don't stomp on the old issue(s))

bsclifton:
Removing milestone as this was reverted with https://github.com/brave/browser-laptop/commit/e23e25112a091a078aa49b810d1d6222c978bc99

PR #7705, opened 2017-03-14
Fix failling test for tab themeColor

cezaraugusto:
Auditors: @luixxiul

Test Plan:
Automated test must pass:
✓ Uses the default tab color when one is not specified
```
npm run test -- --grep="Uses the default tab color when one is not specified"
```


bsclifton:
++

PR #7773, opened 2017-03-17
split Travis into parallel test suites

diracdeltas:
Breaks it up so we can run 9 test suites in parallel. Also caches NPM dependencies so that these are not re-downloaded for every test suite.

Test Plan:
N/A


diracdeltas:
https://travis-ci.org/brave/browser-laptop/builds/212314661 finished in 32 minutes (much of which was waiting for jobs to start) so it's like a 15min improvement

diracdeltas:
@bsclifton i think we will generally only need to clear the cache when changing the runtime environment (ex: upgrading Node), so it should be fine. to be extra safe, we can use shrinkwrap. https://docs.npmjs.com/cli/shrinkwrap

bsclifton:
@diracdeltas ah nice! 😄 In that case, let's merge 🎉

Issue #7794, opened 2017-03-20
travis cache is using the wrong version of Standard


diracdeltas:
we are pinning to 8.1, but Travis's npm cache is using 9.0 so lint is failing. ex: https://travis-ci.org/brave/browser-laptop/jobs/213153792

i think shrinkwrap fixes this

diracdeltas:
i'm going to revert caching for now and then re-enable caching with #7795

PR #7495, opened 2017-03-04
Update lint style for Standard@9.0.0 - first patch

cezaraugusto:
Auditors: @bsclifton, @nejczdovc
Fix #7492

This PR addressed the following new rules:

```
no-multiple-empty-lines
no-useless-return
generator-star-spacing
padded-blocks
```

Test plan (two steps):

1. With current `standard`, run

```
npm run lint
```

- There should be no linting errors

2. Update standard@9.0.0

- Lint errors should not include:
  - `Too many blank lines at the end of file`...
  - `Unnecessary return statement`
  - `Expected indentation of`...
  - `Missing space after *`
  - `Block must not be padded by blank lines`

**Note:** Most files were updated running `standard --fix`. Below files that were updated manually:

```
main.js (code commented as a reference)
settings.js
dataFile.js
```

PR #7795, opened 2017-03-20
Use npm shrinkwrap and travis caching

diracdeltas:
fix #7820 

- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Test Plan:
1. automated tests should not fail with a lint error

PR #7817, opened 2017-03-21
Update Standard@9.0.0

### LINKED ISSUE #7492: Update to Standard@9.0.0 - rules "no-multiple-empty-lines", "no-useless-return", "generator-star-spacing", "padded-blocks"
Version 9.0.0 introduced major changes that changed lint rules for several files.

This work will be split to avoid regressions. From [Standard changelog](https://github.com/feross/standard/blob/v9.0.0/CHANGELOG.md#new-rules) those rules were added:

(note that `changed rules` didn't affect our repo)

* Disallow mixing different operators without parens (no-mixed-operators)
* [x] __Enforce 1 newline at end of file (previously 1 or 2 were ok) (no-multiple-empty-lines)__
* Disallow Unused Expressions (no-unused-expressions)
* [x] __Disallow redundant return statements (no-useless-return)__
* Disallow Incorrect Early Use (no-use-before-define)
* Enforce that Promise rejections are passed an Error object as a reason (prefer-promise-reject-errors)
* Enforce comparing typeof expressions against string literals (valid-typeof)
* [x] __Enforce spacing around * in generator functions (generator-star-spacing)__
* Disallow Unnecessary Labels (no-extra-label)
* Disallow spacing between template tags and their literals (template-tag-spacing)
* [x] __Disallow padding within switch statements and classes (padded-blocks)__
* Enforce that Symbols are passed a description (symbol-description)

This issue addresses rules marked as checked.

cezaraugusto:
Auditors: @bsclifton

Fix #7492
Follow-up of #7495 and #7785

Will also help with standard version inconsistency between repo/travis (#7794)

This PR addressed the following new rules:

```
no-mixed-operators (follow-up with 1 occurency)
no-unused-expressions
prefer-promise-reject-errors
valid-typeof
no-useless-escape
```

Test plan:

1. Run `npm run lint`
2. Should have no linting errors

### DIFF
--- a/app/dataFile.js
+++ b/app/dataFile.js
@@ -55,7 +55,7 @@ module.exports.readDataFile = (resourceName, url) => {
     fs.readFile(storagePath(url), (err, data) => {
       if (err || !data || data.length === 0) {
         // console.log('rejecting for read for resource:', resourceName)
-        reject()
+        reject(new Error('unable to read data file'))
       } else {
         // console.log('resolving for read for resource:', resourceName)
         resolve(data)
--- a/app/dates.js
+++ b/app/dates.js
@@ -4,7 +4,10 @@ var isoWeek = function () {
   date.setHours(0, 0, 0, 0)
   date.setDate(date.getDate() + 3 - ((date.getDay() + 6) % 7))
   var jan4 = new Date(date.getFullYear(), 0, 4)
-  return 1 + Math.round((((date.getTime() - jan4.getTime()) / 86400000) - 3 + ((jan4.getDay() + 6) % 7) / 7))
+  return 1 + Math.round(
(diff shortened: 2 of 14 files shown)


diracdeltas:
Note that this doesn't fully address #7794 because Travis is using latest Standard (9.0.2?) instead of 9.0.0. I believe we still need #7795 to fix that.

cezaraugusto:
how Travis handle deps updates? I think we can update this to match current Travis version

diracdeltas:
@cezaraugusto i think there is a bug in how travis invalidates caches; others report similar issues at https://github.com/travis-ci/travis-ci/issues/3758. i'm not sure if shrinkwrapping helps but it might.

diracdeltas:
anyway i reverted caching on master for now

Issue #7832, opened 2017-03-22
Zoom is not saved when you close the browser and restore on a single site


luixxiul:
**Describe the issue you encountered:**
Zoom is not saved when you close the browser and restore on a single site.

- Platform (Win7, 8, 10? macOS? Linux distro?):

- Brave Version (revision SHA): 0.14.0 Preview 4

- Steps to reproduce:
    1. Test zoom saved when you close the browser and restore on a single site.
    
- Expected result: Zoom should be saved when you close the browser and restore on a single site

- Will the steps above reproduce in a fresh profile? Yes

- Any related issues:

bsclifton:
Closing issue as it was fixed with 5f39fbf; next Muon build will include it

PR #7852, opened 2017-03-23
Fix zoom level on browser restore

diracdeltas:
and remove deprecated zoom level site setting.

fix https://github.com/brave/browser-laptop/issues/7832

Test Plan:
1. launch clean instance of brave
2. go to example.com and change the zoom level
3. close brave
4. reopen brave
5. example.com should load with the previous zoom setting

Test Plan:

### DIFF
--- a/js/components/frame.js
+++ b/js/components/frame.js
@@ -13,6 +13,7 @@ const cx = require('../lib/classSet')
 const siteUtil = require('../state/siteUtil')
 const FrameStateUtil = require('../state/frameStateUtil')
 const UrlUtil = require('../lib/urlutil')
+const {getZoomLevel} = require('../lib/zoom')
 const messages = require('../constants/messages')
 const contextMenus = require('../contextMenus')
 const ipc = require('electron').ipcRenderer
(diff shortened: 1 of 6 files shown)


js/components/frame.js
```diff
@@ -336,7 +337,18 @@ class Frame extends ImmutableComponent {
           this.runOnDomReady()
           delete this.runOnDomReady
         }
+        let zoomCallback = (e) => {
+          if (!e.isMainFrame) {
+            return
+          }
+          this.webview.removeEventListener(e.type, zoomCallback)
+          const zoomPercentage = this.frame && this.frame.get('lastZoomPercentage')
```
bridiver: muon should is supposed to handle this automatically so I think we should fix the problem there. If we need something on the browser-laptop side we should do it in tabs.js because all of the webview methods are deprecated (especially getter methods because they are sync ipc)

diracdeltas:
closing in favor of https://github.com/brave/muon/commit/7baccbe838467bc9e6165d0f79d2f9f60139c938

PR #7887, opened 2017-03-25
Refactor contributionStatement.js with Aphrodite

### LINKED ISSUE #7883: Refactor contributionStatement.js with Aphrodite
## Test plan

https://github.com/brave/browser-laptop/pull/7887#issue-216965677

----

Refactor contributionStatement.js with Aphrodite

Initial work: https://github.com/luixxiul/browser-laptop/commit/4ca21420b62af33694777b1e3484bdafda630194

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:
screenshot 2017-03-25 18 04 06

- Any related issues:

luixxiul:
Closes #7883

- Remove contributionStatement.less
- Replace inlined braveLogo with a PNG file, verifiedIcon with a SVG file
- Introduce constants for margins, borders, paddings, light gray, etc

Auditors:

Test Plan:
1. npm run add-simulated-payment-history
2. Open about:preferences#payments
3. Display the payments history
4. Save the PDF file
5. Open it and make sure it is successfully rendered


luixxiul:
The brave logo PNG file and the verified SVG icon are not displayed on PDF file. Could I know how to display them?

js/about/contributionStatement.js
```diff
@@ -430,4 +424,243 @@ function longFormattedDateFromTimestamp (timestamp) {
...
+    margin: `${containerMargin} ${containerMargin} 0`
+  },
+  pageFooterWrapper__body: {
+    color: globalStyles.color.gray
+  },
+
+  // ContributionStatementPage
+  contributionStatement__page: {
+    display: 'flex',
+    flexFlow: 'column nowrap'
```
luixxiul: Thanks for your feedback @willy-b! I'll ping you again later.

luixxiul:
@willy-b Updated. The files were not displayed because `npm run watch` had been killed. I confirmed it worked as it was.

js/about/contributionStatement.js
luixxiul: Added here: https://github.com/brave/browser-laptop/pull/7887/commits/2f16bbdcd71706a1c29f2fe20183f5835fc93031#diff-dfc3cf3000430cbd1575b0223b06163fR630

js/about/contributionStatement.js
```diff
@@ -430,4 +455,242 @@ function longFormattedDateFromTimestamp (timestamp) {
...
+  },
+  footerWrapper__body: {
+    color: globalStyles.color.gray
+  },
+
+  // ContributionStatementPage
+  statement__page: {
+    display: 'flex',
+    flexFlow: 'column nowrap',
+    pageBrakeAfter: 'always'
```
willy-b: I think you fixed this in next commit which was force pushed over this and the previous one, but `pageBrakeAfter` here should be replaced with `pageBreakAfter`

js/about/contributionStatement.js
willy-b: I think you fixed this in next commit which was force pushed over this and the previous one, but `pageBrakeAfter` here (https://github.com/brave/browser-laptop/pull/7887/commits/2f16bbdcd71706a1c29f2fe20183f5835fc93031#diff-dfc3cf3000430cbd1575b0223b06163fR630) should be replaced with `pageBreakAfter`

luixxiul: Silly mistake.. I updated here: https://github.com/brave/browser-laptop/pull/7887/commits/0164a4df741ce5a9c7489c4a9449e4bbb9e589f4#diff-dfc3cf3000430cbd1575b0223b06163fR630

willy-b:
That's interesting: I still don't see the images in the generated PDF with your latest revision.

The footer text also continues to have extra characters in it, which seem to be have been fixed in your latest screenshot. Are you sure that screenshot is of https://github.com/brave/browser-laptop/pull/7887/commits/0164a4df741ce5a9c7489c4a9449e4bbb9e589f4?

luixxiul:
Would you try `rm -rf ~/.electron && rm node_modules` and `npm i`?

willy-b:
Yes, just re-installed. Still don't see any images in generated PDF (which I can attach if you like).
Though I can confirm that page-breaking is fixed (https://github.com/brave/browser-laptop/pull/7887#discussion_r109281798)

luixxiul:
I'm not sure where the issue is...

1. create payment history
2. npm run watch
3. npm start
4. open about:preferences#payments
5. show payment history
6. click a pdf
7. preview it 

contribution

willy-b:
It looks like you're using the Print dialog manually to print to PDF and open the PDF in preview? That actually works for me too.

But what about the auto-popup for saving the PDF that is triggered by current code? That does not work. (Did you turn this off in your local for making the screencapture?) We could replace that with asking user to use Print -> Print to PDF and choose their own filename.

luixxiul:
> the auto-popup for saving the PDF that is triggered by current code?

Actually I haven't seen it for a while, though I'm running on the latest code. Is it disabled??

willy-b:
I see the auto-popup save dialog in a fresh build from your commit https://github.com/brave/browser-laptop/pull/7887/commits/0164a4df741ce5a9c7489c4a9449e4bbb9e589f4 :
pull-7887-2-trimmed

As you can see in my screencap, the images are missing from the downloaded PDF (opened from downloads bar).

I wonder why you're not seeing the Download PDF auto-popup dialog anymore?

If I do a manual File->Print->Print to PDF of the contribution statement HTML page, as you show in your screencap, then the images are preserved for me too.
We could get rid of the auto-popup and just tell people to do that if you want.

luixxiul:
I'm on macOS. Is it related with Muon or something? / CC @darkdh @bsclifton

darkdh:
I think it related to our 57 upgrade. There will be PREVIEW_FAIL error on both windows and macOS.

luixxiul:
> PREVIEW_FAIL 

Yeah I see it on macOS!

luixxiul:
@willy-b For me, not displaying the pop up is not that bad; in fact I felt quite natural, since I was not forced to save the statement in PDF. I could save, print, copy or just read through the statement of my own will. 

I could imagine some would say, "please do not display the save pop up window as it is annoying."

bsclifton:
@luixxiul is this ready for review? or are you still working out a few things?

luixxiul:
Except inlining the PNG and SVG files, it's ready for review.

If it is impossible to add the external files with printToPDF, I would inline them. / CC @darkdh

willy-b:
it's possible to do this without inline images. you would just need to replace current approach in app/pdf.js with one that doesn't use a data URI. that's totally do-able. I'm sure @darkdh can figure it out (and faster than me) but I can also take a look this evening.

luixxiul:
@willy-b Since I'm not familiar with the process which has leaded to the current approach, I cannot tell, but  if that is the better solution, +1.

darkdh:
I'm working on [fixing printToPDF](https://github.com/brave/browser-laptop/issues/8051)

js/about/contributionStatement.js
```diff
@@ -430,4 +455,242 @@ function longFormattedDateFromTimestamp (timestamp) {
...
+const summaryBoxTableBorder = '10px'
+const summaryBoxTableMargin = '10px'
+const summaryBoxTablePadding = '10px'
+const lightGray = '#f7f7f7'
+const braveLogoWidth = '50px'
+const braveLogoHeight = '50px'
+const braveLogo = require('../../app/extensions/brave/img/braveAbout.png')
+const verifiedIcon = require('../../app/extensions/brave/img/ledger/verified_green_icon.svg')
+
+const styles = StyleSheet.create({
```
NejcZdovc: One suggestion maybe. Instead of having one big object `styles`, maybe create 4 objects:

- styles (general and small styles)
- stylesHeader
- stylesPage
- stylesTable

What do you think?

luixxiul: I'm not sure how worth it is. I left comments and grouped with `__` for each function, and I think it is not so big to separate into the objects.

If we will separate them, let's do so after everything is refactored and we create the rules which decide when to separate objetcs; Otherwise it will definitely lead to inconsistency and we would not like to see that.

luixxiul: Actually if Aphrodite will support nesting like LESS, we do not need to separate the objects. Let's wait and see what will happen.

willy-b:
@luixxiul , I think our current process for saving the PDF without using the print dialog could be re-written to be compatible with non-inlined images, but my vote is to do it in a separate issue/PR.

what are your thoughts on that? (temporarily sticking with inlined images to get the rest of this PR merged now and tackling removing the data URI approach in a follow-up)

luixxiul:
@willy-b ok I see. I will revert the outlined files back to the js file for now, with TODO comments assigning to myself.

luixxiul:
@willy-b OK, I updated.

luixxiul:
After making it possible to use external files, just reverting https://github.com/brave/browser-laptop/pull/7887/commits/eaf340c10d237f20610c627e67c67fa6a7981b3a would work fine.

willy-b:
Just tested, looks good to me!

luixxiul:
@cezaraugusto do you confirm the issue mentioned by @willy-b was fixed?

luixxiul:
If so, the last two commits should simply be dropped.

luixxiul:
Several TODOs exist yet.

willy-b:
The PDF generation triggered auto-save popup still needs inlined images, unless there has been a change. I can make an issue to fix that. You just replaced the commits I reviewed, are we still going with inline images temporarily?

luixxiul:
@willy-b would you please work on the fix? thanks!

cezaraugusto:
@willy-b sorry I missed your last comment before merging. Please feel free to open an issue, not sure which STR to repro error you mentioned

luixxiul:
@cezaraugusto @willy-b ^ https://github.com/brave/browser-laptop/pull/7887#issuecomment-290972828

willy-b:
no worries @cezaraugusto . quick question: does the final version you merged use inline images or not?

cezaraugusto:
no inline-images.

A fix I can think of is to convert images to base64 before rendering, but feel free to apply any method you find best and lmk if you need anything. Thanks for following-up :)

cezaraugusto:
btw auto-popup doesn't show on macOS

luixxiul:
it has been mentioned here: https://github.com/brave/browser-laptop/pull/7887#issuecomment-290968531

willy-b:
okay so we just broke this, just so we could avoid inline images? fair enough.
still seems to me we should have done that in a separate PR.

@darkdh is fixing the popup on MacOS, I think.

darkdh:
It's been fixed in https://github.com/brave/muon/commit/9d32f7fec4f62e72d841e00e223df8a806afa410
Just needs a new muon build

willy-b:
awesome @darkdh. 

so then the popup will be on MacOS and images will be broken in the PDF. you guys sure you don't want to merge @luixxiul's other commit which restores inline images temporarily so this is not broken?

cezaraugusto:
@willy-b I'm looking for an alternative but yes we'll end up inlining images just not very happy with hard-coding it. If time gets shorter and we can't make it we'll eventually go back to @luixxiul fix.

willy-b:
ok, so you're going to restore inline images and reopen https://github.com/brave/browser-laptop/issues/8062  to pull them out once we make them unnecessary?

darkdh:
@willy-b , I think image problem will also be fixed in that commit

willy-b:
unfortunately it won't. the image issue is not in Muon. it's in our use of a data URI here: https://github.com/brave/browser-laptop/commit/7c32ef100f01568d66d3de327d8044d85305024d#diff-dfc3cf3000430cbd1575b0223b06163fR120

PR #7933, opened 2017-03-28
use current, backwards-compatible version of ledger-client

mrose17:
…and ledger-publisher

Test Plan:
- this version of ledger-client is in preparation for the v8 worker code #7334 ... it is backwards-compatible, so a simple independent ledger test should do it!


bbondy:
++

bsclifton:
++

PR #7944, opened 2017-03-29
Update UI spec for private tabs

### LINKED ISSUE #547: replacement icons for private browsing tab icon
@bbondy These will be used in iOS. Please convert current font awesome eyeball to these sunglasses.

privatemode_btn
privatemode_btn 2x
privatemode_btn 3x

### LINKED ISSUE #7943: Update UI spec for private tabs
Per @bradleyrichter request:

1. Private tabs should have its color (purple) only when active, no hoverState
2. Private tabs color should override themeColor
2. Replace 👁 icon with 🕶  

screen shot 2017-03-28 at 11 43 32 pm

cezaraugusto:

Close #7943
Close #547
Auditors: @bsclifton

Test Plan:

* Private tabs should be purple only when active
* Website with themeColor (i.e. YouTube) should have color replaced with private tab color (purple)
* Private icon should be changed from eye to sunglass

screen shot 2017-03-29 at 1 24 11 am


luixxiul:
I added #547 as another issue to be closed with this PR.

cezaraugusto:
nice catch, thanks

NejcZdovc:
@bradleyrichter I personally prefer to have purple background all the time. This ways I can easily and more importantly faster see which tab is private. Other then that, this PR LGTM

bradleyrichter:
@NejcZdovc Did you also find that the colored non-selected private tab made it more difficult see which tab is selected? I think this problem is more severe than knowing if the tab is private before selection. But maybe I need to consider more use cases...

bsclifton:
@bradleyrichter you could create a follow up issue if it made sense- we could have logic which sets a light pastel purple when the private tab is not selected and the regular color purple when it is selected/active.  I think this is a good compromise

PR #7975, opened 2017-03-29
Change inactive private tab background to 30% of original color

NejcZdovc:

### Auditors
@bsclifton @cezaraugusto @bradleyrichter

### Test Plan
- open private tab
- open another private tab
- non active private tab should have light purple background

### DIFF
--- a/app/renderer/components/tabContent.js
+++ b/app/renderer/components/tabContent.js
@@ -223,7 +223,7 @@ class TabTitle extends ImmutableComponent {
   get themeColor () {
     const themeColor = this.props.tab.get('themeColor') || this.props.tab.get('computedThemeColor')
     const activeNonPrivateTab = !this.props.tab.get('isPrivate') && this.props.isActive
-    const activePrivateTab = this.props.tab.get('isPrivate') && this.props.isActive
+    const activePrivateTab = (this.props.tab.get('isPrivate') && this.props.isActive) || this.props.isHover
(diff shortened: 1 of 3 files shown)


bsclifton:
@cezaraugusto did you have input on this? I'll go in and add some tests 😄

app/renderer/components/tabContent.js
```diff
@@ -223,7 +223,7 @@ class TabTitle extends ImmutableComponent {
   get themeColor () {
     const themeColor = this.props.tab.get('themeColor') || this.props.tab.get('computedThemeColor')
     const activeNonPrivateTab = !this.props.tab.get('isPrivate') && this.props.isActive
-    const activePrivateTab = this.props.tab.get('isPrivate') && this.props.isActive
+    const activePrivateTab = (this.props.tab.get('isPrivate') && this.props.isActive) || this.props.isHover
```
cezaraugusto: we have a tab property for that, `this.props.tab.get('hoverState')` which is a prop made for hover events. I think that's the right fit for these changes and you can just remove other mentions for hover you created.

You could also change `const` name to update its meaning for something like `isPrivateTab` since we're adding a rule that's not exclusive of active  privateTabs. i.e.:

```
const activePrivateTab = this.props.tab.get('isPrivate') && (this.props.isActive || this.props.tab.get('hoverState'))
```

PR #8066, opened 2017-04-04
Fixes usage of ledger advance settings in the ledger table

### LINKED ISSUE #7987: toggle excluding a site from ledger sometimes removes it from the list
mac os 0.14 and latest master

steps to repro (starting from a clean profile):
1. enable payments
2. visit nytimes.com for at least 8 seconds or until it appears in the synopsis table
3. set minimum visits to 5 instead of 1
4. exclude nytimes.com by toggling the switch

result:
nytimes.com suddenly disappears from the list when the switch is toggled

expected result:
when minimum visits is set to N, all sites with less than N visits should disappear immediately. the current behavior is very confusing because toggling the exclude switch should not remove the site from the list.

a similar issue exists for increasing the minimum time. if i increase it from 8 seconds to 1 minute, a site with 9 seconds remains in the ledger table. however if i then toggle the include switch on that site, it doesn’t disappear from the list, so the problem isn't as bad.

NejcZdovc:

Resolves #7987

### Auditors
@mrose17 @diracdeltas

### Test Plan
- enable payments
- visit site for at least 8 seconds or until it appears in the synopsis table
- set minimum visits to 5 instead of 1
- site shouldn't be in the ledger table

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -176,7 +176,7 @@ const doAction = (action) => {
         case settings.MINIMUM_VISIT_TIME:
           if (action.value <= 0) break
 
-          synopsis.options.minDuration = action.value
+          synopsis.options.minPublisherDuration = action.value
           updatePublisherInfo()
           break
 
@@ -741,8 +741,9 @@ var enable = (paymentsEnabled) => {
         value = 8 * 1000
         appActions.changeSetting(settings.MINIMUM_VISIT_TIME, value)
       }
+
       // for earlier versions of the code...
-      if ((value > 0) && (value < 1000)) synopsis.options.minDuration = value * 1000
+      if ((value > 0) && (value < 1000)) synopsis.options.minPublisherDuration = value * 1000
 
       value = getSetting(settings.MINIMUM_VISITS)
       if (!value) {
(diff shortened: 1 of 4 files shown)


diracdeltas:
test plan worked, but some synopsis tests may be failing due to this: https://travis-ci.org/brave/browser-laptop/jobs/218608197#L3227. they failed locally for me too.

diracdeltas:
specifically this fails on this branch but passes on master (locally):

```
> npm run test -- --grep='synopsis'

> brave@0.14.1 test /Users/yan/repos/browser-laptop
> cross-env NODE_ENV=test mocha "test/**/*Test.js" "--grep=synopsis"



  synopsis
    ✓ no table if empty synopsis (46ms)
    1) creates synopsis table after visiting a site
    2) can sort synopsis table
    ✓ can disable site (2683ms)

  PublisherToggle component
    default behaviour (when autoSuggest is ON)
      ✓ show as enabled if site is on synopsis list (siteSettings is empty)
      ✓ show as enabled if hostPattern is on siteSettings list (synopsis is empty)
      ✓ show as enabled if url exists in locationInfo (both synopsis and siteSettings are empty)


  5 passing (46s)
  2 failing

  1) synopsis creates synopsis table after visiting a site:
  Error

  2) synopsis can sort synopsis table:
     WaitUntilTimeoutError: element ([data-l10n-id="publisher"]) still not visible after 10000ms
      at isVisible("[data-l10n-id="publisher"]") - waitForVisible.js:37:22
```

NejcZdovc:
@diracdeltas fixed, I needed to add back one property, because it's used in `ledger-publisher`, so we can use this hack for test (visit time is set to 0s when in executing tests)

diracdeltas:
lgtm, thanks for looking into this

PR #8073, opened 2017-04-05
Cache Sync bookmarks and history by objectId

ayumi:
Auditors: @diracdeltas

Fix https://github.com/brave/sync/issues/65

Also fixes Sync bookmarks tests for changed data-test-ids.

Test plan:
1. Point Sync to prod server: In `js/constants/appConfig.js` set `isProduction = true`.
2. Start Brave and join the "forsaken" Sync group.
3. It should seem to do nothing for 10s (downloading data from S3), then lock up for ~20s, then a lot of bookmarks should be synced. Compare with 2-3 minutes of freezing previously.
(Please take care to not clear this sync profile! when you're done, just delete the userData folder.)


js/state/syncUtil.js
```diff
@@ -242,7 +243,7 @@ module.exports.getExistingObject = (categoryName, syncRecord) => {
       item = module.exports.createSiteData(existingObjectData, appState)
       break
     case 'PREFERENCES':
-      const hostPattern = existingObject[0]
+      const hostPattern = existingObjectKeyPath[existingObjectKeyPath.length - 1]
```
diracdeltas: is there a case for `siteSettings` where `existingObjectKeyPath.length !== 1`?

js/state/syncUtil.js
```diff
@@ -261,6 +262,27 @@ module.exports.getExistingObject = (categoryName, syncRecord) => {
 }
 
 /**
+ * Cache a sync object's key path by objectId.
+ * XXX: Currently only caches sites (history and bookmarks).
+ * @param {Immutable.Map} appState The application state's Immutable sync.objectsById Map
+ * @param {Immutable.List} siteDetail
```
diracdeltas: Map, not List ?

js/state/syncUtil.js
```diff
@@ -261,6 +262,27 @@ module.exports.getExistingObject = (categoryName, syncRecord) => {
...
+ * @returns {Immutable.Map} new app state
+ */
+module.exports.updateObjectCache = (appState, siteDetail) => {
```
diracdeltas: maybe rename to `updateSiteCache` for now

diracdeltas: it would be nice if this didn't have to run for users who aren't running sync

diracdeltas:
yay, test plan worked

diracdeltas:
cc @bridiver

js/state/syncUtil.js
ayumi: an existing object's app state key path should be an array of at least length 1. so for siteSettings: ["siteSettings", {hostPattern}]

js/state/syncUtil.js
ayumi: good catch

js/state/syncUtil.js
ayumi: :+1:

diracdeltas:
since the cache isn't updated now when sync is disabled, should it be cleared when sync is toggled off? otherwise the cache will be stale when sync is re-enabled

ayumi:
@diracdeltas good idea
i made it so the cache is recreated every time sync is enabled, after the device's records are uploaded.

PR #8084, opened 2017-04-05
Add ability to move tear off tabs and move them between windows

bbondy:


app/common/commonMenu.js
```diff
@@ -28,6 +28,15 @@ const ensureAtLeastOneWindow = (frameOpts) => {
   }
 }
 
+const getCurrentWindowId = () => {
```
bridiver: maybe `currentWindow` should move to common and this code should go there? I'm going to add the active window to the appState and then everything can read it from the same place

bridiver: actually in this case it's even simpler because you can use -2 to specify the current window (or leave it empty)

app/common/lib/menuUtil.js
```diff
@@ -88,9 +88,13 @@ const createBookmarkTemplateItems = (bookmarks, parentFolderId) => {
...
           if (eventUtil.isForSecondaryAction(e)) {
-            CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_NEW_FRAME, site.get('location'), { openInForeground: !!e.shiftKey }])
+            appActions.createTabRequested({
+              url: site.get('location'),
+              windowId: focusedWindow.id,
+              active: !!e.shiftKey
+            })
           } else {
-            CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_ACTIVE_FRAME_LOAD_URL, site.get('location')])
+            appActions.loadURLInActiveTabRequested(focusedWindow.id, site.get('location'))
```
bridiver: with a small change to `windowState` we can use -2 here as well. Maybe add windowState.WINDOW_ID_CURRENT to match chrome.windows api?

app/filtering.js
```diff
@@ -40,6 +40,7 @@ const beforeSendHeadersFilteringFns = []
 const beforeRequestFilteringFns = []
 const beforeRedirectFilteringFns = []
 const headersReceivedFilteringFns = []
+let partitionsToInitialize = ['default']
```
bridiver: is this still necessary?

bridiver:
would be nice to use `windowState.WINDOW_ID_CURRENT` (-2) for current window (extensions already support this), but that can be a follow-up refactor. Otherwise ++ when the tests are added for moving tabs

PR #8097, opened 2017-04-06
Add support for welcome page

### LINKED ISSUE #7821: Integrate the new welcome screen
@cezaraugusto has worked with @bradleyrichter to deliver a beautiful new welcome screen (feel free to share details here)

The goal for this issue would be to integrate the screen into Brave. We'll want a new `about:welcome` or similarly named page which loads the content in an iframe.

cezaraugusto:

Auditors: @bsclifton
Closes #7821

Test Plan:
```
npm run test -- --grep="AboutWelcome component"
```

QA steps:
* Navigate to about:welcome
* should load content from https://brave.com/welcome.html

### DIFF
--- a/app/extensions/brave/content/styles/defaultStyles.css
+++ b/app/extensions/brave/content/styles/defaultStyles.css
@@ -1,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Use this file as a normalizer
+ * to edit webview for about:pages.
+*/
+
+html, body {
+  font-family: "Helvetica Neue", Arial, sans-serif;
+  font-weight: 400;
+  margin: 0;
+  padding: 0;
+}
+
+html, body, #appContainer, #appContainer > div {
+  height: 100%;
+}
+
+body {
+  font-size: 100%;
+}
+
+.welcomePage {
+  overflow: hidden;
+}
--- a/test/unit/about/welcomePageTest.js
+++ b/test/unit/about/welcomePageTest.js
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+/* global describe, before, after, it */
+
+const mockery = require('mockery')
+const {shallow} = require('enzyme')
+const assert = require('assert')
+let AboutWelcome
+require('../braveUnit')
+
+describe('AboutWelcome component', function () {
+  before(function () {
+    mockery.enable({
+      warnOnReplace: false,
+      warnOnUnregistered: false,
+      useCleanCache: true
+    })
+    AboutWelcome = require('../../../app/renderer/about/welcome')
+    mockery.registerMock('../../../less/about/common.less', {})
+  })
+
+  after(function () {
+    mockery.disable()
+  })
+
+  describe('Rendering', function () {
+    it('renders an iframe', function () {
+      const wrapper = shallow(
+        AboutWelcome
+      )
+      assert.equal(wrapper.find('[data-test-id="welcomeIframe"]').length, 1)
+    })
+  })
+})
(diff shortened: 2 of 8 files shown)


bsclifton:
Moving to 0.14.3

cezaraugusto:
just as a heads-up, blocker for PR fixed by https://github.com/brave/brave-website/commit/774bf249ebc3d356ae52f2d18d6e106cd7d55cf2. Needs landing on master for this being approved.

test/unit/about/welcomePageTest.js
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
cezaraugusto: changed to Enzyme instead of Webdriver

app/extensions/brave/content/styles/defaultStyles.css
```diff
@@ -0,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
cezaraugusto: future refactoring of about pages will benefit from this instead of relying on `common.less` file so I took advantage of this PR to add it

cezaraugusto:
moving to 0.15.4

srirambv:
@cezaraugusto this shows double vertical scroll bar on about:welcome. Is that expected??

cezaraugusto:
thanks @srirambv, fixed. the joy of having it hidden by default :)

bsclifton:
@cezaraugusto @bradleyrichter this is ready for review, right? 😄

cezaraugusto:
@bradleyrichter just pushed website to prod so yes, ready

test/unit/about/welcomePageTest.js
```diff
@@ -0,0 +1,35 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+/* global describe, before, after, it */
+
+const mockery = require('mockery')
+const {shallow} = require('enzyme')
+const assert = require('assert')
+let AboutWelcome
+require('../braveUnit')
+
+describe('AboutWelcome component', function () {
```
luixxiul: Is this intentional? The test plan says `npm run test -- --grep="about:welcome"`, which does not trigger that.

luixxiul:
screenshot 2017-05-18 11 21 19

bradleyrichter:
Yes it would. But we have exhausted ideas for this version unless you have a new idea?

luixxiul:
> unless you have a new idea?

how about moving `import` button on about:preferences#general up so that it would get displayed without scrolling? On MBP you have to scroll to find out the button.

bsclifton:
We could create a deep link using the URL, like `about:preferences#general?import`.  It would take a little work to get that to parse out and then interpreted in the React component, but it's possible

NejcZdovc:
@bsclifton we need to do this anyway for ledger as well. I would suggest that we create a separate issue and do it in 0.15.400 (ledger need it in this version).

bsclifton:
@NejcZdovc good call- can you create the issue (capturing this use case and also the one for ledger)

test/unit/about/welcomePageTest.js
bsclifton: I just updated it 😄

bsclifton:
@cezaraugusto @bradleyrichter is this screen supposed to show on first launch?  When I have a new  session, it still just loads a new tab with the dashboard  (and of course, I didn't see any code changes for this functionality)

bsclifton:
Since there's an iframe inside the page, it's showing a scrollbar which looks kind of weird:
screen shot 2017-05-19 at 11 33 31 am

The iframe should have the scrollbar (this makes sense), but the page itself shouldn't.  This is happening in part because the height is set as 100% (for a quick demo, open dev tools and set to 99%)

bsclifton:
@NejcZdovc I created the issue for the deep link functionality 😄 
https://github.com/brave/browser-laptop/issues/8966

test/unit/about/welcomePageTest.js
cezaraugusto: thanks し( ・∀・)／＼(・∀・) ///

cezaraugusto:
ok rebased & updated

app/extensions/brave/content/styles/defaultStyles.css
```diff
@@ -0,0 +1,27 @@
...
+html, body, #appContainer, #appContainer > div {
+  height: 100%;
+}
+
+body {
+  font-size: 100%;
+}
+
+.welcomePage {
+  overflow: hidden;
```
cezaraugusto: here's our hero for double scrollbars

PR #8107, opened 2017-04-06
Make use of base64 for contributionStatement img

### LINKED ISSUE #8062: Revert the inlined files on contributionStatement.js (follow up to #7887)
## Test plan
See https://github.com/brave/browser-laptop/pull/8107

## Original issue description
Revert the inlined files on contributionStatement.js (follow up to #7887) after making it possible to use the external image files.

- Any related issues: https://github.com/brave/browser-laptop/pull/7887#issuecomment-291491654 #7883

cezaraugusto:
Fix #8062
Auditors: @luixxiul, @willy-b

**Note:** We still can't make use of external files to append images on contributionStatement if we want to save it. 

After component refactor to Aphrodite (#7887) LESS file with hard-coded image string was removed and I missed that case while merging by removing hard-coded string from refactor. 

This PR makes use of `getBase64FromImageUrl` method to solve the issue and avoid hardcoding base64 strings, which could have made style edits harder.

You can follow convo here: https://github.com/brave/browser-laptop/pull/7887#issuecomment-291949576

Test plan:

- Make sure Brave logo and verified icon are base64 strings (devTools -> inspect element)
- Make sure that while saving contributionStatement both Brave logo and verified icon are always shown, even on PDF preview and saved document.

### DIFF
--- a/js/about/contributionStatement.js
+++ b/js/about/contributionStatement.js
@@ -5,8 +5,8 @@
 const React = require('react')
 const ReactDOM = require('react-dom')
 const Immutable = require('immutable')
-const ImmutableComponent = require('../components/immutableComponent')
 const {makeImmutable} = require('../../app/common/state/immutableUtil')
+const {getBase64FromImageUrl} = require('../lib/imageUtil')
 
 const ledgerExportUtil = require('../../app/common/lib/ledgerExportUtil')
 const getTransactionCSVRows = ledgerExportUtil.getTransactionCSVRows
@@ -26,7 +26,10 @@ const ipc = window.chrome.ipcRenderer
 const {StyleSheet, css} = require('aphrodite/no-important')
 const globalStyles = require('../../app/renderer/components/styles/global')
 
-class ContributionStatement extends ImmutableComponent {
+const braveLogo = require('../../app/extensions/brave/img/braveAbout.png')
+const verifiedIcon = require('../../app/extensions/brave/img/ledger/verified_green_icon.svg')
+
+class ContributionStatement extends React.Component {
   constructor () {
     super()
 
@@ -36,7 +39,9 @@ class ContributionStatement extends ImmutableComponent {
       ledgerData: Immutable.Map(),
       publisherSynopsisMap: {},
       synopsis: [],
-      savedPDF: false
+      savedPDF: false,
+      braveLogo: '',
+      verifiedIcon: ''
     }
 
     ipc.on(messages.LEDGER_UPDATED, function (e, ledgerData) {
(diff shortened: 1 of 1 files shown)


willy-b:
tested, lgtm!

js/about/contributionStatement.js
```diff
@@ -160,10 +174,19 @@ class ContributionStatement extends ImmutableComponent {
   }
 
   get ContributionStatementHeader () {
+    const imgStyle = StyleSheet.create({
```
willy-b: curious: why separate the brave logo style here rather than keeping it with the other styles as before (`styles.braveLogo`)?

js/about/contributionStatement.js
```diff
@@ -285,6 +308,17 @@ class ContributionStatement extends ImmutableComponent {
   }
 
   ContributionStatementDetailTable (page, pageIdx, totalPages) {
+    const imgStyle = StyleSheet.create({
```
willy-b: same question as above: why not keep this style with the others as before (`styles.verified`) rather than creating it here?

js/about/contributionStatement.js
```diff
@@ -26,7 +26,10 @@ const ipc = window.chrome.ipcRenderer
 const {StyleSheet, css} = require('aphrodite/no-important')
 const globalStyles = require('../../app/renderer/components/styles/global')
 
-class ContributionStatement extends ImmutableComponent {
+const braveLogo = require('../../app/extensions/brave/img/braveAbout.png')
+const verifiedIcon = require('../../app/extensions/brave/img/ledger/verified_green_icon.svg')
+
+class ContributionStatement extends React.Component {
```
willy-b: why switch from `ImmutableComponent` to `React.Component`?

js/about/contributionStatement.js
luixxiul: +1. If we will do so, why won't we do that for all files in a lump to avoid inconsistency? @cezaraugusto would you mind opening an issue to track that?

js/about/contributionStatement.js
luixxiul: Ditto.

js/about/contributionStatement.js
bsclifton: @willy-b looks like this is now using React's `setState()`. Our `ImmutableComponent` wrapper will only trigger a re-render if properties are updated (not state).  Changing this forces the setState() to re-render

bsclifton: Overall, we want to avoid React.Component in favor of ImmutableComponent... but About pages are a little different. We tend to need to keep state, because of how data gets transferred to it (see Frame.js)

js/about/contributionStatement.js
cezaraugusto: method to transform image to base64 is a thenable, so I had to pass result to a state, and `styles` is not bound to the component, so I can't call `this.state` there. That's why I had to create a separate stylesheet for each image

js/about/contributionStatement.js
cezaraugusto: pls see comment above

js/about/contributionStatement.js
cezaraugusto: I can't manipulate component-level state without using actions/stores with Immutable. Creating an action/store for that is overkill. Since we're on an about:page there's no big deal to use React.Component

bsclifton:
Merging! if we to update based on feedback by @luixxiul or @willy-b, let's do in a follow up. Thanks 😄

js/about/contributionStatement.js
willy-b: Makes sense. Thanks!

PR #8112, opened 2017-04-07
Sync profile devices list

ayumi:
**Requires https://github.com/brave/sync/pull/71**

Fixes https://github.com/brave/sync/issues/41

Test plan:
1. Start Brave and enable Sync, naming the device "Pyramid 1".
2. `npm run start2` Brave and add it to Pyramid 1's Sync profile and name it "Pyramid 2".
3. Pyramid 2's device list should show both pyramids.
4. In Pyramid 1, wait a bit, 30-60s.
5. Pyramid 1's device list should also show both pyramids.

screen shot 2017-04-07 at 07 02 24


app/sync.js
```diff
@@ -337,20 +338,28 @@ module.exports.init = function (appState) {
...
+      deviceIds = deviceIds.add(safeRecord.deviceId)
       const existingObject = syncUtil.getExistingObject(categoryName, record)
       return [safeRecord, existingObject]
     })
     event.sender.send(messages.RESOLVE_SYNC_RECORDS, categoryName, recordsAndExistingObjects)
+    // For each device we saw, update its last record timestamp.
+    // Records' Sync timestamps aren't exposed here, so approximate
+    // with the batch's last timestamp.
+    deviceIds.forEach(deviceId => {
+      appActions.saveSyncDevice(deviceId, {lastRecordTimestamp})
```
diracdeltas: performance-wise, it is better to add all the devices in a single appAction call

ayumi: :+1:

ayumi:
@diracdeltas how about now

diracdeltas:
it feels weird that there is no value in the first column of the table, yet if you click on it, it sorts.

diracdeltas:
i got this error in the page console when i synced to the `forsaken...` group, not sure if it's related

screen shot 2017-04-11 at 12 54 58 am

app/renderer/components/preferences/syncTab.js
```diff
@@ -82,12 +83,47 @@ class SyncTab extends ImmutableComponent {
...
+  }
+
+  get devicesContent () {
+    return <div className='devices'>
+      <div className='sectionTitle' data-l10n-id='syncDevices' data-test-id='syncDevices' />
+      <SortableTable
+        headings={['', 'syncDeviceName', 'syncDeviceLastActive']}
+        defaultHeading='syncDeviceLastActive'
+        defaultHeadingSortOrder='desc'
+        rows={this.props.syncData.get('devices').map((device, id) => [
```
diracdeltas: i think this needs to check that `this.props.syncData.get('devices')` is not null/undefined, may be the cause of the error i saw

ayumi: good catch

ayumi:
@diracdeltas the first column is deviceId, so you would use it to sort devices by original add date. ID probably looks a little better than nothingness.

diracdeltas:
@ayumi yeah i realized that it was deviceID, but it may be confusing for people who did not work on sync code

app/renderer/components/preferences/syncTab.js
```diff
@@ -82,12 +83,53 @@ class SyncTab extends ImmutableComponent {
...
       <SettingItem>
-        <Button l10nId='syncNewDevice' className='whiteButton' onClick={this.props.showOverlay.bind(this, 'syncNewDevice')} />
+        <Button l10nId='syncNewDevice' className='primaryButton' onClick={this.props.showOverlay.bind(this, 'syncNewDevice')} />
       </SettingItem>
     </SettingsList>
   }
 
+  get devices () {
+    let devices = []
+    this.props.syncData.get('devices').forEach((device, id) => {
```
diracdeltas: this should probably also check that devices is non-falsy

diracdeltas: actually is this used at all?

ayumi: thx– it's not; i removed it

PR #8149, opened 2017-04-09
Move top navigation outside the main.js

### LINKED ISSUE #8189: Move navigation logic into new component
Move logic related to top navigation out of `main.js` file into new component. This way it will be ready for redux component pattern

NejcZdovc:

Resolves #8189

### DIFF
--- a/app/common/lib/shieldUtil.js
+++ b/app/common/lib/shieldUtil.js
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const frameStateUtil = require('../../../js/state/frameStateUtil')
+const urlParse = require('../urlParse')
+
+function braveShieldsDisabled (windowStore) {
+  const activeRequestedLocation = frameStateUtil.activeRequestedLocation(windowStore)
+  if (!activeRequestedLocation) {
+    return true
+  }
+
+  const parsedUrl = urlParse(activeRequestedLocation)
+  return parsedUrl.protocol !== 'https:' && parsedUrl.protocol !== 'http:' && (parsedUrl.protocol + parsedUrl.host) !== 'about:safebrowsing'
+}
+
+module.exports = {
+  braveShieldsDisabled
+}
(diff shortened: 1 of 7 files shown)


luixxiul:
Please create an issue for tracking/QA, thanks.

NejcZdovc:
Will do, but first we need to define some more things :smile:

NejcZdovc:
Files move was moved into a separate PR #8187, will rebase to the master after that PR is in. This PR will now focus only on moving logic into new component. For redux implementation I will create separate PR.

app/renderer/components/navigation/navigator.js
```diff
@@ -0,0 +1,324 @@
...
+                ? null
+                : this.extensionButtons
+            }
+            <div className={css(styles.braveMenuContainer)}>
+              <Button iconClass='braveMenu'
+                l10nId='braveMenu'
+                testId='braveShieldButton'
+                className={cx({
+                  navbutton: true,
+                  braveShieldsDisabled: shieldDisabled,
```
bridiver: I think `braveShieldsDisabled` and `braveShieldsDown` can be too easily confused and the button doesn't need to know or care why shields are disabled. I would suggest combining into a single setting and method that handles both cases

NejcZdovc: we are using different css for this two states. Icon is different color if shield is down or disabled

app/common/lib/shieldUtil.js
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const frameStateUtil = require('../../../js/state/frameStateUtil')
```
bridiver: I think we would like to avoid adding new *Util classes. Maybe add this to `frameState.js` and pass in state and frameKey?

NejcZdovc: Yeah I wanted to create util for now and then when moving to redux create shield state. What do you think should we move this into frame or shield state? Keep in mind that we have a lot of logic for the panel

NejcZdovc:
@bridiver will create a new PR, so that branch will have the correct name, also travis is not triggering.

app/renderer/components/navigation/navigator.js
bridiver: ok, the naming is a little confusing, but I guess it matches what is there now

app/common/lib/shieldUtil.js
bridiver: shieldState is fine, but I'd still like to pass in the frame for now vs `activeRequestedLocation`. Can we also change the name to `braveShieldsEnabled`? !braveShieldsDisabled is harder to read

PR #8182, opened 2017-04-10
Covers instances of 'GoogleChrome'

### LINKED ISSUE #8181: Instances of "GoogleChrome" only partially replaced
Yes

The `extensionsUtil.js::bravifyText` method does not cover `GoogleChrome`.

- Steps to reproduce:
    1. Use Pocket on an older version of Brave
    2. Update to 0.14.2
    3. Note presence of "GoogleBrave" in new Extensions page

From session-store-1:
```
"base_path": "…Extensions/niloccemoadcdkdjlinkgdfekeahmflj/2.1.17",
"version": "2.1.17",
"id": "niloccemoadcdkdjlinkgdfekeahmflj",
"description": "__MSG_extDescriptionGoogleChrome__",
```
    
- Actual result:
_GoogleChrome_ is not covered by the `bravifyText` method, resulting in _GoogleBrave_ appearing in the new Extensions page.

- Expected result:
Brave should cover the missing case.

No.

No.

Yes.

- Screenshot if needed:
image

- Any related issues:

jonathansampson:

Fixes #8181

Also removes old stub version of `bravifyText` from _js/about/extensions.js_.

### DIFF
--- a/app/renderer/lib/extensionsUtil.js
+++ b/app/renderer/lib/extensionsUtil.js
@@ -158,4 +158,6 @@ module.exports.isPasswordManager = (extensionId) => {
  * @param {String} text - The text to check against
  * @returns {String} Text with replaced string
  */
-module.exports.bravifyText = (text) => text && text.replace(/Chrome|Google Chrome/g, 'Brave')
+module.exports.bravifyText = (text) => {
+  return text && text.replace(/Chrome|Google ?Chrome/g, 'Brave')
+}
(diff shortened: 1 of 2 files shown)


bsclifton:
@jonathansampson can you update this (and the original issue) to have testing steps? Thanks 😄

jonathansampson:
@bsclifton In this particular case, you would need to have an older session-store-1 file that had Pocket's name/description cached. Upon updating Brave, the cached name/description from session-store-1 would be loaded, and the incorrect value would be shown.

I don't think this will be something we need to watch moving forward.

We could, and probably should, write a few unit tests for `bravifyText` though.

bsclifton:
@jonathansampson ah ok...  unit tests work 😄  Would you be able to make some? (I can help if needed, let me know).  Then we can mark this as no QA needed, since it would be covered by automated tests

jonathansampson:
@bsclifton I still have issues running tests on Windows, but I think Unit Tests will work easier than the Chrome Driver ones. I'll write up a couple tests, and see if they work as expected.

PR #8205, opened 2017-04-10
Moves navigation logic from main.js to new component

NejcZdovc:

Resovles #8189

### Auditor
@bridiver

### DIFF
--- a/app/common/state/shieldState.js
+++ b/app/common/state/shieldState.js
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const frameStateUtil = require('../../../js/state/frameStateUtil')
+const urlParse = require('../urlParse')
+
+function braveShieldsEnabled (frame) {
+  const lastCommittedURL = frameStateUtil.getLastCommittedURL(frame)
+  if (!lastCommittedURL) {
+    return false
+  }
+
+  const parsedUrl = urlParse(lastCommittedURL)
+  return !(parsedUrl.protocol !== 'https:' && parsedUrl.protocol !== 'http:')
+}
+
+module.exports = {
+  braveShieldsEnabled
+}
--- a/app/renderer/components/navigation/navigator.js
+++ b/app/renderer/components/navigation/navigator.js
@@ -1,0 +1,342 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const React = require('react')
+const Immutable = require('immutable')
+const {StyleSheet, css} = require('aphrodite')
+
+// Actions
+const appActions = require('../../../../js/actions/appActions')
+const windowActions = require('../../../../js/actions/windowActions')
+const contextMenus = require('../../../../js/contextMenus')
+const getSetting = require('../../../../js/settings').getSetting
+
+// Components
+const ImmutableComponent = require('../../../../js/components/immutableComponent')
+const NavigationBar = require('./navigationBar')
+const LongPressButton = require('../../../../js/components/longPressButton')
+const Menubar = require('../menubar')
+const WindowCaptionButtons = require('../windowCaptionButtons')
+const Button = require('../../../../js/components/button')
+const BrowserAction = require('../browserAction')
+
+// State
+const tabState = require('../../../common/state/tabState')
+const extensionState = require('../../../common/state/extensionState')
+const siteSettingsState = require('../../../common/state/siteSettingsState')
+
+// Util
+const {currentWindow, isMaximized} = require('../../currentWindow')
+const {makeImmutable} = require('../../../common/state/immutableUtil')
+const platformUtil = require('../../../common/lib/platformUtil')
+const {braveShieldsEnabled} = require('../../../common/state/shieldState')
+const tabUtil = require('../../lib/tabUtil')
+const eventUtil = require('../../../../js/lib/eventUtil')
+const {isNavigatableAboutPage, getBaseUrl} = require('./../../../../js/lib/appUrlUtil')
+const frameStateUtil = require('../../../../js/state/frameStateUtil')
+const siteSettings = require('../../../../js/state/siteSettings')
+const cx = require('../../../../js/lib/classSet')
+
+// Constants
+const settings = require('../../../../js/constants/settings')
+const appConfig = require('../../../../js/constants/appConfig')
+
+class Navigator extends ImmutableComponent {
+  constructor () {
+    super()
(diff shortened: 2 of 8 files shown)


NejcZdovc:
@bridiver as per your comment https://github.com/brave/browser-laptop/pull/8149/files#r110748326 I changed util into state and change `braveShieldsDisabled` to `braveShieldsEnabled`

NejcZdovc:
@bridiver in https://github.com/brave/browser-laptop/pull/8205/commits/d759270e8b69fd09a322f6f5f264db7c9f5aea20 I also changed entry param, so that we only operate with activeFrame, we don't need state for this function

NejcZdovc:
@bsclifton @bridiver can we please merge this one?

app/common/state/shieldState.js
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const frameStateUtil = require('../../../js/state/frameStateUtil')
+const urlParse = require('../urlParse')
+
+function braveShieldsEnabled (frame) {
+  const activeRequestedLocation = frameStateUtil.activeRequestedLocation(frame)
```
bridiver: as a side node, `committed` does not mean that the url has completely loaded, it just means that a load was started and could have been blocked or redirected. Before `load-start` a url would be considered `provisional`

test/unit/app/renderer/components/navigation/navigatorTest.js
```diff
@@ -59,12 +59,32 @@ describe('Main component unit tests', function () {
...
+  const activeTab = appState.getIn(['tabs', 0])
+
   describe('when user has history going forwards and backwards', function () {
     let wrapper
 
     before(function () {
       wrapper = shallow(
-        <Main windowState={windowState} appState={appState} />
+        <Navigator
+          windowState={windowState}
```
bridiver: passing in the entire window and app state as props will cause the Navigator component to re-render every time anything changes in either one of these stores. We need to make these props much more fine grained so they only update when they actually need to

app/renderer/components/navigation/navigator.js
```diff
@@ -0,0 +1,342 @@
...
+    this.onForwardLongPress = this.onForwardLongPress.bind(this)
+    this.onDoubleClick = this.onDoubleClick.bind(this)
+    this.onDragOver = this.onDragOver.bind(this)
+    this.onDrop = this.onDrop.bind(this)
+    this.onBraveMenu = this.onBraveMenu.bind(this)
+  }
+
+  onNav (e, navCheckProp, navType, navAction) {
+    const activeFrame = frameStateUtil.getActiveFrame(this.props.windowState)
+    const activeTabId = tabUtil.activeTabId(this.props.windowState)
```
bridiver: can't we use `props.activeTab` here?

app/renderer/components/navigation/navigator.js
```diff
@@ -0,0 +1,342 @@
...
+    const extensionBrowserActions = enabledExtensions
+      .map((extension) => extensionState.getBrowserActionByTabId(this.props.appState, extension.get('id'), activeTabId))
+      .filter((browserAction) => browserAction)
+
+    let buttons = extensionBrowserActions.map((browserAction, id) =>
+      <BrowserAction
+        browserAction={browserAction}
+        extensionId={id}
+        tabId={activeTabId}
+        popupWindowSrc={this.props.windowState.getIn(['popupWindowDetail', 'src'])} />
```
bridiver: Can we pass in popupWindowSrc as a prop? See my general note below about passing in props for anything that uses this.props.windowState or this.props.appState. When we transition to ReduxComponent we will just move the props from the parent to `mergeProps`

app/renderer/components/navigation/navigator.js
```diff
@@ -0,0 +1,342 @@
...
+    return (blocked === 0)
+      ? false
+      : ((blocked > 99)
+        ? '99+'
+        : blocked)
+  }
+
+  get extensionButtons () {
+    const activeTabId = tabUtil.activeTabId(this.props.windowState)
+    const enabledExtensions = extensionState.getEnabledExtensions(this.props.appState)
```
bridiver: comments about this.props.windowState also apply to anything using this.props.appState

app/renderer/components/navigation/navigator.js
```diff
@@ -0,0 +1,342 @@
...
+                className='navigationButton forwardButton'
+                disabled={!activeTab || !activeTab.get('canGoForward') || activeTabShowingMessageBox}
+                onClick={this.onForward}
+                onLongPress={this.onForwardLongPress}
+              />
+            </div>
+          </div>
+          <NavigationBar
+            navbar={activeFrame && activeFrame.get('navbar')}
+            sites={this.props.appState.get('sites')}
```
bridiver: this is where prop lifting gets really annoying. All of these props need to be passed in to the Navigator component so they can be passed along to NavigationBar. It's very convenient to use this.props.appState instead, but unfortunately it is not good for performance

js/state/siteUtil.js
```diff
@@ -747,3 +749,7 @@ module.exports.setObjectId = (item) => {
   const crypto = require('crypto')
   return item.set('objectId', new Immutable.List(crypto.randomBytes(16)))
 }
+
+module.exports.enableNoScript = (settings, state) => {
```
bridiver: siteUtil.js is actually for `appState.sites` and site settings are in `appState.siteSettings`. The names are similar, but they are two very distinct things. Maybe a new `siteSettingsState` in `app/common/state`?

Also this method sounds like it will enable noScript, but it's actually checking the state of noScript so maybe `noScriptEnabled` or `isNoScriptEnabled`? Can we swap also swap the state and settings parameters to be consistent with other state helpers and maybe move this to `siteSettingsState`?

bridiver:
after discussion with @NejcZdovc all of this stuff was going to be re-rendered anyway as part of Main.js so this doesn't really create any new performance problems.

NejcZdovc:
yes, this PR only moves existing logic from main.js to the new component. I didn't introduce any new logic or optimization. All problems that you mentioned were already there. With this PR I wanted to prepare a good base for the redux implementation into the navigator section.

app/common/state/shieldState.js
```diff
@@ -0,0 +1,20 @@
...
+const urlParse = require('../urlParse')
+
+function braveShieldsEnabled (frame) {
+  const lastCommittedURL = frameStateUtil.getLastCommittedURL(frame)
+  if (!lastCommittedURL) {
+    return false
+  }
+
+  const parsedUrl = urlParse(lastCommittedURL)
+  return !(parsedUrl.protocol !== 'https:' && parsedUrl.protocol !== 'http:')
```
bridiver: merging now, but just a nit that this would be easier to read in the positive form `parsedUrl === 'https'` or better yet use `urlutil.isHttpOrHttps`

bridiver:
++

PR #8220, opened 2017-04-11
remove most references to currentWindow and lazy load the rest

bridiver:
check for automated tests to pass before merging


bridiver:
updated with unit test fix

app/renderer/components/navigation/navigator.js
```diff
@@ -169,7 +169,7 @@ class Navigator extends ImmutableComponent {
     if (!e.target.className.includes('navigatorWrapper')) {
       return
     }
-    return !isMaximized() ? currentWindow.maximize() : currentWindow.unmaximize()
+    return !isMaximized() ? windowActions.shouldMaximize(getCurrentWindowId()) : windowActions.shouldMinimize(getCurrentWindowId())
```
bbondy: I guess should* is kind of like *Requested for tab creation, maybe we should use one or the other.  No need to hold this up though just thinking out loud.

bbondy:
++

app/renderer/components/navigation/navigator.js
bridiver: these are really just temporary to replace the currentWindow.method calls. This should really be `onNavigatorDoubleClick` or something like that

bridiver: see https://github.com/brave/browser-laptop/pull/8220/files#diff-2db1344f65c9be10474c5562bcef4c60R1134

PR #8225, opened 2017-04-11
refactor of Frame (redux component)

### LINKED ISSUE #8392: Redux component - Frame
Transform Frame into Redux component.

NejcZdovc:

Resolves #8392 

Test Plan:
test that the browser starts normally

### DIFF
--- a/app/common/state/siteState.js
+++ b/app/common/state/siteState.js
@@ -1,0 +1,18 @@
+const assert = require('assert')
+const {makeImmutable, isMap, isList} = require('./immutableUtil')
+
+const validateState = function (state) {
+  state = makeImmutable(state)
+  assert.ok(isMap(state), 'state must be an Immutable.Map')
+  assert.ok(isList(state.get('sites')), 'state must contain an Immutable.List of sites')
+  return state
+}
+
+const siteState = {
+  getSites: (state) => {
+    state = validateState(state)
+    return state.get('sites')
+  }
+}
+
+module.exports = siteState
--- a/app/renderer/components/common/messageBox.js
+++ b/app/renderer/components/common/messageBox.js
@@ -3,20 +3,33 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const React = require('react')
-const ImmutableComponent = require('../immutableComponent')
+const {StyleSheet, css} = require('aphrodite')
+
+// Components
+const ReduxComponent = require('../reduxComponent')
 const Dialog = require('../../../../js/components/dialog')
 const Button = require('../../../../js/components/button')
 const SwitchControl = require('../../../../js/components/switchControl')
+
+// Actions
 const appActions = require('../../../../js/actions/appActions')
+
+// Constants
 const KeyCodes = require('../../../common/constants/keyCodes')
-const config = require('../../../../js/constants/config')
+
+// State
+const tabState = require('../../../common/state/tabState')
+const tabMessageBoxState = require('../../../common/state/tabMessageBoxState')
+const frameStateUtil = require('../../../../js/state/frameStateUtil.js')
+
+// Utils
 const {makeImmutable} = require('../../../common/state/immutableUtil')
 
-const {StyleSheet, css} = require('aphrodite')
+// Styles
 const commonStyles = require('../styles/commonStyles')
 const globalStyles = require('../styles/global')
 
-class MessageBox extends ImmutableComponent {
+class MessageBox extends React.Component {
   constructor () {
     super()
     this.onKeyDown = this.onKeyDown.bind(this)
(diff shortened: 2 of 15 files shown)


NejcZdovc:
blocked by #8238

NejcZdovc:
@bsclifton @bridiver @bbondy unit test is still failing, but we need to review this and merge it, otherwise merge conflicts will start popping up

**Unit test problem**
Problem is with calling a function on Frame component that is wrapped into ReduxComponent via connect, so you can’t call function directly like this
https://github.com/brave/browser-laptop/pull/8225/files#diff-a6847547aef4f43c811e9e97dcb0322cL46
or spy on it like this 
https://github.com/brave/browser-laptop/pull/8225/files#diff-a6847547aef4f43c811e9e97dcb0322cL45

luixxiul:
sorry for stepping in but if you have several PRs expecting I think you are expected to work on another base branch like `redux-refactoring` and solve conflicts later by yourself (sorry if that is not a common practice on github). See: https://github.com/brave/browser-laptop/pull/8443 and the PRs referred there. I think it is not a good practice to hurry people to review PRs because they have been already so busy working on something. The reason why I claim so is that I'm afraid of some kind of regressions could happen. Especially without test plans there should be much possibility that regressions would happen.

NejcZdovc:
@luixxiul thank you for feedback. I understand that and I added label for a review only today. I don't expect or demand from them to do it today :smiley:. All three of them already helped me a lot on this PR, for which I am really grateful. 

Why I think we can't go with a base branch is that things couldn't be merged after redux is done. What we are doing with redux components and moving things into state is new way of writing components. Because of that some things are not compatible and we will change a lot functions and how they are called. And if we don't merge this on a frequent basis, other developers would adding code that will be then removed for example or using functions that are not valid anymore. This way they will waste time writing an old code.

Because of that we wanted to approach this redux refactoring with small PR's and not doing everything in one big swing. @bridiver did already a lot of amazing work for url bar (changing to redux and optimizing things) and it's in master already. This way developers can start using this new approach faster.

For test's I didn't add them, because we already have test's for frames and with this PR I didn't implement any new feature, just move things around.

luixxiul:
> I didn't implement any new feature, just move things around.

Sometimes even doing so would cause troubles (slow performance etc.). If you improve something with this PR would you mind adding a manual test plan even if it would be a tiny one? like:

    1. Clear your profile
    2. Open the browser
    3. Make sure the browser starts in/under ... seconds

Otherwise more simply: `test that the browser starts normally`. That would help, thanks.

app/renderer/components/common/messageBox.js
```diff
@@ -3,20 +3,34 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const React = require('react')
-const ImmutableComponent = require('../immutableComponent')
+const {StyleSheet, css} = require('aphrodite')
```
luixxiul: Can't it be aphrodite/no-important?

NejcZdovc: @luixxiul not sure, we need to ask @bsclifton about this code, I just reordered imports.

NejcZdovc:
@luixxiul sure I can do this, not a problem

luixxiul:
@NejcZdovc please make sure to add test steps to the 1st post for fixes of these fails on travis: https://travis-ci.org/brave/browser-laptop/jobs/226054730#L2461

app/renderer/components/common/messageBox.js
bsclifton: @luixxiul maybe? we could do that, but it would require testing. I would vote to leave it as is.  IMO, using `no-important` takes away the benefit we get by using Aphrodite.

We introduced Aphrodite to our code base because we were getting destroyed by regressions and the goal (I thought) was to make styles NOT overridable.  When it's `no-important`, styles might be overridden (which can cause regression).

luixxiul: OK, I understand. I proposed it so we could remove styles in the existing LESS files at the same time. It is fine to replace them with later.

Otherwise: I'm thinking we would set `no-important` to the general components (such as ones in `common` folder) and not to other specific ones, considering [the possibility of removing !important from everything](https://github.com/Khan/aphrodite#todo).

That means:

- `aphrodite` for files in `app/renderer/components/` such as `releaseNotes.js`
- `aphrodite/no-important` for files in `app/renderer/components/common/` such as `dropdown.js`

What do you guys think? @bsclifton @cezaraugusto @NejcZdovc. By doing so we could avoid overwriting `!important` with `!important`, which is obviously not a good habit on CSS.

The thing is that I noticed that it seemed to be impossible to cascade styles with `cx` as we usually do with CSS, which I noticed during refactoring `modalOverlay.js`. eg:

````
<div className={cx({
    [modalOverlay__generalHeaderStyles]: true,
    [css(styles.modalOverlay__paymentHistoryHeaderStyles)]: true
})}
````
I thought that kind of styling should work, overwriting the styles specified with `modalOverlay__generalHeaderStyles`, but I could not find the proper way, instead of creating another wrapper `div` for that `div`. If any, please let me know. @bsclifton @cezaraugusto @NejcZdovc 

We could discuss the thing on another issue if you wish.

Comment (bsclifton) [inline app/renderer/components/common/messageBox.js]: I think I need to understand more about Aphrodite and the use cases for this (I think I'm missing something). Let's definitely chat more about it 😄

Comment (cezaraugusto) [inline app/renderer/components/common/messageBox.js]: yep per @luixxiul example, Aphrodite will concatenate styles and the last would take precedence. I think that doesn't happen with `cx` but definitely happen with standard `css(style1, style2)`.

I'm against `!important` whenever possible. I would compare it to JS `eval()`, it takes precedence whenever it's defined, and that's a bad pattern. It's a feature that might be useful for some edge-cases until we aren't using Aphrodite 100% but long term we should avoid. 

For new components, `!important` is a no-op but this is a huge PR and if that was previously defined we should keep it as-is. As we head closer to 1.0 we should avoid re-work and reduce regression risk at all costs. However, wouldn't mind reviewing a follow-up removing that, but for this PR I think we are fine with what we have.

Comment (luixxiul) [inline app/renderer/components/common/messageBox.js]: > I think that doesn't happen with `cx`

Can we improve it to support cascading styles?

Comment (cezaraugusto) [inline app/renderer/components/common/messageBox.js]: `cx()` is just a lib to apply styles conditionally, so it's affected by the same rules as CSS like deeply nested selectors taking precedence, and `!important` which overrides everything. 

I don't think there's anything we can do with that but just avoid adding new LESS stuff and keep using until it can be removed.

Comment (luixxiul) [inline app/renderer/components/common/messageBox.js]: My solution in mind is:

cascading

I think it might involve to send a patch to Aphrodite, still it seems worth implementing. With that improvement it would be easy to work with common components, enhancing them with more specific ones.

If that kind of cascading works only by improving `cx`, that would be quite fantastic! @cezaraugusto do you think it is possible?

**Comment by NejcZdovc** [comment]
After discussion with @bsclifton unit test was disabled for frame for now. We will address issue of fixing unit tests for redux components in a different PR. We will track all this problems in #8528

File: app/common/state/siteState.js [hunk 2]
```diff
@@ -0,0 +1,18 @@
+const assert = require('assert')
+const { makeImmutable, isMap, isList } = require('./immutableUtil')
```
bsclifton: nitpick- I don't think it's documented anywhere, but we don't have spaces when doing destructuring in most of our code.  So it could look like this instead:
`const {makeImmutable, isMap, isList} = require('./immutableUtil')`

NejcZdovc: Fixed

js/components/main.js
```diff
@@ -939,43 +928,8 @@ class Main extends ImmutableComponent {
...
-                flash={this.props.appState.get('flash')}
-                widevine={this.props.appState.get('widevine')}
-                allSiteSettings={allSiteSettings}
-                frameSiteSettings={this.frameSiteSettings(frame.get('location'))}
-                onFindHide={this.onFindHide}
-                enableNoScript={siteSettingsState.isNoScriptEnabled(this.props.appState, this.frameSiteSettings(frame.get('location')))}
-                braveryDefaults={braveryDefaults}
-                isPreview={frame.get('key') === this.props.windowState.get('previewFrameKey')}
-                isActive={frameStateUtil.isFrameKeyActive(this.props.windowState, frame.get('key'))}
               />)
```
bsclifton: Bye bye code! 🗑  
😄

NejcZdovc:
I will merge this. If we find some more improvements, we can do follow up later on. I need this PR merged so that I can continue with work on #8651. @bsclifton I added note about unit test in #8528.

bsclifton:
Perfect, thanks @NejcZdovc 😄

Issue #8235, opened 2017-04-11
Remove siteDetails.less


luixxiul:
**Describe the issue you encountered:** Remove siteDetails.less

This file is widely used on about pages, and some of them require another LESS file which imports siteDetails.less (X.less requires history.less, which imports siteDetails.less, for example). Creating the framework with Aphrodite reduces such cases and makes it easy to remove siteDetails.less completely.

So far these elements have been created into custom components:

- [x] `<SectionTitle>` (#8237)
- [ ] `<SiteDetailsPageHeader>`
  - [ ] `<SiteDetailsPageHeader  actions>` inside the Header (ref history.js and bookmarks.js)
- [ ] `<SiteDetailsPageContent>`
- [ ] TBA

Any related issues: #8165

PR #8044, opened 2017-04-02
Refactor syncTab.js with Aphrodite

luixxiul:
Closes #8042 
Addresses #7989
Addresses #8054
Addresses #8587

- Unified the modal dialog design of payments and sync
  - Copied the styles of modal dialog of payments to commonStyles.js, adding modalOverlay__dialog and modalOverlay__title to ModalOverlay on syncTab.js
  - Copied panelMargin, panelItemMargin, and panelPadding from payment.js to global.js
  - Added modalVeryLightGray, modalLightGray, and borderRadiusModal to global.js
  - Replaced linkButton with primaryButton
  - Set flex:1 to two elements under postSetupContent
  - TODO: refactor modalOverlay with Aphrodite to remove !important from syncTab.js and commonStyles.js

- Introduced modalOverlay__footer and modalOverlay__footerButton
  - Removed advanceFooter and replaced paymentCommon.advanceFooter with them to make the styling of the modal dialog footer consistent

- Removed styles under .syncContainer, #syncPassphrase, and #syncQR in preferences.less
  - Unified margins between buttons on dialog to 8px with globalStyles.spacing.overlayButtonMargin
  Aligned buttons to the right

- Removed .formControl from forms.less
  - We no longer need it as commonStyles.formControl is applied manually, along with commonStyles.textbox, commonStyles.textbox__outlineable, and commonStyles.textbox__isSettings.

Also:
- Changed settingsListContainerMargin into a variable to apply it to syncOverlayBody__form
- Added "Done" button to the new device sync dialog
- Replaced 'sync' with 'Sync'
- Updated test code

Auditors:

Test plan 1:
1. Open about:preferences#sync
2. Verify you are able to sync two devices using the secret code
3. Visit a site on device 1 and change shield setting, ensure that the saved site preference is synced to device 2
4. Enable Browsing history sync on device 1, ensure the history is shown on device 2
5. Import/Add bookmarks on device 1, ensure it is synced on device 2
6. Ensure imported bookmark folder structure is maintained on device 2
7. Ensure bookmark favicons are shown after sync
8. Open about:preferences#payments
9. Click "Add funds..."
10. Make sure the style of the modal overlay dialog is same as that of sync dialog

Test Plan 2:
1. Automated tests should pass
2. Open about:preferences#sync
3. Make the window small
4. Scroll the page to the bottom
5. Make sure there is 2rem margin between the reset sync button and the the border of the window

Test Plan 3:
1. Open about:preferences#sync
2. Open about:preferences#payments
[...]

PR #8038, opened 2017-04-02
Refactor payment/history.js

luixxiul:
See: cropped picture of the 6th commit on https://github.com/brave/browser-laptop/pull/6047#issue-193827878

Closes #8037

- Changed `aphrodite` to `aphrodite/no-important` (to speed up removing the LESS files)
  Aphrodite without no-important overwrites the existing code, which should have been removed from the LESS files at the same time. It will make it difficult to know which styles should be kept and removed, having confidence in that changes would not cause regressions. It might be slowing down the refactoring work.

- Modified modalOverlay.js by adding `customHeaderClassesStr`
  The intent of the change is to make it possible to apply custom styles to `sectionTitle`. In this case, `braveMediumOrange` to the title.

- Introduced `paymentStylesVariables` and `paymentHistoryTablePadding`
- Appied it to `leftRow` and `paymentHistoryOverlay__title`
  It aligns the padding-left of `sectionTitle` and the most left row.

- Set `position:sticky` to the table header
  It always display the table header. `Sticky` has been available on chromium since a couple of versions.

- Set `columnHeight` and `columnPadding`

- Reverted `commonStyles.primaryButton` to `primaryButton` temporarily
  There have been already `browserButton` and `primaryButton` available with commonStyles.js, still these styles are overwritten by ones which button.less defines. We have to work on refactoring button.less at first.

Also:
- Removed medium
- Removed pull-left, which is no longer used

Auditors:

Test Plan:
1. Open about:preferences#payments
2. The modal dialog looks like the cropped picture of the 6th commit on https://github.com/brave/browser-laptop/pull/6047#issue-193827878

- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Test Plan:

PR #7751, opened 2017-03-16
Restructuring paymentsTab.js

luixxiul:
Closes #7750

Auditors:

- Removed 'pull-left', which is no longer used
- Removed styles.titleBar
- Created flexAlignCenter and flexAlignEnd as general properties
- Replaced margin and padding with position:relative to set space between elements
- Renamed advanceIcon to settingIcon
- Changed the FAQ icons color to that on about pages (cf: about:preferences#plugins)

Also:
- Added globalStyles.fontSize.settingItemSubtext to global.js
- Added commonTextColor to preferences.less
- Added globalStyles.color.commonTextColor to global.js
- Added autoSuggestSwitch__subtext

Test Plan:
1. Open about:preferences#payments
2. Enable the payment
3. Make sure the switches, their labels and the buttons are aligned horizontally
4. Disable the payment
5. Make sure the switches and the labels are aligned

screenshot 2017-04-04 18 18 26

- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Test Plan:

PR #8237, opened 2017-04-11
Create sectionTitle.js

### LINKED ISSUE #7750: restructure paymentTab.js
## Test plan

https://github.com/brave/browser-laptop/pull/7751#issue-214772791

----

screenshot 2017-03-16 20 53 44

Refactored into this: 

screenshot 2017-04-04 18 18 26

- Any related issues: #7348 #7532

### LINKED ISSUE #8165: Update header style on about:preferences pages
When reviewing the Extensions page, I noticed that the header style didn't match the other pages:
prefs-top

The orange style looks great though- we should consider putting this style on the top of each page (possibly with a description of what the page has).  

Here's the page where I noticed the new style (which looks great)
screen shot 2017-04-09 at 10 21 57 pm

cc: @bradleyrichter @cezaraugusto @luixxiul

### LINKED ISSUE #8234: Remove brave.less with sectionTitle.js
**Describe the issue you encountered:** Remove brave.less with sectionTitle.js

- Extra QA steps:
    1.
    2.
    3.

- Any related issues:

### LINKED ISSUE #8349: Add commonForm components to about:styles
**Describe the issue you encountered:** Add `commonForm` components to `about:styles`. 

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues: https://github.com/brave/browser-laptop/issues/7990

### LINKED ISSUE #8350: Add sectionTitle components to about:styles
Add sectionTitle components to about:styles.

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues: #8237

luixxiul:
TODO: Refactor modalOverlay.js to remove `.sectionTitle`. ~~This requires #8038~~ merged

**This PR blocks https://github.com/brave/browser-laptop/pull/8044.**

Closes #7750
Closes #8165
Closes #8234
Addresses #8235
Closes #8349
Closes #8350

Created `TitleWrapper`, `SectionTitle`, and `SubTitle`
 - SectionTitleWrapper and BetaSectionTitleWrapper
 - DefaultSectionTitle, PrefSectionTitle, and AboutPagesSectionTitle
 - AboutPagesSectionSubTitle and SectionSubTitle

Refactored with those components:
 - extensionsTab.js, pluginsTab.js, preferences.js, syncTab.js, and paymentsTab.js
 - about.js, brave.js, bookmarks.js and history.js
 - torrentViewer.js, torrentStatus.js and torrentFileList.js

Added `siteDetailsPageContent` to commonStyles.js and applied:
 - about.js, brave.js, and history.js

Removed styles of sectionTitle from:
 - brave.less
 - siteDetails.less

Auditors:

Test Plan 1:
1. Open about:preferences#search
2. Make sure there is margin under the top header

Test Plan 2:
1. Open about:preferences#sync
2. Make sure "beta" is aligned top
3. Open about:preferences#payments
4. Make sure the labels appear the same row
5. Make sure the switches and the labels are aligned exactly like #7751

Test Plan 3:
1. Open about:about
2. Make sure there is margin under the header
3. Open about:brave, about:bookmarks, and about:history
4. Make sure the header font size is same

Test Plan 4:
1. Open https://webtorrent.io/free-torrents/
2. Click .torrent link
3. Make sure the header font size is same as about:brave
4. Start download
5. Make sure the header font size is same

Test Plan 5:
1. Clear your profile folder
2. Open about:preferences#sync
3. Enable Sync
4. Make sure there is margin under "Devices"

Test Plan 6:
1. Open "Import Browser data" from the menu
2. Make sure the switch is indented

Test Plan 7:
1. Open about:styles
2. Click "commonForm" on TOC
3. Make sure the template of the commonForm is displayed
4. Click "Go back to top"
5. Click "sectionTitle" on TOC
6. Make sure section title components are displayed

Test Plan:


luixxiul:
This PR blocks https://github.com/brave/browser-laptop/pull/8044.

app/renderer/components/preferences/paymentsTab.js
```diff
@@ -264,8 +276,7 @@ const styles = StyleSheet.create({
   paymentsContainer: {
     position: 'relative',
     overflowX: 'hidden',
-    width: '805px',
-    marginTop: '15px'
```
luixxiul: To make the header of payments and sync appear on the same height.

app/renderer/components/sectionTitle.js
```diff
@@ -0,0 +1,131 @@
...
+  render () {
+    return <header className={css(styles.sectionTitleWrapper)}>
+      <h1 className={css(
+        commonStyles.userSelectNone,
+        styles.sectionTitle)} {...this.props} />
+    </header>
+  }
+}
+
+class PrefSectionTitle extends ImmutableComponent {
```
luixxiul: As this is not going to be used, I'm removing it.

js/webtorrent/components/torrentViewer.js
```diff
@@ -9,6 +9,8 @@ const Button = require('../../components/button')
 const TorrentFileList = require('./torrentFileList')
 const TorrentStatus = require('./torrentStatus')
 
+const {AboutPagesSectionTitle} = require('../../../app/renderer/components/sectionTitle')
```
luixxiul: I'll rename the component (webtorrent is not one of about pages)

app/renderer/components/sectionTitle.js
```diff
@@ -0,0 +1,131 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const React = require('react')
+const ImmutableComponent = require('../../../js/components/immutableComponent')
+
+const {StyleSheet, css} = require('aphrodite/no-important')
+const globalStyles = require('./styles/global')
+const commonStyles = require('./styles/commonStyles')
+
+class SectionTitleWrapper extends ImmutableComponent {
```
NejcZdovc: Please split this into single component per file. You should create new folder and place everything there.

luixxiul: I'm afraid doing so would make it hard to maintain the style consistency and enhance components (since you would need to check out every file in the folder not to create duplicates). For me it could cause style scattering.

Actually I prefer calling components with one line like this: https://github.com/brave/browser-laptop/pull/8237/files#diff-943709aef6ee258bb12a2cbb33018988R18, which is the same way as you would load `<commonForm>` components. See: https://github.com/brave/browser-laptop/pull/8011/files#diff-fd618a0193d7534a298420a3d3b4a711R22

I'm cool with splitting anyway. I just would like to know how to avoid styles and components from scattering. (adding components to about:styles would be that way.)

less/modalOverlay.less
```diff
@@ -77,6 +77,7 @@
       transition: @transitionFast;
     }
 
+    /* Copied to siteDetails.js */
```
luixxiul: Will correct the comment -> sectionTitle

app/renderer/components/sectionTitle.js
NejcZdovc: yeah what I would do is add folder for it. This way we know that all style components for about pages are there.

luixxiul: Fine, also let's add the components to about:styles. Splitting and adding them would be perfect.

luixxiul: We could create a sub panel on about:styles for components in the same folder. Creating a long one page is not what we would like to see.

NejcZdovc: what if we create new folder in `renderer` folder named `styles` and we place all styling components into it (one component per file). In there we would keep the same folder structure as we have in `components` folder. This way we keep the context of it. What do you think?

This is one option. Another one is to make exception for styling components and keep them in one file (I still prefer splitting them)

cc @bbondy @bsclifton @cezaraugusto

NejcZdovc: Regarding including, I also like this destructured approach. What I would suggest is something like this.

You are now creating things for `sectionTitle`. Create folder `sectionTitle` and place all components into it. Then in this folder create file `index.js` where you are exporting all this components. This way you can easily use your approach and file structure is still search and time friendly :smiley:

NejcZdovc: Based on this PR I will crate wiki regarding best practices, so that we will all work in the same way. So this PR is really important, we should define and think things in detail. Please don't change things yet until we find and define how should we tackle this one.

luixxiul: Sweet, because I'm still a fan of about:styles, linking to that file from about:styles looks good finally, in my opinion.

Waiting for opinion of other members who will be back online after the holiday.

cezaraugusto: I endorse creating a new folder to include this file but despite being enthusiastic on creating small components, they're still too small to host their own files IMO. 

Since they're just titles, it's very unlikely that we'll add more methods and bulk file with new code.

For the new folder name, I'd suggest something like common/shared to host this, and in future include things like `button` and our already existent `settings` component, so for all highly replicated components. Even if we decide to create an `about` folder -- which I endorse as well -- they should be on that folder

luixxiul:
> mind just moving `sectionTitle` inside a folder called `common` i.e. `app/renderer/components/common/sectionTitle`?

it sound good to me; we would then move other files (dropdown.js, list.js, textbox.js and messageBox.js) into that folder as well.

NejcZdovc:
@cezaraugusto as I said I am partly ok with styling component's to be in the same folder, because I think that we won't write test for them. I would still prefer separate files, but we should only use this approach for styling components and nothing more. All other components should be in a separate files.

luixxiul:
It seems to me that it is not good timing to separate everything rigorously. We even do not know yet how many files and components there will be after refactor work, and how kind of issues there will be coming due to either "many files with a single component" or "many components with a single file". It is not a good idea to spend time on the issue which is yet to exist.

So, how about trying both way to create rules later? I don't think there is the general solution here (as there is not so much projects other than us who adopt Aphrodite), and both way seems to me worth trying to find the better one.

To avoid the chaos, we should create a tracking issue, commenting which components are refactored in which way.

On secitonTitle.js, let me add the sectionTitle components to about:styles precidely and see if that will avoid the issue.

app/renderer/components/sectionTitle.js
```diff
@@ -0,0 +1,131 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const React = require('react')
+const ImmutableComponent = require('../../../js/components/immutableComponent')
```
NejcZdovc: ImmutableComponent was moved into app folder, please update this path

cezaraugusto:
@NejcZdovc sorry if you got me wrong, I agree with you. I just think that for this case, specifically, we should leave as-is. Those titles don't have conditional rendering, nor do anything dynamically that's worth testing. The benefit of having them all together is the easy to change quickly if design spec changes

bsclifton:
@NejcZdovc @cezaraugusto @luixxiul I saw the above- it looks OK to me

Because there are multiple in one file and they are related, it's easy to edit there.  What pops out to me is that none of the components have logic, other than render.  **Since they are related and ONLY have a render, I would argue it's OK to have them in the same file** 😄  Once a component adds any other property or becomes unrelated to the ones there, I believe it should be moved out.  

An example where I myself did this is with the TextBox component:
https://github.com/brave/browser-laptop/blob/master/app/renderer/components/textbox.js

What do you all think?

app/renderer/components/sectionTitle.js
NejcZdovc: @luixxiul please update this path

NejcZdovc:
@luixxiul pleased check tests, because we have some failing tests https://travis-ci.org/brave/browser-laptop/builds/224193912

luixxiul:
I think I moved the file to `common` folder but the commit was dropped accidentally.. I'll fix it soon.

luixxiul:
Updated, and tests should be fine based on the results of my forked repo: https://travis-ci.org/luixxiul/browser-laptop/builds/224445488.

If ok, please add `ready-to-merge` and let me squash the first two commits, thanks. the last one is the fix for a commonForm element which I found on `about:styles`.

NejcZdovc:
@luixxiul still some test are failing, looks like some selectors problem

luixxiul:
damn I'm looking for them

luixxiul:
It turned out the most errors are intermittent and not related with those PRs:

https://travis-ci.org/brave/browser-laptop/jobs/224455046#L2935
https://travis-ci.org/brave/browser-laptop/jobs/224455049#L2288
https://travis-ci.org/brave/browser-laptop/jobs/224455051#L3034

PR #8238, opened 2017-04-11
move about page logic to browser process

bridiver:

Test Plan:


bsclifton:
@bridiver should this be in 0.14.2?

bridiver:
it doesn't have to be, but I just submitted a fix for an error in prefs page

PR #8256, opened 2017-04-12
remove unnecessary `frames` property from `closeWindow`

bridiver:
cc @bbondy

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -77,7 +77,6 @@ const tabUtil = require('../../app/renderer/lib/tabUtil')
 class Main extends ImmutableComponent {
   constructor () {
     super()
-    this.onCloseFrame = this.onCloseFrame.bind(this)
     this.onMouseDown = this.onMouseDown.bind(this)
     this.onClickWindow = this.onClickWindow.bind(this)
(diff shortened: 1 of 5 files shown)


js/components/main.js
```diff
@@ -1002,11 +997,9 @@ class Main extends ImmutableComponent {
             sortedFrames.map((frame) =>
               <Frame
                 ref={(node) => { this.frames[frame.get('key')] = node }}
-                tabData={this.props.appState.get('tabs').find((tab) => tab.get('tabId') === frame.get('tabId'))}
```
bridiver: this was just a random prop that wasn't being used anywhwere

bbondy: strange, I bet it was at some point and not anymore.

bbondy:
++ pls merge as long as tests pass

NejcZdovc:
@bbondy test are failing, but they are failing for master as well, so it's not related to this change

js/components/main.js
bsclifton: @bridiver I don't remember the original use-case, but I did use this when implementing the per-tab alerts (but then removed in favor of a distinct `messageBoxDetail` property)

PR #8258, opened 2017-04-12
change excluded extension state to 'not installed'

### LINKED ISSUE #8233: change extensions "delete" flow to keep ext in the list and reset to "not installed"
current: when using the trash can button (if available) the ext is removed from the browser and also from the list. Unfortunately there is no way, at this time, to reinstall the ext once removed because we don't have an extension store.

image

fix: instead of removing it from the list, just reset to pre-installed state where the slider invokes installation. 
image

cezaraugusto:
- Auditors: @bsclifton
- Close #8233

Test Plan:

1. Delete an extension
2. Extension should not disappear
3. Instead of trash icon you should see "not installed"

### DIFF
--- a/app/renderer/components/preferences/extensionsTab.js
+++ b/app/renderer/components/preferences/extensionsTab.js
@@ -34,6 +34,18 @@ class ExtensionsTab extends ImmutableComponent {
     aboutActions.extensionUninstalled(extensionId)
   }
 
+  onToggleExtension (extension, e) {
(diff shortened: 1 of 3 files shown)


bsclifton:
Moving to 0.14.3 😄   As discussed, there is an outstanding issue where enabling (after excluding/removing) takes a while to reload the extension. Even restarting the browser doesn't always work

Issue #8255, opened 2017-04-12
Source Maps in Releases


jonathansampson:
- Did you search for similar issues before submitting this one?
Yes

- Describe the issue you encountered:
Releases of Brave have developer resources that aren't useful to users, and take up a great amount of space. These should be removed, and only generated when the user is developing Brave locally.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Desktop

- Brave Version (revision SHA):
https://github.com/brave/browser-laptop/commit/f07c1421b982ef66e4cd369e8d928a3f9b19ffd5

- Steps to reproduce:
    1. Navigate to `%appdata%/../local/brave`
    2. Search all nested directories for `*.map`
    3. Note more than 10MB of results
    
- Actual result:
Brave ships development resources to end-users

- Expected result:
Brave is lean, and ships only that which is needed

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
Yes

- Is this an issue in the currently released version?
Yes

- Can this issue be consistently reproduced?
Yes

- Screenshot if needed:

![image](https://cloud.githubusercontent.com/assets/815158/24968101/c123e854-1f71-11e7-9582-a6551237b20e.png)

bsclifton:
Moving to backlog.  This is a good change but needs some more research before we pull it in

PR #8262, opened 2017-04-12
Prevent webpack from generating source maps in production builds

bsclifton:
## Test Plan:
?? - I need help verifying how to test this. I tried `npm run build-package` but didn't see file in either case.

## Description

Fixes https://github.com/brave/browser-laptop/issues/8255

Auditors: @jonathansampson, @bbondy, @bridiver 

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -96,6 +96,7 @@ function development () {  // eslint-disable-line
 
 function production () {  // eslint-disable-line
   var prod = config()
+  delete prod.devtool
   prod.plugins.push(new webpack.optimize.DedupePlugin())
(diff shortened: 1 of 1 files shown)


bridiver:
I'm pretty sure this is not the right change to make. I don't know what the default is, but the docs say `Note: With the next major version the default for -d will change to cheap-module-eval-source-map`
If we don't want the source maps in the release we should leave them out when building the asar file

bridiver:
I'd also like to see what we get in the console without any source maps at all. We don't want even less debugging info than we already have

bsclifton:
@bridiver those are good points- I'm going to go ahead and close this PR out and we can re-assess the issue

PR #8284, opened 2017-04-13
Locks versions of npm packages

### LINKED ISSUE #8283: Lock npm versions
Lock all npm packages to the specific version, so that we are all working on the same dependencies. 

In this issue I am **NOT** updating anything to the mayor version, just settings values that we get now if we run `npm i`

We should update all dependencies with every release.
cc @bbondy @alexwykoff @bsclifton

NejcZdovc:

Resolves #8283

Auditors: @bsclifton

Test Plan:
- test passing


bbondy:
I think we'd have to use npm shrinkwrap or get yarn to work.  This will only lock the top level versions, but each of those packages have fuzzy deps.

NejcZdovc:
Closing in favor of #7795

Issue #8271, opened 2017-04-13
unable to open a partitioned/private link - tab props always false


cezaraugusto:
## Test plan
1. Go to https://slashdot.org/
2. Right click any link and pick (from the context menu) `Open Link in New Private Tab`
3. Confirm tab that opens is private
4. Close private tab, go back to slashdot.org
5. Right click a link and pick (from the context menu) `Open Link in New Session Tab`
6. Go back to slashdot.org and repeat step 5
7. You should now have two session tabs opened, each with a different ID

## Original issue description
- Did you search for similar issues before submitting this one?
yes
- Describe the issue you encountered:
I can't go to a private/partitioned tab by clicking on a link 
- Platform (Win7, 8, 10? macOS? Linux distro?):
any
- Brave Version (revision SHA):
adf4d9a  
- Steps to reproduce:
  1. go to a page
  2. open a link in private/partitioned mode
  3. no icon/always false for props
    
- Actual result:
- If you log `this.props.tab.get('isPrivate')` or `this.props.tab.get('partitionNumber')` on tabs you will always see false/0 respectivelly. Based on that, tab is not private nor partitioned (no icon indicators).

- Expected result:
Should return true for privateTabs and return the partitionNumber for newSession tabs. Indicators for each tab should follow pattern: sunglass for private/person + number for partitioned tabs.

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
yes

- Is this an issue in the currently released version?
Yes, reproducible on v0.14.2dev-Preview3

- Can this issue be consistently reproduced?
yes

cezaraugusto:
worth mentioning: opening a new partitioned/private tab using menu/long press newtab bar and shortcuts works ok. Issue is regarding tabs.

PR #8286, opened 2017-04-13
Fix issue where tabs opened via context menu not opening in a private tab

bsclifton:
## Test Plan
1. Go to https://slashdot.org/
2. Right click any link and pick (from the context menu) `Open Link in New Private Tab`
3. Confirm tab that opens is private
4. Close private tab, go back to slashdot.org
5. Right click a link and pick (from the context menu) `Open Link in New Session Tab`
6. Go back to slashdot.org and repeat step 5
7. You should now have two session tabs opened, each with a different ID

## Description
During app/browser/tab::create(), `isPrivate` is set to true. However, there is no partition set.
If this is missing, I believe the code is checking the opener (which is not correct in this case).

When handler `add-new-contents` fired, the `incognito` property was always set to false. Partition `default`
seems to match picking "New private tab" from file menu and hitting + button > New private tab in tabs bar

After review by @bbondy, this also makes sure to assign partition if isPartitioned/partitionNumber are set (same conditions as checked in getPartition)

Fixes https://github.com/brave/browser-laptop/issues/8271

Auditors: @bbondy, @bridiver

### DIFF
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -85,6 +85,14 @@ const getPartition = (createProperties) => {
   return partition
 }
 
+const needsPartitionAssigned = (createProperties) => {
+  return !createProperties.openerTabId ||
+    createProperties.isPrivate ||
+    createProperties.isPartitioned ||
+    createProperties.partitionNumber ||
+    createProperties.partition
+}
+
 // TODO(bridiver) - refactor this into an action
 ipcMain.on(messages.ABOUT_COMPONENT_INITIALIZED, (e) => {
   const tab = e.sender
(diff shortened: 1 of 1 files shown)


app/browser/tabs.js
```diff
@@ -536,6 +536,9 @@ const api = {
         createProperties.parent_partition = ''
       }
     }
+    if (createProperties.isPrivate && createProperties.partition === undefined) {
```
bbondy: Pls move this into `getPartition` which takes in createProperties and returns the partition.

app/browser/tabs.js
```diff
@@ -85,6 +85,13 @@ const getPartition = (createProperties) => {
   return partition
 }
 
+const needsPartitionAssigned = (createProperties) => {
+  return !createProperties.openerTabId ||
+     (createProperties.isPrivate && createProperties.partition === undefined) ||
+     (createProperties.isPartitioned && createProperties.partition === undefined) ||
+     (createProperties.partitionNumber && createProperties.partition === undefined)
```
bbondy: I think this should be:
```
!createProperties.openerTabId ||
createProperties.isPrivate ||
createProperties.isPartitioned ||
createProperties.partitionNumber ||
createProperties.partition

In particular the last one would also be needed in case someone is creating a tab inside a particular partition and we also want the parent partition to be filled correctly.

bsclifton: fixed!

bbondy:
++ thanks

PR #8260, opened 2017-04-12
Increase tabs bar height by 2px

jamesmudgett:
Fixes https://github.com/brave/browser-laptop/issues/8263

- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

PR #8296, opened 2017-04-13
Revert "Increase tabs bar height by 2px"

luixxiul:
Reverts brave/browser-laptop#8260


jamesmudgett:
Would be nice to leave the new decreased height on bookmarks bar but keep space above address bar, needs new variable.

luixxiul:
let's discuss about it on a new issue :-)

Issue #8323, opened 2017-04-14
fingerprinting brave via extension resources


diracdeltas:
## Test plan

https://github.com/brave/browser-laptop/pull/8325#issue-221868705

----

https://extensions.inrialpes.fr/

seems like it should be blockable

luixxiul:
CC @diracdeltas 

1. Enable LastPass
2. Open https://extensions.inrialpes.fr/ and run test
3. `lastpass-free-password-ma` is detected

If it is expected, please just reclose the issue.

diracdeltas:
@luixxiul that is expected, we don't prevent fingerprinting all extensions, only the Brave-maintained ones.

PR #8325, opened 2017-04-14
remove web_accessible_resources in bundled extensions

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/8323

Test Plan:
1. go to https://extensions.inrialpes.fr/ and test for installed extensions. it should report none.
2. go to https://www.qubes-os.org/downloads/ and download the torrent. it should redirect to webtorrent's page.
3. go to https://en.wikipedia.org/wiki/Magnet_URI_scheme and click any of the magnet links. it should redirect to webtorrent's page.

Test Plan:

### DIFF
--- a/app/browser/webtorrent.js
+++ b/app/browser/webtorrent.js
@@ -1,9 +1,10 @@
 const electron = require('electron')
 const ipc = electron.ipcMain
 const appUrlUtil = require('../../js/lib/appUrlUtil')
+const appActions = require('../../js/actions/appActions')
 const messages = require('../../js/constants/messages')
 const Filtering = require('../filtering')
-const url = require('url')
+const urlParse = require('url').parse
 
 // Set to see communication between WebTorrent and torrent viewer tabs
 const DEBUG_IPC = false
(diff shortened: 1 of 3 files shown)


feross:
Sorry for the delay! Just took a look at the code changes here and everything makes sense. I also confirmed that:

- [x] go to https://extensions.inrialpes.fr/ and test for installed extensions. it should report none.
- [x] go to https://www.qubes-os.org/downloads/ and download the torrent. it should redirect to webtorrent's page.
- [x] go to https://en.wikipedia.org/wiki/Magnet_URI_scheme and click any of the magnet links. it should redirect to webtorrent's page.

I also confirmed that all the torrents on the WebTorrent "Free Torrents" page also work: https://webtorrent.io/free-torrents

There's one issue I ran into: The change seems to break clicks on torrent files. Once a torrent is started, clicking on a video file within the torrent should play it. Instead, the starting page is shown again and the user has to restart the torrent. Once doing so, the file doesn't play. The file list is still shown.

This is due to the ix= parameter not getting processed correctly on `hashchange` after this change. Before, the redirect-based implementation meant we didn't need to listen to `hashchange` since a redirect always happened. Now, a file click actually just adds a hash to the URL and triggers the `hashchange` event.

I'll send a PR to this PR that fixes the issue. Stay tuned.

feross:
Just sent a PR to this PR: https://github.com/brave/browser-laptop/pull/8566 Once that is merged, this PR will update.

PR #8336, opened 2017-04-15
Remove ref dependency from Frame

### LINKED ISSUE #8322: Remove ref dependency from Frame
Because Frame will now be a stateless component, we need to remove dependency on refs from a Frame

NejcZdovc:

Resolves #8322

Test Plan:
- tests are green


app/browser/tabs.js
```diff
@@ -617,6 +618,41 @@ const api = {
       api.createTab(state, action)
     }
     return state
+  },
+
+  goBack: (state, action) => {
+    action = makeImmutable(action)
+    const tab = api.getWebContents(action.get('tabId'))
+    tab.goBack()
```
bbondy: please wrap this with   `if (tab && !tab.isDestroyed()) {` and all below as well.

app/common/state/historyState.js
```diff
@@ -0,0 +1,174 @@
...
+          appActions.createTabRequested({
+            url: 'about:history'
+          })
+          BrowserWindow.getFocusedWindow().webContents.send(messages.HIDE_CONTEXT_MENU)
+        }
+      })
+  }
+
+  if (BrowserWindow.getFocusedWindow()) {
+    BrowserWindow.getFocusedWindow().webContents.send(messages.SHOW_CONTEXT_MENU, {
```
bbondy: Our dispatcher now has the ability to send messages to a particular window.  So it's more preferable to do an app action with a window Id specified.  See for example `newWebContentsAdded`.

bridiver: actually *State.js files specifically should not use any of these apis. They should manipulate state only and avoid other dependencies to make unit testing easier. The logic you have here looks like it probably belongs in a reducer

app/renderer/components/navigation/navigator.js
```diff
@@ -66,7 +65,7 @@ class Navigator extends ImmutableComponent {
         })
       }
     } else {
-      navAction.call(this.activeFrame)
+      navAction.call(this, this.props.activeTab.get('tabId'))
```
bridiver: can use `this.props.tabId` to make the prop a primitive type?

NejcZdovc: we don't have this in yet. I didn't yet implement redux here, so we have what we have for now, but when I will add redux to it, we will have primitives.

NejcZdovc: This PR only removes refs from Frame.

bridiver: ok, no problem

app/browser/tabs.js
NejcZdovc: done

app/common/state/historyState.js
```diff
@@ -0,0 +1,178 @@
...
+  if (tabId > -1 && history && history.entries.length > 0) {
+    const stopIndex = Math.max(((history.currentIndex - config.navigationBar.maxHistorySites) - 1), -1)
+    for (let index = (history.currentIndex - 1); index > stopIndex; index--) {
+      const url = history.entries[index].url
+
+      menuTemplate.push({
+        label: history.entries[index].display,
+        icon: history.entries[index].icon,
+        click: function (e) {
+          if (eventUtil.isForSecondaryAction(e)) {
```
bridiver: you'll have to pass the result of `eventUtil.isForSecondaryAction(e)` to the action because the dom event is not accessible here

app/common/state/historyState.js
```diff
@@ -0,0 +1,178 @@
...
+            appActions.createTabRequested({
```
bridiver: this is also going to be a problem because you can't serialize the method call in the windowAction

js/actions/windowActions.js
```diff
@@ -638,15 +638,15 @@ const windowActions = {
    * @param {Object} frameToSkip - Properties of the frame to keep audio
    */
   muteAllAudioExcept: function (frameToSkip) {
-    let framePropsList = windowStore.getState().get('frames')
+    /* let framePropsList = windowStore.getState().get('frames')
```
bbondy: if this isn't needed anymore can it just be removed?

NejcZdovc: this is still necessary, but I need to refactor it, because if you require `windowStore` in appAction (you are requiring it, because I am are calling windowActions in appStore)  you are getting errors. I will move this logic from windowActions into reducer, so that we don't need store here.

app/browser/tabs.js
```diff
@@ -619,6 +623,87 @@ const api = {
...
+    action = makeImmutable(action)
+    const tab = api.getWebContents(action.get('tabId'))
+    if (tab && !tab.isDestroyed()) {
+      tab.goToIndex(action.get('index'))
+    }
+    return state
+  },
+
+  getWindowId: () => {
+    if (BrowserWindow.getFocusedWindow()) {
```
bridiver: I think this should probably be a `getActiveWindowId` method on windows.js and return `windowState.WINDOW_ID_NONE` instead of -1 just for consistency

app/browser/tabs.js
```diff
@@ -619,6 +623,87 @@ const api = {
...
+        const title = tab.getTitleAtIndex(index)
+
+        let entry = {
+          index: index,
+          url: url,
+          display: title || url,
+          icon: null
+        }
+
+        if (url.startsWith('chrome-extension://')) {
```
bridiver: this should use the appUrlUtil about page methods

app/common/state/contextMenuState.js
```diff
@@ -0,0 +1,159 @@
...
+}
+
+const validateState = function (state) {
+  state = makeImmutable(state)
+  assert.ok(isMap(state), 'state must be an Immutable.Map')
+  return state
+}
+
+const contextMenuState = {
+  setContextMenu: (state, action) => {
```
bridiver: the helpers are meant to be more general purpose so this should probably take `detail` as a parameter instead of `action`

app/common/state/contextMenuState.js
```diff
@@ -0,0 +1,159 @@
...
+      if (!(action.getIn(['detail', 'type']) === 'hamburgerMenu' && state.get('hamburgerMenuWasOpen'))) {
+        state = state.set('contextMenuDetail', action.get('detail'))
+      }
+      state = state.set('hamburgerMenuWasOpen', false)
+    }
+
+    return state
+  },
+
+  onLongBacHistory: (state, action) => {
```
bridiver: I think these two methods probably make more sense in the reducer. The state helpers are mostly just getters/setters for the state. I'm not adamantly opposed to leaving them here, but if they stay the params should be tabId, partitionNumber and history instead of action

js/actions/appActions.js
```diff
@@ -1198,6 +1198,67 @@ const appActions = {
       actionType: appConstants.APP_DRAGGED_OVER,
       draggedOverData
     })
+  },
+
+  /**
+   * Go back in a history for a given tab
+   * @param {number} tabId - Tab id used for an action
+   */
+  onNavigateBack: function (tabId) {
```
bridiver: just a nit, but it would be more consistent to use onGoBack, onGoForward, etc...

app/common/state/contextMenuState.js
bridiver: BacHistory -> BackHistory

app/browser/tabs.js
NejcZdovc: Moved

app/browser/tabs.js
NejcZdovc: not sure which function should be used

app/common/state/contextMenuState.js
NejcZdovc: Done

app/common/state/contextMenuState.js
NejcZdovc: Done

app/browser/tabs.js
bridiver: `isTargetAboutUrl`

NejcZdovc: Done

app/browser/tabs.js
```diff
@@ -619,6 +623,79 @@ const api = {
...
+        } else {
+          if (sites) {
+            const site = sites.find(function (element) { return element.get('location') === url })
+            if (site) {
+              entry.icon = site.get('favicon')
+            }
+          }
+
+          if (!entry.icon) {
+            entry.icon = getDefaultFaviconUrl(url)
```
bridiver: does this work correctly? It looks like we normally get it using `window` which isn't available here

js/lib/urlutil.js
```diff
@@ -353,7 +353,7 @@ const UrlUtil = {
    * @return {string} url The base favicon URL
    */
   getDefaultFaviconUrl: function (url) {
-    if (UrlUtil.isURL(url)) {
+    if (typeof window !== 'undefined' && UrlUtil.isURL(url)) {
```
bridiver: I think we can fix this by using `urlParse` instead of `window.URL`

app/browser/tabs.js
NejcZdovc: you are correct we are not receiving anything here, that's why I added if typeof undefined

js/lib/urlutil.js
NejcZdovc: Done

app/browser/tabs.js
NejcZdovc: Code updated

bsclifton:
++ nice cleanup! 😄

Issue #5438, opened 2016-11-06
Autofill form option blinks and cannot be clicked


tvance929:
**Did you search for similar issues before submitting this one?**
Yes

**Describe the issue you encountered:**
When logging into a web form for the first time, the autofill blinks and cannot be chosen

**Expected behavior:**
Autofill dropdown option does not blink and can be chosen if the user wants to use.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Windows 10

- Brave Version:
0.12.7

- Steps to reproduce:
    1.  Go to Yahoo or Particle.io
    2. Click login and begin filling in username
    3. Autofill options blinks and cannot be chosen

- Screenshot if needed:

- Any related issues:

alexwykoff:
+1 from me, this is painful to observe as a user

bradleyrichter:
I have witnessed this one...

Liunkae:
+1 from #5265 (with GIF example)

Liunkae:
+1 from #5856.

darkdh:
I see the problem now.
STR:
1. You must save username/password by using built-in password manager first
2. Type username other than the username/password in step 1 to make autofill remember
3. Start typing the username in step 2 
4. Autofill and password manager popup will conflict and blink

Liunkae:
+1 from #5144

r3wt:
Similar issue in opensuse. They appear and are selectable, however the arrow keys do not work to select from the choices. should i open a new issue or?

Liunkae:
@r3wt I think you are referring to #1302 (Autofill: can't select popup item using arrow keys), which has already been fixed and will be released with 0.13.1.

bsclifton:
@darkdh was this fixed with:
- [Hide autofill popup only when webcontent is focused #170](https://github.com/brave/muon/pull/170)
- [Follow up of https://github.com/brave/muon/pull/170](https://github.com/brave/browser-laptop/pull/7777)

darkdh:
No, it is planned in my upcoming work: autofill password manager inegration
and btw the fix you mentioned also requires https://github.com/brave/browser-laptop/commit/396cdbcaebc480e488cd3af915a2bb21552b1c2b

PR #8369, opened 2017-04-18
Autofill password manager integration

### LINKED ISSUE #3530: integrate password manager autofill with browser autofill

darkdh:
requires https://github.com/brave/muon/pull/181

fix #3530 #5438 

Test Plan:


app/browser/reducers/passwordManagerReducer.js
```diff
@@ -86,178 +87,145 @@ const init = () => {
...
+  }
+  return state
 }
 
 const passwordManagerReducer = (state, action) => {
   action = makeImmutable(action)
   switch (action.get('actionType')) {
     case appConstants.APP_SET_STATE:
       init()
+      state = migrate(state)
```
diracdeltas: should this only be called if we detect migration hasn't already happened?

app/browser/reducers/passwordManagerReducer.js
```diff
@@ -86,178 +87,145 @@ const init = () => {
...
+const migrate = (state) => {
+  masterKey = masterKey || getMasterKey()
+  if (!masterKey) {
+    console.log('Could not access master password; aborting')
+    return
+  }
+  const passwords = state.get('passwords')
+  if (passwords.size) {
+    passwords.forEach((password) => {
+      let decrypted = CryptoUtil.decryptVerify(password.get('encryptedPassword'),
```
diracdeltas: this should probably abort if decrypted is null

app/browser/reducers/passwordManagerReducer.js
```diff
@@ -86,178 +87,145 @@ const init = () => {
...
+                                               password.get('authTag'),
+                                               masterKey,
+                                               password.get('iv'))
+      let form = {}
+      form['origin'] = password.get('origin')
+      form['signon_realm'] = password.get('origin') + '/'
+      form['action'] = password.get('action')
+      form['username'] = password.get('username')
+      form['password'] = decrypted
```
diracdeltas: Where are passwords stored on disk? I assume Chromium encrypts them using a keychain password somehow

app/browser/reducers/passwordManagerReducer.js
```diff
@@ -86,178 +87,145 @@ const init = () => {
...
-        origin,
-        action,
-        username: username || '',
-        encryptedPassword: encrypted.content,
-        authTag: encrypted.tag,
-        iv: encrypted.iv
-      })
-    }
-  })
+    state = state.set('passwords', new Immutable.List())
```
diracdeltas: it might be a good idea to copy the old passwords somewhere instead of deleting them forever. if the migration doesn't work for some reason, users will be mad if they can't get their passwords back.

js/about/passwords.js
```diff
@@ -63,7 +63,7 @@ class SiteItem extends React.Component {
   }
 
   onDelete () {
-    aboutActions.deletePasswordSite(this.props.site)
+    aboutActions.deletePassword(this.props.site.toJS())
```
diracdeltas: This is supposed to undo the 'never save password' site setting, not delete the password itself

darkdh: sorry for the confusion, new `never save password` list will migrate and save in chromium as the password entry with `blacklisted_by_user = true` attribute.

app/browser/reducers/passwordManagerReducer.js
darkdh: It will be saved in our user dir as `Login Data` sqlite format encrypted by OS keychain.
So for example if you are on Mac, Brave will prompt to ask keychain permission.

app/browser/reducers/passwordManagerReducer.js
darkdh: sure will do

js/about/passwords.js
diracdeltas: it seems that `onDelete` here should then remove the `blacklisted_by_user` attribute, not delete the password

app/browser/reducers/passwordManagerReducer.js
darkdh: good call, will make a copy.

app/browser/reducers/passwordManagerReducer.js
darkdh: If it has already migrated, the original password will be empty and it won't do a second migrate.
(Re. previous comment, will also make a copy of password)

darkdh:
@diracdeltas , I've made some changes according to your feedback in 015e9b31ae8f86825100782279d73a29f3279c28

app/browser/reducers/passwordManagerReducer.js
bridiver: @darkdh maybe the migrate method should be called in sessionStore where we do all the other data migrations?

darkdh: @bridiver I tried to do it in sessionStore at the migration section but our migration requires `electron.session`. And at that time app is not ready so we will have `Error: Session can only be received when app is ready`

app/browser/reducers/passwordManagerReducer.js
```diff
@@ -189,14 +189,17 @@ const migrate = (state) => {
...
+        let form = {}
+        form['origin'] = password.get('origin')
+        form['signon_realm'] = password.get('origin') + '/'
+        form['action'] = password.get('action')
+        form['username'] = password.get('username')
+        form['password'] = decrypted
+        autofill.addLogin(form)
+      }
     })
+    state = state.set('legacy_passwords', state.get('passwords'))
```
diracdeltas: Thanks!
1. This field should be documented in docs/state.md. 
2. should be `legacyPasswords` instead of legacy_passwords
3. i think you can call this once in https://github.com/brave/browser-laptop/pull/8369/commits/015e9b31ae8f86825100782279d73a29f3279c28#diff-6547e1a0bf9650c63b5500045f5ecee5R185 instead of once per password

darkdh: Will do. 
These two state set lines are under `if (passwords.size) {...}` which use to check non-migrated data and I think inside of `passwords.forEach((password) => {...})` is called once per password

diracdeltas: my bad, i didn't notice it was outside of the forEach block

cezaraugusto:
since not included in the last freeze and needing to rebase, moving to 0.15.4. pls change if needed

diracdeltas:
i copied my session-store-1 file from my real profile to my dev profile to test this out. after npm start, i see a bunch of errors in the console:
```
Passwords sqlite error 2067, errno 0: UNIQUE constrai
nt failed: logins.origin_url, logins.username_element, logins.username_value, logins.password_element, logins.si
gnon_realm, sql: INSERT INTO logins (origin_url, action_url, username_element, username_value, password_element,
 password_value, submit_element, signon_realm, preferred, date_created, blacklisted_by_user, scheme, password_ty
pe, possible_usernames, times_used, form_data, date_synced, display_name, icon_url, federation_url, skip_zero_cl
ick, generation_upload_status) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)
```

bridiver:
looks like trying to insert duplicates based on the error

darkdh:
@yan, did you override `session-store-1` at least twice without deleting `Login Data` under user dir?

diracdeltas:
other notes:

* when i delete a password or site setting in about:passwords, it takes a few seconds to disappear from the list (not a blocker)
* when i try to login on a site that already has a login, it prompts me to choose the password for the other login and there is no way to dismiss that prompt except clicking outside the login box. i think the prompt should disappear as soon as the user starts typing.
screen shot 2017-05-12 at 00 17 29
* when i login on https://trac.torproject.org/projects/tor/login, there is no 'remember this password?' message box
* when starting Brave, it shows two prompts to allow keychain access. i guess maybe this is just for the migration?
screen shot 2017-05-12 at 00 09 49
screen shot 2017-05-12 at 00 09 59

bsclifton:
@darkdh I'm having a hard time reproducing the original issue described in https://github.com/brave/browser-laptop/issues/5438.  Do you have a site which has this issue which you used for testing?

darkdh:
@bsclifton  try this:
https://github.com/brave/browser-laptop/issues/5438#issuecomment-266935703

diracdeltas:
> did you override session-store-1 at least twice without deleting Login Data under user dir?

Maybe? In my production profile, I had to override session-store-1 manually once after it was deleted by a Brave update. In my dev-profile, I just copied session-store-1 from production and didn't touch Login Data.

darkdh:
>Maybe? In my production profile, I had to override session-store-1 manually once after it was deleted by a Brave update. In my dev-profile, I just copied session-store-1 from production and didn't touch Login Data.

The errors popup is because we previously migrated passwords into Login Data. So we have manually delete it when testing or we will migrate passwords repeatedly.

darkdh:
> when i delete a password or site setting in about:passwords, it takes a few seconds to disappear from the list. i think it might not be triggering a renderer update.

It took seconds for `GetAutofillableLogins` and `GetBlacklistLogins` to finish. Then callback will be called and the state in about:passwords will get updated and trigger rendering. cc @bridiver

darkdh:
@bsclifton , updated

screen shot 2017-05-12 at 2 46 04 pm

bbondy:
Talked to Anthony, for the remaining issue of delayed showing when deleting, he's going to post a follow up and we're going to merge this now.

app/locale.js
```diff
@@ -215,6 +215,7 @@ var rendererIdentifiers = function () {
     'updateLater',
     'updateHello',
     'notificationPasswordWithUserName',
+    'notificationUpdatePasswordWithUserName',
```
luixxiul: @darkdh we need `notificationUpdatePassword` along with that. Unless:

clipboard01

PR #8436, opened 2017-04-22
Perform drag and drop from the browser end

bbondy:
There was a race condition where drop info was not determined on dragEnd. This gets rid of it by moving the handling in the browser process like it should be anyway.

https://github.com/brave/browser-laptop/issues/8288


bbondy:
@bridiver pls review but I'm merging for now so we can get builds.

js/dnd.js
```diff
@@ -47,9 +29,25 @@ module.exports.onDragStart = (dragType, data, e) => {
   appActions.dragStarted(currentWindowId, dragType, data)
 }
 
+document.addEventListener('keyup', (e) => {
+  if (e.keyCode === 27) {
```
bsclifton: Can this use the constant for ESC? see `app/common/constants/keyCodes.js`

luixxiul: @bbondy ^

bbondy: yep sorry I meant to get back to that before submitting the PR, will do!

PR #8476, opened 2017-04-24
remove isSwipeTrackingFromScrollEventsEnabled

bridiver:
remote.systemPreferences is no longer available and muon now only fires scroll-touch-begin and scroll-touch-end if the system pref is enabled

Test Plan:
Turn on "swipe between pages"
Swipe back/forward should work
Turn off "swipe between pages"
Swipe back/forward should stop working

**** Requires muon 2.56.6 ****
https://github.com/brave/muon/commit/3ee6abd8bbc34f6b092d08bdbd9dcb916d53d616

### DIFF
--- a/js/components/main.js
+++ b/js/components/main.js
@@ -7,7 +7,6 @@ const ImmutableComponent = require('../../app/renderer/components/immutableCompo
 const Immutable = require('immutable')
 const electron = require('electron')
 const ipc = electron.ipcRenderer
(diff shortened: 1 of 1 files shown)


bsclifton:
@bridiver is there an issue for this? Asking since QA uses issues to track 😄

bridiver:
This falls under the general umbrella of removing remote API calls

PR #8520, opened 2017-04-27
Fix stuck test "Syncing then turning it off"

ayumi:


ayumi:
Merging as is change to tests only, and it looks like it helped.

PR #8527, opened 2017-04-27
add help wiki link in Help menu of toolbar Fix #6940

### LINKED ISSUE #6940: Add documentation links to Help menu
From Community:
https://community.brave.com/t/why-help-not-available-from-the-help-menu/954

Presumably Brave now has a large enough set of stable features for work to start on some structured documentation for new users beyond what is available through browsing the Community.

To start with, the Help menu should probably link to key github wikis:
* https://github.com/brave/browser-laptop/wiki
* https://github.com/brave/browser-ios/wiki
* https://github.com/brave/browser-android-tabs/wiki

kumarrishav:

Resolves #6940

Test Plan:
screen shot 2017-04-27 at 5 12 53 pm


kumarrishav:
CC : @bsclifton for review

app/extensions/brave/locales/en-US/menu.properties
```diff
@@ -89,6 +89,9 @@ bookmarksManager=Bookmarks…
 importBrowserData=Import Browser Data…
 exportBookmarks=Export Bookmarks…
 submitFeedback=Submit Feedback…
+braveHelpLaptop=Brave Help Laptop…
+braveHelpIOS=Brave Help IOS…
```
Liunkae: @kumarrishav Could you please change this to `braveHelpIOS=Brave Help iOS…` (correct iOS capitalization)?

kumarrishav:
@Liunkae added the change. 
screen shot 2017-04-28 at 11 25 46 am

kumarrishav:
@Liunkae review?

luixxiul:
I don't think it does not make sense very much to display the link to iOS app on the laptop app for example.

Liunkae:
What if we just keep the laptop help link and rename it `Brave Help…`.

kumarrishav:
I think we should give wiki link to all three platform as they might search
for it from browser. Also, hardly anyone checks help section on
phone(android or iOS).
Or another solution can be a link to page having wiki links to all these
three platform (laptop, ios, android).
We should have links for all three.

Later one looks good to me though.

kumarrishav:
@luixxiul @Liunkae  How should we go on this? 
@bsclifton  Can you give your inputs to this :)

bsclifton:
I think we should ask Alex and Brad (I don't feel confident making those kinds of calls 😄 )
cc: @alexwykoff, @bradleyrichter

alexwykoff:
I think it's a good first effort. What I would do differently is to explore a bit about the problems the user is trying to solve rather than how did they describe the problem. 

More specifically, if a user feels that they need help, how do they want to go about receiving that help and balance that versus what is possible in the organization.  

For laptop, there is significantly more computing power and tools available, so a design is already underway for an about:help page which would have some tools to force a check for the latest update, and diagnose basic networking problems.

If a user has an iOS device or Android @kumarrishav inferred that they do not check help on the device, and maybe that is true, but one should ask why. Is it because prior designs were insufficient?

To circle back to the original problem, should we be adding 3 links to point to each wiki on Github? I would say that we should not do this. Many users will not understand what is Github and why is the information there.  Linking to the community would be more appropriate. Even so, I do not think we would want 3 links as we should assume users are intelligent and can find their niche. 

I hope this helps 😸  I'll copy this answer back to the original issue as well.

bsclifton:
Closing per @alexwykoff's comment; I'll update the original issue with some more notes. Let's talk more there 😄

PR #8532, opened 2017-04-27
Split scroll between titles and folders for BM manager

### LINKED ISSUE #5076: Folders and Titles lists in about:bookmarks should scroll independently
I have several dozen bookmark folders and have several dozen boomarks in each folder. Re-organising bookmarks is impractical and clumsy without being able to scroll the two seperately as is the case in Chromium (and other browsers).

Ubuntu 16.10
Brave: 0.12.5
Electron: 1.4.14
libchromiumcontent: 53.0.2785.143
V8: 5.3.332.47
Node.js: 6.5.0
Update Channel: dev

cezaraugusto:

- Auditors: @bsclifton
- Close #5076

Test Plan:
- Have a lot of bookmarks
- Go to about:bookmarks
- Ensure that:
  - Layout is fixed
  - You can scroll between folders (left-hand side column) and titles (right-hand side column) wouldn't scroll together
  - Less-than-screen-height number of bookmakrs shouldn't let you scroll (layout is fixed)
  - At small sizes we fallback to what we have and you can scroll gracefully


jonathansampson:
@cezaraugusto The headers don't behave the same way. Above the folders, headers don't scroll. Above bookmark items, headers *do* scroll.

bookmarks-headers

cezaraugusto:
@jonathansampson yep that's the ideal but I can't since it uses <table> (`SortableTable` component) which doesn't accept overflow. 

Ideally `sortableTable` should make use of css-grid to keep it's table-like look, but changes would be too bulk and imo deserves a standalone refactor.

++ for ellipsis. I'll do another round on this since Brad don't want custom sidebars and include that, thanks!

jonathansampson:
@cezaraugusto All valid points. I gladly approve, provided we can create an issue to refactor the bookmarks items portion for frozen headers.

bradleyrichter:
This is great to have, even with the unresolved header.

cezaraugusto:
ok based on https://github.com/brave/browser-laptop/pull/8532#issuecomment-298105891 I created an issue to track fixed header for bm items: #8607. Also added ellipsis for long texts.

I also removed completely `ScrollableContent` element since component was created to address #7930 and this issue, and is no longer required.

PR #8556, opened 2017-04-28
fix syncing error when lastAccessedTime is empty

### LINKED ISSUE #8552: Crash on startup (possibly related to recovering tabs)
- Did you search for similar issues before submitting this one? Yes.

- Describe the issue you encountered: Upon launching the browser, it appears briefly, attemtping to re-load previously tabs and then immediately closes.

If ran through the terminal, I receive the following error.

```
jordan@azurite:~$ brave
Crash reporting enabled
ledger balance warning: {
  "message": "Blockonomics: undefined"
}
[8923:8923:0428/154813.756756:ERROR:CONSOLE(18783)] "(node) warning: possible EventEmitter memory leak detected. %d listeners added. Use emitter.setMaxListeners() to increase limit.", source: chrome://brave/usr/share/brave/resources/app.asar/app/extensions/brave/gen/app.entry.js (18783)
Warning: Can't create site data: {"order":1207,"count":15,"partitionNumber":0,"location":"https://parsecpizzadelivery.com/","title":"Parsec Pizza Delivery trailer - A pixelated deliver 'em up","tags":[],"objectId":[0,63,197,156,48,17,112,109,247,175,79,57,151,123,29,198],"themeColor":"rgb(5, 5, 5)"}
Error: Can't create JS from existingObject! [object Object]
    at Object.module.exports.getExistingObject (/usr/share/brave/resources/app.asar/js/state/syncUtil.js:259:11)
    at records.map (/usr/share/brave/resources/app.asar/app/sync.js:379:39)
    at Array.map (native)
    at EventEmitter.ipcMain.on (/usr/share/brave/resources/app.asar/app/sync.js:375:47)
    at emitMany (events.js:127:13)
    at EventEmitter.emit (events.js:204:7)
    at WebContents.<anonymous> (/usr/share/brave/resources/electron.asar/browser/api/web-contents.js:182:13)
    at emitTwo (events.js:106:13)
    at WebContents.emit (events.js:194:7)
/usr/share/brave/resources/app.asar/node_modules/node-anonize2-relic-emscripten/anonize2-jumbo.js:67
            throw ex
            ^

Error: Can't create JS from existingObject! [object Object]
    at Object.module.exports.getExistingObject (/usr/share/brave/resources/app.asar/js/state/syncUtil.js:259:11)
    at records.map (/usr/share/brave/resources/app.asar/app/sync.js:379:39)
    at Array.map (native)
    at EventEmitter.ipcMain.on (/usr/share/brave/resources/app.asar/app/sync.js:375:47)
    at emitMany (events.js:127:13)
    at EventEmitter.emit (events.js:204:7)
    at WebContents.<anonymous> (/usr/share/brave/resources/electron.asar/browser/api/web-contents.js:182:13)
    at emitTwo (events.js:106:13)
    at WebContents.emit (events.js:194:7)
```

- Platform (Win7, 8, 10? macOS? Linux distro?): Linux - Ubuntu 16.04

- Brave Version (revision SHA): v0.15.0 (71fb730)

- Steps to reproduce: Unfortunately, I'm not certain on what precisely causes the browser to get into the state required to trigger this issue.
    
- Actual result: Browser crashes almost immediately after startup.

- Expected result: Browser does not crash and correctly recovers previously used tabs.

No. Renaming the `/home/username/.config/brave/session-store-1` file and starting the browser works fine, excluding the loss of your previous tabs & some browser settings.

- Is this an issue in the currently released version? Yes.

- Can this issue be consistently reproduced? No. As mentioned, I am uncertain what precisely causes the browser to get into the state required to trigger this issue. It is not related to a particular website unfortunately.

- Any related issues: This appears to occur shortly after a seemingly random crash of the browser, but I can not confirm if this is always the case.

diracdeltas:
fix #8552 

Test Plan: unittests and syncing tests should pass

Test Plan:

### DIFF
--- a/js/state/siteUtil.js
+++ b/js/state/siteUtil.js
@@ -600,14 +600,13 @@ module.exports.isImportedBookmark = function (siteDetail) {
  */
 module.exports.isHistoryEntry = function (siteDetail) {
   if (siteDetail && typeof siteDetail.get('location') === 'string') {
-    if (siteDetail.get('location').startsWith('about:')) {
+    const tags = siteDetail.get('tags')
+    if (siteDetail.get('location').startsWith('about:') ||
+      module.exports.isDefaultEntry(siteDetail) ||
+      isBookmarkFolder(tags)) {
       return false
     }
-    if (module.exports.isDefaultEntry(siteDetail)) {
-      // This is a Brave default newtab site
-      return false
-    }
-    return !!siteDetail.get('lastAccessedTime') && !isBookmarkFolder(siteDetail.get('tags'))
+    return !!siteDetail.get('lastAccessedTime') || !tags || tags.size === 0
   }
   return false
 }
--- a/js/state/syncUtil.js
+++ b/js/state/syncUtil.js
@@ -79,7 +79,7 @@ module.exports.getSiteDataFromRecord = (record, appState, records) => {
   let existingObjectData
 
   if (record.action !== writeActions.CREATE) {
-    const existingObject = module.exports.getObjectById(objectId, category,
+    const existingObject = getObjectById(objectId, category,
       appState)
     existingObjectData = existingObject && existingObject[1]
   }
@@ -237,7 +237,7 @@ module.exports.getExistingObject = (categoryName, syncRecord) => {
   const AppStore = require('../stores/appStore')
   const appState = AppStore.getState()
   const objectId = new Immutable.List(syncRecord.objectId)
-  const existingObject = module.exports.getObjectById(objectId, categoryName, appState)
+  const existingObject = getObjectById(objectId, categoryName, appState)
   if (!existingObject) { return null }
 
   const existingObjectKeyPath = existingObject[0]
@@ -256,7 +256,8 @@ module.exports.getExistingObject = (categoryName, syncRecord) => {
       throw new Error(`Invalid category: ${categoryName}`)
   }
   if (!item) {
-    throw new Error(`Can't create JS from existingObject! ${existingObjectData}`)
+    console.log(`Warning: Can't create JS from existingObject! ${JSON.stringify(existingObjectData)}`)
+    return null
   }
   return {
     action: writeActions.CREATE,
@@ -314,7 +315,7 @@ module.exports.updateSiteCache = (appState, siteDetail) => {
  * @param {Immutable.Map=} appState
  * @returns {Array} [<Array>, <Immutable.Map>] array is AppStore searchKeyPath e.g. ['sites', 10] for use with updateIn
  */
-module.exports.getObjectById = (objectId, category, appState) => {
+const getObjectById = (objectId, category, appState) => {
   if (!(objectId instanceof Immutable.List)) {
     throw new Error('objectId must be an Immutable.List')
   }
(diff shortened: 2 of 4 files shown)


js/state/siteUtil.js
```diff
@@ -607,7 +607,7 @@ module.exports.isHistoryEntry = function (siteDetail) {
       // This is a Brave default newtab site
       return false
     }
-    return !!siteDetail.get('lastAccessedTime') && !isBookmarkFolder(siteDetail.get('tags'))
+    return !isBookmarkFolder(siteDetail.get('tags'))
```
ayumi: I think this change causes all bookmarks to show up as History items.
It might be that https://github.com/brave/browser-laptop/blob/master/app/common/lib/historyUtil.js#L19 is now returning every site not a bookmark folder.

Repro:
1. Start pyramid 0, enable Sync and visit archive.org and wikipedia.org
2. Start pyramid 1, join pyramid 0's sync group
3. Examine pyramid 1's history – it has the 2 sites.

Expected: Pyramid 1's history should be empty because history doesn't sync by default.

diracdeltas: oops, good catch

diracdeltas: Hm, actually I can't repro this. Did you start from fresh pyramids?

ayumi: i did

diracdeltas:
@ayumi i may have fixed it

js/state/siteUtil.js
```diff
@@ -600,14 +600,13 @@ module.exports.isImportedBookmark = function (siteDetail) {
...
+      module.exports.isDefaultEntry(siteDetail) ||
+      isBookmarkFolder(tags)) {
       return false
     }
-    if (module.exports.isDefaultEntry(siteDetail)) {
-      // This is a Brave default newtab site
-      return false
-    }
-    return !!siteDetail.get('lastAccessedTime') && !isBookmarkFolder(siteDetail.get('tags'))
+    return !!siteDetail.get('lastAccessedTime') || !tags || tags.size === 0
```
ayumi: @diracdeltas hmm that fixed half of it

try syncing to the `forsaken lindy ...` group. i see a lot of history items.

should this line be
```js
return !!siteDetail.get('lastAccessedTime') && (!tags || tags.size === 0)
```

diracdeltas: @ayumi yeah, that is more restrictive so i'll change it to `&&`

diracdeltas: @ayumi actually the crash was caused by a history entry which did not have `lastAccessedTime`, so changing it to `!!siteDetail.get('lastAccessedTime') && (!tags || tags.size === 0)` would cause it to crash again. 

when i sync to `forsaken...`, the synced bookmarks appear in `about:history`. this seems like the correct behavior to me because every bookmark is by definition also a history entry.

ayumi: i don't think there'd be a crash any more due to the change to downgrade the `throw` into `console.log`.

On bookmarks == history, I think that's not obvious to a user; so if I synced my bookmarks I'd expect history to be empty. It would only make sense if I understood the sites data model. However, this is how it currently happens in `master` so this is probably ok.

diracdeltas: @ayumi right, it would fix the crash, but I think the user's data was a valid history entry and should not be ignored by sync.

i think "make synced bookmarks not part of history" is a separate issue

PR #8558, opened 2017-04-28
Adds List validation for reduxComponent

### LINKED ISSUE #8557: List in redux component
We sometimes need to pass in Immutable List into redux component props, because we need to display a list of X items (for example tabs). For that reason we need to do a List check in `ReduxComponent`.

NejcZdovc:

Resolves #8557

Auditors: @bridiver

Test Plan:
- create redux component
- pass prop as a simple List
- component should not update if props is the same

### DIFF
--- a/app/common/state/immutableUtil.js
+++ b/app/common/state/immutableUtil.js
@@ -17,6 +17,14 @@ const api = {
     return Immutable.List.isList(obj)
   },
 
+  isSameHashCode: (first, second) => {
+    if (first === null && second === null) {
+      return true
+    }
+
+    return second !== null ? first.hashCode() === second.hashCode() : false
+  },
+
   makeImmutable: (obj) => {
     return api.isImmutable(obj) ? obj : Immutable.fromJS(obj)
   }
--- a/app/renderer/components/reduxComponent.js
+++ b/app/renderer/components/reduxComponent.js
@@ -3,6 +3,7 @@ const ImmutableComponent = require('./immutableComponent')
 const React = require('react')
 const windowStore = require('../../../js/stores/windowStore')
 const debounce = require('../../../js/lib/debounce')
+const {isList, isSameHashCode} = require('../../common/state/immutableUtil')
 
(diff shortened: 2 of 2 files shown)


app/common/state/immutableUtil.js
```diff
@@ -17,6 +17,10 @@ const api = {
     return Immutable.List.isList(obj)
   },
 
+  isSame: (first, second) => {
```
bridiver: to avoid confusion I think should probably is `isSameHashCode`

app/common/state/immutableUtil.js
```diff
@@ -17,6 +17,10 @@ const api = {
...
+    return second !== null ? first.hashCode() === second.hashCode() : false
```
bridiver: if both objects are null it should return true

app/renderer/components/reduxComponent.js
```diff
@@ -45,8 +46,16 @@ class ReduxComponent extends ImmutableComponent {
   }
 
   shouldComponentUpdate (nextProps, nextState) {
-    return Object.keys(nextState).some((prop) => nextState[prop] !== this.state[prop]) ||
-      Object.keys(nextProps).some((prop) => nextProps[prop] !== this.props[prop])
+    return Object.keys(nextState).some((prop) => {
+      return isList(nextState[prop])
```
bridiver: since this is a little more complicated now it would probably be better to pass an instance method to both instead of using anonymous functions. It would be slightly faster anyway and avoid the overhead of creating and GCing the closure

app/common/state/immutableUtil.js
NejcZdovc: Done

app/common/state/immutableUtil.js
NejcZdovc: Done

app/renderer/components/reduxComponent.js
NejcZdovc: Done

PR #8566, opened 2017-04-29
Fix broken torrent file view (fix for PR #8325)

feross:
**This is a PR to a PR!** Merge this **BEFORE** merging PR #8325.

Once a torrent is started, clicking on a video file within the torrent
should play it. Instead, the starting page is shown again and the user
has to restart the torrent. Once doing so, the file doesn't play. The
file list is still shown.

This is due to the ix= parameter not getting processed correctly on
hashchange after this change. Before, the redirect-based implementation
meant we didn't need to listen to hashchange since a redirect always
happened. Now, a file click actually just adds a hash to the URL and
triggers the hashchange event.

### DIFF
--- a/js/webtorrent/entry.js
+++ b/js/webtorrent/entry.js
@@ -20,7 +20,7 @@ require('../../node_modules/font-awesome/css/font-awesome.css')
 const store = {
   name: null, // Torrent name
   ix: null, // Selected file index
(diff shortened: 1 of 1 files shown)


diracdeltas:
Test plan worked for me :+1:

PR #8593, opened 2017-05-01
use write.writeImportant for session store

### LINKED ISSUE #7876: When writing session to disk, ensure we do a flush
per discussion with @bridiver: 

We'll want to use `fs.fsync` when saving `session-store-1` to disk. We've had problems with this file being corrupted when a BSOD happens on Windows. The save that happens every 5 mins may not have been flushed.

Notes from the discussion:
- I think you have to use the low-level writeStream methods
- even `writeSync` doesn’t force a sync to disk
- the sequence should be create -> flush -> rename. The atomicity of rename on windows is questionable because of the method that node uses so we might want to consider patching it to use the documented atomic method
- node rename is definitely not atomic on windows and requires a second flush because it doesn’t use the `MOVEFILE_WRITE_THROUGH` flag
- create -> flush -> rename -> flush and then the file is guaranteed to be written to disk

bridiver:
fix #7876

Test Plan:
quit/restart behavior when updating

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -16,7 +16,6 @@ const downloadStates = require('../js/constants/downloadStates')
 const urlParse = require('./common/urlParse')
 const getSetting = require('../js/settings').getSetting
 const appUrlUtil = require('../js/lib/appUrlUtil')
-const promisify = require('../js/lib/promisify')
 const siteSettings = require('../js/state/siteSettings')
 const settings = require('../js/constants/settings')
 const userPrefs = require('../js/state/userPrefs')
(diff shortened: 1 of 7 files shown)


bbondy:
++

app/filtering.js
```diff
@@ -727,25 +726,21 @@ module.exports.isResourceEnabled = (resourceName, url, isPrivate) => {
...
 /**
  * Clears all session cache.
  * @return a promise that always resolves (called on app shutdon so must always)
  */
 module.exports.clearCache = () => {
-  let p = Promise.resolve()
   for (let partition in registeredSessions) {
     let ses = registeredSessions[partition]
-    p = p.then(promisify(ses.clearCache.bind(ses)).catch(() => {}))
+    setImmediate(ses.clearCache.bind(ses))
```
bsclifton: Since there are docs on the Muon side, I think it would be helpful to put a link to the docs as a comment; ex:
https://github.com/brave/muon/blob/master/docs/api/session.md#sesclearcachecallback

(and if needed, update the docs)

app/filtering.js
```diff
@@ -727,25 +726,21 @@ module.exports.isResourceEnabled = (resourceName, url, isPrivate) => {
...
     let ses = registeredSessions[partition]
-    p = p.then(promisify(ses.clearStorageData.bind(ses)).catch(() => {}))
+    setImmediate(ses.clearStorageData.bind(ses))
   }
-  return p
 }
 
 /**
  * Clears all session cache.
  * @return a promise that always resolves (called on app shutdon so must always)
```
bridiver: yep, that's true for several of these now

PR #8605, opened 2017-05-01
use muon.file.writeImportant to write ledger files (#8602)

### LINKED ISSUE #8602: use muon.file.writeImportant to write ledger files
- Did you search for similar issues before submitting this one? yes

- Describe the issue you encountered: replace `atomicWriter` in `ledger.js` with `muon.file.writeImportant`

- Platform (Win7, 8, 10? macOS? Linux distro?): all

- Brave Version (revision SHA): all

- Steps to reproduce:
    1. start browser
    2. visit enough sites for enough time to cause `ledger-synopsis.json` to be written
    
- Actual result: files are written without corruption

- Expected result: same

- Will the steps above reproduce in a fresh profile? no. If not what other info can be added? visit some sites to `ledger-synopsis.json` gets written

- Is this an issue in the currently released version? yes

- Can this issue be consistently reproduced? yes

- Extra QA steps: none

- Screenshot if needed: none

- Any related issues: based on https://github.com/brave/browser-laptop/pull/8593

mrose17:

Test Plan:  no new tests. existing ledger tests should apply.

Fixes #8602

### DIFF
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -31,7 +31,6 @@
 contributeP: (stickyP OR !excluded) AND eligibleP   AND !blockedP
  */
 
-const crypto = require('crypto')
 const fs = require('fs')
 const os = require('os')
 const path = require('path')
@@ -274,8 +273,6 @@ var init = () => {
   try {
     appDispatcher.register(doAction)
     initialize(getSetting(settings.PAYMENTS_ENABLED))
-
-    doneTimer = setInterval(doneWriter, 1 * msecs.minute)
   } catch (ex) { console.log('ledger.js initialization failed: ' + ex.toString() + '\n' + ex.stack) }
 }
 
@@ -283,7 +280,6 @@ var quit = () => {
   quitP = true
   visit('NOOP', underscore.now(), null)
   clearInterval(doneTimer)
-  doneWriter()
 
   if ((!getSetting(settings.PAYMENTS_ENABLED)) && (getSetting(settings.SHUTDOWN_CLEAR_HISTORY))) reset(true)
 }
@@ -723,7 +719,7 @@ var initialize = (paymentsEnabled) => {
               }
               getStateInfo(stateResult)
 
-              atomicWriter(pathName(statePath), stateResult, { flushP: true }, () => {})
+              muonWriter(pathName(statePath), stateResult)
             })
           }
         } catch (ex) {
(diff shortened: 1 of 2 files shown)


app/ledger.js
```diff
@@ -2093,6 +2093,19 @@ var networkConnected = underscore.debounce(() => {
  * low-level utilities
  */
 
+var muonWriter = (path, payload) => {
+  muon.file.writeImportant(path, JSON.stringify(payload, null, 2), (success) => {
+    if (!success) return console.log('write error: ' + path)
+
+    if ((quitP) && (!getSetting(settings.PAYMENTS_ENABLED)) && (getSetting(settings.SHUTDOWN_CLEAR_HISTORY))) {
```
bsclifton: this is kind of weird; so just to confirm... when Brave is exiting (user chose to quit) and payments aren't enabled / history is set to clear, we destroy the file by setting quitP to true and then logging something?  I mention it's weird because the deletion only happens after the file is written to (versus just deleting the file).

app/ledger.js
```diff
@@ -2093,6 +2093,19 @@ var networkConnected = underscore.debounce(() => {
...
+      if (ledgerInfo._internal.debugP) console.log('\ndeleting ' + path)
+      return fs.unlink(path, (err) => { if (err) console.log('unlink error: ' + err.toString()) })
+    }
+
+    if (ledgerInfo._internal.debugP) console.log('\nrenaming ' + path)
```
bsclifton: Is this a stale comment? renaming?

app/ledger.js
mrose17: yes, it's weird: this deals with a race condition that @NejcZdovc found during testing.

it's done that way because the file may be in the process of being written when we want to delete it. in which case, if we unlink the file, we might just then go ahead and write it again. hence, always delete in this case...

app/ledger.js
mrose17: oops! fixed.

app/ledger.js
bsclifton: @mrose17 ah ok- makes sense 😄

mrose17: i'm not sure whether it makes sense. it is certainly counter-intuitive: if you have successfully written the file, and the user is quitting, and the ledger isn't enabled, and history is to be deleted, then get rid of the synopsis file... reminiscent of https://www.youtube.com/watch?v=GoxcyV4DJuY

mrose17:
no more atomicWriter ... https://github.com/brave/browser-laptop/commit/34d7eb94b20e50be5c84a5ad02920904eee81b0f ... ok, i'm done, i hope!

PR #8320, opened 2017-04-14
Add new context menu to entire tab bar (including Reopen last closed tab)

philkloose:
- [X] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

This is my attempt to resolve https://github.com/brave/browser-laptop/issues/8303.
Note -- this is really my first pull request beyond a README grammar change, so it's quite possible I overlooked something.

Test Plan:
Open a new tab, visit a site (i.e. https://google.com), close the tab and then right-click in the space to the right of any open tabs.  You should be presented with a new context menu with an option to reopen the last closed tab.

PR #8623, opened 2017-05-02
Fixes #8540 - Added "Mute All", Fixed "Mute Other Tabs", made "(Un)Mu…

### LINKED ISSUE #8540: Tab context menus should offer "Mute all tabs"
Originally proposed in https://github.com/brave/browser-laptop/pull/8320, right clicking (or cmd + clicking) inside the tab area should bring up a context menu.

This should *include* "Mute All Tabs":
image

This should always mute all of the tabs.  It was attempted in the PR, but @philkloose noticed that the quick implementation I did toggled the mute state (instead of muting all).  When all are muted, we may consider changing the menu to "Unmute all"

cc: @bradleyrichter for more details 😄

philkloose:
Fixes #8540 - Added "Mute All", Fixed "Mute Other Tabs", made "(Un)Mute Tab" always present, changed Mute icon display behavior.

Test Plan:

This PR contains several things, many of which weren't discussed in #8540.  I probably shoved more stuff into this than I should have and I'm happy to have any and all ideas shot down -- I'm mainly doing this to familiarize myself with javascript and github.

1. I added a "Mute All" menu item to the tabsToolbar context menu and verified that it strictly mutes instead of toggling.  I did not implement a "master mute layer" proposed here https://github.com/brave/browser-laptop/pull/8320#issuecomment-297859212 for a few reasons.  One -- I didn't know how :).  Two - I really couldn't come up with a use case for it and it would have been hard to convey the effect of the button to the user succinctly.
2. The "Mute Other Tabs" tab menu item also suffered from the same issue of toggling, so I fixed that too.
3. The "Mute/Unmute Tab" menu item was only conditionally showing when the tab was actively playing audio.  This doesn't make sense to me since I want to be able to mute a tab _before_ I begin playing a video.  This behavior is identical to how Chrome works currently.
4. If a tab is muted, the icon should show regardless of whether or not it is currently playing audio.  I made this change.  Again, this mimics Chrome's behavior.

Lastly, I would suggest like to suggest that the mute icon be made less ambiguous.
image vs image seems obvious until you realize that if the first one isn't present as a comparison that the second one simply looks like a speaker.  The user doesn't know that it's missing the sound waves emanating from the speaker.  I would propose an icon with the speaker slashed out.

### DIFF
--- a/app/renderer/components/tabs/content/audioTabIcon.js
+++ b/app/renderer/components/tabs/content/audioTabIcon.js
@@ -14,7 +14,7 @@ const globalStyles = require('../../styles/global')
 const tabStyles = require('../../styles/tab')
 
 class AudioTabIcon extends ImmutableComponent {
-  get pageCanPlayAudio () {
+  get pageAudioActive () {
     return !!this.props.tab.get('audioPlaybackActive')
   }
 
@@ -24,7 +24,7 @@ class AudioTabIcon extends ImmutableComponent {
   }
 
   get mutedState () {
-    return this.pageCanPlayAudio && !!this.props.tab.get('audioMuted')
+    return !!this.props.tab.get('audioMuted')
   }
 
   get audioIcon () {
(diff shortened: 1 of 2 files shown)


philkloose:
I just installed a Standard JavaScript formatting extension after this commit, so sorry for all the inconsistencies.  I'll be sure to check/resolve them pre-commit going forward.

bsclifton:
@philkloose since your fix for https://github.com/brave/browser-laptop/issues/8303 was reverted, I'm going to close this PR for the moment.  #8303 has notes with some next steps. If you're doing cherry picking, maybe you can include this fix in there too? 😄 

Let me know if you need help

PR #8631, opened 2017-05-02
Fix loading tab regression

### LINKED ISSUE #8550: No progress indicator when reloading page
yes

It's a User Experience issue:  I don't see any indication that a page is loading, while it's loading.  I'm sure there's supposed to be a progress indication someplace, but it's either not working or escaping my attention.

Mac OS X 10.11.6

71fb730

- Steps to reproduce:
    1. Open browser (I've been using it for a number of days, and usually just close the laptop's lid as opposed to shutting down)
    2. Open some website.  
    3. Hit Cmd+r.  Multiple times fi you want.  
    
- Actual result:
I NEVER get a progress indication anywhere.

- Expected result:
Visible indication that the browser has received and understood my keystroke, and that it's now performing some progress - i.e. I can wait for it to complete as opposed to repeating my keystroke.

I'm not sure I know what this question means.  I've only been using Brave for a week or so, but don't remember this being an issue before - if that helps.

Yes, I updated recently (yesterday?  I can't remember exactly)

On _my_ machine... every time.

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues:

cezaraugusto:
Auditors: @bbondy, @bsclifton
Fix #8550

Test Plan: 
loading icon should be visible while reloading a tab

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -324,22 +324,24 @@ const doAction = (action) => {
       break
     case windowConstants.WINDOW_WEBVIEW_LOAD_START:
       {
-        const framePath = ['frames', frameStateUtil.getFramePropsIndex(frameStateUtil.getFrames(windowState), action.frameProps)]
+        const statePath = path =>
+          [path, frameStateUtil.getFramePropsIndex(frameStateUtil.getFrames(windowState), action.frameProps)]
+
         // Reset security state
         windowState =
-          windowState.deleteIn(framePath.concat(['security', 'blockedRunInsecureContent']))
-        windowState = windowState.mergeIn(framePath.concat(['security']), {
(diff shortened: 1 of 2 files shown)


bsclifton:
It might not be straightforward, but I'd love to see tests around this if you are able to @cezaraugusto 😄 

We'd need to mock `AppDispatcher`, from which we can get the reference to `doAction` when register is called.  Then we can call passing in `windowConstants.WINDOW_WEBVIEW_LOAD_START` and verify the result modified the window state appropriately

cezaraugusto:
ok rebased + amended some more tests, ready for re-review

bbondy:
++, took me a minute to see it, but was resulting in `['tabs', 'frames', ...]`

bsclifton:
++! 😄

Issue #2714, opened 2016-07-25
pdfjs needs local file access


diracdeltas:
## Test plan

https://github.com/brave/browser-laptop/pull/8645#issue-225861735

----

the pdfjs extension doesn't have the 'Access File URLs' permission, so it can't browsr `file:///` pdfs. cc @bridiver

willy-b:
hey @diracdeltas, I think you might already have fixed this with #6330.
If not, would you mind pointing me to any resources on how PDFJS is integrated in Brave now? It used to be in `browser-laptop/app/extensions/pdfjs` but has now moved.  Also, is Brave using your fork?

Thanks in advance, I'm looking at this for PR #6385 where I try to open a generated PDF in a new tab

diracdeltas:
@willy-b it's not fixed yet sadly.

pdfjs is now loaded from a remote server. it gets downloaded in ~/Library/Application Support/brave-development/Extensions/jdbefljfgobbmcidnmpjamcbhnbphjnb/ so you can test changes by editing files in there directly

it's using my fork at https://github.com/diracdeltas/pdf.js (built from master using 'gulp chromium')

willy-b:
sweet, thanks for the info

willy-b:
Took a look at this.
Good news: this seems to be a XHR cross-origin issue. Bad news: changing extension permissions didn't seem to fix it (for me) (see next comment). 

I compared Google Chrome, where PDF.js can open local PDFs, with Brave.

You can run something like the following in both browsers to see difference  (from debugger during page load to ensure same context for good measure):
```
var xhr = new XMLHttpRequest()
xhr.open("GET", "file:///home/willy/Downloads/brave-downloads/test-pdf.pdf", true)
xhr.send()
```

The above works on Google Chrome (from within the PDF.js extension in process of rendering a local PDF)
and does NOT work in Brave (**error:** `Cross origin requests are only supported for protocol schemes: [...]`)
I followed in debugger and everything else is identical, right up to and including the XHR.


**notes / details**
- URL bar suggests an origin difference between browsers (but there isn't really)
In Google Chrome the URL bar uses the `chrome-extension:` proto, but Brave shows the `file://` extension.
Chrome: `chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/file:///home/willy/Downloads/brave-downloads/test-pdf.pdf`
vs. Brave: `file:///home/willy/Downloads/brave-downloads/test-pdf.pdf`)
[...]

PR #8645, opened 2017-05-03
allow PDFJS to load local files

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/2714

test plan:
1. start Brave with PDFJS enabled and browse to a local PDF file. it should display correctly. the urlbar should show the 'file:' location.
2. turn off PDFJS and go to the file. it should prompt to download.

Test Plan:

### DIFF
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -400,6 +400,9 @@ module.exports.init = () => {
   }
 
   let loadExtension = (extensionId, extensionPath, manifest = {}, manifestLocation = 'unpacked') => {
+    if (extensionId === config.PDFJSExtensionId) {
+      manifestLocation = 'component'
+    }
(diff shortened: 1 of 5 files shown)


app/extensions.js
```diff
@@ -400,6 +400,9 @@ module.exports.init = () => {
   }
 
   let loadExtension = (extensionId, extensionPath, manifest = {}, manifestLocation = 'unpacked') => {
+    if (extensionId === config.PDFJSExtensionId) {
+      manifestLocation = 'component'
```
diracdeltas: Note: 'component' type extensions are able to access cross-origin 'file:' URLs, but somehow they are not able to see 'file:' URLs in the chrome.webRequest listener.

bbondy:
It would be interesting to dig into chrome.webRequest more, but this is ok for now. Nice to have it working, thanks!

PR #8721, opened 2017-05-05
fallback to local data when extension is disabled

### LINKED ISSUE #8720: Fallback extension to local data always when extension is disabled
@bbondy reported an issue regarding LastPass where the image is broken and description is showing as `__MSG__`:

screenshot 2017-05-05 08 00 33

Issue happens because once enabled for the first time, extension data is got from the manifest, and if you disable chrome-extension URL is no longer valid, and data is lost. 

To fix that we should fallback to local data always when the extension is disabled.

cezaraugusto:
Auditors: @bbondy
Fix #8720

Test Plan:
- Disable an extension
- Check that "version" is set as "latest" (because it's dummy data).
- Enable extension
- Check that "version" match extension version. (because data is fetched from manifest)


bbondy:
++

PR #8813, opened 2017-05-10
sync device list sort by ID order is wrong.

### LINKED ISSUE #8311: sync device list sort by ID order is wrong
if you sort by ID, it sorts lexicographically instead of numerically:
screen shot 2017-04-13 at 8 31 54 pm

kumarrishav:
Submitter Checklist:

Fix #8311
Test Plan:

screen shot 2017-05-10 at 11 28 38 pm
Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/preferences/syncTab.js
+++ b/app/renderer/components/preferences/syncTab.js
@@ -137,7 +137,7 @@ class SyncTab extends ImmutableComponent {
     return devices.map((device, id) => [
       {
         html: id,
-        value: parseInt(id)
+        value: device.get('lastRecordTimestamp')
       },
       {
         html: device.get('name'),


kumarrishav:
@ayumi @diracdeltas This is the best fix i could found for this i.e using timestamp for sorting for ID as both are incrementing accordingly. As per my analysis, this is happening because of a bug in tablesort module (https://github.com/tristen/tablesort/blob/gh-pages/src/tablesort.js#L27). 
It take ID as string and sorts accordingly. So, if number of digits in all IDs are equal then sorting order will be correct. 
Another solution can be : prefix 0's to IDs to make equal number of digits (manipulating the ID by appending zeroes which i didn't find good way)

app/renderer/components/preferences/syncTab.js
```diff
@@ -137,7 +137,7 @@ class SyncTab extends ImmutableComponent {
     return devices.map((device, id) => [
       {
         html: id,
-        value: parseInt(id)
+        value: device.get('lastRecordTimestamp')
```
ayumi: i don't think is correct because if other devices add records out of order, the id list will be even more jumbled.

diracdeltas: i think this could often be out-of-order with the ID. it's probably better to just pad the ID or change `value` to a string of 0's repeating `id` times.

kumarrishav: Yeah, this will be doable.

kumarrishav:
@ayumi To get first row, it should access like this : this.props.rows._root.nodes[0].entry[1]
instead of [this.props.rows[0]](https://github.com/brave/browser-laptop/blob/master/js/components/sortableTable.js#L434) . this.props.rows[0] returns undefined that's why it takes `string` as sort-method (default method). 

Another workaround can be : pass the sort-method (number/string) as props. 
or  @diracdeltas said just pad the ID or change value to a string of 0's repeating id times.

How we should go on this?

NejcZdovc:
I think that we should fix it so that the correct format is passed in. So if the first one in undefined, we should replace it with 0 for example, or query it correctly if this is a problem

kumarrishav:
@NejcZdovc this.props.rows._root.nodes[0].entry[1] gives expected result. should i use that instead?

NejcZdovc:
I didn't wrote this code, so it would be better if @ayumi or @diracdeltas answer this one

ayumi:
I had a WIP PR where I tried to solve this problem by changing the rows array:
https://github.com/brave/browser-laptop/commit/8d8b6e440cc800e468189bef0d70f36b722090f5

It doesn't fix SortableTable but solves the current problem.
It should probably have tests.

kumarrishav:
okay. should i close this PR then? @ayumi

ayumi:
@kumarrishav thanks for your understanding. I've updated #8311 to reference the WIP patch.

Issue #5761, opened 2016-11-21
URL selection on the URL bar conflicts with the URL autocomplete based on history


luixxiul:
**Describe the issue you encountered:** URL selection on the URL bar conflicts with the URL autocomplete based on history.

1 Open `https://github.com/brave/browser-laptop/issues/4852#issuecomment-261872105` in a new tab
2 Delete `#issuecomment-261872105` on the URL bar
3 Hit <kbd>Command</kbd> + a to select `https://github.com/brave/browser-laptop/issues/4852`

Then `#issuecomment-261872105` is filled automatically and selected

![url](https://cloud.githubusercontent.com/assets/3362943/20475587/fac2cbf6-b00f-11e6-8470-648454f3a94c.gif)

**Expected behavior:** `https://github.com/brave/browser-laptop/issues/4852` should be selected 

- Platform (Win7, 8, 10? macOS? Linux distro?): macOS

- Brave Version: master branch

- Any related issues:

patrick99e99:
+1, this happens to me all the time but using keyboard shortcuts:  cmd-l to select url bar, cmd-right-arrow to go to the end of the url, option-left-arrow to move back by path a few times, ctrl-k to delete from where the cursor is at...  result is new text immediately appears where I tried to delete from.

luixxiul:
This seems to have been fixed. Closing.

luixxiul:
Adding the most recent milestone to forward the issue to the QA work.

Issue #5338, opened 2016-11-02
about:newtabs - have relevant results shown (implement frecency)


bsclifton:
cc: @bbondy @bradleyrichter 

With some favorite sites up in the "Top Sites" area, the items are not easy to identify.
![screen shot 2016-11-01 at 5 19 07 pm](https://cloud.githubusercontent.com/assets/4733304/19912411/849dd032-a057-11e6-83ce-d8b684b9523f.png)

We need to find a better way to represent bookmarks which have the same favicon:
- having a bunch of medium.com posts pinned or in favs, which will all have the same icon
- bookmarking multiple parts of brianbondy.com :smile: 
- tweets

luixxiul:
Eventually you will have this one..

<img width="556" alt="screenshot 2016-11-02 11 41 43" src="https://cloud.githubusercontent.com/assets/3362943/19914846/629c6b8e-a0f1-11e6-87f6-f91632949b13.png">

Is this a card game? :-D

bsclifton:
@srirambv brought up a great concern here (similar to @luixxiul's concern) https://github.com/brave/browser-laptop/issues/5347

Multiple sites on the same origin will each take up a tile.  Is this behavior that we want?

bradleyrichter:
We need to dedupe. The example above would simply make GitHub the top site and you would have a link to github.com on a single tile.

bsclifton:
Requirements (please edit this post as needed):
- Only show one link per domain
- Tile can remain as the favicon with the theme color (although, folks can turn off theme color in Preferences > Tabs)
- Mousing over needs to show the title (which it does currently)

bsclifton:
Also- if we wanted nicer high-res non-favicons, we could consider spriting up like a top 20 sheet

bsclifton:
Moving to 0.12.10.  For 0.12.9, the ideal behavior would be to limit folks to only 1 row of icons (until we're able to resolve this issue and the other issues)

cc: @cezaraugusto

Liunkae:
Related issue: #5482 (only one row of topSites)

bbondy:
Moving to 0.12.10 but would consider moving back if a PR is in place.

bbondy:
Please update to use the same algorithm with frecency as is done for urlbar suggestions.

bsclifton:
Note per our convo: we can use the frecency algorithm from suggestions.js, but we'd need to move the code to the renderer process

bradleyrichter:
[...]

PR #8824, opened 2017-05-10
Bloodhound for URL bar suggestions & move suggestions to browser process

### LINKED ISSUE #5878: URL bar selects autocomplete result on keypress
I only did a quick search, there may already be an issue (help appreciated finding)

Auto-complete results are being chosen when I'm doing a very specific set of key presses when focus is in the URL bar (see steps to repro).

I found this when trying my default logic for copy/paste, which is stupid now that I'm typing it out... but it did uncover this.  If you instead use cmd + A to select the contents of the URL bar (instead of what is outlined in step 3), it does work and you can cmd + C it just fine (which is more efficient anyways)

**Expected behavior:**
Auto-complete result should not be picked unless I use up/down arrow

macOS

- Brave Version:
0.12.10

- Steps to reproduce:
    1. Visit https://www.oediscountparts.com/partlocator/index.cfm?siteid=215242
    2. Click inside the URL bar to give it focus and delete the path portion of the URL
      - Hit end (windows) or cmd + right arrow (mac) to move cursor to end
      - hold shift and the modifier key (either ctrl on windows or option on mac) and press left arrow key repeatedly, until you've selected `/partlocator/index.cfm?siteid=215242`
      - Press delete or backspace to remove that part
    3. Try to select the rest of whats left
      - hold shift + modifier key (cmd / ctrl) and then press left arrow key
      - notice that the first result is auto-selected (which then overrides the URL you had in the box)

- Screenshot if needed:

- Any related issues:

### LINKED ISSUE #7453: Performance : urlbar response is slow when searching across many bookmarks
Yes

After importing a large number of bookmarks, the urlbar response becomes considerably slow. It was found that bookmarks suggestions was the controlling factor.

Tested on OS X

0.13.5 (https://github.com/brave/browser-laptop/commit/0e683420964546ac26a8c2903ef871a35a6e71d8)

- Steps to reproduce:
    1. With a clean session, try to visit reddit.com, add a couple bookmarks
    2. Import a large number of bookmarks
    3. With a new tab, try to visit reddit.com
    
- Actual result:
The urlbar becomes considerably laggy.

- Expected result:
The urlbar should be at least as responsive as if the bookmarks were not there. That is, a user should not be penalized for trying to visit a domain root when they have bookmarks to deeper links.

Yes

Yes

Yes

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

[2017-03-02 09-52-15.mp4.zip](https://github.com/brave/browser-laptop/files/814365/2017-03-02.09-52-15.mp4.zip)

- Any related issues:

bbondy:
This PR does several things:
- Makes suggestions code independent of window renderer
- Moves URL bar suggestion creating into the browser process
- Creates a wrapper lib for bloodhound autocomplete results for bookmarks and history
- Uses that wrapper lib instead of using indexOf for faster autocomplete results.
- Moves several windowActions around url bar suggestions into app actions and stops naming them set*
- Dispatcher with queryInfo of windowId specified will now always route to the browser process, before it would skip that if the window renderer was routing an app action to itself.  The browser process in that case will not re-dispatch to the window process.
- Various other perf and refactoring improvements.

Fix #7453
Fix #5878
Fix https://github.com/brave/browser-laptop/issues/8919

Submitter Checklist:

- [x] Better to not rebase in this case, changes are very cleanly defined in various small refactors leading up to using bloodhound for URL suggestions in the browser process.

Test Plan:
- Test around search options in preferences to make sure they work like they used to
- Import lots of history and bookmarks and make sure it's still snappy to get suggestions
- Make sure google.com shows up first when typing google.com
- Make sure things work at least as good as before in terms of getting results you wanted when typing.
- **QA team:** please do a full pass around URL bar suggestions, try to find things that don't work as expected but used to work before.  Please post follow ups for found things within the context of the same .300 milestone.  Please avoid posting things that have always worked that way in previous versions though.

Reviewer Checklist:

Tests

- [x] Tests should be independent and work correctly when run indivually or as a suite [ref](https://github.com/brave/browser-laptop/wiki/Code-Guidelines#test-dependencies)


js/actions/appActions.js
```diff
@@ -1337,6 +1337,21 @@ const appActions = {
       hostPattern: origin,
       key: 'savePasswords'
     })
+  },
+
+  /**
+   * Indicates that the urlbar text has changed, usually from user input
+   *
+   * @param {string} location - The text to set as the new navbar URL input
```
bsclifton: docs need updating to include the windowId / input fields

bsclifton: derp- this is an individual commit, will go back to looking at the all files changed view 😛  Disregard

bsclifton:
I think this issue is fixed: https://github.com/brave/browser-laptop/issues/5761

bsclifton:
The weird behavior I noticed with TAB has been logged as a separate issue: https://github.com/brave/browser-laptop/issues/8919

app/renderer/components/navigation/urlBarSuggestions.js
```diff
@@ -36,12 +38,12 @@ class UrlBarSuggestions extends ImmutableComponent {
 
   render () {
     const suggestions = this.props.suggestionList
-    const bookmarkSuggestions = suggestions.filter((s) => s.type === suggestionTypes.BOOKMARK)
-    const historySuggestions = suggestions.filter((s) => s.type === suggestionTypes.HISTORY)
-    const aboutPagesSuggestions = suggestions.filter((s) => s.type === suggestionTypes.ABOUT_PAGES)
-    const tabSuggestions = suggestions.filter((s) => s.type === suggestionTypes.TAB)
-    const searchSuggestions = suggestions.filter((s) => s.type === suggestionTypes.SEARCH)
-    const topSiteSuggestions = suggestions.filter((s) => s.type === suggestionTypes.TOP_SITE)
+    const bookmarkSuggestions = suggestions.filter((s) => s.get('type') === suggestionTypes.BOOKMARK)
```
bsclifton: Would it be better to just iterate `suggestionList` once (ex: w/ a forEach) and use a switch to push results onto empty array?

bsclifton:
Food for thought (while it's fresh 😄 ): these improvements could be used to show more relevant results (and also in a more efficient way) for the new tab dashboard.  For more info, see https://github.com/brave/browser-laptop/issues/5338

app/sessionStore.js
```diff
@@ -658,7 +658,8 @@ module.exports.defaultAppState = () => {
     httpsEverywhere: {
       count: 0
     },
-    defaultWindowParams: {}
+    defaultWindowParams: {},
+    searchDetail: null
```
bsclifton: don't forget to add to state.md 😄

docs/appActions.md
```diff
@@ -1023,6 +1023,18 @@ Notifies autoplay has been blocked
 
 
 
+### savePassword()
```
bsclifton: Would these be better named as `passwordSaved` / `passwordUpdated`? (to be more consistent with others that you introduced, like urlBarTextChanged)

bsclifton:
Other than the comments above, code and tests look great 😄  I'll keep poking at it for a bit and will post any findings 👍

app/renderer/components/navigation/urlBarSuggestions.js
bbondy: That would be better but not introduced in this commit so leaving as is for now. The right way is to store it separately in state so no filter is needed at all.

app/sessionStore.js
bbondy: done, good catch.

docs/appActions.md
bbondy: This is actually only showing up because I fixed javadoc comment block on some of anthony's actions.  It's not introduced here.

srirambv:
Bookmark items are not listed when `Show history matches` is enabled. Disabling it brings up the `Show bookmark matches` in the suggested items
bloodhound

srirambv:
Auto complete search term shows complete URL as suggested search term. Selecting doesn't load the site rather searches for the URL.
autosearch

PR #8827, opened 2017-05-10
setting the state in componentWillReceiveProps caused current/next co…

bridiver:
…mparisons to return false instead of true

Test Plan:
Automated tests were failing

Reviewer Checklist:

Tests


bsclifton:
this could definitely use some Enzyme tests; you can likely create a demo component and then check if expected result/behavior happens when executing life cycle methods

luixxiul:
Do we need to do QA on this?

PR #8841, opened 2017-05-11
[WIP] make use of mouse velocity for vertical tab preview

cezaraugusto:
Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

## This is a work in progress. However, look-and-feel is already testable if you wish.

Auditors: @bradleyrichter


bradleyrichter:
I found a few bugs to look at:

1. It seems easy to get it into a state where the tab is selected by hover and not mouse click.

2. Possibly related to #1, but the tab can enter a state where the tab text and title bar are correct but the DOM content are from a different tab

example:
image

3. My first tab would not show a preview for a while

4. leaving focus and returning caused tabs to be preview-able but not clickable.

cezaraugusto:
yes as we talked I think this is a different issue and more related to #8842.

cezaraugusto:
after taking this step for ~2 weeks I don't think this will ever work as-is so I'll close. I have a better solution.

PR #8844, opened 2017-05-11
Perf components

bridiver:
intentionally left this as 3 separate commits


app/renderer/components/reduxComponent.js
```diff
@@ -20,20 +20,20 @@ class ReduxComponent extends ImmutableComponent {
     super(props)
     this.componentType = props.componentType
     this.state = {}
```
NejcZdovc: I think that we don't need to set react state anymore

bridiver: it just gets set so it isn't null

Issue #8850, opened 2017-05-12
new tab always shows about:blank


diracdeltas:
- Did you search for similar issues before submitting this one?
yes

- Describe the issue you encountered:
opening a new tab (via cmd+t) shows 'about:blank' instead of blank in the URL bar. I have the 'always show url bar' preference set.

OS X, 0.15.303
![screen shot 2017-05-12 at 05 36 38](https://cloud.githubusercontent.com/assets/549654/25983587/488d6840-36d5-11e7-9735-ddb9f4e59523.png)

bsclifton:
@diracdeltas is this the same as https://github.com/brave/browser-laptop/issues/8525 and/or https://github.com/brave/browser-laptop/issues/8378?

diracdeltas:
I don't think it's the same as https://github.com/brave/browser-laptop/issues/8525 because it happens when i hit cmd+t regardless of whether i have an about: tab open. It may be the same as https://github.com/brave/browser-laptop/issues/8378 except it happens all the time, not sporadically.

alexwykoff:
Initial launch with a clean profile shows the appropriate message of 'Enter a URL or search term' in the urlbar. Subsequent openings of new tabs are as described above.

This would also explain why it's missing a favicon when I open a lot of tabs. 😸

PR #8880, opened 2017-05-16
fix 'about:blank' being displayed permanently on new tabs

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/8850

Test Plan:
1. open several new tabs. they should all show blank in the urlbar instead of about:blank after a short lag.
2. 'new tab button' and 'new tab signal' tests should pass.

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/navigation/urlBar.js
+++ b/app/renderer/components/navigation/urlBar.js
@@ -357,10 +357,13 @@ class UrlBar extends React.Component {
   componentDidUpdate (prevProps) {
     // this.urlInput is not initialized in titleMode
     if (this.urlInput) {
+      const pdfjsEnabled = getSetting(settings.PDFJS_ENABLED)
       if (this.props.activeFrameKey !== prevProps.activeFrameKey) {
         this.keyPressed = false
         // The user just changed tabs
-        this.setValue(this.props.locationValue)
+        this.setValue(this.props.locationValue !== 'about:blank'
+          ? this.props.locationValue
+          : UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))
         // Each tab has a focused state stored separately
         if (this.props.isFocused) {
           this.focus()
@@ -369,7 +372,7 @@ class UrlBar extends React.Component {
         windowActions.setRenderUrlBarSuggestions(false)
       } else if (this.props.location !== prevProps.location) {
         // This is a url nav change
-        this.setValue(UrlUtil.getDisplayLocation(this.props.location, getSetting(settings.PDFJS_ENABLED)))
+        this.setValue(UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))
       } else if (this.props.hasLocationValueSuffix &&
                 this.props.isActive &&
                 this.props.locationValueSuffix !== this.lastSuffix) {
--- a/test/tab-components/tabTest.js
+++ b/test/tab-components/tabTest.js
@@ -5,6 +5,8 @@ const messages = require('../../js/constants/messages')
 const settings = require('../../js/constants/settings')
 const {urlInput, backButton, forwardButton, activeTab, activeTabTitle, activeTabFavicon, newFrameButton, notificationBar, contextMenu, pinnedTabsTabs, tabsTabs} = require('../lib/selectors')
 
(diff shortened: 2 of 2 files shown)


bridiver:
this is a good workaround for now, but I want to capture the two underlying problems:
1. We update the urlbar location for the activeFrame only even though each frame has its own urlbar in the state. The urlbar should track location updates regardless of whether it is visible or not
2. We are updating the location too late on navigation. It should be updated when the provisional load succeeds, but we don't update it until the navigation completes

test/tab-components/tabTest.js
```diff
@@ -89,6 +91,18 @@ describe('tab tests', function () {
         .waitForExist('[data-test-id="tab"][data-frame-key="2"]')
         .waitForTextValue('[data-test-id="tab"][data-frame-key="2"]', 'New Tab')
     })
+
+    it('shows empty urlbar', function * () {
```
bridiver: both of these tests pass even when I remove the urlbar changes so I'm not sure if they really capture the issue

app/renderer/components/navigation/urlBar.js
```diff
@@ -357,10 +357,13 @@ class UrlBar extends React.Component {
   componentDidUpdate (prevProps) {
     // this.urlInput is not initialized in titleMode
     if (this.urlInput) {
+      const pdfjsEnabled = getSetting(settings.PDFJS_ENABLED)
       if (this.props.activeFrameKey !== prevProps.activeFrameKey) {
         this.keyPressed = false
         // The user just changed tabs
-        this.setValue(this.props.locationValue)
+        this.setValue(this.props.locationValue !== 'about:blank'
```
bridiver: going to merge this as a workaround for now, but we want to get away from putting this kind of logic in components. The same workaround could be done in a reducer to update the urlbar location on tab change

test/tab-components/tabTest.js
diracdeltas: they failed for me but maybe i was on a slightly older commit

app/renderer/components/navigation/urlBar.js
```diff
@@ -357,10 +357,13 @@ class UrlBar extends React.Component {
...
+          ? this.props.locationValue
+          : UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))
```
bbondy: I don't understand what this does.

Only in the case that`this.props.locationValue` is `about:blank` does it go to `UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))` and `UrlUtil.getDisplayLocation('about:blank', pdfjsEnabled))` actually returns about:blank.

bbondy: Since this doesn't do anything, I'm changing this to:
```
-        this.setValue(this.props.locationValue !== 'about:blank'
-          ? this.props.locationValue
-          : UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))
+        if (this.props.locationValue !== 'about:blank') {
+          this.setValue(UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))
+        }
```

which I think is the expected functionality.

bbondy: I think in the timing I'm experiencing the change does nothing, but on the timing you're experiencing it does something. Trying to think of the right thing to do here.

bbondy: This is also the line of code that causes typing fast after new tab to get clobbered by the way. Very noticeable if you have lots of bookmarks.

diracdeltas: @bbondy the issue was that when i opened a new tab, this.props.locationValue is about:blank (and wouldn't change unless i typed in the urlbar) but this.props.location is about:newtab. So previously, when i opened a new tab, it would show about:blank *forever*. this seemed to only happen to some users, as noted in the original issue.

It seems like your change would cause the issue to re-appear.

diracdeltas: my understanding was that we should always be showing `getDisplayLocation(this.props.location)` unless the user is modifying the urlbar. That would be the case here if the user switched to a tab where they had been modifying the urlbar, but not when opening a new tab.

PR #8894, opened 2017-05-16
Add locationSiteKeyCache and use for siteUtil.isBookmarked

### LINKED ISSUE #8703: URL bar, on every keystroke the browser re-checks if the current location is bookmarked
With 10,000 bookmarks, this adds 80ms to every keystroke.

Yes

- Steps to reproduce:
    1. Instrument `siteUtil.isSiteBookmarked()` by adding to the beginning `const startTime = new Date().getTime()` and to the end console.log(`isSiteBookmarked() ${new Date().getTime() - startTime}`).
    2. Start a new profile and import 10,000 bookmarks.
    3. Type in the URL bar and observe logs.
    
- Actual result:
2 calls for 40 ms / keystroke

- Expected result:
No calls. (It should only check if the location changes.)

Yes

Yes

Yes

- Extra QA steps:
    1. The "forsaken" sync group has a lot of bookmarks. It might help you if you're testing.

- Any related issues:
#7453

ayumi:
Fix #8703

Submitter Checklist:

Test Plan:
- Make sure tests are passing.

Reviewer Checklist:

Tests

### DIFF
--- a/js/about/newtab.js
+++ b/js/about/newtab.js
@@ -96,6 +96,7 @@ class NewTabPage extends React.Component {
     }).size > 0
   }
   isBookmarked (siteProps) {
+    // XXX: Fixme, not passing state in!
     return siteUtil.isSiteBookmarked(this.topSites, siteProps)
   }
   get gridLayout () {
--- a/js/state/siteCache.js
+++ b/js/state/siteCache.js
@@ -1,0 +1,90 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+const Immutable = require('immutable')
+const siteUtil = require('./siteUtil')
+const UrlUtil = require('../lib/urlutil')
+
+const createLocationSiteKeysCache = (state) => {
+  state = state.set('locationSiteKeysCache', new Immutable.Map())
+  state.get('sites').forEach((site, siteKey) => {
+    const location = siteUtil.getLocationFromSiteKey(siteKey)
+    if (!location) {
+      return
+    }
+    state = addLocationSiteKey(state, location, siteKey)
+  })
+  return state
+}
+
+module.exports.loadLocationSiteKeysCache = (state) => {
+  const cache = state.get('locationSiteKeysCache')
+  if (cache) {
+    return state
+  }
+  return createLocationSiteKeysCache(state)
+}
+
+/**
+ * Given a location, get matching appState siteKeys based on cache.
+ * Loads cache from appState if it hasn't been loaded yet.
+ * @param state Application state
+ * @param location {string}
+ * @return {Immutable.List<string>|null} siteKeys including this location.
+ */
+module.exports.getLocationSiteKeys = (state, location) => {
+  const normalLocation = UrlUtil.getLocationIfPDF(location)
+  return state.getIn(['locationSiteKeysCache', normalLocation])
+}
+
+/**
+ * Given a location, add appState siteKey to cached siteKeys list.
+ * Returns new state.
+ * @param state Application state
+ * @param location {string}
+ * @param siteKey {string}
+ */
+const addLocationSiteKey = (state, location, siteKey) => {
+  if (!siteKey || !location) {
+    return state
+  }
+  const normalLocation = UrlUtil.getLocationIfPDF(location)
+  const cacheKey = ['locationSiteKeysCache', normalLocation]
+  const siteKeys = state.getIn(cacheKey)
+  if (!siteKeys) {
+    return state.setIn(cacheKey, new Immutable.List([siteKey]))
+  } else {
+    if (siteKeys.includes(siteKey)) {
+      return state
+    }
+    return state.setIn(cacheKey, siteKeys.push(siteKey))
+  }
+}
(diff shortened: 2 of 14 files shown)


js/state/siteUtil.js
```diff
@@ -84,7 +85,26 @@ module.exports.getSiteKey = function (siteDetail) {
...
+  const splitKey = siteKey.split('|', 2)
+  if (typeof splitKey[0] === 'string' && typeof splitKey[1] === 'string') {
+    return splitKey[0]
+  }
+  return null
+}
+
+/**
+ * Checks if a siteDetail has the specified tag.
+ * O(n) warning; this iterates through sites.
```
bbondy: I think this should either be removed or it should just access in O(1) to check via site key.

js/state/siteUtil.js
```diff
@@ -107,6 +127,28 @@ module.exports.isSiteBookmarked = function (sites, siteDetail) {
...
+  const sites = state.get('sites')
+  const siteKeys = siteCache.getLocationSiteKeys(state, location)
+  if (!siteKeys || siteKeys.length === 0) {
+    return false
+  }
+  return siteKeys.some(key => {
+    const site = sites.get(key)
+    if (!site) {
+      return false
+    }
```
bbondy: some will still iterate over everything, I think it would be best to just make the cache related to bookmarks, and if the key is in there then return in O(1) if it is bookmarked.

app/browser/reducers/tabsReducer.js
```diff
@@ -55,6 +56,16 @@ const tabsReducer = (state, action, immutableAction) => {
...
       state = tabState.maybeCreateTab(state, action)
+      const url = action.getIn(['tabValue', 'url'])
+      if (url) {
+        let tabValue = immutableAction.get('tabValue')
+        if (!tabValue) {
+          break
+        }
+        const bookmarked = isLocationBookmarked(state, url)
+        tabValue = tabValue.set('bookmarked', bookmarked)
+        state = tabState.updateTabValue(state, tabValue)
```
bbondy: I think this should all be removed and then just add the bookmarked property in `getTabValue` and it'll work better.

ayumi: bookmarking a page doesn't seem to trigger this function. how would that state change get reflected here?

js/state/siteUtil.js
ayumi: This should only iterate over a median of 1 or 2 keys.

We could have a location-to-bookmarked cache if we assume a location can be bookmarked only once.

bbondy: gotcha ok

app/browser/reducers/tabsReducer.js
bbondy: I think I'd handle `appConstants.APP_ADD_SITE` in the `tabsReducer`, the only problem is that doesn't tell you which tab to update. You could either find the tabs with that location or pass in the info with ADD_SITE, that might be better if you can fix all the callers.

Then you could just call`updateTab` after getting the tab by tabId

js/about/newtab.js
```diff
@@ -96,6 +96,7 @@ class NewTabPage extends React.Component {
     }).size > 0
   }
   isBookmarked (siteProps) {
+    // XXX: Fixme, not passing state in!
```
bbondy: Could you check how often this is called? If it is often could you put this in a different component passing in only the URL? It would ensure each is only rendered / called once.

test/unit/state/siteCacheTest.js
```diff
@@ -0,0 +1,145 @@
+/* global describe, it */
+
+const siteTags = require('../../../js/constants/siteTags')
+const siteCache = require('../../../js/state/siteCache')
+const siteUtil = require('../../../js/state/siteUtil')
+const assert = require('assert')
+const Immutable = require('immutable')
+// const mockery = require('mockery')
+// const settings = require('../../../js/constants/settings')
```
bbondy: nit these can be removed in a follow up

test/unit/state/siteCacheTest.js
```diff
@@ -0,0 +1,145 @@
...
+
+describe('siteCache', function () {
```
bbondy: nice job on these tests 💯

js/state/siteCache.js
```diff
@@ -0,0 +1,90 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+const Immutable = require('immutable')
+const siteUtil = require('./siteUtil')
+const UrlUtil = require('../lib/urlutil')
+
+const createLocationSiteKeysCache = (state) => {
```
bbondy: In a follow up after this is merged, please move this to `app/common/state`.  Basically no files should be in js/* anymore and we're migrating away from that.

test/unit/state/siteCacheTest.js
```diff
@@ -0,0 +1,145 @@
+/* global describe, it */
```
bbondy: Please also move this test to `test/unit/app/common/state/`

bbondy:
Some minor follow ups to do in a new commit but merging this.

js/about/newtab.js
ayumi: i think it's called once per site icon shown on newTab, when you open a new tab.
it seems to be pretty fast.

Issue #8895, opened 2017-05-16
Spacing is missing between text for new private tab


cezaraugusto:
This landed in #8074 but somehow regressed

luixxiul:
https://github.com/brave/browser-laptop/commit/6ec02d8b4e7a0452240cf3a29c7a463d9a150989#commitcomment-22169274

Issue #8340, opened 2017-04-15
Discussion: CSS methodologies (OOCSS/BEM/SMACSS/...) in creating/refactoring with Aphrodite


luixxiul:
- Describe the issue you encountered:
https://github.com/brave/browser-laptop/commit/234cfba499ababd498b73ca31065219f9e369d4c#commitcomment-21773452

> Adding styles without semantic rules is a quite bad habit, which actually has resulted in creating CSS preprocessors like LESS and SCSS, and I notice that such a case started happenning on our project as well. We should consider the way of avoiding it somehow. I'm afraid of it that otherwise it would be very difficult to keep style consistency.

@cezaraugusto and @bsclifton agree with me that some kind of methodology to avoid styles from scattering and keep them consistent should be adopted, until Aphrodite will start supporting nesting styles like we do with LESS. By creating components with one of the methodologies it will be easy to nest the code blocks later.

At first we need to discuss which methodology we are going to use. At a glance BEM will be suited for us. See: https://github.com/brave/browser-laptop/pull/7831/files#diff-fa59d1e0221a5bee94fd2abd9b96dcfdR158

cezaraugusto:
I think BEM fill our needs nicely. Defining block/element/modifier for our codebase seems straightforward:

```css
.site-search {} /* Block */
.site-search__field {} /* Element */
.site-search--full {} /* Modifier */
```

and 

```css
.person {}
.person__hand {}
.person--female {}
.person--female__hand {}
.person__hand--left {}
```

Seems easy to apply that pattern to a component such as buttons:

```js
browserButton: {}, // block
browserButton__icon: {}, // element
browserButton--primary: {} // modifier
browserButton--primary__icon: {} // mixing it all
```

Looks good, improves readability and is easy enough to get started without digging too much on the methodology. 

I found this article a great resource on how to get started:
https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

++ from me

cezaraugusto:
I started to wonder if we could use a custom BEM style, so instead of using dashes we could use only one underline for modifiers? This way we could get ride of quotes. Sorry if bikeshedding just want to make sure we have a definitive guide for this.

```js
browserButton: {}, // block
browserButton__icon: {}, // element
browserButton_primary: {} // modifier
browserButton_primary__icon: {} // mixing it all
```

NejcZdovc:
@cezaraugusto yeah I like this one. I hate strings for properties

[...]

PR #8898, opened 2017-05-17
Fix regression: privateTabPaddingHorizontal

luixxiul:
Addresses #8895

Auditors: @cezaraugusto

Test Plan:
1. new private tab paragraphs should be properly spaced

Submitter Checklist:

Reviewer Checklist:

Tests

### DIFF
--- a/js/about/newprivatetab.js
+++ b/js/about/newprivatetab.js
@@ -88,11 +88,11 @@ const styles = StyleSheet.create({
   title: {
     color: globalStyles.color.white100,
     fontSize: '30px',
-    marginBottom: '20px'
+    marginBottom: globalStyles.spacing.privateTabPaddingHorizontal
   },
 
   text: {
-    paddingBottom: '20px',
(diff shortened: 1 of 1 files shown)


bsclifton:
Here before
screen shot 2017-05-17 at 9 20 22 am

and after
screen shot 2017-05-17 at 9 28 55 am

cezaraugusto:
thanks but I think as we move more things to Aphrodite we should avoid using global styles for spacing and keep that file specific for future-friendly theming

luixxiul:
For the sake of that I have added the prefix *privateTab* to the var :-)

NejcZdovc:
I agree here with @cezaraugusto, because this will probably only used in private tab, we don't need it in global. We should keep global as slick as possible.

luixxiul:
My thought on ^: https://github.com/brave/browser-laptop/issues/8340#issuecomment-298558379

PR #8915, opened 2017-05-17
fix 'Error: ENOTEMPTY: directory not empty' test errors

diracdeltas:
Test Plan:
1. Run a test such as `npm run test -- --grep='Syncing then'`. You should not see any ENOTEMPTY errors.
2. Travis logs should not contain a lot of ENOTEMPTY errors.

### DIFF
--- a/test/lib/brave.js
+++ b/test/lib/brave.js
@@ -7,7 +7,7 @@ require('./coMocha')
 const series = require('async/series')
 
 const path = require('path')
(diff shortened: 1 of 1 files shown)


diracdeltas:
Actually this doesn't get rid of all the errors because `rm -rf` doesn't work if a file is still being used (https://github.com/isaacs/rimraf/issues/25). But it does seem to reduce them by a lot.

Issue #8769, opened 2017-05-09
Add ability to update to preview releases (with advanced option)


aekeus:
This issue tracks the upgradeable preview browser project. A separate issue will track the server side requirements.

Jacalz:
This will be great 👍Will this also fix that it asks you to update to the stable channel when using the preview version? @aukeus

luixxiul:
How about letting users manage profiles or creating a incremental backup of the profile per each preview for users to restore it in case the browser preview crashes suddenly and the data is lost?

Users would like to test previews and send a bug report, as long as they know data would be kept safe... I feel it is a bit harsh to say "you should test on your own responsibility."

alexwykoff:
Test Plan : 
We should be able to update to preview builds from here on.

bsclifton:
@aekeus to add steps 😄

PR #8940, opened 2017-05-18
Add ability to update to a preview release

aekeus:
* Add advanced preference to update to preview releases
  * Add parameter sent to the updater service to allow preview releases
  * Re-factor method to retrieve version for the current browser

Auditors: @bbondy, @bsclifton

Test plan:

This pull request has been extensively tested with a manual
parallel update system. It has been tested on both windows and
mac with the following scenarios:

  a) preview off - confirmed update to subsequent release version
  b) preview off - confirmed no update to preview version
  c) preview on - confirmed update to subsequent preview version

This change set is dependent on having the most current version
of the updater, with preview support, installed.

Relates to #8769

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -58,7 +58,6 @@ const appConfig = require('../js/constants/appConfig')
 const appActions = require('../js/actions/appActions')
 const SessionStore = require('./sessionStore')
 const AppStore = require('../js/stores/appStore')
-const PackageLoader = require('./package-loader')
 const Autofill = require('./autofill')
 const Extensions = require('./extensions')
 const TrackingProtection = require('./trackingProtection')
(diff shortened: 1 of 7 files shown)


NejcZdovc:
@aekeus I know it's still WIP, but just a friendly reminder that we should add note on about brave page as well when they are on a preview verison.

app/index.js
```diff
@@ -458,33 +457,33 @@ app.on('ready', () => {
...
+      )
 
-      // Setup the auto updater, check the env variable first because it's
-      // used to check the update channel before releases.
-      Updater.init(process.platform, process.arch, process.env.BRAVE_UPDATE_VERSION || pack.version)
-
-      // This is fired by a menu entry (for now - will be scheduled)
+      // This is fired by a menu entry
       process.on(messages.CHECK_FOR_UPDATE, () => Updater.checkForUpdate(true))
-      ipcMain.on(messages.CHECK_FOR_UPDATE, () => Updater.checkForUpdate(true))
```
bbondy: I think this line is still needed.

app/updater.js
```diff
@@ -84,21 +85,26 @@ var scheduleUpdates = () => {
...
+  var query = { accept_preview: updateToPreviewReleases ? 'true' : 'false' }
 
   if (baseUrl) {
     debug(`updateUrl = ${baseUrl}`)
     scheduleUpdates()
     // This will fail if we are in dev
     try {
-      autoUpdater.setFeedURL(baseUrl)
+      // add the preview flag to the base feed url
+      autoUpdater.setFeedURL(`${baseUrl}?${querystring.stringify(query)}`)
```
bbondy: Should this be wrapped with encodeURIComponent?

bbondy: or even cleaner just ?accept_preview=true or ?accept_preview=false, but that might require server change.

aekeus:
@NejcZdovc Thanks for the suggestion. Right now the browser has no indication that it is a preview release. This was done so that we can then promote a preview/rc to full status without issuing a new binary.

app/index.js
aekeus: Great find. This broke the hamburger menu item. Fixing now, thanks.

app/updater.js
bbondy: OK cool re Slack, querystring encodes it as normal params.

bbondy:
++

luixxiul:
just a question: is it possible to downgrade the preview? or do we have to re-install the release by manual?

aekeus:
@bsclifton The changes to the server side, vault-updater, have been tested and deployed. As you commented, the accept_preview flag is used to determine if metadata is returned when an update request is made.

aekeus:
@luixxiul It is not possible to downgrade a preview release. A re-install is required.

luixxiul:
> @luixxiul It is not possible to downgrade a preview release. A re-install is required.

I think that is worth mentioning as a note for users.

PR #8943, opened 2017-05-18
WIP: fix #3227

### LINKED ISSUE #3227: Intermittent failure on travis with fingerprinting 2 not 3
There's some kind of problem with travis and fingerprinting test with a permfail on the `blocks fingerprinting` test. This value should be 3 but it returns 2 always on travis.

For now I check for either of them but we should fix this or adjust the test to account for what's wrong. I think it's the latter.
#2626

diracdeltas:
opening a PR to check whether it passes on Travis


diracdeltas:
looks like it didn't work

PR #8992, opened 2017-05-21
Fixes tab title on tab update

### LINKED ISSUE #8908: Title is not updated when using navigation buttons
yes
Title is not updated when using navigation buttons
macOS
8bb62f7
- Steps to reproduce:
    1. go to site clifton.io
    2. create a history
    3. click on back/forward buttons

    
- Actual result:
sometimes tab title is not display (location is shown)
- Expected result:
title should be displayed
yes
no
not all the time but with few clicks you can see a problem

NejcZdovc:
Submitter Checklist:

Resolves #8908

Test Plan:
- go to site clifton.io
- create a history
- click on back/forward buttons
- check that title is always correct

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/reducers/frameReducer.js
+++ b/app/renderer/reducers/frameReducer.js
@@ -92,7 +92,8 @@ const frameReducer = (state, action, immutableAction) => {
         }
       }
 
-      const title = immutableAction.getIn(['changeInfo', 'title'])
+      // TODO fix race condition in Muon more info in #9000
+      const title = immutableAction.getIn(['tabValue', 'title'])
       if (title != null) {
         state = state.setIn(['frames', index, 'title'], title)
       }


app/renderer/reducers/frameReducer.js
```diff
@@ -92,7 +92,7 @@ const frameReducer = (state, action, immutableAction) => {
         }
       }
 
-      const title = immutableAction.getIn(['changeInfo', 'title'])
+      const title = immutableAction.getIn(['tabValue', 'title'])
```
bridiver: this may help, but won't be a complete fix because it doesn't address the race condition. Getting the title from `tabValue` will only fix the title when the tab state changes for some other reason. I think this change is ok for now, but we should add a comment/TODO that the race condition with appState is the real problem

bridiver: we also shouldn't mark this as resolving the issue for the same reason

NejcZdovc: Currently the only problem with the title not updating is when you use navigation buttons, so when tab state changes. What should I wrote in todo? `TODO fix race condition with appState`

Issue #8974, opened 2017-05-20
Control+W does not work on tabs opened by clicking go back button


luixxiul:
## Test plan
    1. Open about:preferences#tabs
    2. Change "When closing an active tab" setting to `Select the last viewed tab`
    3. Open a new tab
    1. Visit https://github.com on the tab
    2. Click 'Features' on the menu
    3. Click Go back button **two times** with the middle mouse button on the navigation bar
    4. Select the last tab
    5. <kbd><kbd>Control</kbd> + <kbd>W</kbd></kbd>
    6. Repeat ^

Focus should go form last tab to the parent tab (since it was last active).

Since the fix involved refactoring the logic for close tab, you may wish to test that all 3 of these options work as expected:
![screen shot 2017-05-25 at 8 54 51 am](https://cloud.githubusercontent.com/assets/4733304/26458365/dafede82-4127-11e7-8b3a-340ae1d224c6.png)

## Description
**Describe the issue you encountered**: By changing the tab setting to `Select the last closed tab`, Control+W does not work on tabs opened by clicking go back button.

- Platform (Win7, 8, 10? macOS? Linux distro?): tested on debian

- Brave Version (revision SHA): 0.15.306

- Steps to reproduce:
    1. Open about:preferences#tabs
    2. Change "When closing an active tab" setting to `Select the last closed tab`
    3. Open a new tab
    1. Visit https://github.com on the tab
    2. Click 'Features' on the menu
    3. Click Go back button **two times** with the middle mouse button on the nagivation bar
    4. Select the last tab
    5. <kbd><kbd>Control</kbd> + <kbd>W</kbd></kbd>
    6. Repeat ^
    
- Actual result: only 1 tab can be closed

- Expected result: both tabs should be closed

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:
  

- Any related issues:

bsclifton:
When the tab is closed, we do calculate a new active frame (and this is correct).  However, the tab state is never updated (ex: `appState.get('tabs')`.  When any action happens which tries to use the `TAB_ID_ACTIVE`, it will then fail (because no tabs are marked as active).

Working on a fix...

bsclifton:
Findings I had while making a work-around:

 The underlying problem has these steps:
 - tab is cloned two times
 - focus is given to the 2nd cloned tab
 - ctrl + w happens; this triggers `APP_TAB_CLOSE_REQUESTED`
   - window is closed as expected
[...]

PR #9023, opened 2017-05-24
Active tab update issue

bridiver:
Submitter Checklist:

Fixes https://github.com/brave/browser-laptop/issues/8974

Test Plan:

Reviewer Checklist:


bsclifton:
I think a new Muon version is needed (build is in progress).  This includes the following patch:
https://github.com/brave/muon/commit/4fdf666cf6b8b0492dbf7f5fbdbb9d9f99e1d4bb

When using this code as-is, I was able to cause an issue by setting tab close behavior to the default (parent).  Will retest when new Muon is available 😄

bsclifton:
After the Muon PR, the issues I saw were fixed 👍 

I think he only remaining issue is:
Behaviors for close (when you pick one of the three options) don't seem to work as described
screen shot 2017-05-25 at 8 54 51 am

bsclifton:
My setup:
2 pinned tabs (https://slashdot.org/, https://www.roboform.com/filling-test-all-fields)

`select the last viewed tab` - this always seems to be choosing the 2nd pinned tab, no matter what I do

`select the next tab` - this behavior is inconsistent. Sometimes, it picks the previous tab, sometimes it picks what seems to be a random tab (maybe the previous index?).  If I have two tabs open and try to close the first, it selects the first pinned tab.

`select its parent tab` - this seems to be working for the most part. It doesn't always select the parent though.  If I open the same link via middle clicking 3 times, closing one of them chooses another one (not necessarily the opener)

bridiver:
I pushed a fix for "last active", but I can't replicate any issues with "next tab". "parent" should only go back to the parent if it is also the last active tab, otherwise it should go to "next"

PR #9051, opened 2017-05-25
Converts urlBarSuggestions into redux

NejcZdovc:
Submitter Checklist:

Auditors: @bsclifton @bridiver @bbondy

Test Plan:
- test if suggestion list is displayed
- test if hover and click is working on suggestion list

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/navigation/urlBarSuggestions.js
+++ b/app/renderer/components/navigation/urlBarSuggestions.js
@@ -3,112 +3,103 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const React = require('react')
-const ImmutableComponent = require('../immutableComponent')
+const Immutable = require('immutable')
 
+// Components
+const ReduxComponent = require('../reduxComponent')
 const UrlBarSuggestionItem = require('./urlBarSuggestionItem')
-const windowActions = require('../../../../js/actions/windowActions')
+
+// Actions
 const appActions = require('../../../../js/actions/appActions')
-const suggestionTypes = require('../../../../js/constants/suggestionTypes')
+
+// Utils
 const cx = require('../../../../js/lib/classSet')
 const locale = require('../../../../js/l10n')
-const {isForSecondaryAction} = require('../../../../js/lib/eventUtil')
+const suggestions = require('../../../common/lib/suggestion')
+const frameStateUtil = require('../../../../js/state/frameStateUtil')
 const {getCurrentWindowId} = require('../../currentWindow')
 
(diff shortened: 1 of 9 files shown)


app/renderer/components/navigation/urlBarSuggestions.js
```diff
@@ -3,112 +3,103 @@
...
+    addToItems(list.topSiteSuggestions, 'topSiteTitle', locale.translation('topSiteSuggestionTitle'), 'fa-link')
+
+    return items
   }
 
-  componentDidMount () {
+  mergeProps (state, dispatchProps, ownProps) {
+    const currentWindow = state.get('currentWindow')
+    const activeFrame = frameStateUtil.getActiveFrame(currentWindow) || Immutable.Map()
+    const urlBar = activeFrame.getIn(['navbar', 'urlbar']) || Immutable.Map()
```
bridiver: just fyi - you can use `activeFrame.getIn(['navbar', 'urlbar'], Immutable.Map())` which is a little safer because it properly handles falsy values when applicable

app/renderer/components/navigation/urlBarSuggestions.js
```diff
@@ -3,112 +3,103 @@
...
+    const documentHeight = Number.parseInt(
+      window.getComputedStyle(document.querySelector(':root')).getPropertyValue('--navbar-height'), 10
```
bridiver: we should add a TODO to take a look at this because I suspect getting the computed styles for the entire document is very slow

js/actions/appActions.js
```diff
@@ -1394,6 +1394,23 @@ const appActions = {
   },
 
   /**
+   * Indicates URL bar selected index
+   *
+   * @param {number} windowId - the window ID
+   * @param {number} selectedIndex - The index for the selected item (users can select items with down arrow on their keyboard)
+   */
+  urlBarSelectedIndexChanged: function (windowId, selectedIndex) {
```
bridiver: I like this change to pass the index instead of calculating the selections in the component

app/renderer/components/navigation/urlBarSuggestions.js
NejcZdovc: nice, fixed

app/renderer/components/navigation/urlBarSuggestions.js
NejcZdovc: I will fix all computed styles in #9176, so I think there is no need for TODO

Issue #8978, opened 2017-05-20
Loading about:safebrowsing after blocking phishing site displays the original URL of the tab for an instant


luixxiul:
**Describe the issue you encountered:** Loading about:safebrowsing after blocking a phishing site displays the original URL of the tab for an instant.

![clipboard01](https://cloud.githubusercontent.com/assets/3362943/26275959/c22ae51e-3da7-11e7-8ffb-b9c32eaab572.gif)

- Platform (Win7, 8, 10? macOS? Linux distro?): Tested on Debian

- Brave Version (revision SHA): 0.15.306

- Steps to reproduce:
    1. Load downloadme.org on a GitHub page
    
- Actual result: The GitHub page's URL is displayed for an instant

- Expected result: It should not be displayed.

- Is this an issue in the currently released version?
  No.
  ![clipboard01](https://cloud.githubusercontent.com/assets/3362943/26275974/302ad100-3da8-11e7-94f4-6687cee60751.gif)

- Extra QA steps:
    1.
    2.
    3.

- Any related issues:

diracdeltas:
i think this is not a blocker but will look at it

diracdeltas:
this is tricky to fix. i think it is best to just wait until the workaround in ea4f72a1edbfb35512d8338ba689ad97933065ed is no longer needed, then this issue will automatically go away.

luixxiul:
This can no longer be reproduced :-)

Issue #9056, opened 2017-05-25
about:safebrowsing should work regardless of how you got there


alexwykoff:
- Did you search for similar issues before submitting this one?
yes

- Describe the issue you encountered:
when reviewing https://github.com/brave/browser-laptop/issues/8905 it was discovered that cmd+click opening a link to downloadme.org resulted in a blank page

- Platform (Win7, 8, 10? macOS? Linux distro?):
os x

- Brave Version (revision SHA):
0.15.3xx

- Steps to reproduce:
    1. visit http://alexwykoff.com/tests/testlinks.html
    2. cmd+click on the downloadme.org link

    
- Actual result:
a blank page is displayed

- Expected result:
All manner of display for about:safebrowsing should work. This includes but is not limited to:
Clicking on the link
Cmd+clicking on the link
Directly typing into the urlbar
Bookmarks

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
Yes

- Is this an issue in the currently released version?
No

- Can this issue be consistently reproduced?
Yes

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues:

diracdeltas:
checked that all of those methods worked except cmd-click/rightclick to open in new tab

PR #9058, opened 2017-05-25
Temporary fix for #9056

diracdeltas:
Test Plan:
1. go to http://alexwykoff.com/tests/testlinks.html
2. open downloadme.org directly in the urlbar, via a link, via right-click to open in new tab, and via cmd+click. all should eventually redirect to the safebrowsing page, and the safebrowsing text should be visible.
3. safebrowsing travis test should still pass
4. go to about:preferences#shields and disable safebrowsing
5. go back to  http://alexwykoff.com/tests/testlinks.html and cmd+click on downloadme.org. the site should now load.

this is a temporary workaround. it causes the safebrowsing page to flicker while loading in some cases.

### DIFF
--- a/app/adBlock.js
+++ b/app/adBlock.js
@@ -32,6 +32,10 @@ module.exports.safeBrowsingResourceName = 'safeBrowsing'
  *   for example safe browsing checks main frame URLs but ad block checks do not.
  */
 const startAdBlocking = (adblock, resourceName, shouldCheckMainFrame) => {
+  if (resourceName === module.exports.safeBrowsingResourceName) {
+    // Needed for #9056
+    module.exports.safeBrowsingInstance = adblock
+  }
   Filtering.registerBeforeRequestFilteringCB((details) => {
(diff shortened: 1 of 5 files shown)


alexwykoff:
I tested vs a clean profile. cmd+click did not appear to work
safebrowsing

diracdeltas:
@alexwykoff does right-click to open in new tab work? does it work intermittently?

diracdeltas:
nvm  i can repro

diracdeltas:
@alexwykoff please try this new fix

@luixxiul i think this also fixes #8978 or at least turns it into a different bug. :)

alexwykoff:
++ lgtm

diracdeltas:
@alexwykoff added tests

bbondy:
@diracdeltas I'm going to merge but please do a follow up to remove that code from tabs.js.  Alex mentioned that you can't reproduce this in 0.15.2 so I'd start with doing a regression range find to see.

PR #9061, opened 2017-05-25
keep redux component state between renders

bridiver:
Submitter Checklist:

Test Plan:
Automated tests

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/reduxComponent.js
+++ b/app/renderer/components/reduxComponent.js
@@ -16,13 +16,13 @@ const buildPropsImpl = (props, componentType) => {
 }
 
 class ReduxComponent extends ImmutableComponent {
-  constructor (props) {
+  constructor (componentType, props) {
     super(props)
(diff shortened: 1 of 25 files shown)


bridiver:
also `props` should be passed in `constructor` and `super` for `React.Component` so I updated all of them

app/renderer/components/reduxComponent.js
```diff
@@ -72,8 +72,5 @@ class ReduxComponent extends ImmutableComponent {
 }
 
 module.exports.connect = (componentType) => {
```
bridiver: this is the actual fix, but passing `props` in the `super` call is required

NejcZdovc:
I will merge it because I need it to test it out in the other PR and it's working correctly. Only one thing is that I am not sure. @bridiver why did you added props into constructor for non redux components? It was not there before and it was working correctly?

Issue #4318, opened 2016-09-27
Menu Bar Obscures Caption Buttons


jonathansampson:
**Did you search for similar issues before submitting this one?**

Yes

**Describe the issue you encountered:**

At certain widths, the Menu Bar will push the Window Caption Buttons out of frame.

**Expected behavior:**

The Caption Buttons and Menu Bar can be shown together, without obscuring either.

- Platform (Win7, 8, 10? macOS? Linux distro?): Windows 10

- Brave Version: 0.12.2

- Steps to reproduce:
    1. Toggle "Hide the menu bar" off in Settings
    2. Begin reducing the width of the Brave window
    3. Note that at two (at least) breakpoints, the Window Close button is not visible

- Screenshot if needed:

![image](https://cloud.githubusercontent.com/assets/815158/18863524/8f1f9e32-8446-11e6-98a8-f56d59989a10.png)

- Any related issues:

luixxiul:
Too much menus on the bar? Here at least "Debug" could be removed but I think it does not solve the issue completely as the length of strings of menus are different among languages. We should investigate the situation on each language.

luixxiul:
a way to solve it could be to set ``overflowIndicator`` just like we do in bookmarksToolbar.js

luixxiul:
I'm digging this issue on Windows 7 and tend to think it is impossible to solve this only with CSS. By removing most of the properties under `navbarMenubarFlexContainer`, I thought setting `white-space: nowrap` to `.menuBar` would solve the issue, but it doesn't because the width of `html` is not affected by any change inside it.

Without `white-space: nowrap`
![clipboard01](https://cloud.githubusercontent.com/assets/3362943/21078788/fd68c5ea-bfbe-11e6-8e36-b57216203eb3.png)

With `white-space: nowrap`
![clipboard01](https://cloud.githubusercontent.com/assets/3362943/21078790/0e5af67a-bfbf-11e6-9c58-16538776eca2.png)

I'm wondering if there would be a way of changing the width of `html` dynamically based on the width inside it, in order to display all elements without overflowing.

bsclifton:
Moved to backlog

cezaraugusto:
@luixxiul didn't get in this code but what about having ellipsis for the menu bar, would that work? not sure per comments if we can manipulate something in there but we could at least have a wrapper around.

[...]

PR #9093, opened 2017-05-27
Add "view all" option to download bar

### LINKED ISSUE #9090: "Show all" button  on download bar to view all downloads
Yes

This is a feature request. There should be a download button on download bar like in chrome to view all downloads.

Manjaro Linux 17.0.1 Gellivara

0.14.1

- Steps to reproduce:
    1. Download something
    
- Actual result:
No "Show all" button
2017-05-27-120205_1365x63_scrot

- Expected result:
"Show all" button
2017-05-27-120352_1363x56_scrot

abs-zero:
Resolve #9090

Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues/9090) for my issue if one did not already exist.

Test Plan:

Reviewer Checklist:

Tests

### Description
This tries to resolve issue #9090. The solution is still incomplete.
**Before:**
2017-05-27-184118_1356x52_scrot

**After:**
peek 2017-05-27 17-57

### Problem
However, a new problem arose with responsiveness.
not-responsive

I do not have any idea where in the code and how this hiding of download items is being done. So, I hope I could use some help from here.

### DIFF
--- a/app/renderer/components/download/downloadsBar.js
+++ b/app/renderer/components/download/downloadsBar.js
@@ -27,7 +27,6 @@ class DownloadsBar extends ImmutableComponent {
     super()
     this.onHideDownloadsToolbar = this.onHideDownloadsToolbar.bind(this)
     this.onShowDownloads = this.onShowDownloads.bind(this)
-    this.downloadBarButtonsNode = null
   }
   onHideDownloadsToolbar () {
     windowActions.setDownloadsToolbarVisible(false)
@@ -40,17 +39,12 @@ class DownloadsBar extends ImmutableComponent {
     })
     windowActions.setDownloadsToolbarVisible(false)
   }
-
   render () {
     const getComputedStyle = require('../../getComputedStyle')
     const downloadItemWidth = Number.parseInt(getComputedStyle('--download-item-width'), 10)
     const downloadItemMargin = Number.parseInt(getComputedStyle('--download-item-margin'), 10)
     const downloadBarPadding = Number.parseInt(getComputedStyle('--download-bar-padding'), 10)
-
-    const downloadBarButtons = this.downloadBarButtonsNode !== null
-                            ? this.downloadBarButtonsNode.getBoundingClientRect().width
-                            : 0
-
+    const downloadBarButtons = Number.parseInt(getComputedStyle('--download-bar-buttons'), 10)
     const numItems = Math.floor((this.props.windowWidth - (downloadBarPadding * 2) - downloadBarButtons) / (downloadItemWidth + downloadItemMargin))
     return <div className='downloadsBar'
       onContextMenu={contextMenus.onDownloadsToolbarContextMenu.bind(null, undefined, undefined)}>
@@ -68,13 +62,10 @@ class DownloadsBar extends ImmutableComponent {
                 downloadsSize={this.props.downloads.size} />)
         }
       </div>
-      <div
-        className={cx({
-          downloadBarButtons: true,
-          [css(styles.downloadsBar__downloadBarButtons)]: true
-        })}
-        ref={refNode => { this.downloadBarButtonsNode = refNode }}
-        >
+      <div className={cx({
+        downloadBarButtons: true,
+        [css(styles.downloadsBar__downloadBarButtons)]: true
+      })}>
         <BrowserButton secondaryColor
           l10nId='downloadViewAll'
           testId='downloadViewAll'
--- a/app/renderer/components/download/downloadsBar.js
+++ b/app/renderer/components/download/downloadsBar.js
@@ -10,7 +10,7 @@ const Button = require('../common/button')
 const BrowserButton = require('../common/browserButton')
 const DownloadItem = require('./downloadItem')
 
-const {StyleSheet} = require('aphrodite/no-important')
+const {StyleSheet, css} = require('aphrodite/no-important')
 
 // Actions
 const windowActions = require('../../../../js/actions/windowActions')
@@ -19,6 +19,8 @@ const appActions = require('../../../../js/actions/appActions')
 
 // Utils
 const contextMenus = require('../../../../js/contextMenus')
+
+const cx = require('../../../../js/lib/classSet')
 
 class DownloadsBar extends ImmutableComponent {
   constructor () {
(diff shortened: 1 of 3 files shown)


luixxiul:
@abs-zero I opened a PR on your repo here: https://github.com/abs-zero/browser-laptop/pull/1

yet I'm not sure how to solve the problem you mentioned, so I left it as-is.

cezaraugusto:
hi @abs-zero I think @luixxiul covered what was missing, is this ready for review? thanks!

luixxiul:
@cezaraugusto actually not completely. the timing (viewport width) when the downloaded item should be hidden needs fixing. I just fixed the alignment of the close and show all button.

abs-zero:
Yes, the problem I mentioned with a gif still has not been fixed. It would help if I knew where the logic of unmounting of download items are written.

cezaraugusto:
so calculation to show/hide items takes into account the downloadBar button -- in fact the close (`x`) button in the right. It is set to `22px` which is a little more than the button size (`18px`).

With the change proposed here, we should take into account the `view all` button as well, so we need to update variable size to take the new button size as well. For reference, code is described here: https://github.com/brave/browser-laptop/blob/master/less/downloadBar.less#L11

For me, the total size including view all option seems to be around `123px`. @abs-zero can you test if changing var value to that fix the issue? seems to fix for me. If so pls update PR so we can accept it. btw thanks for the help, PR is looking great!

cezaraugusto:
btw calculation is defined here: https://github.com/brave/browser-laptop/blob/master/app/renderer/components/download/downloadsBar.js#L34 but I don't think there's no need to change anything there

abs-zero:
Thanks for the pointing out the direction. I will test it.

app/renderer/components/download/downloadsBar.js
```diff
@@ -7,24 +7,39 @@ const React = require('react')
...
     windowActions.setDownloadsToolbarVisible(false)
     webviewActions.setWebviewFocused()
   }
+  onShowDownloads () {
+    appActions.createTabRequested({
+      activateIfOpen: true,
+      url: 'about:downloads'
+    })
+    windowActions.setDownloadsToolbarVisible(false)
+    webviewActions.setWebviewFocused()
```
cezaraugusto: btw I think this method is not needed

abs-zero: you mean these two lines: 
```
windowActions.setDownloadsToolbarVisible(false)
webviewActions.setWebviewFocused()
```
?

abs-zero:
All the problems have been fixed now. Thanks for all the help.

luixxiul:
> For me, the total size including view all option seems to be around 123px

the thing is the value varies on every language. the total width requires l10n viewpoint.

app/renderer/components/download/downloadsBar.js
cezaraugusto: only `webviewActions.setWebviewFocused()`. For parity, we should auto-hide the downloads bar after clicking "show all", as you did above.

cezaraugusto:
@luixxiul yes that's correct. 

We'll need then to get `downloadBarButtons` width first, no matter the language. I would go with `getBoundingClientRect().width` method, so we can find the button wrapper width, and with this value, replace `downloadsBarButtons` so `numItems` could make the correct calculation.

@abs-zero is this do-able to do? don't want to add extra work so ok if not I can jump in and add these changes. Let me know if you need something from us.

NejcZdovc:
@cezaraugusto we should really avoid using `getBoundingClientRect`, because we would be calling it on every resize as well if I understand correctly. We are doing this already on a lot of places. Can we use ` flex` and say that left part where we have items should grow automatically and the left part where we have the button should be 1?

abs-zero:
@cezaraugusto Yup, I have pushed a commit using `getBoundingClientRect`. @NejcZdovc I am sorry I dont understand what you said about `flex`

luixxiul:
`Cannot read property 'getBoundingClientRect' of null`

cezaraugusto:
@NejcZdovc your assumption is correct re method will be called each time window is resized. But the way downloadsBar items behave is for parity with Chrome. There's no CSS-only way to emulate the same mechanism without cutting off some element -- which would lead us to bad UI.

I think this addition is very cool, even more coming from a contributor. While I agree with you that adding more and more listeners for window resize would lead us to perf decrease I think it is still too small for this change and if I'm wrong, we can always roll back or even ask Brad for a new spec in downloadsBar.

NejcZdovc:
@cezaraugusto We are now fixing a lot of perf problems and we don't want to introduce new ones. Maybe it looks like a small one, but adding a lot of small ones we get a big problem. `getBoundingClientRect` is really expensive. Can you please write a quick summery what you want to achieve and I will take a look. Maybe I find a way with CSS only :smiley:

luixxiul:
@NejcZdovc in that case please don't forget to change lang settings to see your fix is avoiding l10n issues. thanks!

I tried to find a pure CSS solution but I couldn't. I assume removing the fixed width of a downloaded item would be that way.

NejcZdovc:
@luixxiul I am just finishing pause download PR and I will take a look at this one. Thanks for the heads up regarding l10n.

cezaraugusto:
@NejcZdovc it should work the same way as currently but including a new element ("view all" button). You can also take a look at how Chrome does it, which is the same way as we do.

The first fix I suggested in https://github.com/brave/browser-laptop/pull/9093#issuecomment-305708781 was simple and inexpensive, css-only. However @luixxiul reminded that locale may change button's size, so any value we add would be just a guess.

luixxiul:
@cezaraugusto I had faced the similar issue on https://github.com/brave/browser-laptop/issues/4318, which I couldn't find a way of fixing it with CSS only. Do you think that issue needs the same solution you mentioned above?

abs-zero:
If perf is going to be a problem, can we not call `getBoundingClientRect` only once when the download bar is set visible or when the language setting is changed and cache the `width`. This way we do not have to call `getBoundingClientRect` on every resize.

abs-zero:
@luixxiul Sorry for the broken commit earlier. I did not encounter that error before maybe because a cached bundle was being used.

cezaraugusto:
hey @abs-zero thanks for fixing that.

can you revert 3219cf27321e63cad77b6421d048f0f931a06f5a (but keeping `webviewActions.setWebviewFocused()` removed) and also 8da955c7fc39115f5b9618f5ee0c9457ed1783f1? We just need to make a tweak on `--download-bar-buttons` variable size like I said before and we're good to go. It is the same thing you did in a94b8ed6369e4d12887e801f839e4d2e0054932f but just update it from 127px to 140px. It works ok in that size, sorry for confusion. 

That said, once done that change it looks good for me, thanks!

abs-zero:
Done. :smiley:

bsclifton:
Thanks for working hard on this @abs-zero (and huge thanks @cezaraugusto and @luixxiul for the assists).  This will be a great touch when it's all wrapped up 😄 👍

PR #8926, opened 2017-05-18
use navigation state for urlbar display value

bridiver:
Submitter Checklist:

- [ ] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [ ] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Test Plan:

Reviewer Checklist:

Tests

- [ ] Adequate test coverage exists to prevent regressions
- [ ] Tests should be independent and work correctly when run individually or as a suite [ref](https://github.com/brave/browser-laptop/wiki/Code-Guidelines#test-dependencies)
- [ ] New files have MPL2 license header

PR #9154, opened 2017-05-30
Navigation state

bridiver:
replacement for https://github.com/brave/browser-laptop/pull/8926

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests


app/browser/reducers/flashReducer.js
```diff
@@ -20,7 +20,7 @@ const flashReducer = (state, action, immutableAction) => {
         const location = action.get('location')
         const settings = siteSettings.getSiteSettingsForURL(state.get('siteSettings'), location)
         if (!(settings && ['boolean', 'number'].includes(typeof settings.get('flash')))) {
-          flash.showFlashMessageBox(location, action.get('senderTabId'))
+          flash.showFlashMessageBox(location, action.get('tabId'))
```
bbondy: nice catch

app/common/actions/tabActions.js
```diff
@@ -0,0 +1,40 @@
...
+
+'use strict'
+
+const dispatchAction = require('../../common/dispatcher/dispatchAction')
+
+const tabActions = {
+  // TODO(bridiver) - this is an action anti-pattern and the code that uses it should be refactored
+  // to be declarative
+  reload: function tabActionsReload (tabId, ignoreCache) {
+    dispatchAction(tabActions.reload.name, {
```
bbondy: I like that it's split out into a different action file like we previously discussed; however, I think that we'll eventually hit an unintentional bug because of not namespacing on the object or module and using just the function name.  I won't block the review on that but I think we should change it in the near future since it'll be prone to collisions.

app/browser/tabs.js
```diff
@@ -283,6 +284,57 @@ const updateAboutDetails = (tab, tabValue) => {
...
+
+  return navigationEntry
+}
+
+const createNavigationState = (navigationHandle, controller) => {
+  return Immutable.Map({
+    hasCommitted: navigationHandle.hasCommitted(),
+    isErrorPage: navigationHandle.isErrorPage(),
+    netErrorCode: navigationHandle.getNetErrorCode(),
+    activeEntry: fixDisplayURL(controller.getActiveEntry(), controller),
```
bbondy: This would be a good place for some comments about what each *Entry property is for.  Please add in a follow up.

app/common/dispatcher/dispatchAction.js
```diff
@@ -0,0 +1,17 @@
+module.exports = function (name, action) {
```
bbondy: I wonder if it might have been better to just re-use dispatch from appDispatcher but I guess that might be a little confusing.

app/common/actions/tabActions.js
```diff
@@ -0,0 +1,40 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const dispatchAction = require('../../common/dispatcher/dispatchAction')
+
+const tabActions = {
+  // TODO(bridiver) - this is an action anti-pattern and the code that uses it should be refactored
+  // to be declarative
+  reload: function tabActionsReload (tabId, ignoreCache) {
```
bbondy: it's fine for this one but I prefer something like reloadRequested when it's a temporary thing that we will change later.

app/renderer/components/navigation/urlBar.js
```diff
@@ -364,60 +357,46 @@ class UrlBar extends React.Component {
...
 
   componentDidUpdate (prevProps) {
-    // this.urlInput is not initialized in titleMode
-    if (this.urlInput) {
-      const pdfjsEnabled = getSetting(settings.PDFJS_ENABLED)
-      if (this.props.activeFrameKey !== prevProps.activeFrameKey) {
-        // The user just changed tabs
-        this.setValue(this.props.locationValue !== 'about:blank'
-          ? this.props.locationValue
-          : UrlUtil.getDisplayLocation(this.props.location, pdfjsEnabled))
```
bbondy: yay for this being removed.

app/renderer/reducers/frameReducer.js
```diff
@@ -64,6 +66,19 @@ const closeFrame = (state, action) => {
   return state
 }
 
+const getLocation = (location) => {
+  location = location.trim()
+  location = getSourceAboutUrl(location) ||
```
bbondy: Does this need a case for pdf as well?

js/lib/appUrlUtil.js
```diff
@@ -209,7 +209,6 @@ module.exports.isTargetMagnetUrl = function (input) {
  * @param {string} input
  */
 module.exports.isUrl = function (input) {
```
bbondy: This is probably the funniest part of our code :S

bbondy:
Could you double check on this test, it could possibly be new:
```
  2) urlBar tests loading same URL as current page with changed input "before all" hook:
     element (#titleBar) still not visible after 10000ms
  Error
```

app/common/actions/tabActions.js
bridiver: yea, I had actually commented on this in a previous version of the PR and said I wasn't super happy with it, but trying to find some way to avoid the constants file

app/common/dispatcher/dispatchAction.js
bridiver: in the short term it's confusing, but we need to split out the actual dispatcher functionality from the code that uses it so we can include it in different contexts like worker threads. I'm open to name changes if you have thoughts...

bridiver:
I've seen that test fail intermittently and I checked all the failures, but that was several rebases ago

app/common/dispatcher/dispatchAction.js
bridiver: also wanted to remove the appStore dependency for actions because it causes a lot of circular dependency issues

app/common/actions/tabActions.js
bridiver: any thoughts/ideas?

PR #9176, opened 2017-05-31
Saves computed styles in the state

### LINKED ISSUE #9149: Move getComputedStyle into store
Move `window.getComputedStyle` into app store. We should have const file where we define all variables that are used in `getComputedStyle` and when app is mounted or resized we will automatically change this values.

NejcZdovc:
Submitter Checklist:

Resolves #9149

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/windows.js
+++ b/app/browser/windows.js
@@ -55,7 +55,7 @@ const getWindowValue = (windowId) => {
 }
 
 const updateWindow = (windowId) => {
-  let windowValue = getWindowValue(windowId)
+  const windowValue = getWindowValue(windowId)
   if (windowValue) {
     appActions.windowUpdated(windowValue)
   }
--- a/app/common/lib/windowsUtil.js
+++ b/app/common/lib/windowsUtil.js
@@ -4,10 +4,14 @@
 
 const Immutable = require('immutable')
 
-module.exports.getPinnedSiteProps = site => {
+const getPinnedSiteProps = site => {
   return Immutable.fromJS({
     location: site.get('location'),
     order: site.get('order'),
     partitionNumber: site.get('partitionNumber') || 0
   })
+}
+
+module.exports = {
+  getPinnedSiteProps
 }
--- a/app/renderer/components/bookmarks/bookmarksToolbar.js
+++ b/app/renderer/components/bookmarks/bookmarksToolbar.js
@@ -34,6 +34,7 @@ const {calculateTextWidth} = require('../../../../js/lib/textCalculator')
 
 // Styles
 const globalStyles = require('../styles/global')
+const domUtil = require('../../lib/domUtil')
 
 class BookmarksToolbar extends ImmutableComponent {
   constructor () {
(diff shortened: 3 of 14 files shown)


app/renderer/components/window.js
```diff
@@ -93,11 +96,14 @@ class Window extends React.Component {
 
   componentDidMount () {
     appActions.windowReady(getCurrentWindowId())
+    this.updateComputedStyles()
+    window.addEventListener('resize', this.updateComputedStyles)
```
bridiver: we already have a window resize handler in windows.js that triggers `appActions.windowUpdated` and it is debounced to prevent a flood of actions during resize. I think it would be better to send that action to the window using queryInfo (see `urlBarTextChanged` for an example) and update the computed styles in response to that

app/renderer/components/window.js
```diff
@@ -45,6 +47,7 @@ class Window extends React.Component {
 
     this.onChange = this.onChange.bind(this)
     this.onAppStateChange = this.onAppStateChange.bind(this)
+    this.updateComputedStyles = throttle(this.updateComputedStyles.bind(this), 66)
```
bridiver: I think we can get rid of this and rely on the existing `windowUpdated` debounce

js/actions/appActions.js
```diff
@@ -1411,6 +1411,14 @@ const appActions = {
     dispatch({
       actionType: appConstants.APP_UPDATE_LOG_OPENED
     })
+  },
+
+  updateComputedStyles: function (windowId, values) {
```
bridiver: I think this should actually be in windowActions. We are generally moving away from windowActions and appActions, but we will keep windowActions for actions that are related to the window component. It can still be handled in the browser process

bridiver: also should be declarative `updateComputedStyles` -> `computedStylesChanged` or something similar

app/common/state/windowState.js
```diff
@@ -148,6 +148,13 @@ const api = {
     return state.set('windows', windows.delete(index).insert(index, windowValue))
   },
 
+  updateComputedStyles: (state, action) => {
```
bridiver: I think `setComputedStyles` and we are moving away from passing actions into state helpers so I think this should take `windowId` and `computedStyles`

app/common/lib/windowsUtil.js
```diff
@@ -3,11 +3,47 @@
...
+    computedStyles[group].forEach(prop => {
+      const value = `--${group}-${prop}`
+      result[value] = root.getPropertyValue(value).trim()
+    })
+  })
+
+  return result
+}
+
+const getComputedProperty = (state, property) => {
```
bridiver: I think this should be replaced by a `getComputedStyles` method in `windowState`. We don't want anything outside of windowState to be tightly coupled to the storage path

bridiver:
I like the general idea, just needs some tweaks

app/renderer/components/window.js
NejcZdovc: Moved into `window.js`

app/common/state/windowState.js
NejcZdovc: Done

app/common/lib/windowsUtil.js
NejcZdovc: Moved

js/actions/appActions.js
NejcZdovc: Done

app/renderer/components/window.js
NejcZdovc: Done

app/browser/windows.js
```diff
@@ -55,9 +55,11 @@ const getWindowValue = (windowId) => {
 }
 
 const updateWindow = (windowId) => {
-  let windowValue = getWindowValue(windowId)
-  if (windowValue) {
+  const win = BrowserWindow.fromId(windowId)
+  if (win) {
+    const windowValue = getWindowValue(windowId)
     appActions.windowUpdated(windowValue)
+    updateComputedStyles(win)
```
NejcZdovc: The only problem with calling this function from this file is that we don't have document, so you can't do `document.querySelector`. Any idea how to fix this? @bridiver

NejcZdovc:
@bridiver can you please checkout my review comment

app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -118,19 +119,16 @@ class BookmarksToolbar extends ImmutableComponent {
...
-      this.toolbarPadding = Number.parseInt(this.root.getPropertyValue('--bookmarks-toolbar-padding'), 10)
-      this.bookmarkItemMargin = Number.parseInt(this.root.getPropertyValue('--bookmark-item-margin'), 10) * 2
-      // No margin for show only favicons
-      this.chevronMargin = Number.parseInt(this.root.getPropertyValue('--bookmark-item-chevron-margin'), 10)
-      this.fontSize = this.root.getPropertyValue('--bookmark-item-font-size')
-      this.fontFamily = this.root.getPropertyValue('--default-font-family')
-      this.chevronWidth = this.chevronMargin + Number.parseInt(this.fontSize)
-    }
+    this.maxWidth = windowsUtil.getCSSElementProperty('bookmark-item-max-width')
+    this.padding = windowsUtil.getCSSElementProperty('--bookmark-item-padding') * 2
```
bridiver: these shouldn't have dashes, right?

NejcZdovc: You are correct. Fixed

app/common/constants/computedStyles.js
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const computedStyles = {
```
bridiver: should this go under `renderer` now instead of `common`?

bridiver: `computed` also probably doesn't make sense anymore since these are static values

app/common/lib/windowsUtil.js
```diff
@@ -4,10 +4,22 @@
...
+
+const getPinnedSiteProps = site => {
   return Immutable.fromJS({
     location: site.get('location'),
     order: site.get('order'),
     partitionNumber: site.get('partitionNumber') || 0
   })
 }
+
+const getCSSElementProperty = (prop) => {
```
bridiver: maybe `renderer/styleUtil` ? `getStyleValue` ?

app/common/constants/computedStyles.js
bridiver: oh, almost forgot. We should comment this file and the less files to make sure they don't get out-of-sync

bridiver: don't want anyone to think they can only update one or the other

NejcZdovc: I would leave it here, because we have all constants files here and this are constants

NejcZdovc: Comment added

bridiver:
please add comments to `styleValues.js` and relevant `less` files in a follow-up. Thanks!

PR #9177, opened 2017-05-31
Add style standard to style doc

### LINKED ISSUE #8340: Discussion: CSS methodologies (OOCSS/BEM/SMACSS/...) in creating/refactoring with Aphrodite

cezaraugusto:
Auditors: @luixxiul
Close #8340

We have standardjs to take care of coding practices but we're still loose about how we deal with styles. As codebase grows it gets harder to follow a convention. This update adds new coding practices while writing styles and enforce using our adapted BEM pattern.

### DIFF
--- a/docs/style.md
+++ b/docs/style.md
@@ -25,10 +25,13 @@ Also, note that this style guide was re-made after some code has being refactore
     * [Styles specific for a given component](#styles-specific-for-a-given-component)
   * [Naming styles](#naming-styles)
     * [Naming styles objects](#naming-styles-objects)
+    * [How we name styles](#how-we-name-styles)
+    * [Defining our blocks, elements, and modifiers](#defining-our-blocks-elements-and-modifiers)
     * [Naming styles themselves](#naming-styles-themselves)
   * [Styles are the last thing](#styles-are-the-last-thing)
   * [Important is evil](#important-is-evil)
 * [Best practices](#best-practices)
+  * [Always add space between styles](#always-add-space-between-styles)
   * [Dealing with pseudo-states](#dealing-with-pseudo-states)
   * [Dealing with media-queries](#dealing-with-media-queries)
   * [Dealing with keyframes](#dealing-with-keyframes)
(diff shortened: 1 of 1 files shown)


NejcZdovc:
@cezaraugusto can we put this into linter? eslint custom rule maybe?

cezaraugusto:
I was thinking about that yesterday, a custom eslint rule would be ideal. We can take `StyleSheet.create` as the identifier and work from there looking for spaces/quotes and other rules we want to check.

Didn't have time to look closer but worth considering, so ++. I created an issue to track this #9181

luixxiul:
I would add the reason why we are going to choose BEM over other methodologies (1. following BEM should avoid visual regressions nicely, 2. BEM makes it possible for us to make preparations for style nesting in advance, though it is not sure if it could be implemented to `Aphrodite`).

Also I found a typo on the file. I'm going to fix it and push the fix.

luixxiul:
This is the fix: https://github.com/luixxiul/browser-laptop/commit/0bfcb6ad7e1156b2e71b35bc2cdd4bbffcdcff93

please cherry-pick and squash it, thanks.

cezaraugusto:
thanks, added some reasons on why we choose BEM, feel free to edit if you have more/better ones

PR #9261, opened 2017-06-05
CodeCov support

NejcZdovc:
Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ notifications:
 env:
   - CXX=g++-4.8 NODE_ENV=test TEST_DIR=lint
   - CXX=g++-4.8 NODE_ENV=test TEST_DIR=unit
+  - CXX=g++-4.8 NODE_ENV=test TEST_DIR=codecov
   - CXX=g++-4.8 NODE_ENV=test TEST_DIR=about
   - CXX=g++-4.8 NODE_ENV=test TEST_DIR=app
   - CXX=g++-4.8 NODE_ENV=test TEST_DIR=bookmark-components
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 [![Misc Tests](https://badges.herokuapp.com/travis/brave/browser-laptop?env=TEST_DIR=misc-components&label=misc-components-tests)](https://travis-ci.org/brave/browser-laptop)
 [![Navbar Tests](https://badges.herokuapp.com/travis/brave/browser-laptop?env=TEST_DIR=navbar-components&label=navbar-components-tests)](https://travis-ci.org/brave/browser-laptop)
 [![Tab Tests](https://badges.herokuapp.com/travis/brave/browser-laptop?env=TEST_DIR=tab-components&label=tab-components-tests)](https://travis-ci.org/brave/browser-laptop)
+[![codecov.io](https://codecov.io/github/brave/browser-laptop/coverage.svg?branch=master)](https://codecov.io/gh/brave/browser-laptop?branch=master)
 
 # Brave Browser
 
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -6,10 +6,15 @@ sudo: required
...
   - sh -e /etc/init.d/xvfb start
+install:
+  - npm install -g codecov
+  - npm install
 before_script:
   - npm run download-sync-client
 script:
   - npm run testsuite
+after_script:
+  - npm run unittest-cov && codecov
```
diracdeltas: is this supposed to be run after every test in the suite or only after the entire suite is done? right now it seems to be running after every test suite.

NejcZdovc: Yeah this is the last cycle in build lifecycle, so this should be executed when scripts are finished.

.travis.yml
```diff
@@ -6,10 +6,16 @@ sudo: required
...
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+install:
+  - npm install -g codecov
+  - npm install
 before_script:
   - npm run download-sync-client
 script:
   - npm run testsuite
+after_success:
```
diracdeltas: i don't think this is what you want, since it will run the coverage scripts once for every successful test in the suite. if all else fails you can add a new TEST_DIR value for it and add it as the last "test" in the testsuite, similar to what is done for TEST_DIR=lint

diracdeltas:
this approach looks good to me but the codecov job is failing: https://travis-ci.org/brave/browser-laptop/jobs/241067943

NejcZdovc:
Yeah, still working on it with codecov support

stevepeak:
I'm still seeing the bash uploader used here: https://travis-ci.org/brave/browser-laptop/jobs/241067943#L2980

1. The parentheses appear to be the issue therefore I suggest using the node uploader only.
2. Your original setup was installing the node uploader then using bash, this was wrong. What you can do is install the node uploader then call it. Or just call the bash uploader 👍

NejcZdovc:
@stevepeak It took me a while to get back on this task, but now it's finally working. Thank you for your inputs :+1:

NejcZdovc:
@stevepeak I do have one more question. Right now report will be added only if the CI is passing. We still have some intermittent tests, so travis is currently failing. Nevertheless we would still like to see codecov report in the checks section. I notice this sections in your [docs](https://docs.codecov.io/docs/merging-reports#section-why-did-i-not-receive-a-notification-yet-), which I also [added to this PR](https://github.com/brave/browser-laptop/pull/9261/files#diff-8bc9db6adba7208e25c96be322351063R4), but report is not added to the section. I think that this flag will completely ignore travis hook now.

Any suggestions how to solve this one?

stevepeak:
@NejcZdovc (sorry for delay)

You can customize how Codecov behaves:

```yaml
codecov:
  notify:
    # after_n_builds: 1  # notify after N builds (default null)
    # wait_for_ci: yes  # disable waiting for CI (default: yes)
    require_ci_to_pass: no  # require CI to pass
```

NejcZdovc:
Ready for review, now everything is working. Thank you again @stevepeak

diracdeltas:
1. do we need codecov jobs for both patch and project, or just project?
2. should this only run for PRs? looks like you can specify `only_pulls`

NejcZdovc:
1) I would keep both, so that you know how your PR is doing separately
2) we can run it only for [the master branch](https://github.com/codecov/support/wiki/Filtering-Branches) and add a tag to readme

wdyt?

diracdeltas:
sgtm!

PR #9275, opened 2017-06-05
allow unloading of sync extension

bridiver:
don’t use process.on/emit internally

Just something I came across while making some other changes

Test Plan:
Enable sync from a fresh profile

Reviewer Checklist:

Tests

### DIFF
--- a/app/sync.js
+++ b/app/sync.js
@@ -9,10 +9,12 @@ const electron = require('electron')
 const qr = require('qr-image')
 const ipcMain = electron.ipcMain
 const locale = require('./locale')
-const messages = require('../js/constants/sync/messages')
+const messages = require('../js/constants/messages')
+const syncMessages = require('../js/constants/sync/messages')
 const categories = require('../js/constants/sync/proto').categories
 const writeActions = require('../js/constants/sync/proto').actions
 const config = require('../js/constants/appConfig').sync
+const syncExtensionId = require('../js/constants/config').syncExtensionId
 const appActions = require('../js/actions/appActions')
 const syncConstants = require('../js/constants/syncConstants')
 const appDispatcher = require('../js/dispatcher/appDispatcher')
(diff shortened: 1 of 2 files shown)


diracdeltas:
i think this breaks some of the sync tests (`npm run test -- --grep='Syncing'`) https://travis-ci.org/brave/browser-laptop/jobs/239762817

also i'm unable to setup sync from a fresh profile in about:preferences#sync

app/sync.js
```diff
@@ -307,20 +309,19 @@ module.exports.onSyncReady = (isFirstRun, e) => {
...
   const reset = () => {
     log('Resetting browser local sync state.')
     appActions.changeSetting(settings.SYNC_ENABLED, false)
     appActions.changeSetting(settings.SYNC_DEVICE_NAME, undefined)
     appActions.resetSyncData()
   }
   // sent by about:preferences when sync should be reloaded
-  ipcMain.on(RELOAD_MESSAGE, () => {
-    process.emit(RELOAD_MESSAGE)
+  ipcMain.on(messages.RELOAD_SYNC_EXTENSION, () => {
```
diracdeltas: `messages` here is `js/constants/sync/messages`, but it should actually be `js/constants/messages`

app/sync.js
```diff
@@ -307,20 +309,19 @@ module.exports.onSyncReady = (isFirstRun, e) => {
...
+    console.log('reloading sync')
+    extensions.reloadExtension(syncExtensionId)
   })
   // sent by about:preferences when resetting sync
-  ipcMain.on(RESET_SYNC, (e) => {
+  ipcMain.on(messages.RESET_SYNC, (e) => {
```
diracdeltas: same as above

app/sync.js
bridiver: ah, that's probably the issue then

bridiver: I ran into something similar when trying to use config.syncExtensionId

bridiver: it would be helpful if we could use `dispatch-action` from the sync extension instead of using ipc handlers. The ipc handlers will be problematic when we start converting things to worker threads

bridiver:
updated

bridiver:
wanted to check on window.open test failures before merging, but that was an unrelated issue

Issue #8541, opened 2017-04-27
Application slowness (when opening new tabs, loading a website)


bsclifton:
Reported from one of our users [via Twitter](https://twitter.com/fafneri/status/857645991012564992):

> hey. I've been having some problems on loading new tabs/opening websites, they take so long. Is there anything I can do?

> when I try to open something through the new tab page, it takes ~15s. when I try to open a website, ~10.

luixxiul:
After opening the browser several hours I found the tab was not smoothly changed. It took a couple of seconds to change the tab by clicking.

![lag](https://cloud.githubusercontent.com/assets/3362943/25710594/a6436542-3127-11e7-8f51-5f8ee7277a7e.gif)

ashkitten:
Same here, and the entire browser freezes while loading a webpage.

Southern-Discomfort:
Exactly my problem, the browser is extremely laggy on Windows 10 x64. It is unusable at the moment. It literally takes 3 or 4 seconds to open a new tab. Crashes are common. Something has happened...I get it is only beta.

pjv:
Until recent updates i couldn't use brave (Macos Sierra, on a very beefy new iMac) because it was too slow to tolerate. A couple days ago I downloaded the latest release:

![2017-06-02 at 10 17 am](https://cloud.githubusercontent.com/assets/327716/26732308/cb32163a-477c-11e7-8bbe-81b6f05f4f44.png)

And wonder of wonders, you guys have improved things pretty dramatically. It's actually tolerable (though still sluggish) to launch, and quite zippy opening tabs and loading pages.

But... overall moving around in the app it still has a sluggish feeling to it and closing a tab is still completely absurd. I hit command-W or click the X to close a tab and I count slowly:

- one one thousand
- two one thousand
- three one thousand

... and somewhere around three one thou the tab finally closes.

It's fine if you don't have things fully optimized. Extended beta's are great. But I think you guys should stop calling Brave "faster" than the alternatives because since (as of now) it actually feels materially slower than the alternatives in many ways, people who download it and run it and experience it as slow are going to just assume that nothing you say is credible.

I understand the principle on which you are claiming speed and I am not arguing with that. Not downloading all the crap that Brave blocks will certainly speed things up. But if the app itself performs so slowly that downloading junk is not the bottleneck in perceived snappiness in a user's experience, then marketing the browser as "fast" is going to sound untrustable.

[...]

PR #9279, opened 2017-06-06
In siteUtil.updateSiteFavicon use location cache

### LINKED ISSUE #9276: Every URL navigation takes 1000+ ms in updateSiteFavicon (With 10K bookmarks)
Yes

Navigating to URLs and clicking links is very slow with 10K bookmarks.

MacOS

e14a9dbc7806a44913f51117e33ed59aac7cbf89

- Steps to reproduce:
    1. Instrument `siteUtil.updateSiteFavicon()` to measure its runtime.
    2. Load 10K bookmarks.
    3. Go to https://github.com/brave/browser-laptop and see how long updateSiteFavicon takes.
    4. Click the H1 link "browser-laptop" to navigate to the same page and see how long updateSiteFavicon takes.
    
- Actual result:
First navigation takes 1000 ms.
Clicking the link to itself takes 1000 * 2 ms.

- Expected result:
< 100 ms

Yes; with 10K bookmarks

Yes

Yes

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues:
https://github.com/brave/browser-laptop/issues/7162

ayumi:
Fix #9276

Note: This slightly changes the function behavior because it removes use of [normalizeUrl](https://github.com/sindresorhus/normalize-url) when comparing locations. By default the library removes `www` from URLs:
```
'https://www.example.com/about.html#test'
=> 'https://example.com/about.html#test'
```

We had been testing this case, assuming that we had both `https://www.brave.com/` and `https://brave.com`, calling updateSiteFavicon with `https://brave.com` and expected both sites to be updated.

I don't think this case actually happens (please correct me if I'm wrong), and the siteKey cache does normalize location when [setting](https://github.com/brave/browser-laptop/blob/master/app/common/state/siteCache.js#L52) and [getting](https://github.com/brave/browser-laptop/blob/master/app/common/state/siteCache.js#L37) so it seems safe.

Auditors: @bsclifton @bbondy

Test Plan:
1. load 10K bookmarks
1. Enable bookmarks bar favicons via Preferences > General > "Bookmarks Bar" => "Text and Favicons".
2. Create a bookmark folder.
3. Navigate to https://archive.org
4. Bookmark page.
5. Navigate to https://wikipedia.org
6. Bookmark page into the folder.
7. Examine bookmarks toolbar; there should be favicons for archive and wikipedia.
8. Open the Bookmarks Manager and confirm there are favicons.

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

Related: #8541

### DIFF
--- a/js/state/siteUtil.js
+++ b/js/state/siteUtil.js
@@ -3,7 +3,6 @@
 
 'use strict'
 const Immutable = require('immutable')
-const normalizeUrl = require('normalize-url')
 const siteCache = require('../../app/common/state/siteCache')
 const siteTags = require('../constants/siteTags')
 const settings = require('../constants/settings')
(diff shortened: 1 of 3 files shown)


luixxiul:
@ayumi just a head-up: could I know how this issue should solve #8541? I couldn't tell based on the STR in the 1st post.

ayumi:
@luixxiul I think STR would be to first load 10K bookmarks, then navigate some websites. The STR for https://github.com/brave/browser-laptop/issues/9276 is related.

PR #9282, opened 2017-06-06
Increase urlbarForm input height to the equal value as 'urlbarFormHeight'

### LINKED ISSUE #4922: double click in addressbar maximizes/restores the window instead of selecting the address
**Did you search for similar issues before submitting this one?** Yes

**Describe the issue you encountered:** Sometimes when I place the mouse pointer in the bottom part of the addressbar and double-click to select the address, rather than selecting the address, the window is maximized/restored. This is because the pointer tip can be several pixels into the bottom of the addressbar and still targeting the toolbar below the addressbar. 

**Expected behavior:** When the pointer is visibly inside the addressbar and clicked it should focus the addressbar or turn into an i-beam cursor for text selection. 
- Brave Version: 0.12.5
- Steps to reproduce:
  1. with the window maximized move the mouse pointer into the bottom 20% or so of the addressbar
  2. double click
  3. see the window restored

luixxiul:
Also:
- Set white background to .urlbarForm inside navigator
- Remove redundant specifications (the specified background is inherited to all of the elements inside .urlbarForm)
- Remove duplicate variables

Fixes #4922
Addresses #9283

Auditors:

Test Plan:
1. Make sure that clicking the area between the top and bottom edge of the URL bar selects the URL address

Submitter Checklist:

Reviewer Checklist:

Tests

Assigning @cezaraugusto & @bsclifton for review - it would be appreciated if one of you could check if the fix should work as expected on Windows.

### DIFF
--- a/less/navigationBar.less
+++ b/less/navigationBar.less
@@ -642,7 +642,6 @@
   justify-content: center;
   height: @urlbarFormHeight;
   padding: 0 10px 0 3px;
-  background: @navigationBarBackgroundActive;
   display: flex;
   flex-grow: 1;
   min-width: 0%; // allow the navigator to shrink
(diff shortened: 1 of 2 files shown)


luixxiul:
I know that making one tiny change on `navigationBar.less` (which is messy to be honest) could lead to visual regressions, while this case I think it would not.

I will start working on that file soon, after browserButton is mostly refactored.

PR #9414, opened 2017-06-13
Refactors TabPage and TabPages into redux

### LINKED ISSUE #9339: Redux component - Tab page
## Test plan
See https://github.com/brave/browser-laptop/pull/9414#issue-235456487

## Original issue description
Convert `TabPage` and `TabPages` into redux component.

NejcZdovc:
## Test Plan
- try to open enough tabs to trigger tab page (default is 20)
- see if audio indicator is displayed if you have video playing in tab page
- check if you can close tab page (note about #9420)
- check if you can mute/unmute tab page

## Description
Submitter Checklist:

Resolves #9339

Reviewer Checklist:

### DIFF
--- a/app/renderer/components/tabs/tabPage.js
+++ b/app/renderer/components/tabs/tabPage.js
@@ -3,9 +3,10 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const React = require('react')
+const Immutable = require('immutable')
 
 // Components
-const ImmutableComponent = require('../immutableComponent')
+const ReduxComponent = require('../reduxComponent')
 
 // Actions
 const appActions = require('../../../../js/actions/appActions')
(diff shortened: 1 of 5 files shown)


app/renderer/components/tabs/tabPage.js
```diff
@@ -44,21 +50,20 @@ class TabPage extends ImmutableComponent {
...
-    const moveToFrame = this.props.tabPageFrames.get(0)
     const sourceDragData = dndData.getDragData(e.dataTransfer, dragTypes.TAB)
     const sourceDragFromPageIndex = this.props.sourceDragFromPageIndex
     // This must be executed async because the state change that this causes
     // will cause the onDragEnd to never run
     setTimeout(() => {
       // If we're moving to a right page, then we're already shifting everything to the left by one, so we want
       // to drop it on the right.
-      windowActions.moveTab(sourceDragData.get('key'), moveToFrame.get('key'),
+      windowActions.moveTab(sourceDragData.get('key'), this.props.moveToFrameKey,
```
bsclifton: Should we still be using windowActions / frameKey? I think it would make more sense to use appAction / tabId

bsclifton: Although, that change might be quite a bit larger. I'd be OK with us doing that later 😄

app/renderer/components/tabs/tabPages.js
```diff
@@ -3,39 +3,43 @@
...
+const {getSetting} = require('../../../../js/settings')
+
+class TabPages extends React.Component {
+  mergeProps (state, ownProps) {
+    const currentWindow = state.get('currentWindow')
+    const frames = frameStateUtil.getNonPinnedFrames(currentWindow) || Immutable.List()
+    const tabsPerPage = Number(getSetting(settings.TABS_PER_PAGE))
+
+    const props = {}
+    props.tabPageCount = Math.ceil(frames.size / tabsPerPage)
```
bsclifton: can this but put into a property? that way we can unit test it 😄

NejcZdovc: You mean into function?

app/renderer/components/tabs/tabPage.js
NejcZdovc: Yeah when I am doing refactors I try to keep it as close as possible, because this way we are less error prone and we will have a few more iterations of redux components anyway when we will start moving into apps state (state helpers)

app/renderer/components/tabs/tabPages.js
bsclifton: @NejcZdovc yeah- basically the ES6 classes have "properties" when they're exposed using a getter.  Like:
```javascript
class TabPage extends React.Component {
  //..
  get tabPageCount () {
    return Math.ceil(this.frames.size / this.tabsPerPage)
  }
  //..
```

versus a typical method or function
```javascript
class TabPage extends React.Component {
  //..
  tabPageCount () {}
}

function tabPageCount() {}
```

bsclifton: Getters are very easy to test using Enzyme 😄

NejcZdovc: Problem is that you can't use getter in mergeProps, because you don't have access to the component. So what we can do is create helper in util file

bsclifton: ahh that's right! I had forgotten 😄  I saw you use getters in https://github.com/brave/browser-laptop/pull/9413, but I now see those are in `render`.  Util method would be great 😄

NejcZdovc: Yes you can use getters for html content so that renderer is not that big, but one thing that you need to be carful is that you always trigger re-render when needed. That means that you shouldn't do any calculation inside the getter that can change getter return.

NejcZdovc: Moved into util file and added unit tests

PR #9457, opened 2017-06-14
Converts ClearBrowsingDataPanel into redux component

### LINKED ISSUE #9440: Redux component - ClearBrowsingDataPanel
## Test plan

https://github.com/brave/browser-laptop/pull/9457#issue-235903786

----

Convert `ClearBrowsingDataPanel` into redux component.

NejcZdovc:
Submitter Checklist:

Resolves #9440

Auditors: @bridiver @bsclifton

Test Plan:
- check if you can clear history
- check if toggles are preserved as you click clear
- toggle one switch, click cancel and check that toggle was not preserved when you open it next time
- check if ledger is still working correctly: ledger initi, ledger backup, ledger recovery

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/menu.js
+++ b/app/browser/menu.js
@@ -693,11 +693,11 @@ const doAction = (action) => {
       }
       break
     case appConstants.APP_ON_CLEAR_BROWSING_DATA:
-      if (action.clearDataDetail.get('browserHistory')) {
-        appDispatcher.waitFor([appStore.dispatchToken], () => {
+      appDispatcher.waitFor([appStore.dispatchToken], () => {
+        if (appStore.getState().getIn(['clearBrowsingDataDefaults', 'browserHistory'])) {
           createMenu()
-        })
-      }
+        }
+      })
       break
     case windowConstants.WINDOW_CLICK_MENUBAR_SUBMENU:
       appDispatcher.waitFor([appStore.dispatchToken], () => {
--- a/app/ledger.js
+++ b/app/ledger.js
@@ -56,7 +56,6 @@ const uuid = require('uuid')
 const appActions = require('../js/actions/appActions')
 const appConfig = require('../js/constants/appConfig')
 const appConstants = require('../js/constants/appConstants')
-const appDispatcher = require('../js/dispatcher/appDispatcher')
 const messages = require('../js/constants/messages')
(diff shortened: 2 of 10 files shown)


app/ledger.js
```diff
@@ -168,7 +168,8 @@ const doAction = (action) => {
       break
 
     case appConstants.APP_ON_CLEAR_BROWSING_DATA:
-      if (action.clearDataDetail.get('browserHistory') && !getSetting(settings.PAYMENTS_ENABLED)) reset(true)
+      // TODO
```
NejcZdovc: @bsclifton @bridiver I would need your help here. What we need to add here is state (now we only get action) so that you can check it the same way we do it for sites. How should we address this problem?

NejcZdovc:
Except this ledger problem everything else is done.

app/ledger.js
NejcZdovc: I was thinking to add state directly in same as we do with other reducers. One thing that we can also do is get state only in this case with `getState`

NejcZdovc:
Don't squash commits, because second one is quite special and I would keep it separated.

app/browser/menu.js
```diff
@@ -693,11 +693,11 @@ const doAction = (action) => {
       }
       break
     case appConstants.APP_ON_CLEAR_BROWSING_DATA:
-      if (action.clearDataDetail.get('browserHistory')) {
-        appDispatcher.waitFor([appStore.dispatchToken], () => {
+      appDispatcher.waitFor([appStore.dispatchToken], () => {
+        if (appStore.getState().getIn(['clearBrowsingDataDefaults', 'browserHistory'])) {
```
bridiver: why the change here from action.clearDataDetail to checking the state?

NejcZdovc: Because we are not sending anything in the action anymore. We have everything that we need directly in the store already. We are saving data here already https://github.com/brave/browser-laptop/pull/9457/files#diff-23ca389e2bcb77191b5a9c10900eb3a3R697

NejcZdovc: actions has no params https://github.com/brave/browser-laptop/pull/9457/files#diff-5fea3af223318beea03d50c8ee3de64dR591

PR #9475, opened 2017-06-14
dont’t call to get site settings outside of mergeProps

bridiver:
Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/frame/frame.js
+++ b/app/renderer/components/frame/frame.js
@@ -91,15 +91,6 @@ class Frame extends React.Component {
     windowActions.closeFrame(this.props.frameKey)
   }
 
-  getFrameBraverySettings (props) {
-    props = props || this.props
-    const frameSiteSettings =
-      siteSettings.getSiteSettingsForURL(props.allSiteSettings, props.location)
-    return Immutable.fromJS(siteSettings.activeSettings(frameSiteSettings,
-                                                        appStoreRenderer.state,
-                                                        appConfig))
-  }
-
   isAboutPage () {
     return aboutUrls.get(getBaseUrl(this.props.location))
   }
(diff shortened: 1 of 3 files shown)


diracdeltas:
flash allow-once doesn't seem to work.

1. go to http://homestarrunner.com/
2. right click to allow flash once
3. close the tab then go back to http://homestarrunner.com/
4. flash is still allowed

NejcZdovc:
@bridiver can you please rebase? Is scenario by @diracdeltas fixed?

bridiver:
@NejcZdovc @diracdeltas updated to fix that issue

app/renderer/components/frame/frame.js
```diff
@@ -112,58 +103,44 @@ class Frame extends React.Component {
...
-      appActions.removeSiteSetting(origin, 'noScript', this.props.isPrivate)
+    if (props.noScript === 0) {
+      appActions.removeSiteSetting(props.origin, 'noScript', props.isPrivate)
     }
-    const noScriptExceptions = activeSiteSettings.get('noScriptExceptions')
-    if (noScriptExceptions) {
-      appActions.noScriptExceptionsAdded(origin,
-        noScriptExceptions.map(value => value === 0 ? false : value))
+    if (props.noScriptExceptions) {
+      appActions.noScriptExceptionsAdded(props.origin, this.props.noScriptExceptions.filter((value, host) => value !== 0 ? false : value))
```
diracdeltas: i think this line is wrong - should be the same as the original

diracdeltas: also i think it should be props.noScriptExceptions instead of this.props.noScriptExceptions

bridiver: probably doesn't work because of that. I'll take a look

app/renderer/components/frame/frame.js
```diff
@@ -112,58 +103,44 @@ class Frame extends React.Component {
...
-      appActions.removeSiteSetting(origin, 'noScript', this.props.isPrivate)
+    if (props.noScript === 0) {
+      appActions.removeSiteSetting(props.origin, 'noScript', props.isPrivate)
     }
-    const noScriptExceptions = activeSiteSettings.get('noScriptExceptions')
-    if (noScriptExceptions) {
-      appActions.noScriptExceptionsAdded(origin,
-        noScriptExceptions.map(value => value === 0 ? false : value))
+    if (props.noScriptExceptions) {
+      appActions.noScriptExceptionsAdded(props.origin, props.noScriptExceptions.filter((value, host) => value !== 0 ? false : value))
```
diracdeltas: this should still be `this.noScriptExceptions.map(value => value === 0 ? false : value))`

bridiver:
should be good now

diracdeltas:
@bridiver did you see my comment above ^

bridiver:
no, looks like I mixed it up on the rebase

bridiver:
updated

app/renderer/components/frame/frame.js
```diff
@@ -112,58 +103,44 @@ class Frame extends React.Component {
...
-      appActions.removeSiteSetting(origin, 'noScript', this.props.isPrivate)
+    if (props.noScript === 0) {
+      appActions.removeSiteSetting(props.origin, 'noScript', props.isPrivate)
     }
-    const noScriptExceptions = activeSiteSettings.get('noScriptExceptions')
-    if (noScriptExceptions) {
-      appActions.noScriptExceptionsAdded(origin,
-        noScriptExceptions.map(value => value === 0 ? false : value))
+    if (props.noScriptExceptions) {
+      appActions.noScriptExceptionsAdded(props.origin, props.noScriptExceptions.filter((value, host) => value === 0 ? false : value))
```
diracdeltas: this is still the incorrect version, it should be the version above using `map` not `filter`

bridiver:
ok, this time I copy/pasted it so it better be right :)

bridiver:
test added for `getVisibleOrigin`, but otherwise it's a refactoring so should be covered by existing tests

Issue #9017, opened 2017-05-23
IME does not work properly on input form


luixxiul:
## Test plan

1. Enable IME (Japanese/Chinese/Korean/...) on your system
2. Open about:autofill
3. Select 'Add address'
4. Hit some keys
5. Make sure you are seeing the pulldown
6. Hit <kbd>Tab</kbd>
7. Make sure a conversion candidate is selected
8. Hit <kbd>Enter</kbd>
9. Make sure the candidate is decided

----

**Describe the issue you encountered:** IME does not work properly on input form. You cannot select a conversion candidate. 

- Platform (Win7, 8, 10? macOS? Linux distro?): macOS

- Brave Version (revision SHA): https://github.com/brave/browser-laptop/compare/v0.15.303dev...v0.15.304dev

- Steps to reproduce:
    1. Enable Japanese lang input system
    2. Open about:autofill
    3. Select 'Add address'
    4. Hit some keys
- Actual result:

   ![autofill](https://cloud.githubusercontent.com/assets/3362943/26372992/d32887a0-403a-11e7-82fd-7568ecc15575.gif)
  The form is not properly styled as I tried that by reverting it to https://github.com/brave/browser-laptop/blob/c719806d2265434b7e74fce0ef9a7bd061ddcc64/js/components/autofillCreditCardPanel.js to make sure CommonForm is not causing the issue.

  You could confirm the issue on Bookmark Hanger, which appears by clicking the star icon.
  Yet the issue cannot be reproduced on about:preferences#sync.

  ![autofill3](https://cloud.githubusercontent.com/assets/3362943/26373538/d9d8ece6-403c-11e7-8335-bd032d7ea848.gif)

- Expected result: on 0.15.2

  ![autofill2](https://cloud.githubusercontent.com/assets/3362943/26373228/a9d63644-403b-11e7-9285-74753cc846cf.gif)

- Is this an issue in the currently released version?
  No.

- Extra QA steps:
    1.
    2.
    3.

- Any related issues: #

bsclifton:
@luixxiul is this on master? or the last release? because they are very different at the moment.  If this is on master only, please remove from this milestone.  Thanks 😄

darkdh:
confirm it occurs on autofill and bookmark dialog of 0.15.307

luixxiul:
@bsclifton I'm trying to find out since which release it is broken.

luixxiul:
[...]

PR #9490, opened 2017-06-15
Converts AutofillAddressPanel into redux component

### LINKED ISSUE #9444: Redux component - AutofillAddressPanel
## Test plan

https://github.com/brave/browser-laptop/pull/9490#issue-236161281

----

Convert `AutofillAddressPanel` into redux component.

NejcZdovc:
Submitter Checklist:

Resolves #9444

Auditors: @bsclifton @bridiver @darkdh 

Test Plan:
- try to add autofill address
- try to delete it
- try to edit it

Test Plan (l10n):
1. Install Google IME (CJK);
2. Enable it (Hiragana `ひらがな` mode)
3. Open `about:autofill`
4. Input `brave`
5. It should display the conversion candidate like this:
  screenshot 2017-06-20 0 51 32

Reviewer Checklist:

Tests

### DIFF
--- a/app/autofill.js
+++ b/app/autofill.js
@@ -13,18 +13,18 @@ module.exports.init = () => {
   })
 }
 
-module.exports.addAutofillAddress = (detail, guid) => {
+module.exports.addAutofillAddress = (detail) => {
   session.defaultSession.autofill.addProfile({
-    full_name: detail.name,
-    company_name: detail.organization,
-    street_address: detail.streetAddress,
-    city: detail.city,
-    state: detail.state,
-    postal_code: detail.postalCode,
-    country_code: detail.country,
-    phone: detail.phone,
-    email: detail.email,
-    guid: guid
+    full_name: detail.get('name'),
+    company_name: detail.get('organization'),
+    street_address: detail.get('streetAddress'),
+    city: detail.get('city'),
+    state: detail.get('state'),
+    postal_code: detail.get('postalCode'),
+    country_code: detail.get('country'),
+    phone: detail.get('phone'),
+    email: detail.get('email'),
+    guid: detail.get('guid')
   })
 }
 
--- a/app/renderer/components/autofill/autofillAddressPanel.js
+++ b/app/renderer/components/autofill/autofillAddressPanel.js
@@ -3,16 +3,13 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const React = require('react')
-const ImmutableComponent = require('../immutableComponent')
+const Immutable = require('immutable')
+const {css} = require('aphrodite/no-important')
+
+// Components
+const ReduxComponent = require('../reduxComponent')
 const Dialog = require('../common/dialog')
 const Button = require('../common/button')
-const windowActions = require('../../../../js/actions/windowActions')
-const appActions = require('../../../../js/actions/appActions')
-const KeyCodes = require('../../../common/constants/keyCodes')
-
-const {css} = require('aphrodite/no-important')
-const commonStyles = require('../styles/commonStyles')
-
 const {
   CommonFormLarge,
   CommonFormSection,
@@ -21,6 +18,15 @@ const {
   commonFormStyles
 } = require('../common/commonForm')
 
+// Actions
+const windowActions = require('../../../../js/actions/windowActions')
+const appActions = require('../../../../js/actions/appActions')
+
+// Constants
+const KeyCodes = require('../../../common/constants/keyCodes')
+
+// Styles
+const commonStyles = require('../styles/commonStyles')
 const commonForm = css(
   commonStyles.formControl,
   commonStyles.textbox,
@@ -29,9 +35,9 @@ const commonForm = css(
   commonFormStyles.input__box
 )
 
-class AutofillAddressPanel extends ImmutableComponent {
-  constructor () {
-    super()
+class AutofillAddressPanel extends React.Component {
+  constructor (props) {
+    super(props)
     this.onNameChange = this.onNameChange.bind(this)
     this.onOrganizationChange = this.onOrganizationChange.bind(this)
     this.onStreetAddressChange = this.onStreetAddressChange.bind(this)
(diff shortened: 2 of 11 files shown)


app/renderer/components/autofill/autofillAddressPanel.js
```diff
@@ -45,118 +51,118 @@ class AutofillAddressPanel extends ImmutableComponent {
...
         break
       case KeyCodes.ESC:
-        this.props.onHide()
+        this.onHide()
         break
     }
   }
+
   onSave () {
-    appActions.addAutofillAddress(this.props.currentDetail, this.props.originalDetail)
```
darkdh: why removing `this.props.currentDetail`? This will cause save not working

app/renderer/components/autofill/autofillAddressPanel.js
```diff
@@ -45,118 +51,118 @@ class AutofillAddressPanel extends ImmutableComponent {
...
+    windowActions.setAutofillAddressDetail()
+  }
+
+  mergeProps (state, ownProps) {
+    const currentWindow = state.get('currentWindow')
+    const currentDetail = currentWindow.getIn(['autofillAddressDetail', 'currentDetail'], Immutable.Map())
+    const originalDetail = currentWindow.getIn(['autofillAddressDetail', 'originalDetail'], Immutable.Map())
+
+    const props = {}
+    // Used in renderer
```
darkdh: you will also need to set 
```
props.currentDetail = currentDetail
props.originalDetail = originalDetail
```

NejcZdovc: we don't want to do that, let's discuss it more on the call

app/renderer/components/autofill/autofillAddressPanel.js
NejcZdovc: because this is this WIP and I need to check some things before finishing this PR

app/autofill.js
```diff
@@ -15,16 +15,16 @@ module.exports.init = () => {
 
 module.exports.addAutofillAddress = (detail, guid) => {
```
darkdh: `guid` arg here can be removed now

test/contents/autofillTest.js
```diff
@@ -152,8 +153,10 @@ describe('Autofill', function () {
         .keys(Brave.keys.END)
         .typeText(emailInput, 'mm', email)
         .click(saveAddressButton)
+        .windowByUrl(Brave.browserWindowUrl)
         .waitUntil(function () {
           return this.getAppState().then((val) => {
+            console.log(val.value.autofill.addresse)
```
darkdh: forgot to remove?

app/autofill.js
NejcZdovc: Removed

test/contents/autofillTest.js
NejcZdovc: removed

NejcZdovc:
> This change will introduce the problem of #9017

Not sure what you meant by that

darkdh:
just like this 
https://cloud.githubusercontent.com/assets/3362943/26372992/d32887a0-403a-11e7-82fd-7568ecc15575.gif
The Chinese IME I use will have composition state same as Japanese.
And this will cause I couldn't select the word I need
it should let user to select word like this
https://cloud.githubusercontent.com/assets/3362943/26373228/a9d63644-403b-11e7-9285-74753cc846cf.gif

NejcZdovc:
@darkdh if I understand correctly this PR will cause a regression for thing that was fixed in #9017, because I removed this part https://github.com/brave/browser-laptop/commit/4102e95eb87ad53c72b2b7eea2a2e6f423cb1391#diff-590c7f19ab835ddb769d5c3e5b227ab5R51 ?

darkdh:
yeah, but in this PR. Adding it back doesn't solve the issue because lots of things also changed

NejcZdovc:
Yeah we can't just add that part back, because we are connected to the store now and it will not work, so we need to find another approach of how to fix this.

NejcZdovc:
Will try to find a way how to fix it, added back label WIP

NejcZdovc:
@darkdh can you please provide STR how to trigger this problem, because I tried it on my machine and I can't reproduce it, autofill dialog is not auto-dismissing

NejcZdovc:
@luixxiul I saw that you tested out #9017, can you please checkout out this PR and see if you can reproduce this problem or not. I couldn't, but I am not sure that I did correct STR

darkdh:
@NejcZdovc you need an input method which requires selecting character for a certain symbol combination like Chinese or Japanese.
So the problem is it won't allow us to select the character. If we are in selecting stage, there will be underline on that character(s)

darkdh:
I use `Zhuyin` on Mac. and you can have`測試` by input `hk4g4` and you can see underline on these two characters and you should be able to change character by arrow left/right to select character you want to change and arrow down is to change the character

luixxiul:
Yes the issue gets back.

bbb

app/renderer/components/autofill/autofillAddressPanel.js
bridiver: what is the current status of this?

NejcZdovc: this is done, but we need to find a way to fix IME problem

NejcZdovc:
@luixxiul @darkdh thank you very much for STR. I finally managed to reproduce it. I also pushed the fix for it, so can you please try it again?

app/renderer/components/autofill/autofillAddressPanel.js
```diff
@@ -173,62 +161,99 @@ class AutofillAddressPanel extends ImmutableComponent {
...
+                  commonStyles.textbox,
+                  commonStyles.textbox__outlineable,
+                  commonFormStyles.input__box
+                )}
                 data-test-id='nameOnAddress'
-                spellCheck='false' onKeyDown={this.onKeyDown} onChange={this.onNameChange}
-                ref={(nameOnAddress) => { this.nameOnAddress = nameOnAddress }} />
+                spellCheck='false'
+                ref={(ref) => { this.nameOnAddress = ref }}
+                defaultValue={this.props.name}
```
darkdh: neat pick. The order is different than others

NejcZdovc: Done

luixxiul:
One slight issue on deleting the candidate.

STR:
1. Enable the Japanese IME
2. Input `brave`
3. Make sure you get `ｂらう゛ぇ`
4. Hit backspace 3 times
5. Make sure you only have `ｂ`
6. Hit that one more time to delete `ｂ`

Actual result: just conversion candidate pulldown disappears and `ｂ` is still there

Expected result: `ｂ` should be removed

@NejcZdovc would you mind solving this issue as it is a regression? thanks!

Issue #9306, opened 2017-06-07
Closing Tabs Results in White Page


jonathansampson:
## Test plan
See https://github.com/brave/browser-laptop/pull/9539

## Original issue description
- Did you search for similar issues before submitting this one?
Yes

- Describe the issue you encountered:
Upon closing 2 tabs, the user may be presented with a white page.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Desktop

- Brave Version (revision SHA):
https://github.com/brave/browser-laptop/commit/75ffa36130b1ad33c6f311e5e16ed7838b0b50a0

- Steps to reproduce:
    1. Open 7 tabs (7th tab selected)
    2. Click x on 6th tab (don't move mouse)
    3. Click x on 7th tab (should move under your mouse)
    
- Actual result:
Focused tab is white.

- Expected result:
Focused tab is visible to the user.

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
Yes

- Is this an issue in the currently released version?
Yes

- Can this issue be consistently reproduced?
Yes

![white-tab](https://user-images.githubusercontent.com/815158/26891366-9e849c02-4b69-11e7-8a79-8ff012152697.gif)

bridiver:
`may` means it is intermittent? Or can you reliably repro?

jonathansampson:
@bridiver I was able to reproduce it 100% of the time (as was @NejcZdovc I believe) when I carefully followed the steps above.

cezaraugusto:
so issue happens for any prefs config for "When closing an active tab" option

issue reported above (https://github.com/brave/browser-laptop/issues/9306#issuecomment-308908661) is different than the main issue. I created a new one for that #9536 with fix under #9537. /cc @jonathansampson 

Seems that issue happened after 039109635a173acddc326dce45916ef00d738409. For me, following the STR it is always falling in under this conditional https://github.com/brave/browser-laptop/commit/039109635a173acddc326dce45916ef00d738409#diff-585bc5df4db53a44a36b51f6bd4e0bc3R36. Given next frame is a user decision, I'm not sure what is expected in this case.

@bridiver do you mind jumping in to see what's going wrong? I created a failing test case in #9539.

bridiver:
thanks for the test case @cezaraugusto ! I have an idea of what the issue might be and should be able to track it down pretty quickly

bridiver:
Haven't checked the test run yet and probably needs a unit test update, but the new test passes

PR #8852, opened 2017-05-12
make tab previews based on mouse idle time

cezaraugusto:
# This PR changed its purpose, please see description

Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave/browser-laptop/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- [x] Ran `git rebase -i` to squash commits (if needed).

Close #8860

## Note for reviewers:

Previews are now happening based on mouse idle time. Which means that it will only be fired if you stay the mouse over a tab for a given amount of time -- which is defined in prefs->tabs.

Once the mouse is long enough over a tab and you wish to preview another tab, the preview is fired immediately. Once you leave the tab to another component, the event is canceled and you'll only be able to preview again if the mouse is idle again.

Test Plan:
### Intentional preview
1. Open three tabs
2. Hover over one inactive tab, let your mouse stopped for 2s (this is the default for now)
3. Tab should preview
4. Hover over third tab
5. Third tab should preview

### Unintentional preview/intentional preview
1. Open three tabs
2. Keep hovering from top/bottom, left/right
3. No tab should preview
4. Stop at a given tab
5. It should preview
6. Hover over something other than another tab
7. Hover over a tab again
8. It should not preview
9. Let your mouse stopped over a tab
10. It should preview

### Dealing with options
1. Go to prefs->tabs
2. Disable tab preview
3. Tab preview timing selection should be hidden
4. Enable again
5. Switch to "normal" (set to 1s) or "short" (set to 500ms)
6. Open three tabs
7. Hover over one inactive tab, let your mouse stopped for 2s (this is the default for now)
8. Tab should preview
9. Hover over third tab
10. Third tab should preview faster than previous STR

Reviewer Checklist:

Tests

- [ ] Adequate test coverage exists to prevent regressions
- [ ] Tests should be independent and work correctly when run indivually or as a suite [ref](https://github.com/brave/browser-laptop/wiki/Code-Guidelines#test-dependencies)

PR #9542, opened 2017-06-18
Issue 9306 related

bridiver:
Submitter Checklist:
@cezaraugusto I did some related changes for #9306 that turned out not to be necessary, but are still good to have and can go into 0.19. Still need to check tests, but wanted to get this out there to look at

Test Plan:

Reviewer Checklist:

Tests


app/browser/reducers/tabsReducer.js
```diff
@@ -67,6 +67,11 @@ const updateActiveTab = (state, closeTabId) => {
     }
   }
 
+  // fall back to first tab
+  if (nextTabId === tabState.TAB_ID_NONE) {
+    nextTabId = tabState.get('tabs').first()
+  }
+
   if (nextTabId !== tabState.TAB_ID_NONE) {
```
bsclifton: When I wrote the unit tests for this, I was unable to find a way to hit this condition.  When you originally pulled the logic out of frameStateUtil and moved here, you had a similar block.  Since it wasn't possible to hit, I had removed it.  Did you see a case where this runs?

bridiver:
This PR was changes I made that turned out not to be necessary for the fix and this is just leftover. It's actually not correct anyway because it should get the first tab in the window and that is already covered by "next" so this isn't needed

cezaraugusto:
@bridiver are we following the dumb/smart components convention where util files are made to host the logic, leaving the main component just for presentational purposes? 

I'm changing the way we preview tabs in #8852 and if this turns to be accepted first I'll need to re-work the logic

bridiver:
the logic should be in the reducers (which can use state/util files) and yes, the components should be as "dumb" as possible. Ideally action should be things like `tabClicked` or `tabMouseEnter` letting the reducer to decide what actually needs to be done in response

bridiver:
@cezaraugusto it looks like this overrides some recent changes you made to the component. Can you make sure this covers whatever changes you made?

cezaraugusto:
@bridiver yes, thanks for the reminder. I'll make sure to update your changes. 

#8852 will also change it so keeping an eye anyway on which goes first.

bridiver:
I think #8852 is actually moving in the wrong direction with `setPreviewMode`. "setter" actions should be deprecated and we don't want to call multiple actions together. `setPreviewMode` doesn't provide any new information that `setPreviewFrame` doesn't already provide. I think better to merge this first (assuming there are no issues with it) and then make other changes

bridiver:
there is an issue with this PR that I am fixing right now

bridiver:
updated @cezaraugusto 
I really think we should merge this one first because it's only going to get more complicated to merge the functionality if we keep adding things to the component

cezaraugusto:
@bridiver btw once we're good to go I'm going to accept this one before the land of #8852 as per your request. That PR will be included in 0.18.x per last weekly triage meeting, so I'm going to move this to the same milestone as well.

app/renderer/components/tabs/tabPage.js
```diff
@@ -32,23 +32,6 @@ class TabPage extends React.Component {
     this.onMouseLeave = this.onMouseLeave.bind(this)
```
cezaraugusto: I think those should be removed too

cezaraugusto: maybe remove it all given other handlers are not bound to the constructor but curious if that affects reduxComponent

bridiver: actually I'm surprised this doesn't cause an error, but I'll remove it

cezaraugusto: btw there's no listener for tab page previews anymore and I can't preview a tab set.

bridiver: ah, I think I accidentally grouped them together. Is there a test for it that is failing or did you just find it by manual testing?

cezaraugusto: yes, you can try `npm run test -- --grep="tab pages tab page previews hovering over a tab page changes it"`

luixxiul:
@cezaraugusto I added the QA steps for the tab preview issue to the 1st post based on your feedback here: https://github.com/brave/browser-laptop/pull/9542#pullrequestreview-45567818. Please correct them if they are not right (I replaced `fourth tab` with `3rd tab` as there should be only 3 tabs opened).

cezaraugusto:
@luixxiul awesome, thanks for that

Issue #8671, opened 2017-05-04
Extension: MetaMask


jonathansampson:
Extension: [MetaMask](https://chrome.google.com/webstore/detail/metamask/nkbihfbeogaeaoehlefnkodbefgpgknn)

![image](https://cloud.githubusercontent.com/assets/815158/25688786/9f7f8c76-3047-11e7-8070-7d66c9145c5f.png)

For development purposes, the background page for this extension can be accessed by navigating to the following URL: chrome-extension://dkjoiblfkdpnahoolgkbmgihdfkflnkm/_generated_background_page.html

## Needed APIs:

**chrome.browserAction**

- [x] #5367 chrome.browserAction.setBadgeBackgroundColor ([docs](https://developer.chrome.com/extensions/browserAction#method-setBadgeBackgroundColor))
- [x] #5366 chrome.browserAction.setBadgeText ([docs](https://developer.chrome.com/extensions/browserAction#method-setBadgeText))

**chrome.extension**

- [x] #8688 chrome.extension.getURL ([docs](https://developer.chrome.com/extensions/extension#method-getURL), _deprecated_)

**chrome.runtime**

- [x] #8682 chrome.runtime.connect ([docs](https://developer.chrome.com/extensions/runtime#method-connect))
- [x] #8683 chrome.runtime.getManifest ([docs](https://developer.chrome.com/extensions/runtime#method-getManifest))
- [x] #8684 chrome.runtime.onConnect.addListener ([docs](https://developer.chrome.com/extensions/runtime#event-onConnect))
- [x] #8685 chrome.runtime.onInstalled.addListener ([docs](https://developer.chrome.com/extensions/runtime#event-onConnect))(https://github.com/brave/muon/pull/344)
- [x] ~#8686 chrome.runtime.reload ([docs](https://developer.chrome.com/extensions/runtime#method-reload))~

**chrome.tabs**

- [x] #8680 chrome.tabs.create ([docs](https://developer.chrome.com/extensions/tabs#method-create))

**chrome.windows**

- [x] #6789 chrome.windows.create ([docs](https://developer.chrome.com/extensions/windows#method-create))(https://github.com/brave/muon/pull/295)
- [x] #8677 chrome.windows.getAll ([docs](https://developer.chrome.com/extensions/windows#method-getAll))
- [x] #8678 chrome.windows.remove ([docs](https://developer.chrome.com/extensions/windows#method-remove))
[...]

PR #9647, opened 2017-06-21
add metamask to available extensions

kevinlawler:
brave/browser-laptop#8671

already added to vault manifest & live on server
brave/vault-updater#43

Auditors:
@bridiver @bsclifton @jonathansampson


bridiver:
the resizing is normal, but we could probably eliminate it by debouncing `setPopupWindowDetail` in `popupWindow.js`

bsclifton:
@bridiver I wonder if that same logic is affecting LastPass (the popup is sized incorrectly when DPI is increased)

bsclifton:
Moving to 0.19.x because there is code this depends on (see https://github.com/brave/browser-laptop/pull/9585)

luixxiul:
Does this close #8671?

bsclifton:
@luixxiul I don't think so- cc: @jonathansampson

PR #9678, opened 2017-06-23
Converts Main into redux component

### LINKED ISSUE #9452: Redux component - Main
Convert `Main` into redux component.

NejcZdovc:
Submitter Checklist:

Resolves #9452

Auditors: @bbondy @bridiver @bsclifton

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/main/main.js
+++ b/app/renderer/components/main/main.js
@@ -3,7 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const React = require('react')
-const ImmutableComponent = require('../immutableComponent')
+const ReduxComponent = require('../reduxComponent')
 const Immutable = require('immutable')
 const electron = require('electron')
 const urlResolve = require('url').resolve
@@ -55,7 +55,6 @@ const siteUtil = require('../../../../js/state/siteUtil')
 const searchProviders = require('../../../../js/data/searchProviders')
 const defaultBrowserState = require('../../../common/state/defaultBrowserState')
 const shieldState = require('../../../common/state/shieldState')
-const siteSettingsState = require('../../../common/state/siteSettingsState')
 const menuBarState = require('../../../common/state/menuBarState')
 const windowState = require('../../../common/state/windowState')
 const updateState = require('../../../common/state/updateState')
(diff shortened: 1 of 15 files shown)


app/renderer/components/main/main.js
```diff
@@ -402,12 +384,11 @@ class Main extends ImmutableComponent {
...
 
-    const self = this
     ipc.on(messages.BLOCKED_RESOURCE, (e, blockType, details) => {
-      const frameProps = frameStateUtil.getFrameByTabId(self.props.windowState, details.tabId)
-      frameProps && windowActions.setBlockedBy(frameProps, blockType, details.url)
+      windowActions.setBlockedBy(this.props.activeFrameKey, blockType, details.url)
     })
 
+    // TODO can we remove this?
     ipc.on(messages.BLOCKED_PAGE, (e, blockType, details) => {
```
NejcZdovc: @cezaraugusto You commented this block, so can we remove it?

@diracdeltas Do we need this ipc call at all, now that is commented out?

cezaraugusto: it was listening to a message but taking no action when fired iirc. I can't recall why I didn't remove but given it is doing nothing seems like a leftover so I'm fine removing.

NejcZdovc: Ok, thank you. Let's wait for @diracdeltas conifmation as well

diracdeltas: Not sure. The code that was commented out was added by @bridiver in a12cac496fc3dcbb59591c4e760ad0fde87476e0.

bridiver: not sure what it would do if it blocked the whole page. We also have to be very careful about what we do inside `onBeforeRequest` because it blocks resource loading. Created https://github.com/brave/browser-laptop/pull/9702 to make it more efficient, but we should be saving the data in appState and reading the blocked resources from there

NejcZdovc: So right now we have only comment in this call. We need to determinate what to do with it. Currently we are not doing anything. Now if you don't have any objections I will remove it otherwise we need to add some logic into this function. In this state it's waste of space :smiley:

bridiver: I'm fine with removing it

NejcZdovc: Thank you for your inputs, function removed

js/actions/windowActions.js
```diff
@@ -216,13 +231,15 @@ const windowActions = {
...
    * Dispatches a message to the store when the frame is active and the window is focused
    *
-   * @param {Object} frameProps - the frame properties for the webview in question.
+   * @param {Object} location - location for the webview in question.
+   * @param {Object} tabId - tabId for the webview in question.
    */
-  setFocusedFrame: function (frameProps) {
-    if (frameProps) {
+  setFocusedFrame: function (location, tabId) {
+    if (location) {
```
bridiver: I think this logic should be in the reducer. `setFocusedFrame` should only need `tabId`

bridiver:
some actions are being changed to `tabId` and others to `frameKey`. I think we should prefer `tabId` for anything that we are changing

js/actions/windowActions.js
NejcZdovc: I refactored `menu.js` so that now uses correct implementation of the reducer and have access to the sate. But still we can't remove location, because `eventStore` don't have access to the store and I don't feel comfortable refactoring it in this PR.

NejcZdovc: I am talking about this line https://github.com/brave/browser-laptop/pull/9678/files#diff-bc67a4599e41618040dc83b94144397aR99

NejcZdovc:
@bridiver I changed all frameKeys into tabIds.

PR #9692, opened 2017-06-23
Prevent Tab Close Button from allowing tab-tearoff

### LINKED ISSUE #9511: Prevent Tab Close Button from allowing tab-tearoff
As a user, I may start to click a tab close button and then change my mind not to close it. I need to be able to mouse-out from the action while dragging as a means of escape/cancel.

Currently the drag results in a tear-off.

cezaraugusto:
Auditors: @luixxiul
Close #9511

Test Plan:
1. Tear-off a tab by dragging any part of the tab other than close button
2. It should tear-off
3. Try to tear-off a tab dragging from tab close button
4. You can't

### DIFF
--- a/app/renderer/components/tabs/content/closeTabIcon.js
+++ b/app/renderer/components/tabs/content/closeTabIcon.js
@@ -31,6 +31,7 @@ class CloseTabIcon extends React.Component {
   constructor (props) {
     super(props)
     this.onClick = this.onClick.bind(this)
+    this.onDragStart = this.onDragStart.bind(this)
   }
 
   get frame () {
@@ -46,6 +47,10 @@ class CloseTabIcon extends React.Component {
       })
       appActions.tabCloseRequested(this.props.tabId)
     }
+  }
+
+  onDragStart (event) {
+    event.preventDefault()
   }
 
   mergeProps (state, ownProps) {
@@ -76,6 +81,8 @@ class CloseTabIcon extends React.Component {
       className={css(this.props.showCloseIcon && styles.closeTab)}
       l10nId='closeTabButton'
       onClick={this.onClick}
+      onDragStart={this.onDragStart}
+      draggable='true'
     />
   }
 }
--- a/app/renderer/components/tabs/content/tabIcon.js
+++ b/app/renderer/components/tabs/content/tabIcon.js
@@ -33,6 +33,8 @@ class TabIcon extends ImmutableComponent {
     return <div
       className={this.props.className}
       data-test-favicon={this.props['data-test-favicon']}
+      onDragStart={this.props.onDragStart}
+      draggable={this.props.draggable}
       onClick={this.props.onClick}>
       {
         this.props.symbol


NejcZdovc:
@cezaraugusto let's try not to pass functions as params, because this will cause re-renders (immutable component compares params with simple `!==` check)

image

cezaraugusto:
@NejcZdovc thanks just bound to the constructor, lmk if that works.

cezaraugusto:
feel free to check if changes work too

luixxiul:
@cezaraugusto is this ready for manual testing?

cezaraugusto:
@luixxiul yes 👍

NejcZdovc:
@cezaraugusto did you test if this is not triggering re-renders? I think it still does. You can check this if you add shouldUpdate function in `TabIcon` and check if comparison between prev and next props is ok.

cezaraugusto:
@NejcZdovc while I know that you're worried about perf decrease, I don't know what is expected here. I guess this is somehow related to the work in ReduxComponent, which I'm not familiar with and iirc there's no reference about your claim anywhere in docs nor in other Redux refactoring.

TabIcon is stateless, and has an onClick handler already. I guess this already triggers a re-render too? What is the difference between onClick and onDragStart on such cases? Are you saying that we should add `shouldComponentUpdate` somewhere? If so I can't see a big perf improvement on using it and also `shouldComponentUpdate` is hard to maintain from my experience.

Please note I'm not saying that your claim is wrong, I'm ok with anything to make our app better. I'm just wanting to know how it should work and then if that's a new standard it should be written somewhere to raise awareness. Thanks for bringing that to my attention.

NejcZdovc:
@cezaraugusto let me do a little test and will explain everything

luixxiul:
@NejcZdovc you could create a document to explain what Redux is, why we need the refactoring work, which major changes are there, which rules we should follow, how to contribute the work, following the style guidelines for Aphrodite: https://github.com/brave/browser-laptop/blob/master/docs/style.md

That would help a lot people to understand what is going on.

NejcZdovc:
@luixxiul yes this is completely on me. I have this on my todo from our redux call, but I was 100% focused on 0.19 version and porting everything into redux. Will do as soon as I finish 0.19

luixxiul:
If the document will not be ready soon, until then please consider to push a proposal fix, give someone an advice in detail or create an issue to track that issue, as you know better than anyone else what should be done there. thanks!

bsclifton:
It'll be a while before we're starting the release process for 0.19.x... but I agree, it would be nice to see a demo.  I added you as a reviewer @NejcZdovc, so that you can share the information and any proposals to fix 😄

cezaraugusto:
> Will do as soon as I finish 0.19

we are still on 0.17.x which is not even released yet. Please consider doing that as a priority. 

Otherwise, every PR sent is potentially wrong and will potentially lead the committer to re-work or follow-up something that they thought was right, and we lose productivity doing that. 

Making a reference guide means that both committers and reviewers can concentrate on the real bugs rather than random guessing if a piece of code fits or not the new best practices.

NejcZdovc:
Will do it tomorrow, but what I am talking here is not related to redux refactoring. We are talking about passing data into Immutable component, which is nothing new.

cezaraugusto:
well you are wrong, this is new for me. What data are you talking about? 

This is just preventing a default behavior. Just another handler I'm introducing. I can't see any reference for your claim anywhere. I'm talking about that documentation. 

```
# Best practices while working on the renderer code

1. Don't pass data into Immutable component
1.1 Why?
1.2 Bad/Good examples
```

that would be enough for half my problems

bsclifton:
I think the best action for now is to accept this (if it meets our guidelines) and follow up with a fix (if needed).  @NejcZdovc since you're aware of the performance issues, can you create an issue and self-assign?

I think it's unfair to ask folks to follow best practices when there isn't enough documentation / resources available to fulfill those best practices.  This issue (as it's written) also captures a specific problem and the PR is solving that problem. The performance seems (as a reviewer) to be negligible from my point of view, so I think a follow up is best

NejcZdovc:
Ok, let's take a step back and merge this one :) Like I said I will create docs for redux tomorrow and we can finalize it together.

Little back story why I said in the begging to remove function as it was, was because we had this problems when comparing things inside `mergeProps` for redux. That's why we only want primitives inside `mergeProps`. And the same principle I wanted to add here. You added a fix for it and only thing that I asked is can you check if you add `shouldComponentUpdate` and see what is going on with params. Are they handled correctly now or not? Will passing function into the component trigger re-renders or not? We had this problem for mergeProps, but I wasn't sure if the same thing applies for ImmutableComponent as well. That's why I said that I need to test it. And to my last point, why I said that this is not new is, because I am talking about primitives for quite some time now.

Ok end of the story, now let's get back to the present :) I know that we don't have documentation for it and I will fix this tomorrow. I merged this because I did my test and `shouldComponentUpdate` returns false when comparing this two functions (probably because of references) and will not trigger re-renders. So we are good :champagne: And that's everything that I wanted from you, to double check after you did `bind` in the constructor.

I also tested this PR and it does what it suppose to do :1st_place_medal:

NejcZdovc:
I also opened a discuss about it on the slack so we can see what is going on with this function

luixxiul:
I'd open it on an issue for open discussion (unless it is confidential).

PR #9700, opened 2017-06-25
(Drag and drop) get full path from custom data in dataTransfer.items from https://git…

### LINKED ISSUE #9699: Dragging and dropping a file onto the browser does not work
yes
Drag and drop is not working, because path is not in drop event.
all
435fa34
- Steps to reproduce:
    1. dnd one file to tabs toolbar
    
- Actual result:
White tab with infinite loader
- Expected result:
File is displayed
yes
yes
yes
- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues:

bridiver:
…hub.com/brave/muon/pull/228

Submitter Checklist:

Fixes #9699

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/bookmarks/bookmarksToolbar.js
+++ b/app/renderer/components/bookmarks/bookmarksToolbar.js
@@ -86,8 +86,9 @@ class BookmarksToolbar extends ImmutableComponent {
       }
     }
     if (e.dataTransfer.files.length > 0) {
-      Array.from(e.dataTransfer.files).forEach((file) =>
-        appActions.addSite({ location: file.path, title: file.name }, siteTags.BOOKMARK))
+      Array.from(e.dataTransfer.items).forEach((item) => {
+        item.getAsString((name) => appActions.addSite({ location: item.type, title: name }, siteTags.BOOKMARK))
+      })
       return
     }
 
--- a/app/renderer/components/navigation/navigator.js
+++ b/app/renderer/components/navigation/navigator.js
@@ -32,7 +32,7 @@ const {getCurrentWindowId, isMaximized, isFullScreen, isFocused} = require('../.
 const {isWindows} = require('../../../common/lib/platformUtil')
 const {braveShieldsEnabled} = require('../../../common/state/shieldState')
 const eventUtil = require('../../../../js/lib/eventUtil')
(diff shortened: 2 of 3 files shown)


app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -86,8 +86,9 @@ class BookmarksToolbar extends ImmutableComponent {
       }
     }
     if (e.dataTransfer.files.length > 0) {
```
NejcZdovc: Here we should probably change from `e.dataTransfer.files.length` into `e.dataTransfer.items.length`

app/renderer/components/navigation/navigator.js
```diff
@@ -115,9 +115,10 @@ class Navigator extends React.Component {
 
   onDrop (e) {
     if (e.dataTransfer.files.length > 0) {
```
NejcZdovc: ditto

app/renderer/components/tabs/tabs.js
```diff
@@ -98,9 +98,10 @@ class Tabs extends React.Component {
     }
 
     if (e.dataTransfer.files) {
```
NejcZdovc: ditto

app/renderer/components/bookmarks/bookmarksToolbar.js
bridiver: no, we specifically don't want to do that because we only want these items when there are files

NejcZdovc: so we didn't replace files object with items, but we only added items into dataTransfer?

bsclifton:
pulling into 0.17.x since we need a new Muon 4.1.x version RE: extensions

app/renderer/components/bookmarks/bookmarksToolbar.js
bridiver: files only contains the filename, not the full path. We added the full path for files into items when files are present

bridiver: there are limitations on what we can do with this api without breaking compatibility so rather than change what is contained in file, we added the extra data to items

NejcZdovc: make sense, then we don't need to change this line :smiley:

PR #9713, opened 2017-06-26
Removes all non-primitive types from BookmarksToolbar and BookmarkTool…

### LINKED ISSUE #9712: Only primitives in BookmarksToolbar and BookmarkToolbarButton
## Test plan

https://github.com/brave/browser-laptop/pull/9713#issue-238516609

----

Only primitives in merge props. Because of that we need to move context menu functions into the reducer.

NejcZdovc:
…barButton

Submitter Checklist:

Resolves #9712

Auditors: @bsclifton

Test plan:

**Note**: Reordering in context menu is not working. Will create issue for it.

1)
- right click on a blank bookmark toolbar
- context menu should be displayed

2)
- right click on a bookmark
- context menu should be displayed

3)
- right click on a bookmark folder
- context menu should be displayed

4)
- left click on a bookmark folder
- context menu should be displayed with items in the folder

5)
- create so many bookmark items that you get overflow menu
- check that overflow menu is correctly displayed 

6)
- check if bookmarks are displayed and updated correctly

7) DND test
- try to reorder bookmarks on the bookmark toolbar
- try to add bookmark into a folder
- try to bookmark a tab with dragging lock icon on the bookmark toolbar

Reviewer Checklist:

Tests

### DIFF
--- a/app/common/lib/bookmarkUtil.js
+++ b/app/common/lib/bookmarkUtil.js
@@ -41,8 +41,8 @@ const isBookmarkNameValid = (detail, isFolder) => {
   const title = detail.get('title') || detail.get('customTitle')
   const location = detail.get('location')
   return isFolder
-    ? (title != null && title.trim().length > 0)
-    : (location != null && location.trim().length > 0)
+    ? (title != null && title !== 0) && title.trim().length > 0
+    : location != null && location.trim().length > 0
 }
 
 const showOnlyFavicon = () => {
@@ -56,24 +56,19 @@ const showFavicon = () => {
     btbMode === bookmarksToolbarMode.FAVICONS_ONLY
 }
 
-const getDNDBookmarkData = (state, bookmark) => {
+const getDNDBookmarkData = (state, bookmarkKey) => {
   const data = (state.getIn(['dragData', 'dragOverData', 'draggingOverType']) === dragTypes.BOOKMARK &&
     state.getIn(['dragData', 'dragOverData'], Immutable.Map())) || Immutable.Map()
 
-  if (data.get('draggingOverKey') == null) {
-    return Immutable.Map()
-  }
-
-  // TODO (nejc) this is slow, replace with simple ID check - we need to add id into bookmark object
-  return (Immutable.is(data.get('draggingOverKey'), bookmark)) ? data : Immutable.Map()
+  return data.get('draggingOverKey') === bookmarkKey ? data : Immutable.Map()
 }
 
 const getToolbarBookmarks = (state) => {
   const sites = state.get('sites', Immutable.List())
 
   const noParentItems = siteUtil.getBookmarks(sites)
-    .sort(siteUtil.siteSort)
     .filter((bookmark) => !bookmark.get('parentFolderId'))
+    .sort(siteUtil.siteSort)
   let widthAccountedFor = 0
   const overflowButtonWidth = 25
   const onlyFavicon = showOnlyFavicon()
@@ -119,9 +114,9 @@ const getToolbarBookmarks = (state) => {
   }
 
   return {
-    visibleBookmarks: noParentItems.take(i),
+    visibleBookmarks: noParentItems.take(i).map((item, index) => index).toList(),
     // Show at most 100 items in the overflow menu
-    hiddenBookmarks: noParentItems.skip(i).take(100)
+    hiddenBookmarks: noParentItems.skip(i).take(100).map((item, index) => index).toList()
   }
 }
 
--- a/app/renderer/components/bookmarks/bookmarkToolbarButton.js
+++ b/app/renderer/components/bookmarks/bookmarkToolbarButton.js
@@ -15,23 +15,16 @@ const windowActions = require('../../../../js/actions/windowActions')
 const appActions = require('../../../../js/actions/appActions')
 const bookmarkActions = require('../../../../js/actions/bookmarkActions')
 
-// Store
-const windowStore = require('../../../../js/stores/windowStore')
-
 // Constants
 const dragTypes = require('../../../../js/constants/dragTypes')
 const {iconSize} = require('../../../../js/constants/config')
-const {bookmarksToolbarMode} = require('../../../common/constants/settingsEnums')
-const settings = require('../../../../js/constants/settings')
 
 // Utils
 const siteUtil = require('../../../../js/state/siteUtil')
 const {getCurrentWindowId} = require('../../currentWindow')
 const dnd = require('../../../../js/dnd')
 const cx = require('../../../../js/lib/classSet')
-const {getSetting} = require('../../../../js/settings')
 const frameStateUtil = require('../../../../js/state/frameStateUtil')
-const contextMenus = require('../../../../js/contextMenus')
 const bookmarkUtil = require('../../../common/lib/bookmarkUtil')
 
 // Styles
@@ -56,10 +49,6 @@ class BookmarkToolbarButton extends React.Component {
 
   componentDidMount () {
     this.bookmarkNode.addEventListener('auxclick', this.onAuxClick)
-  }
-
-  get activeFrame () {
-    return windowStore.getFrame(this.props.activeFrameKey)
   }
 
   onAuxClick (e) {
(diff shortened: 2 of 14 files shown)


app/common/lib/bookmarkUtil.js
```diff
@@ -119,9 +119,9 @@ const getToolbarBookmarks = (state) => {
   }
 
   return {
-    visibleBookmarks: noParentItems.take(i),
+    visibleBookmarks: noParentItems.take(i).map((item, index) => index).toList(),
```
bridiver: not sure I understand this change

js/actions/bookmarkActions.js
```diff
@@ -4,15 +4,18 @@
...
 const appActions = require('./appActions')
 const windowActions = require('./windowActions')
 const eventUtil = require('../lib/eventUtil')
+const appStoreRenderer = require('../stores/appStoreRenderer')
 const {SWITCH_TO_NEW_TABS} = require('../constants/settings')
 const getSetting = require('../settings').getSetting
 
 const bookmarkActions = {
-  openBookmarksInFolder: function (allBookmarkItems, folderDetail) {
+  openBookmarksInFolder: function (folderDetail) {
```
bridiver: this one should also be converted to an action. Not sure why this is called `bookmarkActions`, probably just some legacy reason because it's not an action the way we define them now

bridiver: I think you can probably dump this whole file, make a real action and move the logic to the reducer

app/renderer/components/bookmarks/bookmarkToolbarButton.js
```diff
@@ -146,30 +135,40 @@ class BookmarkToolbarButton extends React.Component {
...
+    const rectLeft = e.target.getBoundingClientRect()
+    const rectBottom = e.target.parentNode.getBoundingClientRect()
+    const left = (rectLeft.left | 0) - 2
+    const top = (rectBottom.bottom | 0) - 1
+
+    if (e && e.stopPropagation) {
+      e.stopPropagation()
+    }
+
+    windowActions.onShowBookmarkFolderMenu(this.props.bookmarkKey, left, top)
```
bridiver: can you add a TODO to merge `setBookmarksToolbarSelectedFolderId` into `onShowBookmarkFolderMenu`. We should never call more than one action at a time

app/renderer/reducers/contextMenuReducer.js
```diff
@@ -33,6 +54,403 @@ const validateState = function (state) {
...
+      faIcon = isFolder ? 'fa-folder-o' : 'fa-file-o'
+    }
+    const templateItem = {
+      bookmark: site,
+      draggable: true,
+      label: site.get('customTitle') || site.get('title') || site.get('location'),
+      icon: showFavicon ? site.get('favicon') : undefined,
+      faIcon,
+      contextMenu: function (e) {
+        onSiteDetailMenu(state, item[0], e)
```
bridiver: I think this is supposed to be `windowActions. onSiteDetailMenu` ?

bridiver: is there a test that covers this? Looks like it should be failing if there is

app/common/lib/bookmarkUtil.js
NejcZdovc: Instead of sending the whole object we are sending only keys. So we have list of keys and this way we are not triggering re-renders. We send this key to the child components which gets the whole object from the state.

bridiver: I see, makes sense

js/actions/bookmarkActions.js
NejcZdovc: I agree that we need to convert this, but not sure if this PR is the right one to do it in. One thing that we also need to discuss is how can we call other actions inside the reducers. One case that I had was calling an actions that was then handled in multiple reducers. So we need to define this

app/renderer/components/bookmarks/bookmarkToolbarButton.js
NejcZdovc: Done

app/renderer/reducers/contextMenuReducer.js
NejcZdovc: Fixed

js/actions/bookmarkActions.js
bridiver: we should not be calling actions from reducers, that breaks the uni-directional flow that is the whole point of flux/redux. The only time we are doing that now is for some special cases in the windowStore and maybe a few legacy calls. If you have a specific example to look at we can talk about options

bridiver: as far as converting it goes, this is just like the others. It is accessing `appStoreRenderer` and that is problematic. Why is this any different than the others that you already converted?

NejcZdovc: For example this refactoring
https://github.com/brave/browser-laptop/pull/9678/files#diff-de5d4f1b019deecc5b1926970450546cR392

https://github.com/brave/browser-laptop/pull/9678/files#diff-733d30fbf34816322d3f25dd59c1eb99R747

I know that is in anti-pattern, but I am not sure how to achieve it otherwise (I already talked with @bsclifton about it), so I am opened for any suggestions how to fix this and have only one action per call. I didn't find any good approach with the current implementation.

bridiver: so that is one of the temporary cases where we are firing appActions from windowActions, but the long-term fix is to just handle it in the browser process and the window updates based on state, not an action

NejcZdovc: yes and this is what I am doing in a lot of PR's right now. Moving actions into the reducers, so that I can remove state from components to have only primitives.

bridiver: but we still have the problem here that we are accessing `appStoreRenderer` for state in this "action"

NejcZdovc: what I can do is move this into the new app action and call other actions out of this new action. What do you think?

bridiver: we want to avoid calling actions from actions. I still don't understand why this is any different from the ones you already changed?

NejcZdovc: I can create a new app action, but we will be calling `appActions.createTabRequested` from it, so it will be again action inside an action.

bridiver: an appAction doesn't need to call `createTabRequested` it can just call `tabs.create`

NejcZdovc: I am not sure that that's enough. So if you check out  `createTabRequested` you will see that we have two places where we call `APP_CREATE_TAB_REQUESTED`.

First: https://github.com/brave/browser-laptop/blob/master/app/browser/reducers/tabsReducer.js?utf8=%E2%9C%93#L138

Second: https://github.com/brave/browser-laptop/blob/master/app/browser/reducers/sitesReducer.js?utf8=%E2%9C%93#L158

Now if instead of calling this app actions I just do `tabs.create` some logic will be lost, which would be called otherwise. This is the main reason why I keep using actions, because they are used in different places and I could regress some logic if I just pick one. What I don't want to do is duplicating a code.

bridiver: the handler in sitesReducer should be changed to `APP_TAB_CREATED`. `APP_CREATE_TAB_REQUESTED` is a temporary workaround until we move things to the browser process so nothing else should handle it

bridiver: in any case if `tabs.create` doesn't encapsulate all the logic for creating a tab then we need to fix that because the api is not useful if it can only be called through `APP_CREATE_TAB_REQUESTED`

bridiver: for instance `activateIfOpen` shouldn't be an action property because it always applies to specific urls

NejcZdovc: Based on our conversation yesterday I would suggest that we merged this PR. This file `bookmarkActions.js` was not really the focus in this PR, so I would suggest that we create a follow up where we would refactor only this file.

NejcZdovc:
re-run `TEST_DIR=bookmark-components npm run testsuite` 4 times locally and it's passing every single time

ayumi:
what's the status on this one? currently UI lag is immense in master, encumbering other performance work – i'm really looking forward to this patch!

app/renderer/reducers/contextMenuReducer.js
```diff
@@ -89,6 +101,414 @@ const onTabPageMenu = function (state, action) {
   tabPageMenu.popup(getCurrentWindow())
 }
 
+const openInNewTabMenuItem = (url, isPrivate, partitionNumber, openerTabId) => {
```
bsclifton: Each of these menu types should be in a file (I'd suggest commonMenu- see https://github.com/brave/browser-laptop/blob/2bd7726b6d6ea83f41fbebdfab6b4141f8a5a78b/app/common/commonMenu.js). This would make it easy to unit test each method.  A quick example of something easy to test: you can check each method that's exported to make sure each has a label field that called `locale.translation()`.

bsclifton: Would be great in a follow up 😄  Everything else looks good

PR #9729, opened 2017-06-27
Follow up of Welcome page #9480

cezaraugusto:
**Manual test plan:**

Please see https://github.com/brave/browser-laptop/pull/9480#issue-236069494

**Test Plan:**

```
npm run test -- --grep="tabs API unit tests"
npm run test -- --grep="APP_WINDOW_READY"
```

**QA notes:**

1) This just moves the logic to the correct place to keep updated with our app best practices and needs no QA. The test plan is already specified in https://github.com/brave/browser-laptop/pull/9480#issue-236069494. Labeling as no-qa-needed and release-notes-exclude. Please take the PR link above and related issue as the release notes.

2) The welcome screen will only be shown in production mode

**Release team notes:**

https://github.com/brave/browser-laptop/pull/9480#issue-236069494 is tagged for 0.18.x. This reverts mentioned PR commit and adds a new one with updated logic. I'm going to move this too to avoid inconsistency across branches.


app/browser/reducers/tabsReducer.js
```diff
@@ -337,6 +337,31 @@ const tabsReducer = (state, action, immutableAction) => {
...
+        }
+      }
+
+      const welcomeScreenProperties = {
+        'url': 'about:welcome',
+        'windowId': action.getIn(['createProperties', 'windowId'])
+      }
+
+      // tests should not be aware of the welcome screen
+      const shouldShowWelcomeScreen = process.env.NODE_ENV !== 'test' &&
```
bridiver: we should probably have a way to override this so we can test it. Maybe it would be better to set the defaultState for about.welcome.showOnLoad in sessionStore.js to `true` except for production and then we set it to `false` when we want to test?

app/browser/reducers/tabsReducer.js
```diff
@@ -337,6 +337,31 @@ const tabsReducer = (state, action, immutableAction) => {
         }
       }
       break
+    case appConstants.APP_WINDOW_READY: {
+      console.log(action)
```
bridiver: please removing logging :)

cezaraugusto: oops 🙈

bsclifton:
Thanks for following up with this, @cezaraugusto 😄 👍  (and thanks for helping w/ the refactor @bridiver!)

test/unit/app/browser/reducers/tabsReducerTest.js
```diff
@@ -608,6 +609,80 @@ describe('tabsReducer unit tests', function () {
     })
   })
 
+  describe('APP_WINDOW_READY', function () {
```
cezaraugusto: cc @bsclifton

app/sessionStore.js
```diff
@@ -658,7 +658,7 @@ module.exports.defaultAppState = () => {
         pinnedTopSites: pinnedTopSites
       },
       welcome: {
-        showOnLoad: true
+        showOnLoad: !['test', 'development'].includes(process.env.NODE_ENV)
```
bsclifton: This is perfect! good call setting here (rather than putting the check into the code) 😄

PR #9754, opened 2017-06-28
UrlBarIcon uses now only primitives

### LINKED ISSUE #9753: Only primitives in UrlBarIcon
## Test plan

https://github.com/brave/browser-laptop/pull/9754#issue-239150822

----

Remove `activeFrame` prop from merge props. Replace it with activeFrameKey and create new window action for `prepareBookmarkDataFromCompatible`

NejcZdovc:
Submitter Checklist:

Resolves #9753

Test Plan:
- try to dnd lock image from url bar to the bookmark toolbar
- new bookmark should be added

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/navigation/urlBarIcon.js
+++ b/app/renderer/components/navigation/urlBarIcon.js
@@ -111,7 +111,9 @@ class UrlBarIcon extends React.Component {
 
   onDragStart (e) {
     dndData.setupDataTransferURL(e.dataTransfer, this.props.location, this.props.title)
-    dndData.setupDataTransferBraveData(e.dataTransfer, dragTypes.TAB, this.props.activeFrame)
+    dndData.setupDataTransferBraveData(e.dataTransfer, dragTypes.TAB, {
+      tabId: this.props.tabId
+    })
   }
 
   mergeProps (state, ownProps) {
@@ -138,7 +140,7 @@ class UrlBarIcon extends React.Component {
 
     // used in other functions
     props.title = activeFrame.get('title', '')
-    props.activeFrame = activeFrame // TODO (nejc) only primitives
+    props.tabId = activeFrame.get('tabId')
 
     return props
   }
--- a/js/dnd.js
+++ b/js/dnd.js
@@ -7,8 +7,6 @@ const appActions = require('./actions/appActions')
 const ReactDOM = require('react-dom')
 const dndData = require('./dndData')
 const dragTypes = require('./constants/dragTypes')
-const siteTags = require('./constants/siteTags')
-const siteUtil = require('./state/siteUtil')
 const appStoreRenderer = require('./stores/appStoreRenderer')
 const {getCurrentWindowId} = require('../app/renderer/currentWindow')
 const {ESC} = require('../app/common/constants/keyCodes.js')
@@ -137,11 +135,8 @@ module.exports.isMiddle = (domNode, clientX) => {
 module.exports.prepareBookmarkDataFromCompatible = (dataTransfer) => {
   let bookmark = dndData.getDragData(dataTransfer, dragTypes.BOOKMARK)
   if (!bookmark) {
-    const frameProps = dndData.getDragData(dataTransfer, dragTypes.TAB)
-    if (frameProps) {
-      bookmark = siteUtil.getDetailFromFrame(frameProps, siteTags.BOOKMARK)
-      appActions.addSite(bookmark, siteTags.BOOKMARK)
-    }
+    const dragData = dndData.getDragData(dataTransfer, dragTypes.TAB)
+    windowActions.onFrameBookmark(dragData.get('tabId'))
   }
   return bookmark
 }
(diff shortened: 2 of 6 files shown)


app/renderer/components/navigation/urlBarIcon.js
```diff
@@ -111,7 +111,9 @@ class UrlBarIcon extends React.Component {
 
   onDragStart (e) {
     dndData.setupDataTransferURL(e.dataTransfer, this.props.location, this.props.title)
-    dndData.setupDataTransferBraveData(e.dataTransfer, dragTypes.TAB, this.props.activeFrame)
+    dndData.setupDataTransferBraveData(e.dataTransfer, dragTypes.TAB, Immutable.fromJS({
```
bridiver: conversion to/from JS is a relatively expensive operation, why do we need to send an object instead of just the activeFrameKey value

NejcZdovc: Because this function expect data to Immutable https://github.com/brave/browser-laptop/blob/master/js/dndData.js#L26. What I can do is add a check in this function and if data is Immutable convert it, otherwise leave it as it is.

bridiver: that function turns it right back into js again. I think we should stick with regular js

NejcZdovc: Fixed

js/dnd.js
```diff
@@ -137,11 +135,8 @@ module.exports.isMiddle = (domNode, clientX) => {
 module.exports.prepareBookmarkDataFromCompatible = (dataTransfer) => {
   let bookmark = dndData.getDragData(dataTransfer, dragTypes.BOOKMARK)
   if (!bookmark) {
-    const frameProps = dndData.getDragData(dataTransfer, dragTypes.TAB)
-    if (frameProps) {
-      bookmark = siteUtil.getDetailFromFrame(frameProps, siteTags.BOOKMARK)
-      appActions.addSite(bookmark, siteTags.BOOKMARK)
-    }
+    const frame = dndData.getDragData(dataTransfer, dragTypes.TAB)
+    windowActions.onFrameBookmark(frame.get('activeFrameKey'))
```
bridiver: this change doesn't make sense to me. There's no `state` that is needed here so I don't see any reason to create a window action that forwards to an app action. We are using that anti-pattern in select places where there isn't a good workaround right now, but this doesn't seem necessary. What I think we should do (in another PR) is convert the `drop` methods to actions and move this code inside a (preferably browser process) reducer. It also seems like a bad idea to hide an action call inside this method and it should probably just return the bookmark from dndData and do nothing else. I think we're adding a layer of indirection here that is just going to be removed later anyway

NejcZdovc: Why I moved this into separate action is because we need to get detail of the frame so we can add it into sites.

Here we need to have access to the window state: https://github.com/brave/browser-laptop/pull/9754/files/e0827f5b092eecfbe915b23838fb9dccc8fdcc5b#diff-be01d0957f991e2400072d82d108450fL142

bridiver: I think the confusing part here is that you're getting `frame` from `dndData.getDragData`, but apparently that isn't a frame so I think like the other one we should just pass `frameKey`. I just noticed this calls `frame.get('activeFrameKey')` which makes it even more confusing

NejcZdovc: Probably the naming is wrong here, so I fixed it. We send this data from this line https://github.com/brave/browser-laptop/pull/9754/files/e0827f5b092eecfbe915b23838fb9dccc8fdcc5b#diff-51a114fecaac13cd41162f4c956e7508R114 and then get it back here as Immutable Map with this function https://github.com/brave/browser-laptop/blob/26a580bc3148ec427465c3e266e6bd42778da786/js/dndData.js#L11

app/renderer/components/navigation/urlBarIcon.js
```diff
@@ -111,7 +111,9 @@ class UrlBarIcon extends React.Component {
 
   onDragStart (e) {
     dndData.setupDataTransferURL(e.dataTransfer, this.props.location, this.props.title)
-    dndData.setupDataTransferBraveData(e.dataTransfer, dragTypes.TAB, this.props.activeFrame)
+    dndData.setupDataTransferBraveData(e.dataTransfer, dragTypes.TAB, {
+      activeFrameKey: this.props.activeFrameKey
```
bridiver: sorry to do this to you, but can we use `tabId` instead of `frameKey`? If we had the `tabId` it would be easy to convert this to a browser process action (please add TODO for that) with a `siteUtil.getDetailFromTabId` method. I think we should drop `active` as well once it gets to this point

NejcZdovc: I totally agree and I forget about this one too frequently. Fixed

NejcZdovc:
comment added

bsclifton:
++

PR #9757, opened 2017-06-28
Removes non-primitive types from UrlBar

### LINKED ISSUE #9756: Only primitives in UrlBar
## Test plan

https://github.com/brave/browser-laptop/pull/9757#issue-239222274

----

We need to remove `suggestionList` and `onStop` props from mergeProps.

NejcZdovc:
Submitter Checklist:

Resolves #9756

Auditors: @bridiver @bbondy

Test Plan:
- general test of url bar
- focus on suggestions

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/navigation/navigationBar.js
+++ b/app/renderer/components/navigation/navigationBar.js
@@ -82,18 +82,9 @@ class NavigationBar extends React.Component {
   }
 
   onStop () {
+    // TODO (bridiver) - remove shortcut
     ipc.emit(messages.SHORTCUT_ACTIVE_FRAME_STOP)
-    if (this.props.navbar.getIn(['urlbar', 'focused'])) {
-      windowActions.setUrlBarActive(false)
-      const shouldRenderSuggestions = this.props.navbar.getIn(['urlbar', 'suggestions', 'shouldRender']) === true
-      const suggestionList = this.props.navbar.getIn(['urlbar', 'suggestions', 'suggestionList'])
-      if (!shouldRenderSuggestions ||
-          // TODO: Once we take out suggestion generation from within URLBarSuggestions we can remove this check
-          // and put it in shouldRenderUrlBarSuggestions where it belongs.  See https://github.com/brave/browser-laptop/issues/3151
-          !suggestionList || suggestionList.size === 0) {
-        windowActions.setUrlBarSelected(true)
-      }
-    }
+    windowActions.onStop(this.props.isFocused, this.props.shouldRenderSuggestions)
   }
 
   componentDidMount () {
@@ -144,7 +135,8 @@ class NavigationBar extends React.Component {
     props.showHomeButton = !props.titleMode && getSetting(settings.SHOW_HOME_BUTTON)
 
     // used in other functions
-    props.navbar = navbar // TODO(nejc) remove, primitives only
+    props.isFocused = navbar.getIn(['urlbar', 'focused'], false)
+    props.shouldRenderSuggestions = navbar.getIn(['urlbar', 'suggestions', 'shouldRender']) === true
     props.sites = state.get('sites') // TODO(nejc) remove, primitives only
     props.activeTabId = activeTabId
     props.showHomeButton = !props.titleMode && getSetting(settings.SHOW_HOME_BUTTON)
@@ -208,10 +200,7 @@ class NavigationBar extends React.Component {
           : null
         }
       </div>
-      <UrlBar
-        titleMode={this.props.titleMode}
-        onStop={this.onStop}
-        />
+      <UrlBar titleMode={this.props.titleMode} />
       {
         this.props.showPublisherToggle
         ? <div className='endButtons'>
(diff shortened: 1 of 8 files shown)


app/renderer/components/navigation/navigationBar.js
```diff
@@ -144,7 +139,8 @@ class NavigationBar extends React.Component {
     props.showHomeButton = !props.titleMode && getSetting(settings.SHOW_HOME_BUTTON)
 
     // used in other functions
-    props.navbar = navbar // TODO(nejc) remove, primitives only
+    props.isFocused = navbar.getIn(['urlbar', 'focused'])
```
bsclifton: should this have a default (in case null, etc)? ex:
`props.isFocused = navbar.getIn(['urlbar', 'focused'], false)`

app/renderer/components/navigation/urlBar.js
```diff
@@ -125,11 +118,10 @@ class UrlBar extends React.Component {
           const isLocationUrl = isUrl(location)
           if (!isLocationUrl && e.ctrlKey) {
             appActions.loadURLRequested(this.props.activeTabId, `www.${location}.com`)
-          } else if (this.shouldRenderUrlBarSuggestions &&
-              ((typeof this.activeIndex === 'number' && this.activeIndex >= 0) ||
+          } else if (this.props.showrUrlBarSuggestions &&
```
NejcZdovc: done

app/renderer/components/navigation/urlBar.js
```diff
@@ -125,11 +118,10 @@ class UrlBar extends React.Component {
...
+              ((typeof this.props.activeIndex === 'number' && this.props.activeIndex >= 0) ||
               (this.props.urlbarLocationSuffix && this.props.autocompleteEnabled))) {
             // Hack to make alt enter open a new tab for url bar suggestions when hitting enter on them.
-            const isDarwin = process.platform === 'darwin'
             if (e.altKey) {
               if (isDarwin) {
```
NejcZdovc:
```
isDarwin = () => {
  return process.platform === 'darwin' ||
    navigator.platform === 'MacIntel'
}
```

it first checks `process.platform === 'darwin'` same as we did in the removed code. Isn't then this the same or can `navigator.platform === 'MacIntel'` lead into wrong determination of OS?

app/renderer/components/navigation/navigationBar.js
NejcZdovc: We only use it in if statement, so we don't need it, but I added it anyway :smiley:

app/renderer/components/navigation/urlBar.js
bsclifton: @NejcZdovc the check in platformUtil has been tested well in both the browser and renderer process.  Typically, if you're in the renderer, `process.platform` will return `undefined`- hence the need to also check `navigator.platform` (which IS set in the renderer)

bsclifton: Using this utility method is encouraged over duplicating the code (especially since it has tests 😄 )

NejcZdovc: fixed

NejcZdovc:
@bsclifton can we merged this one?

app/renderer/components/navigation/urlBar.js
```diff
@@ -428,6 +398,16 @@ class UrlBar extends React.Component {
     contextMenus.onUrlBarContextMenu(e)
   }
 
+  onStop () {
+    ipc.emit(messages.SHORTCUT_ACTIVE_FRAME_STOP)
```
bridiver: I don't like duplicating this logic here. Can we create `windowActions.onStop` and have both methods call that instead?

NejcZdovc: So just move the whole content of this function into another window action?

bridiver: yea, that will have the added benefit of calling one windowAction instead of 2. You can refactor `WINDOW_SET_URL_BAR_SELECTED` and `WINDOW_SET_URL_BAR_ACTIVE` handlers in urlBarReducer into methods so you can call them from `WINDOW_ON_STOP`. In a follow-up we can get rid of `SHORTCUT_ACTIVE_FRAME_STOP` and replace it with a `WINDOW_ON_STOP` handler in the browser process

bridiver: `sendToFocusedWindow` also needs to die :)

NejcZdovc: Done

app/renderer/components/navigation/navigationBar.js
```diff
@@ -83,17 +83,7 @@ class NavigationBar extends React.Component {
 
   onStop () {
     ipc.emit(messages.SHORTCUT_ACTIVE_FRAME_STOP)
```
bridiver: I think we should put this ipc message in there as well. When we get rid of it, we'll just switch over to the browser process handling WINDOW_ON_STOP

NejcZdovc: problem is that `ipc.emit` is not working in `urlBarReducer`

NejcZdovc: as pre our discussion, I added TODO's, so that we remove this shortcut.

bsclifton:
++

PR #9792, opened 2017-06-30
Removes non-primitive types from NavigationBar

### LINKED ISSUE #9791: Only primitives in NavigationBar
Remove `sites` from `mergeProps`. `navbar` will be removed in #9757

NejcZdovc:
Submitter Checklist:

Resolves #9791

Auditors: @bridiver @bsclifton

Test Plan:
- try to add a bookmark (star)
- try to remove bookmark

Reviewer Checklist:

Tests

### DIFF
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -14,7 +14,7 @@ const frameStateUtil = require('../state/frameStateUtil')
 const ipc = require('electron').ipcRenderer
 const messages = require('../constants/messages')
 const debounce = require('../lib/debounce')
-const getSetting = require('../settings').getSetting
+const {getSetting} = require('../settings')
 const UrlUtil = require('../lib/urlutil')
 const {l10nErrorText} = require('../../app/common/lib/httpUtil')
 const { makeImmutable } = require('../../app/common/state/immutableUtil')
@@ -23,6 +23,8 @@ const assert = require('assert')
 const contextMenuState = require('../../app/common/state/contextMenuState')
 const appStoreRenderer = require('./appStoreRenderer')
 const windowActions = require('../actions/windowActions')
+const siteUtil = require('../state/siteUtil')
+const siteTags = require('../constants/siteTags')
 
 let windowState = Immutable.fromJS({
   activeFrameKey: null,
@@ -748,6 +750,22 @@ const doAction = (action) => {
           })
           appActions.loadURLRequested(action.tabId, 'about:certerror')
         }
+        break
+      }
+    case windowConstants.WINDOW_ON_TOGGLE_BOOKMARK:
+      {
+        // show the AddEditBookmarkHanger control; saving/deleting takes place there
+        const activeFrame = frameStateUtil.getActiveFrame(windowState)
+        let siteDetail = siteUtil.getDetailFromFrame(activeFrame, siteTags.BOOKMARK)
+        const key = siteUtil.getSiteKey(siteDetail)
+
+        if (key !== null) {
+          const site = appStoreRenderer.state.getIn(['sites', key], Immutable.Map())
+          siteDetail = siteDetail.set('parentFolderId', site.get('parentFolderId'))
+          siteDetail = siteDetail.set('customTitle', site.get('customTitle'))
+        }
+        siteDetail = siteDetail.set('location', UrlUtil.getLocationIfPDF(siteDetail.get('location')))
+        windowActions.setBookmarkDetail(siteDetail, siteDetail, null, action.isBookmarked, true)
         break
       }
     default:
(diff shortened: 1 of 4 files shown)


js/stores/windowStore.js
```diff
@@ -750,6 +752,22 @@ const doAction = (action) => {
...
         break
       }
+    case windowConstants.WINDOW_ON_TOGGLE_BOOKMARK:
+      {
+        // show the AddEditBookmarkHanger control; saving/deleting takes place there
+        const activeFrame = frameStateUtil.getActiveFrame(windowState)
+        let siteDetail = siteUtil.getDetailFromFrame(activeFrame, siteTags.BOOKMARK)
+        const key = siteUtil.getSiteKey(siteDetail)
+
+        if (key !== null) {
```
bsclifton: this logic here would be better if it was contained inside `siteUtil.getDetailFromFrame` method, IMO (especially since it can be tested).  There are other places that call this method and they too should be having the parentFolderId and customTitle set (in fact, if the other calls are not setting this, it may be causing a bug / bugs)

NejcZdovc: Not sure why this logic should be in `siteUtil.getDetailFromFrame`. In this `if` statement we are getting data from `appStore` and adding them to the `siteDetail`. Not sure where else would we be using this? We use this function in 13 different places. We only want to access `appStore` in reducers and not in components.

js/stores/windowStore.js
```diff
@@ -750,6 +752,22 @@ const doAction = (action) => {
...
+          const site = appStoreRenderer.state.getIn(['sites', key], Immutable.Map())
+          siteDetail = siteDetail.set('parentFolderId', site.get('parentFolderId'))
+          siteDetail = siteDetail.set('customTitle', site.get('customTitle'))
+        }
+        siteDetail = siteDetail.set('location', UrlUtil.getLocationIfPDF(siteDetail.get('location')))
+        windowActions.setBookmarkDetail(siteDetail, siteDetail, null, action.isBookmarked, true)
```
NejcZdovc: Yeah, we want to remove this, but sadly we need to keep it for now. You will see this pattern in other PR's as well. For this particular case we will remove it in #9710, because we will remove `setBookmarkDetail` as per our discussion.

js/stores/windowStore.js
bsclifton: the data it gets from `appStore` is the site which matches... `getDetailFromFrame` should be returning the entry from the sites map (which is what `siteUtil` does; it's the wrapper around the sites map which is stored in `appStore`)

In this case, `getDetailFromFrame` is just taking data from the frame and returning it... rather than doing a lookup and fetching the already existing one.  This code here then copies the data from the existing one (which is a mistake IMO, since it could have just used the existing reference)

bsclifton: Here's a quick explanation of why this is a problem:
if you remove this line:
`siteDetail = siteDetail.set('parentFolderId', site.get('parentFolderId'))`

The code will create a NEW sites entry (which is wrong).  The reason why it creates a new one is because the key has the folderId in it.  The windowStore shouldn't care about this logic- it should just use the object it gets back.  Instead, we are having business logic in the windowStore about a siteDetails entry to avoid a bug (instead of returning an existing reference- the one which the key matches)

bsclifton: this would fix the problem, although it's a little more sloppy:
```js
siteDetail = appStoreRenderer.state.getIn(['sites', key], Immutable.Map())
// siteDetail = siteDetail.set('parentFolderId', site.get('parentFolderId'))
// siteDetail = siteDetail.set('customTitle', site.get('customTitle'))
```

NejcZdovc: Not sure that we can to that, because now in your example you overrode sites from window state, which means that bookmark will not have favicon, tags, themeColor etc. It will only have `parentFolderId`, `title` and `customTitle`

NejcZdovc: we are only missing `parentFolderId` in this example, I just checked and everything else is there. But I think this whole PR, will be redundant when I do #9710 :smiley:

bsclifton: that's true- maybe it's OK as-is (for this PR) and that would be the place to tackle it 😄 

It should have everything (the example above) because (in the current code) whatever is found is copied from appState.  So instead of copying it, just use the entire object (which like you said, has favicon, tags, themeColor, etc).  This is great because if we needed to expose those we could.  By not using the real object, we're opening ourselves up to bugs.  For example, what if `parentFolderId` is changed to `parentDirectoryId`. If we properly handled it, we'd only need to update in `siteUtil`.  But because this code is copying the object, we'd have to update in 2 (or more) places

NejcZdovc:
Closing in favor of #9710

PR #9809, opened 2017-06-30
Only update tabPageIndex if user is not hovering other tab

### LINKED ISSUE #9779: Changing tabset count messes up tab favicon and close button
## Test plan

https://github.com/brave/browser-laptop/pull/10134#issue-245571404

----

Yes

Changing tabset count messes up tab favicon and close button

Windows 10 x64

Brave	0.17.12
rev	0095c35
Muon	4.1.6

- Steps to reproduce:
    1. Set tab count to 20 and  set homepage to reddit.com/r/random
    2. Open more than 100 tabs by pressing ctrl+t, ensure there are 5-6 tabset
    3. With all these tabs set tabcount to 100, close button and favicon on tabs gets messed up as the breakpoint is not properly calculated
    
- Actual result:
Changing tabset count messes up tab favicon and close button

- Expected result:
Tab breakpoint should be properly calculated

Yes

Yes

Yes

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:
image
- Any related issues:
cc: @cezaraugusto

### LINKED ISSUE #9895: Change per-frame hoverState to a single hoverTabIndex key
Based on feedback left in https://github.com/brave/browser-laptop/pull/9809#discussion_r125720809.

If #9809 is made on time it should be shipped in 0.17.14. This is split of work makes it easier for me to keep track of changes. This is just for code improvement. Setting to the same milestone.

For QA interest:

This affects tab's hover state and tab previews and we have tests for all scenarios changed. For clarification manual test plan is:

1. Private/new session tabs should not show icon when tab is hovered;
2. Close tab icon should show on tab hover;
3. Tab preview should show on tab hover

cezaraugusto:
Auditors: @bridiver, @NejcZdovc, @bsclifton, @bbondy
Fix #9779
Fix #9895

### Reviewer note:
This PR includes work of PR #9896 (first commit). If this PR goes merged without it, please do not squash and I'll close the other. Thanks!

STR: 

1. Set tabs to 6 to make it faster
2. Open one more to have 2 tab pages
3. In the last opened tab, visit a slow website (shields on helps), i.e. like wired.com
4. Before page is loaded, set first tabPage as active.
5. Hover over a tab
6. Wait slow website to load (you can't see it but in wired it fully loads after 12 ads blocked)
7. TabPage should not jump

### DIFF
--- a/app/renderer/reducers/frameReducer.js
+++ b/app/renderer/reducers/frameReducer.js
@@ -37,7 +37,7 @@ const closeFrame = (state, action) => {
   }
 
   const frameProps = frameStateUtil.getFrameByKey(state, action.frameKey)
-  const hoverState = state.getIn(['frames', index, 'hoverState'])
+  const hoverState = frameStateUtil.getTabHoverState(state, action.frameKey)
 
   state = state.merge(frameStateUtil.removeFrame(
     state,
@@ -117,12 +117,13 @@ const frameReducer = (state, action, immutableAction) => {
         state = state.setIn(['frames', index, 'title'], title)
       }
 
+      const hoverState = frameStateUtil.getTabHoverState(state, frame.get('key'))
       // TODO fix race condition in Muon more info in #9000
       const active = immutableAction.getIn(['tabValue', 'active'])
       if (active != null) {
         if (active) {
           state = state.set('activeFrameKey', frame.get('key'))
-          if (frame.get('hoverState')) {
+          if (hoverState) {
             state = state.set('previewFrameKey', null)
           }
           if (frame.getIn(['ui', 'tabs', 'hoverTabPageIndex']) == null) {
--- a/app/renderer/reducers/frameReducer.js
+++ b/app/renderer/reducers/frameReducer.js
@@ -117,20 +117,15 @@ const frameReducer = (state, action, immutableAction) => {
         state = state.setIn(['frames', index, 'title'], title)
       }
 
-      const hoverState = frameStateUtil.getTabHoverState(state, frame.get('key'))
       // TODO fix race condition in Muon more info in #9000
       const active = immutableAction.getIn(['tabValue', 'active'])
       if (active != null) {
         if (active) {
           state = state.set('activeFrameKey', frame.get('key'))
-          if (hoverState) {
-            state = state.set('previewFrameKey', null)
-          }
           if (frame.getIn(['ui', 'tabs', 'hoverTabPageIndex']) == null) {
             state = state.deleteIn(['ui', 'tabs', 'previewTabPageIndex'])
           }
           state = state.setIn(['frames', index, 'lastAccessedTime'], new Date().getTime())
-
           state = frameStateUtil.updateTabPageIndex(state, frame)
         }
       }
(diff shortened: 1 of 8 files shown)


app/renderer/reducers/frameReducer.js
```diff
@@ -128,7 +128,14 @@ const frameReducer = (state, action, immutableAction) => {
           }
           state = state.setIn(['frames', index, 'lastAccessedTime'], new Date().getTime())
 
-          state = frameStateUtil.updateTabPageIndex(state, frame)
+          // Only update tabPage if user is not hovering other page
+          // Otherwise each time the active tab is updated the tabPage
+          // will be updated as well. While we want that behavior not necessarily
+          // the user will be on the same tabPage as the active tab. See bug #9779
+          const isPreviewing = state.get('frames').some(frame => frame.get('hoverState'))
```
bridiver: I think this check should be in `updateTabPageIndex` to make the functionality consistent wherever it is called, but there are also a few issues here I think. The first one is that we still want to updated the tab page index, but without clearing the `previewTabPageIndex`. Moving the check inside `updateTabPageIndex`, right? If the check is inside `updateTabPageIndex` that is easily accomplished. The second is that I think we want to check hover state for the tab pages and tabs `['ui', 'tabs', 'hoverTabPageIndex']`? The last is that we want to avoid iterating over frames and I can't see any reason to have a per-frame `hoverState` prop. Can we change it to `['ui', 'tabs, 'hoverTabIndex']`?

bridiver: actually is it even necessary to clearn `previewTabPageIndex` inside `updateTabPageIndex`? I think it's covered by hoverTabPageIndex now. Maybe we can just remove it? The same probably applies to `previewFrameKey` above

cezaraugusto: per-frame hoverState is needed to check whether or not the close tab icon should be shown -- a workaround I did for Aphrodite that can't handle children classes of pseudo-state. Needs to be per-frame.

Are you suggesting to add a new state that checks whether or not a tab is being hovered, like a global _clone_ of hoverState to avoid iterating over frames? I think so but I'm often confused by state names regarding tabs.

bridiver: When I say we don't need a per-frame value, I'm suggesting that we can have a single key to keep track of which frame is hovered. 
```
frames.1.hoverstate = false
frames.2.hoverstate = true
frames.3.hoverstate = false
```
changes to `ui.tabs.hoverTabIndex = 2`

cezaraugusto:
ok feedback addressed, ready for re-review

js/state/frameStateUtil.js
```diff
@@ -460,15 +460,17 @@ function isPinned (state, frameKey) {
...
 
   if (index === -1) {
     return state
   }
 
-  state = state.setIn(['ui', 'tabs', 'tabPageIndex'], index)
-  state = state.deleteIn(['ui', 'tabs', 'previewTabPageIndex'])
-
-  return state
+  // Do not update tabPageIndex if user is in hover mode
```
bbondy: Do you know why it even attempts to do this if in hover state mode?

bbondy: The concern here is that updateTabPageIndex should just do what it says and not ignore for `isTabInHoverState`.

Instead we should just call `updateTabPageIndex` when the active state of a tab changes. 
Because otherwise you could be on tab page 2 with an active tab on tab page 3. And as you're looking at tab page 2, audio could start playing on tab page 3 and it would change the active tab. 

What you did mostly fixes it, but there are some edge cases that are missed. 

I posted this issue for tracking the work, I'll merge this in the meantime:
https://github.com/brave/browser-laptop/issues/9911

You can do that task out of line with lower priority.

PR #9853, opened 2017-07-04
allow website access with long press on backbutton

### LINKED ISSUE #9852: Unable to go back to previous website with long press on back button
STR: 
1. Open github.com
2. select any link
3. press the back button long
3. select github from pulldown

Actual result: nothing happens

cezaraugusto:
Auditors: @bridiver, @luixxiul
Fix #9852

STR: 

1. Open github.com
2. select any link
3. press the back button long
4. select github from pulldown
5. Actual result: page should load

### DIFF
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -798,7 +798,8 @@ const api = {
 
   goToIndex: (tabId, index) => {
     const tab = getWebContents(tabId)
-    if (tab && !tab.isDestroyed() && tab.canGoToIndex(index)) {
+    const validIndex = index >= 0 && index < tab.getEntryCount()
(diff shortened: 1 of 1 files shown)


app/browser/tabs.js
```diff
@@ -798,7 +798,7 @@ const api = {
 
   goToIndex: (tabId, index) => {
     const tab = getWebContents(tabId)
-    if (tab && !tab.isDestroyed() && tab.canGoToIndex(index)) {
+    if (tab && !tab.isDestroyed() && tab.goToIndex(index)) {
```
bridiver: we don't want to call `goToIndex` twice, but we should check that index >= 0 and < the number of entries

cezaraugusto: ok, updated

PR #9867, opened 2017-07-05
Don't overwrite existing files when DOWNLOAD_ALWAYS_ASK is false

### LINKED ISSUE #7764: Downloading the same name file disabling the save file dialog should change the filename
## Test plan

https://github.com/brave/browser-laptop/pull/9867#issue-240628027

----

Downloading the same name file disabling the save file dialog should change the filename, like Edge does. For now the file is overwritten, which users do not always want to see.

- Brave Version (revision SHA): 0.14.0 Preview

- Steps to reproduce:
    1. Open about:preferences
    1. Specify a download path
    2. Disable "Always ask me where to save files"
    1. Download a file of a release from https://github.com/brave/browser-laptop/releases (eg RELEASES-ia32 of the Preview3)
    2. Wait until the download ends
    3. Download another file which has the same name of a **different version**  (eg RELEASES-ia32 of the **Preview2**)
    
- Actual result: the file is overwritten

- Expected result: the file should be saved with the filename changed, eg RELEASES-ia32(1)

- Any related issues: #2110

Marking this with impact/high as the users could lose their files by accident because of this setting, which behaves slightly different than the other browsers.

timborden:
Fixes #7764

Submitter Checklist:

Test Plan:
- Tested that the file was downloaded to the location specified in the settings
- Tested that a second downloaded file would not replace an existing file of the same name

Reviewer Checklist:

Tests

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -27,6 +27,7 @@ const dialog = electron.dialog
 const app = electron.app
 const uuid = require('uuid')
 const path = require('path')
+const fs = require('fs')
 const getOrigin = require('../js/lib/urlutil').getOrigin
 const {adBlockResourceName} = require('./adBlock')
 const {updateElectronDownloadItem} = require('./browser/electronDownloadItem')
@@ -541,8 +542,22 @@ function registerForDownloadListener (session) {
       itemFilename = item.getFilename()
     }
 
-    const defaultPath = path.join(getSetting(settings.DOWNLOAD_DEFAULT_PATH) || getSetting(settings.DEFAULT_DOWNLOAD_SAVE_PATH) || app.getPath('downloads'), itemFilename)
-    const savePath = ((process.env.SPECTRON || (!getSetting(settings.DOWNLOAD_ALWAYS_ASK) && !item.promptForSaveLocation())) ? defaultPath : dialog.showSaveDialog(win, { defaultPath }))
+    const defaultDir = (getSetting(settings.DOWNLOAD_DEFAULT_PATH) || getSetting(settings.DEFAULT_DOWNLOAD_SAVE_PATH) || app.getPath('downloads'))
+    let savePath
+    if (process.env.SPECTRON || (!getSetting(settings.DOWNLOAD_ALWAYS_ASK) && !item.promptForSaveLocation())) {
+      let willOverwrite = true
+      let matchedFilenames = 0
+      while (willOverwrite) {
+        savePath = path.join(defaultDir, (matchedFilenames ? `${itemFilename.replace(new RegExp(`${path.extname(itemFilename)}$`), '')} (${matchedFilenames})${path.extname(itemFilename)}` : itemFilename))
+        if (!fs.existsSync(savePath)) {
+          willOverwrite = false
+        } else {
+          matchedFilenames++
+        }
+      }
+    } else {
+      savePath = dialog.showSaveDialog(win, { defaultPath: path.join(defaultDir, itemFilename) })
+    }
 
     // User cancelled out of save dialog prompt
     if (!savePath) {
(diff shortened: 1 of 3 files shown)


app/filtering.js
```diff
@@ -520,14 +521,29 @@ function registerForDownloadListener (session) {
     }
 
     const defaultPath = path.join(getSetting(settings.DOWNLOAD_DEFAULT_PATH) || getSetting(settings.DEFAULT_DOWNLOAD_SAVE_PATH) || app.getPath('downloads'), itemFilename)
-    const savePath = ((process.env.SPECTRON || (!getSetting(settings.DOWNLOAD_ALWAYS_ASK) && !item.promptForSaveLocation())) ? defaultPath : dialog.showSaveDialog(win, { defaultPath }))
+    let savePath = ((process.env.SPECTRON || (!getSetting(settings.DOWNLOAD_ALWAYS_ASK) && !item.promptForSaveLocation())) ? defaultPath : dialog.showSaveDialog(win, { defaultPath }))
```
bridiver: I think there should be a single `if` statement here that wraps all of this, but can you switch `!getSetting(settings.DOWNLOAD_ALWAYS_ASK) && !item.promptForSaveLocation()` to `getSetting(settings.DOWNLOAD_ALWAYS_ASK)  || item.promptForSaveLocation()`? The former is confusing

app/filtering.js
```diff
@@ -520,14 +521,29 @@ function registerForDownloadListener (session) {
...
 
     // User cancelled out of save dialog prompt
     if (!savePath) {
       event.preventDefault()
       return
     }
 
+    if (!getSetting(settings.DOWNLOAD_ALWAYS_ASK) && !item.promptForSaveLocation()) {
```
bridiver: please merge into `else` for `if` above

app/filtering.js
```diff
@@ -520,14 +521,29 @@ function registerForDownloadListener (session) {
...
+      let willNotOverwrite = false
```
bridiver: double negative checks are also hard to understand `!willNotOverwrite` -> `willOverwrite`

app/filtering.js
```diff
@@ -520,14 +521,29 @@ function registerForDownloadListener (session) {
...
+      let matchedFilenames = 0
+      let querySavePath
+      while (!willNotOverwrite) {
+        querySavePath = (matchedFilenames ? savePath.substring(0, savePath.lastIndexOf('.') >= 0 ? savePath.lastIndexOf('.') : savePath.length) + ` (${matchedFilenames})` + savePath.substring(savePath.lastIndexOf('.') >= 0 ? savePath.lastIndexOf('.') : savePath.length) : savePath)
```
bridiver: I think this will break things like `.tar.gz`. Maybe get the first index?

timborden:
Thanks for the feedback @bridiver 

I've refactored the code based on your suggestions.

I also dug into the extension name edge cases (like `tar.gz`). As a result I switched from the `substring()` approach to using `path.extname()`. Sadly, it didn't solve the issue, but the code is shorter and more readable.

I did some searching for possible solutions. One thought was to use the MIME:
```
> var mime = require('mime-types')
undefined
> mime.lookup('package.json')
'application/json'
> mime.extension(mime.lookup('package.json'))
'json'
```
....sadly, it didn't work for `tar.gz` files
```
> mime.lookup('v0.17.13dev.tar.gz')
false
```

It seems that the most common approach is to split on the last "."...even the OS X save prompt defaults to using the last ".":
screen shot 2017-07-06 at 11 01 42

Short of rolling with a whitelist of file extensions, which feels like overkill....it looks like this may be the best option.

timborden:
Just added tests....FYI

luixxiul:
@timborden would you please rebase on the latest master? thanks!

timborden:
Hey @luixxiul ...sorry for the relay (just changed my GitHub settings so I don't miss these notifications)

I just rebased....FYI

luixxiul:
@timborden thanks! I'd like to see you keep up the nice work :-)

bsclifton:
@bridiver did you want to re-review before this gets merged?

bsclifton:
Merged 😄 👌

timborden:
Awesome....thanks @bsclifton, @bridiver and @luixxiul

PR #9912, opened 2017-07-08
Fix chance of having a lost window

### LINKED ISSUE #9806: Session restore can lead to lost windows if there is a hung or slow Window on shutdown
Pasting the issue that I described on Slack here:

First something things by design:
- Session-store-1 is saved every 5 minutes.
- If there is a crash on minute 4, data is restored as it was 4 minutes ago.  If this is a first time run, then no save is done it just restores to the same way they started it the first time.
- How the shutdown process works is the browser process sends a message to all windows to report their state to the browser process.  Browser process upon getting all state writes it all out into session-store-1

[3:48] 
the problem comes into play if your computer is very very slow, or if you have a hung window.

[3:49] 
In which case the hung window will be cut out of the session store.

[3:49] 
https://github.com/brave/browser-laptop/blob/master/app/index.js#L188 <- read comment there, it is the issue

[3:49] 
instead of using that window's last known good state

[3:49] 
we just save without that window

[3:50] 
I think we can store each window's last state and include at least that in the case of a hung window.

[3:50] 
and that would be a quick fix we could include in 0.18.x (or a 0.17.x hotfix if you really want)

[3:52] 
if you have a hung window in that case, whether you had a 5-minute auto save or a Quit operation, then it would lead to the cut out window issue.

[3:53] 
if all windows are hung for some reason, then all windows are lost.

[3:53] 
note that the save in that line of code was meant to protect against a hung window, so that we wouldn't lose app state. It does that job well, app state is not lost.  But it could be better by keeping last known reported window state.

bbondy [4:00 PM] 
https://github.com/brave/browser-laptop/blob/master/app/index.js#L180 <- we reset the number of windows to 0 on each initiate save there

bbondy:
Fix #9806

Submitter Checklist:

Test Plan:
`npm run unittest -- --grep="sessionStoreShutdown unit tests"`

`npm run watch-test`
`npm run test -- --grep="sessionStore test"`

Reviewer Checklist:

Tests

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -48,16 +48,15 @@ if (process.platform === 'win32') {
 
 const electron = require('electron')
 const app = electron.app
-const BrowserWindow = electron.BrowserWindow
 const ipcMain = electron.ipcMain
 const Immutable = require('immutable')
-const Updater = require('./updater')
+const updater = require('./updater')
 const Importer = require('./importer')
 const messages = require('../js/constants/messages')
-const appConfig = require('../js/constants/appConfig')
 const appActions = require('../js/actions/appActions')
 const SessionStore = require('./sessionStore')
-const AppStore = require('../js/stores/appStore')
+const {startSessionSaveInterval} = require('./sessionStoreShutdown')
+const appStore = require('../js/stores/appStore')
 const Autofill = require('./autofill')
 const Extensions = require('./extensions')
 const TrackingProtection = require('./trackingProtection')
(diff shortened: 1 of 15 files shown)


bbondy:
Ready for review pls

test/app/sessionStoreTest.js
```diff
@@ -1,42 +1,35 @@
...
 
 const Brave = require('../lib/brave')
-const Immutable = require('immutable')
 const {urlInput, navigatorBookmarked, navigatorNotBookmarked, pinnedTabsTabs, tabsTabs} = require('../lib/selectors')
 const siteTags = require('../../js/constants/siteTags')
-const siteUtil = require('../../js/state/siteUtil')
+const settings = require('../../js/constants/settings')
 
-describe('sessionStore', function () {
+describe('sessionStore test', function () {
```
bsclifton: These tests are failing for me every time ☹️  ...but it looks good in Travis-ci? 🤔

bsclifton: OK just ran them again (3rd or 4th attempt).  Worked great 😄 👍

bsclifton:
This cherry-picked into 0.19.x without issues... but I had trouble cherry-picking into 0.18.x 😦 

@bbondy can you help me resolve the conflict

app/index.js
```diff
@@ -99,105 +86,8 @@ const prefsRestartLastValue = {}
...
-      }, appConfig.quitTimeout)
-    } else {
-      saveAppState()
-    }
-  })
-}
-
 // exit cleanly on signals
-['SIGTERM', 'SIGHUP', 'SIGINT', 'SIGBREAK'].forEach((signal) => {
+;['SIGTERM', 'SIGHUP', 'SIGINT', 'SIGBREAK'].forEach((signal) => {
```
NejcZdovc: Yeah, I think it is. Because we don't use `;` and you have `}` in the last line and this line starts with `[`, so this could lead into errors

bbondy: yep it's a good practice to start lines like that with a semicolon so you don't need to worry about moved code and the preceding line of code.

bbondy:
merged to 0.18.x

Issue #5389, opened 2016-11-04
Tracking: Disabled tests


bbondy:
# Overview:
This issue tracks the disabled tests. It should always appear in the current milestone so that it's a constant reminder for us to work at re-enabling them. This will help us get green so we can better determine when to merge and when not to merge things.

# Other info:
- If a test is failing, the code that introduced it should be backed out if the test can’t be fixed right away.
- If a test is failing intermittently, it should be added to this list.  If it is newly intermittent, the task that introduced it should be backed out if it can be determined. 
- Everyone should try to get all of these tests below down to 0.
- When a test is fixed check it off and move it to the fixed intermittents section.

# Disabled tests:
 
- [ ] findBar findbar input remembered but no active ordinals after navigation until RETURN key:
  test/about/componentUpdaterTest.js
  test/components/findbarTest.js
  more info: Disabled for https://github.com/brave/browser-laptop/issues/6147
  AssertionError: '0 matches' == '1 of 1'
  expected - actual
  -0 matches
  +1 of 1

- [ ] Google Widevine can be enabled and installed
  test/about/componentUpdaterTest.js
  more info: This is only disabled on Linux and it's because the feature is not available there yet.

- [ ] should attempt a reload when reload is clicked
  test/components/errorPageTest.js
  more info: need a better way to test this
- [x] Limit effect of running insecure content in private frame
  test/components/navigationBarTest.js
  more info: Unknown.
  Owner: @darkdh
- [ ] Parses theme-color meta tag when one is present
  test/components/navigationBarTest.js
  more info: Unknown
  Owner: @bbondy 
- [ ] Obtains theme color from the background
  test/components/navigationBarTest.js
  more info: Unknown
  Owner: @bbondy
- [ ] Obtains theme color from a top header and not background
  test/components/navigationBarTest.js
  more info: Unknown
  Owner: @bbondy
- [ ] focuses the opener
  test/components/windowTest.js
  more info: Unknown
- [ ] blurs the opener
  test/components/windowTest.js
  more info: Unknown
- [ ] setting observer unit test
  test/unit/app/browser/settingObserverTest.js
  more info: I think this can just be removed along with the code since I don't think it's used.  Needs to grep for the module though to be sure.
- [ ] passes NIST Test Case 13
  test/unit/lib/cryptoUtilTest.js
  more info: fails with Error: Unsupported state or unable to authenticate data
- [x] homepage displays punycode
  test/about/preferencesTest.js
[...]

PR #9919, opened 2017-07-08
Comment unused require for rimraf to fix missing module error

### LINKED ISSUE #9918: missing rimraf dep makes build unable to launch
Rimraf was used as a dep to exclude an extension. Extension exclusion is not possible anymore but still host some of its code. At some point we removed rimraf from package.json and that is throwing an error in extensions regarding missing dep.

This happens in 0.17.x up to master. Seems that error in dist builds is silent.

STR:

1. Have a clean install
2. Start Brave (npm run watch/npm start)
3. Can't launch Brave

cezaraugusto:
This was blocking builds after a clean install.
Fix #9918 

This also happens from 0.17.x to master, but given the feature is not enabled anymore, error in dist build is silent. 

Error exists because "exclude extension" option was calling `rimraf` as a global dep but at some point we removed it. Given we can't exclude an extension via extensions panel anymore, this became a leftover and requiring a missing module was stopping build from running. 

STR:

1. Have a clean install
2. Start Brave (npm run watch/npm start)
3. Should start

### DIFF
--- a/app/browser/reducers/extensionsReducer.js
+++ b/app/browser/reducers/extensionsReducer.js
@@ -4,11 +4,12 @@
 
 'use strict'
 
-const electron = require('electron')
-const app = electron.app
+// const electron = require('electron')
+// const app = electron.app
 
-const path = require('path')
-const rimraf = require('rimraf')
+// const path = require('path')
+// TODO @cezaraugusto enable again once we have a GUI to exclude an extension
+// const rimraf = require('rimraf')
 
 const extensionState = require('../../common/state/extensionState')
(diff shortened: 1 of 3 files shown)


cezaraugusto:
I had this same error again so I'm going ahead and merge. This has no impact and only left commented a package we don't require anymore. Please ping me if there's something wrong and I'll revert and investigate. thanks.

cezaraugusto:
also moving to 0.21.x

bsclifton:
++

NejcZdovc:
@cezaraugusto please add this disabled tests onto this list #5389. Thank you

PR #9921, opened 2017-07-09
bad tabByIndex behavior

cezaraugusto:
pls see convo in automated-tests for context

**this pr is a proof of concept and should not be merged.**


bsclifton:
@cezaraugusto please re-open if we need this still

PR #9966, opened 2017-07-11
remove calls to `focusedWindow.webContents.send`

bridiver:
we don't want any references to `focusedWindow.webContents` that might be called from the renderer process

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/js/contextMenus.js
+++ b/js/contextMenus.js
@@ -1080,7 +1080,7 @@ function mainTemplateInit (nodeProps, frame, tab) {
             enabled: tab.get('canGoBack'),
             click: (item, focusedWindow) => {
               if (focusedWindow) {
-                focusedWindow.webContents.send(messages.SHORTCUT_ACTIVE_FRAME_BACK)
+                CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_ACTIVE_FRAME_BACK])
               }
             }
           }, {
@@ -1088,7 +1088,7 @@ function mainTemplateInit (nodeProps, frame, tab) {
             enabled: tab.get('canGoForward'),
             click: (item, focusedWindow) => {
               if (focusedWindow) {
-                focusedWindow.webContents.send(messages.SHORTCUT_ACTIVE_FRAME_FORWARD)
+                CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_ACTIVE_FRAME_FORWARD])
               }
             }
           }, {
@@ -1096,7 +1096,7 @@ function mainTemplateInit (nodeProps, frame, tab) {
             accelerator: 'CmdOrCtrl+R',
             click: (item, focusedWindow) => {
               if (focusedWindow) {
-                focusedWindow.webContents.send(messages.SHORTCUT_ACTIVE_FRAME_RELOAD)
+                CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_ACTIVE_FRAME_RELOAD])
               }
             }
           },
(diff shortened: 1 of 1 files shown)


luixxiul:
this was moved to the latest milestone based on https://github.com/brave/browser-laptop/pull/9966#event-1159570344

NejcZdovc:
@luixxiul why was this moved from 0.20 to 0.21? Link that you shared is linking to this PR :smiley:

luixxiul:
@NejcZdovc because it had been added to Nightly, so re-added to the latest Nightly.

NejcZdovc:
@luixxiul not sure if I understand. We change Nightly with every release

luixxiul:
If you think we don't need tests and feedbacks from the users for this PR as much as you usually do for PRs tagged with Nightly, please revert my change.

NejcZdovc:
So we set milestones based on a current master. For example this PR is quite old and when @bridiver crated this PR 0.20 was Nightly. In the meantime we released 0.18 and that means that 0.20 was moved to the developer channel and 0.21 become Nightly. Even though that we changed names of the milestones, that doesn't mean that this effects already existing milestones that are set for existing PRs.

js/contextMenus.js
```diff
@@ -472,7 +472,7 @@ function tabTemplateInit (frameProps) {
         label: locale.translation('reloadTab'),
         click: (item, focusedWindow) => {
           if (focusedWindow) {
-            focusedWindow.webContents.send(messages.SHORTCUT_FRAME_RELOAD, frameKey)
+            CommonMenu.sendToFocusedWindow(focusedWindow, messages.SHORTCUT_FRAME_RELOAD, frameKey)
```
NejcZdovc: I also see that you are passing `frameKey` and `tabId` into `CommonMenu.sendToFocusedWindow`. which only accepts `focusedWindow` and `message`

luixxiul:
> this PR is quite old

Isn't what matters more how much the change has been tested? I've believed that was why we released the previews to the users for testing on various environments. sorry if I'm quite wrong.

bsclifton:
@bridiver I get the same error as @NejcZdovc; would you be able to check this out?

bsclifton:
@bridiver can we get a rebase?  Also, see the notes above

js/contextMenus.js
bridiver: Indeed. I was counting on ci to find any problems and it looks like this needs some rethinking

bridiver: I'll just remove the calls that pass extra params for now

bridiver:
updated

NejcZdovc:
@bridiver it's still not working.

STR:
1) go to some page
2) right click on it and select `View page source` or `Print`

Result:
you get this error:
```
(WEBUI context) http://localhost:8080/gen/app.entry.js:30082: Uncaught TypeError: message.splice is not a function{TypeError: message.splice is not a function
    at Object.module.exports.sendToFocusedWindow (http://localhost:8080/gen/app.entry.js:30082:13)
    at click (http://localhost:8080/gen/app.entry.js:21986:24)
    at CallbacksRegistry.apply (extensions::callbacks-registry:61:39)
    at EventEmitter.<anonymous> (extensions::remote:284:21)
    at emitThree (extensions::event_emitter:115:13)
    at EventEmitter.emit (extensions::event_emitter:196:7)
    at EventEmitter.ipcRenderer.emit (extensions::ipc_utils:35:22)}
```

NejcZdovc:
@bridiver I pushed fix for this error. Can you please check if everything is ok and I can merge it then

js/contextMenus.js
```diff
@@ -1080,23 +1080,23 @@ function mainTemplateInit (nodeProps, frame, tab) {
             enabled: tab.get('canGoBack'),
             click: (item, focusedWindow) => {
               if (focusedWindow) {
-                CommonMenu.sendToFocusedWindow(focusedWindow, messages.SHORTCUT_ACTIVE_FRAME_BACK)
+                CommonMenu.sendToFocusedWindow(focusedWindow, [messages.SHORTCUT_ACTIVE_FRAME_BACK])
```
bridiver: odd that these are arrays. Seems like `sendToFocusedWindow` should handle `...args` and while I'm looking at it, it also seems odd that it's called `sendToFocusedWindow`, but takes a window argument. Seems like it should be `sendToWindow` or not take the `focusedWindow` as an arg

Issue #9000, opened 2017-05-22
Fix race condition for the tab title (follow up of #8908)


NejcZdovc:
**Describe the issue you encountered**
We need to fix the race condition that is caused in Muon. Because of that title is not set correctly when you are using navigation buttons. This is a follow up of #8908 where we did a workaround for it.

NejcZdovc:
fixed with #10008

bridiver:
I'm reopening this because active status is not properly handled in #10049 with respect to active state changes. #10049 does not change anything in regards to the race condition

bridiver:
the race condition itself was fixed https://github.com/brave/browser-laptop/commit/f54270f158b42544d748f6a20107eb72740eb3ff and https://github.com/brave/muon/commit/be74ba73ca8b456a039bd06e34cb925602020409
but we need to change the handling of `active` to only use `tab` if `changeInfo` is an empty object

cezaraugusto:
ok thanks for the heads-up. I'll switch back the assignment

bridiver:
yea, the active change can be handled separately

bsclifton:
Milestone removed

Issue #8429, opened 2017-04-21
tab page selection reverts to previous tab page on its own


bradleyrichter:
v0.15.0dev-RC2

Also exists in 0.15.314.

My experience is that the set changes even during tab-previewing while hunting for a tab.

1. start in the last set.
2. click on the 2nd set button.
3. start hovering over tabs to hunt for a tab.
4. see tab set selection revert back to last, and likewise, tab previews are now changed to last set.

easier to show than explain:

![tabset_bounceout](https://cloud.githubusercontent.com/assets/13509546/25298328/cc1d3808-26a8-11e7-974b-2f7f0f731cd7.gif)

jonathansampson:
Doesn't appear to repro for me:

Brave Version: https://github.com/brave/browser-laptop/commit/958aab0b1bcd460e9ffdfb7b1fc55d6d9a204b62
Operating System: Windows 10 (15063)

![ezgif com-video-to-gif](https://cloud.githubusercontent.com/assets/815158/25397228/2fc64f28-29ad-11e7-8bf6-719cd1529a30.gif)

cezaraugusto:
* issue happens when page a is still loading when you switch tab pages. 
* once page is loaded it focus on the loaded tab and set it as active, instead of keeping user choice. * Not always the loaded tab is set as active visually but tab page is always set.

bradleyrichter:
I had to switch to 100 tabs per page yesterday because otherwise can’t get work done with this bug.

cezaraugusto:
closed by https://github.com/brave/browser-laptop/commit/b397e9d592a1c7db77470afd0fdf96bc9f56d130 in #9809. I referenced the wrong commit by mistake.

PR #10008, opened 2017-07-17
Update tabPage based on active tab state

cezaraugusto:
Submitter Checklist:

Test Plan:
1. Open enough tabs to create a new tab page
2. max tabs per page + 1 should redirect to second tab page
3. Close the last tab page
4. You should be redirected to the first tab page
(below is for bug #8429)
5. Open another page so you can be in tab page 2 again
6. Go to wired.com
7. Before page fully loads, switch to tab page 1 (without setting another tab to active)
8. Tab pages shouldn't jump

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/tabs/content/tabIcon.js
+++ b/app/renderer/components/tabs/content/tabIcon.js
@@ -30,12 +30,22 @@ class TabIcon extends ImmutableComponent {
       }
     })
 
+    let altProps
+    if (!this.props.symbol) {
+      altProps = {
+        'data-test-id': this.props['data-test-id'],
+        'data-test2-id': this.props['data-test2-id']
+      }
+    }
+
     return <div
       className={this.props.className}
       data-test-favicon={this.props['data-test-favicon']}
       onDragStart={this.props.onDragStart}
       draggable={this.props.draggable}
-      onClick={this.props.onClick}>
+      onClick={this.props.onClick}
+      {...altProps}
+    >
       {
         this.props.symbol
           ? <span
--- a/app/renderer/reducers/frameReducer.js
+++ b/app/renderer/reducers/frameReducer.js
@@ -83,7 +83,10 @@ const getLocation = (location) => {
 const frameReducer = (state, action, immutableAction) => {
   switch (action.actionType) {
     case appConstants.APP_TAB_UPDATED:
+      // This case will be fired for both tab creation and tab update.
+      // being `tabValue` set for tab creation and `changeInfo` set for tab update
       const tab = immutableAction.get('tabValue')
+      const changeInfo = immutableAction.get('changeInfo')
       if (!tab) {
         break
       }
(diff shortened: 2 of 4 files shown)


app/renderer/components/tabs/content/tabIcon.js
```diff
@@ -30,12 +30,22 @@ class TabIcon extends ImmutableComponent {
...
+    }
+
     return <div
       className={this.props.className}
       data-test-favicon={this.props['data-test-favicon']}
       onDragStart={this.props.onDragStart}
       draggable={this.props.draggable}
-      onClick={this.props.onClick}>
+      onClick={this.props.onClick}
+      {...altProps}
```
cezaraugusto: @NejcZdovc I cherry-picked this from #9923 but I'm unsure why do we need it?

NejcZdovc: this is needed, because we add `data-test-id` to span element, but click is added to div element and if you want to do `click` in the test you need to have `data-test-id` on div as well.

app/renderer/reducers/frameReducer.js
```diff
@@ -117,22 +117,32 @@ const frameReducer = (state, action, immutableAction) => {
...
+      // when a tab is set to active, and will not be updated for the same tab
+      // as long as it remains active.
+      // This likely deprecates tabValue but still lacks state for fresh profile.
+      // TODO check if changeInfo is a better fit than tabValue and make use
+      // of it app-wise and deprecate the latter.
+      const active = immutableAction.getIn(['changeInfo', 'active'])
+
+      // This is a workaround for changeInfo which do not provide
+      // active state on a fresh profile.
+      if (active == null) {
```
bridiver: Actually we want to check to make sure that `changeInfo` is not null. If changeInfo.active is null that just means that the active status hasn't changed

app/renderer/reducers/frameReducer.js
```diff
@@ -94,7 +97,7 @@ const frameReducer = (state, action, immutableAction) => {
       }
 
       const index = frameStateUtil.getIndexByTabId(state, tabId)
-      const pinned = immutableAction.getIn(['changeInfo', 'pinned'])
+      const pinned = tabUpdated.get('pinned')
```
bridiver: tabUpdated could potentially be undefined for all of these

bridiver: actually this comment is wrong because it defaults to an empty object

app/renderer/reducers/frameReducer.js
```diff
@@ -112,26 +115,28 @@ const frameReducer = (state, action, immutableAction) => {
       }
 
       // TODO fix race condition in Muon more info in #9000
```
NejcZdovc: this TODO can be removed now. Can you please remove it in the follow up and you can close #9000 with it? :smiley:

app/renderer/reducers/frameReducer.js
```diff
@@ -112,26 +115,28 @@ const frameReducer = (state, action, immutableAction) => {
...
-      const title = immutableAction.getIn(['tabValue', 'title'])
+      const title = tab.get('title')
       if (title != null) {
         state = state.setIn(['frames', index, 'title'], title)
       }
 
       // TODO fix race condition in Muon more info in #9000
```
NejcZdovc: this TODO can be removed now. Can you please remove it in the follow up and you can close #9000 with it? :smiley:

bridiver: @NejcZdovc why would we remove the TODO? This doesn't change anything

bridiver: I believe the race condition no longer exists, but `active` should be changed to check `changeInfo.active` if `changeInfo` is not empty and if `changeInfo` is empty (no attributes at all) it should check `tab`

NejcZdovc: yeah I think that race condition is no longer present, that's why I suggested to do a follow up, remove todos and close issue #9000

bridiver: the race condition no longer exists, but this code does not fix #9000

PR #10010, opened 2017-07-17
Optimize getToolbarBookmarks

### LINKED ISSUE #10009: Optimize getToolbarBookmarks
getToolbarBookmarks is expensive and it's currently called in mergeProps which would be the reason why things have gotten worse with perf.

screen shot 2017-07-16 at 10 51 30 pm

After optimizing:
screen shot 2017-07-16 at 11 16 45 pm

bbondy:
Fix #10009

Submitter Checklist:

  - Already covered by existing tests

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/common/lib/bookmarkUtil.js
+++ b/app/common/lib/bookmarkUtil.js
@@ -65,8 +65,19 @@ const getDNDBookmarkData = (state, bookmarkKey) => {
   return data.get('draggingOverKey') === bookmarkKey ? data : Immutable.Map()
 }
 
+let oldSites
+let lastValue
+let lastWidth
 const getToolbarBookmarks = (state) => {
   const sites = state.get('sites', Immutable.List())
+  const windowWidth = window.innerWidth
+  if (sites === oldSites &&
+      lastWidth === windowWidth &&
+      lastValue) {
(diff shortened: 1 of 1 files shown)


bbondy:
Found this is new with redux so I moved it to 0.19.x

app/common/lib/bookmarkUtil.js
```diff
@@ -65,8 +65,18 @@ const getDNDBookmarkData = (state, bookmarkKey) => {
   return data.get('draggingOverKey') === bookmarkKey ? data : Immutable.Map()
 }
 
+let oldSites
+let lastValue
+let lastWidth
 const getToolbarBookmarks = (state) => {
   const sites = state.get('sites', Immutable.List())
+  if (sites === oldSites &&
+      lastWidth === window.innerWidth &&
```
bridiver: window width is available in appstate.windows, but at the very least we should call this once and save the value because each call forces a layout recalc in webkit

bbondy: good catch, updated PR to only do it once for now, which makes it the same number as before (once).

PR #10047, opened 2017-07-18
change about:brave incorrect windows platform info

### LINKED ISSUE #8933: about:brave shows win32 for all Windows platforms (even x64)
If you look at about:brave on 64 bit Windows, you'll notice it reports `win32` as the platform.  [This value is coming from node.js](https://nodejs.org/api/os.html#os_os_platform)
screen shot 2017-05-18 at 9 31 28 am

This is confusing because it looks like something is 32 bit, even though we are clearly running the 64 bit version of the software and Windows is 64 bit.

The solution would be to do a string replacement. When `win32` is returned, we can substitute in a string like "Microsoft Windows".  We could put this logic into the [platformUtil module](https://github.com/brave/browser-laptop/blob/7452b7ec81f8b501a0b747829fa81090a67becee/app/common/lib/platformUtil.js) and update the part of the code which currently updates the values on about:brave:
https://github.com/brave/browser-laptop/blob/7452b7ec81f8b501a0b747829fa81090a67becee/app/sessionStore.js#L417

baoqchau:
Submitter Checklist:
Fix #8933

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/common/lib/platformUtil.js
+++ b/app/common/lib/platformUtil.js
@@ -47,3 +47,16 @@ module.exports.isLinux = () => {
   return !module.exports.isDarwin() &&
     !module.exports.isWindows()
 }
+
+module.exports.formatOsPlatform = (os) => {
+  let platformMapping = {
+    'aix': 'AIX',
+    'darwin': 'macOS',
+    'freebsd': 'FreeBSD',
+    'linux': 'Linux',
+    'openbsd': 'OpenBSD',
+    'sunos': 'sunOS',
+    'win32': 'Microsoft Windows'
+  }
+  return Object.keys(platformMapping).includes(os) ? platformMapping[os] : os
+}
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -15,6 +15,7 @@ const Immutable = require('immutable')
 const fs = require('fs-extra')
 const path = require('path')
 const electron = require('electron')
+const os = require('os')
 const app = electron.app
 const locale = require('./locale')
 const UpdateStatus = require('../js/constants/updateStatus')
@@ -32,6 +33,7 @@ const { makeImmutable } = require('./common/state/immutableUtil')
 const tabState = require('./common/state/tabState')
 const windowState = require('./common/state/windowState')
 
+const platformUtil = require('./common/lib/platformUtil')
 const getSetting = require('../js/settings').getSetting
 const sessionStorageName = `session-store-${sessionStorageVersion}`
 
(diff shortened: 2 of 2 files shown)


bsclifton:
@baoqchau instead of doing the formatting there, what do you think about adding a new method to the platformUtil helper? Maybe something like `formatOsPlatform`

It could be called like:
`platformUtil.formatOsPlatform(require('os').platform)`

That way, if we'd like to make more readable names for macOS or Linux, we could do that too 😄

baoqchau:
Thanks for your suggestion. Sounds good, I will do it.

bsclifton:
@baoqchau thanks for taking this on 😄 

The changes you're proposing look good, with a two exceptions:
`'darwin': 'macOS'`
`'sunos': 'SunOS'`

If we wanted to get fancy, we could take all three parts (arch, platform, version) and build a string like so:
`Microsoft Windows (64-bit, 10.0.14393)`

app/common/lib/platformUtil.js
```diff
@@ -47,3 +50,16 @@ module.exports.isLinux = () => {
...
+  let platformMapping = {
+    'aix': 'AIX',
+    'darwin': 'macOS',
+    'freebsd': 'FreeBSD',
+    'linux': 'Linux',
+    'openbsd': 'OpenBSD',
+    'sunos': 'sunOS',
+    'win32': 'Microsoft Windows'
+  }
+  return platformMapping[os]
```
NejcZdovc: This will be a good one ++. Can you just please handle null scenario as well. Let's say that os is not on our list we would like to return just os value as is and not undefined/null. Thank you

app/common/lib/platformUtil.js
```diff
@@ -1,4 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
+ *
+ *
+ *
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
```
bsclifton: whoops- I think the above white space wasn't intended

app/sessionStore.js
```diff
@@ -66,8 +68,7 @@ const getStoragePath = () => {
  * @param {object} payload - Application state as per
  *   https://github.com/brave/browser/wiki/Application-State
  *   (not immutable data)
- * @return a promise which resolves when the state is saved
- */
+ * @return a promise which resolves when the state is saved */
 module.exports.saveAppState = (payload, isShutdown) => {
```
bsclifton: same here- can you leave the `*/` on the next line? this matches the style we have in the rest of the app 😄

NejcZdovc:
@baoqchau can you please squash this into one commit? Then I think we are ready to go :smiley:

bsclifton:
Merged! Congrats on your first contribution @baoqchau 😄  Please hit us up anytime if you need help with an issue or if you need help finding an issue 😄

PR #10052, opened 2017-07-19
Sync: Add fetch delay to mitigate S3 inconsistency

ayumi:
Related to https://github.com/brave/sync/issues/139

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/sync.js
+++ b/app/sync.js
@@ -349,7 +349,8 @@ module.exports.onSyncReady = (isFirstRun, e) => {
       }
     }
     e.sender.send(syncMessages.FETCH_SYNC_RECORDS, categoryNames, startAt)
-    startAt = syncUtil.now()
+    // Reduce syncUtil.now() by this amount to compensate for records pending S3 consistency. See brave/sync #139
+    startAt = syncUtil.now() - config.fetchOffset
     appActions.saveSyncInitData(null, null, startAt)
     resendPendingRecords()
   }
--- a/js/constants/appConfig.js
+++ b/js/constants/appConfig.js
@@ -116,6 +116,7 @@ module.exports = {
     testS3Url: 'https://brave-sync-test.s3.dualstack.us-west-2.amazonaws.com/',
     s3Url: isProduction ? 'https://brave-sync.s3.dualstack.us-west-2.amazonaws.com' : 'https://brave-sync-staging.s3.dualstack.us-west-2.amazonaws.com',
     fetchInterval: isProduction ? (1000 * 60 * 3) : (1000 * 60),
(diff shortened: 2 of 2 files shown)


bsclifton:
I'll have to dig into this some more... but in browser => browser testing, I'm seeing bookmarks not sync when both are open. However, after I quit/relaunch, it syncs up just fine. 

My steps:
1. (pyramid1) `npm start` with a session that has a decent # bookmarks
2. (pyramid2) `npm run start2` with a different session (has leftover bits from last sync, before I reset it)
3. enable sync on pyramid1
4. use keywords and join pyramid 1's group on pyramid2
5. Add some folders + bookmarks on pyramid1
6. it doesn't get synced
7. exit both browser instances
8. relaunch (step 1 + 2), it syncs up

app/sync.js
```diff
@@ -349,7 +349,8 @@ module.exports.onSyncReady = (isFirstRun, e) => {
       }
     }
     e.sender.send(syncMessages.FETCH_SYNC_RECORDS, categoryNames, startAt)
-    startAt = syncUtil.now()
+    // Reduce syncUtil.now() by this amount to compensate for records pending S3 consistency. See brave/sync #139
+    startAt = syncUtil.now() - config.fetchOffset
```
bsclifton: is this correct? or should it be `+ config.offset`?

ayumi: this is correct– this is "Find records whose timestamps are after 30s ago", looking in the past because records created 0–30s ago could be inconsistent.

ayumi:
@bsclifton that is weird– when the sync is failing, do you get anything in the sync extension console log (go to chrome-extension://cjnmeadmgmiihncdidmfiabhenbggfjm/_generated_background_page.html and open dev tools)?

(I tried your test plan twice and it seemed ok for me.)

LaurenWags:
Post-Sync import Laptop-Laptop scenario (complex case) - this scenario https://github.com/brave/sync/issues/146#issuecomment-316423654 with different pyramids:
Pyramid 0 was Linux/Ubuntu VM (without the PR)
Pyramid 1 was MacOS laptop (with the PR)

No bookmarks/folders were lost and both pyramids were open during sync. Approx 6k bookmarks were synced.

app/sync.js
bridiver: shouldn't you get the new startAt time (in a tmp var) before sending the `FETCH_SYNC_RECORDS` request? It's extremely unlikely now that you've added the offset, but in a resource starved machine there is still a potential for a gap between the sync end/next start times

SergeyZhukovsky: it is a good point, we do it before on Android

Issue #8761, opened 2017-05-08
History entries often cannot be deleted


Liunkae:
- Did you search for similar issues before submitting this one?
Yes.

- Describe the issue you encountered:
Often when using `Delete History Entry`, the history entry's timestamp is removed but the item is not deleted.

- Platform (Win7, 8, 10? macOS? Linux distro?):
Found on macOS 10.12.x.

- Brave Version (revision SHA):
0.15.2 (af7ef42)

- Steps to reproduce:
    1. Open a new Brave tab and visit github.com (make sure you are logged out).
    2. Visit each of the options at the top (Features, Business, Explore, and Pricing).
    3. Open about:history.
    4. Right-click one of the entries (e.g., Pricing) and click Delete History Entry.
    
- Actual result:
The history entry is moved to a section with no timestamp and is not deleted (see screenshot).

- Expected result:
The history entry should be completely deleted.

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
Not sure.

- Is this an issue in the currently released version?
Yes, 0.15.2.

- Can this issue be consistently reproduced?
Yes, with the steps above. It doesn't happen with every single time, though.

- Extra QA steps:
    1. Open a new Brave tab and visit github.com (make sure you are logged out).
    2. Visit each of the options at the top (Features, Business, Explore, and Pricing).
    3. Open about:history.
    4. Right-click one of the entries (e.g., Pricing) and click Delete History Entry.
    5. Make sure the history item is deleted.

- Screenshot if needed:
![image](https://cloud.githubusercontent.com/assets/19424103/25813437/2c04d968-33df-11e7-85d5-d7e20e21680b.png)

- Any related issues:
None.

eljuno:
It happens to me, too, 0.15.2 (af7ef42).
I reported in community few days ago. https://community.brave.com/t/long-history-list-with-no-time-and-can-not-be-deleted/2693?u=eljuno

For now, I can delete it by clearing browsing data (`Ctrl+Shift+Del`). The history completly vanish. But when I manualy delete the entry, it repeat again.

Here's an additional record:
![brave-history-bug](https://cloud.githubusercontent.com/assets/25360853/25823233/9c607290-3465-11e7-8014-417d982d70b7.gif)

Liunkae:
@eljuno Thanks for the screen recording! Looks like it's happening on Windows too.

eljuno:
@Liunkae ah, yes. I'm on Windows 8.1.

[...]

PR #10069, opened 2017-07-19
Fix deleting history entry

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/8761

test plan:
1. open a new tab, go to some site
2. close the tab
3. open about:history. right-click to delete the site. it should disappear from the list
4. open the History menu. you should no longer see the site under 'Recently Closed'.
5. repeat above steps but opening/closing two tabs and selecting both of them to delete instead of only one

I tried to write a test for this but couldn't find a way to get Webdriver to click on a context menu item in about:history.

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/js/state/siteUtil.js
+++ b/js/state/siteUtil.js
@@ -417,6 +417,11 @@ module.exports.removeSite = function (state, siteDetail, tag, reorder = true, sy
     site = site.set('tags', tags)
     return state.setIn(stateKey, site)
   } else {
+    const siteDetailTags = siteDetail.get('tags')
+    if (!tag && (!siteDetailTags || siteDetailTags.size === 0)) {
+      // Delete the site from history
+      return state.deleteIn(stateKey)
+    }
     site = site.set('lastAccessedTime', undefined)
     return state.setIn(stateKey, site)
   }
--- a/app/browser/menu.js
+++ b/app/browser/menu.js
@@ -673,8 +673,15 @@ const doAction = (state, action) => {
       }
     case windowConstants.WINDOW_CLEAR_CLOSED_FRAMES:
       {
-        closedFrames = new Immutable.OrderedMap()
-        lastClosedUrl = null
+        if (!action.location) {
+          closedFrames = new Immutable.OrderedMap()
+          lastClosedUrl = null
+        } else {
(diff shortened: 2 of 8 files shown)


js/state/siteUtil.js
```diff
@@ -417,6 +417,10 @@ module.exports.removeSite = function (state, siteDetail, tag, reorder = true, sy
     site = site.set('tags', tags)
     return state.setIn(stateKey, site)
   } else {
+    if (!tag && module.exports.isHistoryEntry(siteDetail)) {
+      // Delete the site from history
+      return state.deleteIn(stateKey)
```
darkdh: We don't delete history for a reason which is to prevent https://github.com/brave/browser-laptop/issues/7238 from happening. This line will cause the regression of #7238.
Maybe we should revisit this after #10136 landed

diracdeltas: @darkdh `module.exports.isHistoryEntry` verifies that the site is not a bookmark or a pinned tab before deleting it. but you're right that #10136 probably affects this PR

diracdeltas: #10136 is in the 0.21.x milestone whereas this issue should be fixed in 0.19.x

darkdh: but I tried this PR with STR in #7238. bookmark will disappear and pinned tab will also disappear after relaunch

diracdeltas:
@darkdh good catch, i think it's fixed now

diracdeltas:
Merging. need to make sure this works after #10296

luixxiul:
I added `reverted` but cannot remember which commit reverted this one.

diracdeltas:
@luixxiul should https://github.com/brave/browser-laptop/issues/8761 and https://github.com/brave/browser-laptop/issues/10328 be reopened then?

bsclifton:
@luixxiul was this reverted?  Per @diracdeltas comment above, should the two issues she mentioned be reopened?

luixxiul:
I think I misunderstood something. please ignore https://github.com/brave/browser-laptop/pull/10069#event-1196295632.

Issue #5614, opened 2016-11-14
OpenSuse brave package doesn't include all of the required requirements


posix4e:
OpenSuse brave requires that you install
gconf2-devel and libgnome-keyring-devel

to work correctly

mati865:
It must have been something different than `libgnome-keyring-devel`:
```
	linux-vdso.so.1 (0x00007ffc987e7000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f91c6126000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f91c5f22000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f91c5d1a000)
	libgmodule-2.0.so.0 => /usr/lib64/libgmodule-2.0.so.0 (0x00007f91c5b16000)
	libgobject-2.0.so.0 => /usr/lib64/libgobject-2.0.so.0 (0x00007f91c58c2000)
	libglib-2.0.so.0 => /usr/lib64/libglib-2.0.so.0 (0x00007f91c55aa000)
	libfontconfig.so.1 => /usr/lib64/libfontconfig.so.1 (0x00007f91c5366000)
	libdbus-1.so.3 => /lib64/libdbus-1.so.3 (0x00007f91c5115000)
	libnss3.so => /usr/lib64/libnss3.so (0x00007f91c4dec000)
	libnssutil3.so => /usr/lib64/libnssutil3.so (0x00007f91c4bbe000)
	libsmime3.so => /usr/lib64/libsmime3.so (0x00007f91c4996000)
	libnspr4.so => /usr/lib64/libnspr4.so (0x00007f91c4755000)
	libgio-2.0.so.0 => /usr/lib64/libgio-2.0.so.0 (0x00007f91c43b5000)
	libexpat.so.1 => /usr/lib64/libexpat.so.1 (0x00007f91c4182000)
	libX11.so.6 => /usr/lib64/libX11.so.6 (0x00007f91c3e40000)
	libX11-xcb.so.1 => /usr/lib64/libX11-xcb.so.1 (0x00007f91c3c3e000)
	libxcb.so.1 => /usr/lib64/libxcb.so.1 (0x00007f91c3a14000)
	libXcomposite.so.1 => /usr/lib64/libXcomposite.so.1 (0x00007f91c380f000)
	libXcursor.so.1 => /usr/lib64/libXcursor.so.1 (0x00007f91c3604000)
[...]

PR #10107, opened 2017-07-24
Add alternative dependencies for RPM

### LINKED ISSUE #6946: Brave will not install on OpenSuse

https://github.com/brave/browser-laptop/issues/5614
https://github.com/brave/browser-laptop/issues/5580

Yast complained that there was no GConf2 provider available.

OpenSuse 42.2

0.13.0 (edit and 0.13.1)

- Steps to reproduce:
   1. Download brave-0.13.0.x86_64.rpm
    2. Select in Dolphin file manager and pick Install from the right click menu
    3. Provide root password to Yast
    4. Yast shows error message.

brave install fail

I have the packages in related reports loaded and 0.12.15 is running with no issues.

gconf2

keyring

----------------------

Trying the alternate install option via the command line I get these errors:

`p3620:/home/data/download # rpm -i brave-0.13.0.x86_64.rpm
error: Failed dependencies:
        GConf2 is needed by brave-0.13.0-1.x86_64
        libXScrnSaver is needed by brave-0.13.0-1.x86_64
`

mati865:
Addresses #6946

RPM 4.13.0 was released on 2016-11-03, and it is available on [all supported Fedora versions](https://apps.fedoraproject.org/packages/rpm) and openSUSE Tumbleweed. It added support for boolean dependencies, which we can use to resolve the installation issue on openSUSE Tumbleweed.

This PR does not fix the installation issue on openSUSE Leap. openSUSE Leap users will have to install [the unofficial update](https://build.opensuse.org/package/show/home%3ALedest%3Arpm/rpm).

Submitter Checklist:

Test Plan:
 - Make sure this command installs brave on openSUSE Tumbleweed: `LANG=C sudo zypper in '~/.brave-0.20.0.x86_64.rpm'`
- Make sure this command installs brave on Fedora: `LANG=C sudo dnf install '~/.brave-0.20.0.x86_64.rpm'`

Reviewer Checklist:

Tests

### DIFF
--- a/res/linuxPackaging.json
+++ b/res/linuxPackaging.json
@@ -18,8 +18,8 @@
     "x-scheme-handler/ftp"
   ],
   "requires": [
-    "GConf2",
-    "libXScrnSaver",
+    "(GConf2 or gconf2)",
+    "(libXScrnSaver or libXss1)",
     "lsb"
   ],
   "depends": [


mati865:
Should I also include fix for https://github.com/brave/browser-laptop/issues/5614 here?

luixxiul:
@mati865 not necessary, still if possible, that would be appreciated :-)

mati865:
@luixxiul I cannot reproduce https://github.com/brave/browser-laptop/issues/5614 with Tumbleweed GNOME or KDE virtual machines after fix.

luixxiul:
@mati865 if this PR fixes #5614, then please include `Fixes #5614` inside the commit message or in the 1st post of this PR. thanks.

mati865:
@luixxiul looks like #5614 was already invalid before my PR.

luixxiul:
@mati865 so do you mean this PR does not fix #5614, right?

mati865:
@luixxiul this issue was either fixed without being noticed or it was invalid.

To install Brave on openSUSE one have to ignore it's dependencies and installing `*-devel` packages mentioned in  #5614 could pull those dependencies but it's overkill for users. So marking it as invalid fits more.

res/linuxPackaging.json
```diff
@@ -18,8 +18,8 @@
     "x-scheme-handler/ftp"
   ],
   "requires": [
-    "GConf2",
-    "libXScrnSaver",
+    "(GConf2 or gconf2)",
+    "(libXScrnSaver or libXss1)",
```
Garbee: These kinds of things won't work with the current OpenSUSE Leap for example. It is only running RPM Version 4.11.2 by default. It needs a specific package built for it.

mati865: I mentioned it fixes only on Fedora and oS Tumbleweed.

Garbee: We need a solution that resolves Leap as well. Otherwise that's the entire stable desktop platform of openSUSE left out. And in resolving it for Leap, that will resolve it for Tumbleweed as well.

mati865: I'll close this PR if someone opens PR with better solution.

For now let's put it on hold.

Garbee: There is an issue open for *discussing* solutions before working on the actual code side. That way when work is done we are sure what happens is going to at least have a chance.

mati865: Alternatively I'll remove "Fixes" keyword so this can be merged as fix for Tumbleweed and issue will remain open for Leap users.

Garbee: What effect does this change have on leap though? Would the build's output RPM even continue to work with the new keyword on an older system?

Having *one* solution that works for more systems is better than rolling out two solutions that work in different ways. Hence why I am proposing in the main issue having a second build just for OpenSUSE. I am waiting on input from the team as to what would need to be done and allowed before putting in the work on something that may get rejected since it isn't setup how they'd prefer it get hosted.

mati865: Nothing has changed for Leap:
```
Problem: nothing provides (GConf2 or gconf2) needed by brave-0.20.0-1.x86_64
 Solution 1: do not install brave-0.20.0-1.x86_64
 Solution 2: break brave-0.20.0-1.x86_64 by ignoring some of its dependencies

Choose from above solutions by number or cancel [1/2/c] (c):
```

Considering this issue is 6 months old I still think fixing it for Tumbleweed is better than having it broken on both.

luixxiul: @Garbee do you mean this PR **breaks** the installation on the current stable OpenSUSE version, or the installation has never worked on OpenSUSE at all?

If former, we should not merge the PR.
If latter, I think the PR is still acceptable.
If this PR does not fix anything, it should be closed.

Do you have a better solution for the leap? If so, would you mind letting us know what it is like? thanks!

luixxiul:
@mati865 I updated the test plan on the 1st post to make it simple for people who are not familiar with OpenSUSE like me :-D

res/linuxPackaging.json
mati865: @luixxiul it never worked on both Leap (stable) and Tumbleweed (rolling).
With this PR Tumbleweed is fixed and Leap remains unchabged (broken).

Maybe I should remove "fixes" tag?

mati865:
@luixxiul my bad, I haven't made it clear enough.

res/linuxPackaging.json
luixxiul: @mati865 Yes. I replaced it with `Addresses` on the 1st post for you. Would you mind updating the commit message? thanks!

mati865: @luixxiul no problem, I'll rebase it when I return home.

mati865:
@posix4e seems to be working on better solution, closing.

posix4e:
@mati865  it's true!

PR #10161, opened 2017-07-26
Don't use the previous URL if current virtual URL is safebrowsing

bsclifton:
Fixes https://github.com/brave/browser-laptop/issues/10143

Auditors: @diracdeltas, @alexwykoff

Test Plan:
1. open new tab
2. navigate to downloadme.org
3. notice that URL is properly "about:safebrowsing#downloadme.org"

Submitter Checklist:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -302,7 +302,8 @@ const fixDisplayURL = (navigationEntry, controller) => {
     navigationEntry.virtualURL = getSourceAboutUrl(navigationEntry.virtualURL)
   }
 
-  if (isIntermediateAboutPage(navigationEntry.virtualURL)) {
+  if (isIntermediateAboutPage(navigationEntry.virtualURL) &&
+    !navigationEntry.virtualURL.startsWith('about:safebrowsing#')) {
     const previousEntry = controller.getEntryAtOffset(-1)
     if (!controller.canGoForward() && previousEntry) {
       navigationEntry.virtualURL = previousEntry.virtualURL


app/browser/tabs.js
```diff
@@ -302,7 +302,8 @@ const fixDisplayURL = (navigationEntry, controller) => {
     navigationEntry.virtualURL = getSourceAboutUrl(navigationEntry.virtualURL)
   }
 
-  if (isIntermediateAboutPage(navigationEntry.virtualURL)) {
+  if (isIntermediateAboutPage(navigationEntry.virtualURL) &&
+    navigationEntry.virtualURL.indexOf('about:safebrowsing#') === -1) {
```
diracdeltas: if this is an issue for about:safebrowsing, shouldn't it also be an issue for about:error, about:certerror, etc?

diracdeltas: maybe not because those errors fire after the page has started loading already but about:safebrowsing fires before

diracdeltas: i think this should do a more strict check, like `!virtualUrl.startsWith...`

bsclifton: @diracdeltas I'm not sure, maybe you can help me understand

If you open a new tab and then nav to downloadme.org, filtering picks it up via https://github.com/brave/browser-laptop/blob/0.18.x/app/filtering.js#L150 and restarts the nav process with "about:safebrowsing#downloadme.org", causing the previous page to be the full chrome-extension://about-newtab.html

If you navigate to a site and turn off your network card while the connection is happening, I would expect the browser to have an error page... but have the proper URL in the url bar (ex: the site I'm trying to reach). I was under the impression that a redirect happens after the site is unreachable (so the previous entry would be the correct site)

bsclifton: doh- missed your previous comments 😛  (needed to refresh). Let me implement a better check 😄

bsclifton:
@diracdeltas updated 😄

Issue #10062, opened 2017-07-19
Update aphrodite to 1.2.4


luixxiul:
## Test plan
https://github.com/brave/browser-laptop/issues/10234
1. Make sure the issue reported on #10029 does not happen again

----

See https://github.com/Khan/aphrodite/blob/master/CHANGELOG.md for changelog

bsclifton:
Removing milestone and marking as `reverted`. This was removed with  which was needed to fix https://github.com/brave/browser-laptop/issues/10029.

We should be able to consider upgrading again once https://github.com/brave/browser-laptop/issues/10133 is finished

Please note that per semver, the ^ means any compatible version.  Having `^1.0.0` was already grabbing the latest 1.0 release.  More info available at https://docs.npmjs.com/files/package.json

luixxiul:
why don't we report the bug to the upstream? :-)

cezaraugusto:
@luixxiul actually that's not a bug with Aphrodite. It's not clear where it was added but Aphrodite checks for production env var and concatenates strings. It should be nothing for us be we still rely on legacy code that depends on className. 

For example in main.js we are checking matches against `classList` like `node.matches('[class^="popupWindow"]')`. It works in dev but as soon as build is ready, there's no class starting with `popupWindow`, so we always return false for such checks in production.

The fastest take to unblock next release was to freeze Aphrodite before that addition. But short term is to eliminate all class-dependent code as they're fragile and just re-use Aphrodite at latest.

bsclifton:
@luixxiul if you upgrade, please be sure not to use the ^ in the version 😄  (ex: prefer "1.2.3" over "^1.2.3")

luixxiul:
Here's a discussion if another variable should be applied to minify classname: https://github.com/Khan/aphrodite/pull/246#issuecomment-319047097

bsclifton:
Removing milestone.  We can attempt this once https://github.com/brave/browser-laptop/issues/10551 is completed

luixxiul:
aphrodite was updated to `1.2.4`.

https://github.com/Khan/aphrodite/blob/master/CHANGELOG.md#124

PR #10165, opened 2017-07-27
do not rely on classList for event target checks

### LINKED ISSUE #10133: Rework code to not use class names
## Test plan

https://github.com/brave/browser-laptop/pull/10165#issue-245912630

----

Discovered by @jonathansampson and @kevinlawler while troubleshooting https://github.com/brave/browser-laptop/issues/10029

### Our current theory on the root cause:
The Aphrodite library just introduced a minify option (specifically with https://github.com/Khan/aphrodite/commit/7f9f6028ec9600d8f20f39ac54431d8b801029d9).  Unfortunately, we have checks which rely on the class name and we DO set NODE_ENV = production while doing a build (which triggers this option OR a similar option).

### Example of problem
Here's an example of the issue, showing the difference in the same element between two versions of Brave (notice the `class` attribute is different):
0.17.16:
> `<div class="popupWindow_yfkq6x-o_O-reverseExpand_142fl4c" style="height: 459px; top: 42px; width: 352px; right: 1em;">...</div>`

0.17.17:
> `<div class="_yfkq6x-o_O-_142fl4c" style="height: 459px; top: 44px; width: 352px; right: 1em;">...</div>`

### The proposed fix:
Update the following code to check using attributes, not class names:
https://github.com/brave/browser-laptop/blob/2e92a017151b77d87031fef4e030ced9dd33f00f/app/renderer/components/main/main.js#L481

cezaraugusto:
Submitter Checklist:

Auditors: @bsclifton, @jonathansampson
Close #10133

Test Plan:
```
npm run test -- --grep="eventUtil"
```

Related: #10062 

Manual test plan (cc @luixxiul):

1. Update Aphrodite to `1.2.3` in package.json
2. `rm -rf node_modules/ && rm -rf ~/.electron && npm i`
3. `NODE_ENV=production npm run watch`
4. `npm start`
5. Open Devtool with Shift+F8.
6. Check that styles for tabs should show the style concatenated
7. You should be able to click extension button and open extensions popup, which should be persistent;
8. Bookmark hanger should work as expected;
9. Right click over any of the above should not close the dialog.

Reviewer Checklist:

Tests

### DIFF
--- a/js/lib/eventUtil.js
+++ b/js/lib/eventUtil.js
@@ -10,6 +10,11 @@ module.exports.isForSecondaryAction = (e) =>
   e.button === 1
 
 module.exports.eventElHasAncestorWithClasses = (e, classesToCheck) => {
+  // DO NOT ADD NEW CHECKS USING THIS METHOD
+  // classNames are changed from dev to prod by Aphrodite
+  // and new code will not work. Consider using dataset attribute instead.
+  // ....
+  // TODO deprecate this method.
   let node = e.target
 
   while (node) {
@@ -27,4 +32,21 @@ module.exports.eventElHasAncestorWithClasses = (e, classesToCheck) => {
   }
 
   return false
+}
+
+/**
+ * Checks if a given node dataset matches a given dataset or array of datasets
+ * @param {Object} The node to check if dataset is included
+ * @param {Array|String} the dataset value to check against
+ * @returns {Boolean} Whether or not the given dataset is included in the check
+ */
+module.exports.elementHasDataset = (node, datasetArray) => {
+  if (!node.dataset) {
+    return false
+  }
+
+  const datasetToMatch = Array.isArray(datasetArray) ? datasetArray : [datasetArray]
+  const elementDataset = Object.keys(node.dataset)
+
+  return elementDataset.some(dataset => datasetToMatch.includes(dataset))
 }
--- a/js/lib/eventUtil.js
+++ b/js/lib/eventUtil.js
@@ -11,8 +11,9 @@ module.exports.isForSecondaryAction = (e) =>
 
 module.exports.eventElHasAncestorWithClasses = (e, classesToCheck) => {
   // DO NOT ADD NEW CHECKS USING THIS METHOD
-  // classNames are changed from dev to prod by Aphrodite
+  // classNames are changed from dev to prod by Aphrodite est. v1.2.3
   // and new code will not work. Consider using dataset attribute instead.
+  // See issue #10029 for a breaking example.
   // ....
   // TODO deprecate this method.
   let node = e.target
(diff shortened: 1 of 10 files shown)


luixxiul:
@cezaraugusto to test the effectiveness of this PR, Aphrodite should be updated to `1.2.3`, where we had the issue on the packaged build (because on that version Aphrodite started shrinking the classnames when NODE_ENV = production is set).

luixxiul:
@cezaraugusto @jonathansampson @bsclifton I updated the manual test plan. Would you mind confirming that it works? thanks!

js/lib/eventUtil.js
```diff
@@ -10,6 +10,11 @@ module.exports.isForSecondaryAction = (e) =>
   e.button === 1
 
 module.exports.eventElHasAncestorWithClasses = (e, classesToCheck) => {
+  // DO NOT ADD NEW CHECKS USING THIS METHOD
+  // classNames are changed from dev to prod by Aphrodite
```
luixxiul: @cezaraugusto mind specifying Aphrodite version (`1.2.3`) and adding the number of the issue (#10029)?

luixxiul:
this should fix #10104 as well

test/unit/js/lib/eventUtilTest.js
```diff
@@ -0,0 +1,34 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+/* global describe, it */
+
+const eventUtil = require('../../../../js/lib/eventUtil')
+const assert = require('assert')
+
+require('../../braveUnit')
+
+describe('eventUtil', function () {
```
bsclifton: @cezaraugusto can you add a test for a partial match?  like `nespressoPods` and/or `love_nespresso`?

bsclifton:
@luixxiul @cezaraugusto can you confirm this also closes https://github.com/brave/browser-laptop/issues/10104?

test/unit/js/lib/eventUtilTest.js
cezaraugusto: added `can not accept partial string match`

luixxiul:
could I know how to test the binary build with `NODE_ENV=production` for manual QA? See: https://github.com/brave/browser-laptop/pull/10165#issue-245912630

cezaraugusto:
@luixxiul pls run `NODE_ENV=production npm run watch` when running the app. Notice you'll need to shift+f8 to open devtools. Checking styles for tabs should show the style concatenated

luixxiul:
@cezaraugusto I tried `NODE_ENV=production npm run watch` and it did not work for me. I modified `watch` and `start` inside package.json by adding `cross-env NODE_ENV=production` and that did not work either :-( do you have any tip?

luixxiul:
`NODE_ENV=production npm run watch` does not seem to work because of this error:

````
    ERROR in webtorrentPage.entry.js from UglifyJs
    SyntaxError: Name expected [./app/renderer/components/styles/commonStyles.js:5,6]
webpack: Failed to compile.
````

PR #10181, opened 2017-07-28
add 3rd party fingerprinting block option

### LINKED ISSUE #9029: Fingerprinting protection should distinguish between first and third party
Fingerprinting protection should have 3 options: block all, block third-party, and none. In the future, we should enable 'block third party' by default for all users. Users who currently have FP set to 'on' will be transitioned to 'block all'. 

This supercedes https://github.com/brave/browser-laptop/issues/8142 and is related to https://github.com/brave/browser-laptop/issues/4621 / https://github.com/brave/browser-laptop/issues/2965.

cc @bradleyrichter on design for the Shields panel

diracdeltas:
fix #9029

regular panel:
screen shot 2017-08-02 at 12 09 21 am

compact:
screen shot 2017-08-02 at 12 09 45 am

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests


docs/state.md
```diff
@@ -69,6 +69,12 @@ AppStore
   cookieblockAll: {
     enabled: boolean // enable all cookie/referer blocking
   },
+  fingerprintingProtection: {
```
NejcZdovc: Please sort this alphabetically

less/switchControls.less
```diff
@@ -85,3 +85,6 @@
     }
   }
 }
+.pullRight {
```
luixxiul: I have a fix which is much more straightforward. please wait for a moment.

luixxiul:
CC @bradleyrichter for review:

screenshot 2017-08-03 5 16 28

We could align the switches in a left row like this:

screenshot 2017-08-03 5 20 27

Which one is better?

diracdeltas:
the fp setting was right-aligned because that is where it was before (so people might be used to clicking in that area), but i don't have a strong opinion. actually i think 3 rows with 2 items each would look best.

luixxiul:
that would be look like this:

screenshot 2017-08-03 11 10 43

diracdeltas:
@luixxiul i like the last one the most because it matches with our existing placement

less/switchControls.less
diracdeltas: ++ please feel free to push to this branch (but do not rebase or squash commits)

bradleyrichter:
I'll need to try a different layout. Or we need to give up the wide view.

luixxiul:
@diracdeltas @bradleyrichter OK, so should we tag this PR as WIP for now?

diracdeltas:
This is blocking some work that others are doing on improving fingerprinting protection, so I would prefer it not be pushed to a later milestone. Why can't we just do compact-panel only?

diracdeltas:
Can i get this approved and then @bradleyrichter open up a follow-up issue for the UI tweak?

app/renderer/components/styles/global.js
```diff
@@ -241,21 +241,22 @@ const globalStyles = {
     prefsPanelHeading: '23px'
   },
   appIcons: {
```
luixxiul: elements below this are alphabetized.

app/renderer/components/main/braveryPanel.js
```diff
@@ -560,26 +558,19 @@ class BraveryPanel extends React.Component {
...
-                  'fa-question-circle': true
-                })}
-                  title={config.fingerprintingInfoUrl}
-                  onClick={this.onInfoClick}
-                />
+                  styles.braveryPanel__body__advanced__control__fpWrapper,
+                  styles.braveryPanel__body__advanced__control__forms__title
+                )}>
+                  <span data-l10n-id='fingerprintingProtection' />
+                  <span className={globalStyles.appIcons.question}
```
luixxiul: `braveryPanel__body__advanced__control__fpWrapper ` is cleated to wrap the l10n string and the question mark to align them with flex.

Jacalz:
From a users perspective I think this looks best:
!(https://user-images.githubusercontent.com/3362943/28893023-f05599e6-780a-11e7-94fc-c565bb5459ec.png)
This makes it possible to add two more things to the panel in the future, one with a drop down menu and also one with a switch 🙂

less/switchControls.less
luixxiul: Done https://github.com/brave/browser-laptop/pull/10181/commits/bcdfa30d9cc3fb7db30bcc9fd93334043e4225ee

app/sessionStore.js
```diff
@@ -675,6 +675,28 @@ module.exports.runPreMigrations = (data) => {
 }
 
 module.exports.runPostMigrations = (data) => {
+  // fingerprinting protection migration
```
bsclifton: Not that we have to do it now... but I'm Curious what you think @diracdeltas and @darkdh 

Similar to Ruby and Python, maybe we could create a `/migrations/` folder and have a single file (ex: `0-20-0_1st_3rd_fingerprint`) with a two possible functions (`postMigrate`, `preMigrate`) for each type of migration that we'd like to run. 

We could then call them like so:
```
migrations = [
  require('../../migrations/0-12-0_search_xml'),
  require('../../migrations/0-14-0_autofill_data'),
  require('../../migrations/0-20-0_1st_3rd_fingerprint')
]
migrations.runPostMigrations()
```

NejcZdovc: I really like this idea ++

diracdeltas: i'm for a migrations-migration. :)

NejcZdovc: We were talking with @bbondy when can we remove some old migrations from session store file and this solution would solve it. What I would suggest as well is to create an upgrade script. So it automatically check if file for that version exists or not. For example: let's say that you are upgrading from 0.10 to 0.15.4. When we start the upgrade process we determinate that we need to check versions 0.10.x to 0.15.4 (it can be many versions in between, hot fixes). Then we check if file `migration/0.10.15.js` (hot fix) exist and if so run both pre and post migration. Then you check if `migration/0.11.js`exist and so on. I would have one file per version and it would be standardized with predefined functions. This way we know exactly what was added/removed in which version.

darkdh: ++ for aggregated migrations. We have many migration scattered around now and some of them are made by me :stuck_out_tongue_closed_eyes:

diracdeltas:
This has been updated to use  @Jacalz's recommended layout. @bsclifton can you review?

bradleyrichter:
I think what @Jacalz added above will be fine for now. I am adding another switch so I will re-examine the layout very soon.

@luixxiul Can you adjust the layout as shown above?

diracdeltas:
@bradleyrichter see my comment above yours. no need to adjust since i already did.

luixxiul:
Let me confirm the adjustment by @diracdeltas.

bradleyrichter:
oops. sorry @diracdeltas. Thanks...

luixxiul:
LGTM!

bsclifton:
I created https://github.com/brave/browser-laptop/issues/10488 so that we don't lose the ideas mentioned here 😄

PR #10192, opened 2017-07-28
Removes unnecessary setFocusedFrame calls

### LINKED ISSUE #10186: Ledger getPublisher perf is bad even if ledger is off
All typing and mouse movement call into getPublisher which can be expensive. 

screen shot 2017-07-28 at 10 47 40 am

See also: #9311

I think mainly the problem is when publisher info isn't found it keeps looking for it. We could store that it was looked up but not found and then not re-lookup in that case. But also I think it's not needed to be even called into for cases like mouse move and key down.

NejcZdovc:
Submitter Checklist:

Resolves #10186

Auditors: @bbondy

Test Plan:
- run a profiler and check that ledger is not called when hovering over links or typing in find bar

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/frame/frame.js
+++ b/app/renderer/components/frame/frame.js
@@ -180,8 +180,8 @@ class Frame extends React.Component {
     }
   }
 
-  onPropsChanged () {
-    if (this.props.isActive && isFocused()) {
+  onPropsChanged (prevProps = {}) {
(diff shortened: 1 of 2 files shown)


NejcZdovc:
before:
screen shot 2017-07-28 at 23 40 10

after:
screen shot 2017-07-28 at 23 38 25

PR #10286, opened 2017-08-03
Optimizes fs-extra `require`.

luzmcosta:
To improve performance at load time, fs-extra, which
has been taking 40 ms to require, is loaded at the time
it’s needed, no sooner.

The fs-extra module was also referenced where fs can
do the job, so fs-extra has been replaced by fs in that
instance.


luixxiul:
@luzmcosta would you rebase the PR? thanks!

PR #10293, opened 2017-08-04
rely on compositor properties for tab previews

### LINKED ISSUE #10291: tab previews: stick to changing properties that can be handled by the compositor alone
this avoids layout and paint steps and let the task for the compositor

cezaraugusto:
Auditors: @bsclifton, @luixxiul
close #10291

**before checking out this PR**

1. clear session
2. open two tabs
3. open devtools (shift+f8)
4. go to performance tab, record profile
5. back to tabs, hover over one tab until preview is fired
6. cancel preview
7. stop recording profile
8. go to the profile timeline and select the frames right before and right after preview happens
9. go to the event log
10. sort by total time
11. Keep record (print?) of those numbers

check out this PR and repeat above operation.

numbers should be lower

**before**
screen shot 2017-08-04 at 12 03 55 am

**after** 
screen shot 2017-08-04 at 12 01 36 am

### DIFF
--- a/less/window.less
+++ b/less/window.less
@@ -83,6 +83,7 @@ html,
     height: 100%;
     width: 100%;
     &.isPreview {
+      will-change: opacity;
       opacity: 0.5;
       animation: tabFadeIn 0.75s ease-in-out;
       animation-fill-mode: forwards;


less/window.less
```diff
@@ -83,6 +83,7 @@ html,
     height: 100%;
     width: 100%;
     &.isPreview {
+      will-change: opacity;
```
luixxiul: nice find! There could be more elements which can be optimized with `will-change`.

cezaraugusto: yes the browser reserves a chunk of memory for that so they're costly if done several times but new tab load and tab's loading animation can benefit too. lmk if you know some other areas. it applies for `opacity` and `transform` properties.

PR #10325, opened 2017-08-07
Moves bookmark text calculation into the state

### LINKED ISSUE #9517: Move bookmark text calculation into the state
## Test plan

https://github.com/brave/browser-laptop/pull/10325#issue-248455456

----

Move calculation of bookmark width into state and calculate it only when bookmark is added/edited or favicon setting for bookmark toolbar is changes

#9030

NejcZdovc:
Submitter Checklist:

Resolves #9517

Auditors:

Test Plan:

0) add so many bookmarks that you see bookmark overflow icon 
1) check if overflow is displayed correctly
2) open two windows and try to resize them, check if overflow is displayed correctly
3) try reorder bookmarks from overflow to toolbar and see if overflow is still displayed correctly

Reviewer Checklist:

Tests


luixxiul:
let's remove `package-lock.json` for now.

NejcZdovc:
@luixxiul good catch, thank you

NejcZdovc:
Tests added. Tests for

- `bookmarksState`
- `bookmarkFoldersState`

will be handled in #10310

NejcZdovc:
I see that some new tests are failing, so will take a look at them

NejcZdovc:
Update on tests: Only tests that are failing are failing because we are doing something in the background (`textCalcUtil`) and because of that window handler is lost until we finish that background process. Will try to solve this with @bbondy 's help

NejcZdovc:
Tests are fixed now, thank you for the idea @bbondy

app/common/lib/bookmarkFoldersUtil.js
```diff
@@ -117,11 +118,32 @@ const isMoveAllowed = (folderList, sourceDetail, destinationDetail) => {
...
+  folderDetails = makeImmutable(folderDetails)
+  let key = folderDetails.get('folderId')
+
+  if (!folderDetails.has('folderId')) {
+    key = getNextFolderId(folders)
+  }
+
+  return makeImmutable({
+    title: folderDetails.get('title'),
+    folderId: ~~key,
```
cezaraugusto: nit: I know it was similar before this change but this double `not` in bitwise is hard to read. I don't think we need to floor numbers in here so my guess is that we use it only to coerce from string to number? if so a more readable way would be to explicitly coerce to number like you did in #10610.

NejcZdovc: Correct this is used to convert from possible string to integer. This conversion is a lot faster then `parseInt` or `parseFloat`. Why I changed this conversion in #10610 is because we need negative number (other bookmarks is -1) and this conversion only converts in positive integer.

bsclifton: While it may be faster, it's definitely less readable ☹️  How much of a difference (in terms of performance) are we talking about here?

NejcZdovc: jsperf is not working right now, so I can share specific numbers, but if this is less readable I can change it to `Number()` conversion. Not a problem

NejcZdovc: Changed `~~` to `Number`. Note that I didn't changed `parentId` conversion, because I already changed that in `Other bookmarks` PR

cezaraugusto:
hey @NejcZdovc can you rebase this? thanks!

NejcZdovc:
rebased

app/browser/tabs.js
```diff
@@ -336,6 +336,20 @@ const createNavigationState = (navigationHandle, controller) => {
   })
 }
 
+let backgroundProcess = null
+let backgroundProcessTimer = null
+const runScript = (win, debug, script, cb) => {
```
bbondy: Please add a comment for why debug exists (to keep alive for tests I think?)

app/browser/tabs.js
```diff
@@ -742,20 +756,26 @@ const api = {
...
-      show: false,
-      webPreferences: {
-        partition: 'default'
-      }
-    })
-    win.webContents.on('did-finish-load', (e) => {
-      win.webContents.executeScriptInTab(config.braveExtensionId, script, {}, (err, url, result) => {
-        cb(err, url, result)
-        setImmediate(() => win.close())
+  executeScriptInBackground: (script, cb, debug = false) => {
```
bbondy: ditto about `debug`

app/browser/tabs.js
NejcZdovc: debug was added so that window is not closed and you can view dev tools

app/browser/windows.js
```diff
@@ -61,6 +61,14 @@ const updateWindow = (windowId, updateDefault = false) => {
   }
 }
 
+const onWindowResize = (windowId) => {
+  const windowValue = getWindowValue(windowId)
+  if (windowValue) {
+    appActions.onWindowResize(windowValue)
+    windowActions.onWindowUpdate(windowId, true)
```
bbondy: btw just noting that appActions with a windowId get routed to that window.

NejcZdovc: not sure that I follow, are you saying that I should add windowId to the app action and this way I don't need separate windowAction?

app/common/lib/bookmarkFoldersUtil.js
```diff
@@ -26,7 +27,7 @@ const getNextFolderIdItem = (folders) => {
 }
 
 const getNextFolderId = (folders) => {
-  const defaultFolderId = 0
+  const defaultFolderId = 1
```
bbondy: why does the default folder Id have to change here?

NejcZdovc: because it was wrong. Default folderId can't be 0, because 0 is bookmark toolbar and -1 is other bookmarks

js/lib/textCalculator.js
```diff
@@ -1,9 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this file,
- * You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-module.exports.calculateTextWidth = (text, font = '11px Arial') => {
```
bbondy: 👍

app/common/lib/bookmarkFoldersUtil.js
NejcZdovc: I noticed this when I was adding `buildFolder` function and adding unit tests for it

NejcZdovc: and if don't have any folders that means that it should be 1 and not 0

app/common/lib/textCalcUtil.js
```diff
@@ -0,0 +1,134 @@
...
+      const result = {
+        bookmarks: bookmarks,
+        folders: folders
+      }
+  
+      return JSON.stringify(result)
+    })()
+  `
+
+  tabs.executeScriptInBackground(param, (err, url, result) => {
```
bbondy: I think it's a problem that this calls `tabs.executeScriptInBackground` but that the file exists in common. It needs browser process so the file should be located in the browser please.

app/common/lib/bookmarkFoldersUtil.js
NejcZdovc: this was not a problem before (and it still isn't) because we are passing at least empty Map into this function and that's why this if is not triggered. What we could do is to check if map is  empty and then trigger this if. But in unit test first thing that you do is pass null into it :smile:

app/common/lib/bookmarkUtil.js
```diff
@@ -260,18 +154,87 @@ const getKey = (siteDetail) => {
...
+      if (!topSite.isEmpty()) {
+        dataItem = topSite
+      }
+    }
+  }
+
+  return makeImmutable({
+    title: bookmarkDetail.get('title', ''),
+    location: bookmarkDetail.get('location'),
+    parentFolderId: ~~bookmarkDetail.get('parentFolderId', 0),
```
bbondy: Noticed that below you used Number instead of `~~` here. Maybe use `~~` as well here? It's a bit diff in that `~~` is like `Math.floor`. But I think this is always an integer.

NejcZdovc: there was a discussion about `~~` vs `Number` here https://github.com/brave/browser-laptop/pull/10325#pullrequestreview-57656377 and because it was not readable I changed it to Number in some parts. Other `~~` are fixed here https://github.com/brave/browser-laptop/pull/10610/files

bbondy:
@kjozwiak could you or someone in QA test bookmark toolbar and this PR before we merge.  Make sure it fills up the toolbar fully, that it shows the overflow indicator, etc.   Also test different modes like favicon only, text only and text+icon mode for bookmarks toolbar.

app/browser/windows.js
bbondy: or just pass it into the dispatched action ya and then handle the app action resize constant in the window store. It's a bit abnormal to use windowActions. I see them as more only within a single window renderer. I think it's only used 1 other place.

luixxiul:
@bbondy @NejcZdovc does this PR close https://github.com/brave/browser-laptop/issues/9939 as well?

app/browser/tabs.js
bbondy: that's fine just add a comment to explain it thx

app/common/lib/bookmarkFoldersUtil.js
bbondy: ok thanks.

app/browser/reducers/bookmarkToolbarReducer.js
```diff
@@ -0,0 +1,37 @@
...
+
+// State
+const bookmarksState = require('../../common/state/bookmarksState')
+const bookmarkFoldersState = require('../../common/state/bookmarkFoldersState')
+
+// Util
+const {makeImmutable} = require('../../common/state/immutableUtil')
+const textCalcUtil = require('../../common/lib/textCalcUtil')
+
+const tabsReducer = (state, action, immutableAction) => {
```
bbondy: copy paste error: s/tabsReducer/bookmarkToolbarReducer

app/browser/reducers/bookmarkToolbarReducer.js
```diff
@@ -0,0 +1,37 @@
...
+        if (bookmarkFolders.first() && !bookmarkFolders.first().has('width')) {
+          textCalcUtil.calcTextList(bookmarkFolders.toList())
+        }
+      }
+      break
+  }
+  return state
+}
+
+module.exports = tabsReducer
```
bbondy: update to new name

app/common/lib/bookmarkFoldersUtil.js
```diff
@@ -117,11 +118,32 @@ const isMoveAllowed = (folderList, sourceDetail, destinationDetail) => {
...
+
+  if (!folderDetails.has('folderId')) {
+    key = getNextFolderId(folders)
+  }
+
+  return makeImmutable({
+    title: folderDetails.get('title'),
+    folderId: Number(key),
+    key: key.toString(),
+    parentFolderId: ~~folderDetails.get('parentFolderId', 0),
```
bbondy: ditto not sure which you prefer but either `~~` or `Number`

app/common/lib/bookmarkUtil.js
bbondy: I don't care which way it was just odd to me to see `~~` being changed to `Number` in some places and then also new `~~` being used.

app/common/lib/textCalcUtil.js
```diff
@@ -0,0 +1,134 @@
...
+          parentFolderId: item.get('parentFolderId'),
+          width: result[0]
+        }
+      ]))
+    }
+  })
+}
+const calcTextList = (list) => {
+  const take = 500
+  if (!list.toJS) {
```
bbondy: use immutable helper here for readability pls

app/common/lib/bookmarkFoldersUtil.js
NejcZdovc: I prefer `~~`, but please check this comment https://github.com/brave/browser-laptop/pull/10325#discussion_r137261789

NejcZdovc:
@luixxiul not sure, but we can test it out

app/common/lib/bookmarkUtil.js
NejcZdovc: yeah after this PR and #10610 are merged there will be no more `~~`

app/browser/tabs.js
NejcZdovc: done

app/browser/tabs.js
NejcZdovc: done

app/common/lib/textCalcUtil.js
NejcZdovc: moved

app/browser/reducers/bookmarkToolbarReducer.js
NejcZdovc: done

app/browser/reducers/bookmarkToolbarReducer.js
NejcZdovc: nice catch, done

app/common/lib/textCalcUtil.js
NejcZdovc: done

app/browser/windows.js
NejcZdovc: done
+

NejcZdovc:
@kjozwiak can you please try again

bbondy:
@kjozwiak this is ready for you to re-review next week.

kjozwiak:
@bbondy sounds good! I started testing this yesterday but transitioned into the `0.18.36` once the builds were ready. I'll get this finished Morning morning.

Just an FYI, the overflow icon is now appearing correctly but I still wanted to go through a few more tests to make sure there's no regressions.

NejcZdovc:
@kjozwiak I tried first bug and favicons are not imported on master as well, so it's not limited to this PR. Bug 2 was fixes for folders and bookmarks

PR #10445, opened 2017-08-11
replace node classList with dataset

### LINKED ISSUE #10443: replace node classList instances with dataset
This unblocks #10234

cezaraugusto:
fix #10443 
unblocks #10234

1. have enough bookmarks to show the overflow indicator
2. click the overflow, should open
3. open a bookmark pull down with children
4. you should be able to click the second level child
5. second level child should be nested visually with its parent


srirambv:
Not seeing any second level child due to https://github.com/brave/browser-laptop/issues/12052

srirambv:
@cezaraugusto Is this expected?

cezaraugusto:
can't see a bug in the screenshot what's missing?

srirambv:
The nested items open on to the left used to be right

cezaraugusto:
ya it's expected because the parent menu is on the right. Chrome does the same

PR #10489, opened 2017-08-15
Polish bravery panel

### LINKED ISSUE #5726: Follow-up to #5724: the bravery panel should be much wider
## Test plan

1. Change the lang setting
2. Restart the browser
3. Open bravery panel on a site
4. Make sure the strings are not wrapped

----

#5724 #5715 

on macOS

1. Set your lang setting to Japanese
2. Restart the browser
2. Open https://github.com/brave/browser-laptop/issues/new

screenshot 2016-11-18 19 38 20

The panel should be much wider. It is too small for Japanese, at least. 

By setting ``width: 540px`` you would get:

screenshot 2016-11-18 19 43 56

I have not tested it on other language settings, so I am not sure if it could work globally.

luixxiul:
Closes #5726
Follow-up to #10181

- Remove fixed width; now that we have the compact bravery panel, we would not have to take the possibility into consideration that the panel would cover the UI.
- Increase padding-right from 1.5em to 2em on the dropdown to add 1em space between the option and the down arrow (See `Block 3rd Party Fingerprinting` option below).
- Decrease the margin-bottom of the dropdown and Increase margin between switches

From:

screenshot 2017-08-15 16 29 57

To:

screenshot 2017-08-15 16 30 47

Here is an extreme example (which would not happen in reality, unless the font rendering is quite poor + screen resolution is very low).

screenshot 2017-08-15 16 43 10

That panel can be scrolled horizontally.

Auditors:

Test Plan:
1. Change lang setting
2. Restart the browser
3. Open Bravery panel on an external site
4. Make sure the strings are not wrapped

Submitter Checklist:

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/main/braveryPanel.js
+++ b/app/renderer/components/main/braveryPanel.js
@@ -528,6 +528,7 @@ class BraveryPanel extends React.Component {
                 <SwitchControl className={css(
                   !this.props.isCompactBraveryPanel && gridStyles.row6col1,
                   this.props.isCompactBraveryPanel && gridStyles.row8col1,
+                  !this.props.isCompactBraveryPanel && styles.braveryPanel__body__advanced__control__switchControl_noScript,
                   this.props.isCompactBraveryPanel && styles.braveryPanel_compact__body__advanced__control__switchControl
                 )}
                   onClick={this.onToggleNoScript}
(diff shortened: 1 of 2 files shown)


app/renderer/components/main/braveryPanel.js
```diff
@@ -993,6 +994,7 @@ const styles = StyleSheet.create({
   braveryPanel__body__advanced__control: {
     display: 'grid',
     gridColumnGap: '1rem',
+    gridTemplateColumns: 'max-content max-content',
```
luixxiul: This prevents the strings inside the dropdowns and the labels from wrapping, without forcing the same width for each column.

Issue #10226, opened 2017-08-01
Remove keytar dependency


luixxiul:
**Describe the issue you encountered:** 

Remove keytar dependency. See: https://github.com/brave/browser-laptop/pull/10227#issuecomment-319519959

- Platform (Win7, 8, 10? macOS? Linux distro?): All

- Brave Version (revision SHA): master branch

- Extra QA steps:
    1.
    2.
    3.

- Any related issues:

luixxiul:
Renamed the title based on the discussion here: https://github.com/brave/browser-laptop/pull/10227#issuecomment-319426917

Issue #10448, opened 2017-08-12
Brave prompts to create new keychain named "Default Keychain"


evq:
- Did you search for similar issues before submitting this one?
Yes

- Describe the issue you encountered:

With the [recent muon change to encrypt cookie data at rest](https://github.com/brave/muon/pull/262), Brave now has a hard dependency on gnome-keyring on linux (before it was only needed if you opted to use the builtin password manager). As a result, if you do not already have a keychain, Brave will prompt you to create a new one. I cannot confirm, but I would imagine that if there is an existing keychain, the user would be prompted to allow access. If possible we should use a keychain name that better indicates why Brave is asking for keychain access.

Furthermore, despite being used to protect user data - having to allow keychain access will be seen by some users as concerning. Some users will want a way to stop all requests for keychain access.

See a similar discussion about Chrome asking for access:
https://productforums.google.com/forum/#!topic/chrome/BPLVDSeTmgI

- Platform (Win7, 8, 10? macOS? Linux distro?):
Linux

- Brave Version (revision SHA):
0.19.2

- Screenshot if needed:
![keychain](https://user-images.githubusercontent.com/8796196/29239183-758f84ea-7f37-11e7-8b94-0ecfe11f2db1.png)

bsclifton:
cc: @darkdh

darkdh:
libgnome-keyring is almost deprecated by libsecret
https://bugs.chromium.org/p/chromium/issues/detail?id=466975

luixxiul:
Is it OK to close this?

darkdh:
need investigation, I will check with @evq

PR #10514, opened 2017-08-15
Remove keytar and gnome-keyring deps

diracdeltas:
Give up on migrating old passwords for users who are updating from <0.15.300 to 0.21 and above.

Fix https://github.com/brave/browser-laptop/issues/10226

Test Plan:
npm run test -- --grep='passwords'

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests


posix4e:
This will make 32 bit builds on linux way easier as well. keytar is the last real dep that needs weird pkg_config options when doing the 32 bit linux build. Perfect timing.

app/browser/reducers/passwordManagerReducer.js
```diff
@@ -190,53 +125,15 @@ const clearPasswords = () => {
...
-  }
-  return state
-}
-
 const passwordManagerReducer = (state, action, immutableAction) => {
   action = immutableAction || makeImmutable(action)
   switch (action.get('actionType')) {
     case appConstants.APP_SET_STATE:
       init()
-      state = migrate(state)
```
darkdh: I think we should also check if `legacyPasswords` exists or not. If so, we should delete it to prevent our session data from getting bloated.

.travis.yml
```diff
@@ -37,4 +37,3 @@ addons:
     packages:
       - xvfb
       - g++-4.8
-      - libgnome-keyring-dev
```
evq: gnome-keyring is needed at runtime by muon, not sure if removing this will cause issues. 
see: https://github.com/brave/browser-laptop/issues/10448

darkdh: gnome-keyring is deprecated by libsecret

darkdh: unless you happened to be the 3%
https://bugs.chromium.org/p/chromium/issues/detail?id=466975

darkdh: And I think this is the deps for keytar use not for muon which will be precompiled before browser-laptop can use it.

app/browser/reducers/passwordManagerReducer.js
diracdeltas: done!

PR #10580, opened 2017-08-18
Windows are restored with the correct z-index

### LINKED ISSUE #10473: Preserve z-index of windows when restoring from the session
## Test plan
1. Launch brave and open a few tabs
  - facebook.com
  - twitter.com
  - brave.com
2. Resize the main window to a very specific size (ex: 800x600)
3. Tear off the twitter and brave tabs, each into their own windows
4. Resize those windows and stack them in an order that you can remember
  - put facebook on bottom
  - put twitter over the facebook window
  - put brave over the twitter window
5. Quit Brave and relaunch
6. Z-order should be preserved exactly (brave on top, then twitter, then facebook)

## Original issue description
When you close multiple windows with Menu->Quit Brave and you start browser again, windows don't have the same z-index as they had before.

NejcZdovc:
Submitter Checklist:

Resolves #10473 

Test Plan:
See https://github.com/brave/browser-laptop/issues/10473

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/reducers/windowsReducer.js
+++ b/app/browser/reducers/windowsReducer.js
@@ -278,14 +278,23 @@ const createWindow = (state, action) => {
       const appStore = require('../../../js/stores/appStore')
       win.webContents.setZoomLevel(zoomLevel[toolbarUserInterfaceScale] || 0.0)
 
+      const position = win.getPosition()
+      const size = win.getSize()
+
       const mem = muon.shared_memory.create({
         windowValue: {
           disposition: frameOpts.disposition,
-          id: win.id
+          id: win.id,
+          focused: win.isFocused(),
+          left: position[0],
+          top: position[1],
+          height: size[1],
+          width: size[0]
         },
         appState: appStore.getLastEmittedState().toJS(),
         frames: frames.toJS(),
-        windowState: (restoredImmutableWindowState && restoredImmutableWindowState.toJS()) || undefined})
+        windowState: (restoredImmutableWindowState && restoredImmutableWindowState.toJS()) || undefined
+      })
 
       e.sender.sendShared(messages.INITIALIZE_WINDOW, mem)
       if (action.cb) {
--- a/app/index.js
+++ b/app/index.js
@@ -188,9 +188,23 @@ app.on('ready', () => {
         appActions.newWindow()
       }
     } else {
-      loadedPerWindowImmutableState.forEach((wndState) => {
-        appActions.newWindow(undefined, undefined, wndState)
-      })
+      loadedPerWindowImmutableState
+        .sort((a, b) => {
+          let comparison = 0
+          const aTime = a.getIn(['windowInfo', 'focusTime'], 0)
+          const bTime = b.getIn(['windowInfo', 'focusTime'], 0)
+
+          if (aTime > bTime) {
+            comparison = 1
+          } else if (aTime < bTime) {
+            comparison = -1
+          }
+
+          return comparison
+        })
+        .forEach((wndState) => {
+          appActions.newWindow(undefined, undefined, wndState)
+        })
     }
     process.emit(messages.APP_INITIALIZED)
 
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* global performance */
+
 const appDispatcher = require('../dispatcher/appDispatcher')
 const EventEmitter = require('events').EventEmitter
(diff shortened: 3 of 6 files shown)


NejcZdovc:
this PR is blocked by #10458

cezaraugusto:
@NejcZdovc please confirm if this is ready for review and tag reviewers to check. thanks!

NejcZdovc:
@cezaraugusto this is ready for a review

NejcZdovc:
@cezaraugusto it's ready for a review

js/actions/windowActions.js
```diff
@@ -1211,6 +1211,16 @@ const windowActions = {
       },
       windowValue
     })
+  },
+
+  onWindowFocus: function (windowId, windowValue) {
```
bbondy: Can you move this to appAction?

app/browser/reducers/windowsReducer.js
```diff
@@ -332,6 +333,7 @@ const windowsReducer = (state, action, immutableAction) => {
       break
     case appConstants.APP_WINDOW_CREATED:
       state = windowState.maybeCreateWindow(state, action)
+      windowActions.onWindowFocus(action.getIn(['windowValue', 'windowId']), action.get('windowValue'))
```
bbondy: I don't think we can assume that app window created means always focused.  You could maybe check the window here and if focused do that if for some reason the `focus` event isn't firing.

cezaraugusto:
I'm not sure about the test plan but I did the following:

1. 3 tabs with a, b, c websites
2. tear-off each tab in that order. Have 3 windows with a, b, c (current visible window)
3. Shut down, open again
4. Window with tab c should show

is that an accurate test plan?

cezaraugusto:
for the above windows opened in the opening order so a, b, c instead of c, b, a. Not sure if that's the expected behavior.

cezaraugusto:
for the above windows opened in the reverse order so a, b, c instead of c, b, a. Not sure if that's the expected behavior.

app/browser/reducers/windowsReducer.js
NejcZdovc: done

js/actions/windowActions.js
NejcZdovc: done

NejcZdovc:
@cezaraugusto yeah I didn't try this scenario, was using keyboard shortcuts only. Nice scenario. This use case should be fixed now. Can you please re-review?

@bbondy I rewrote this PR a little bit, so we don't need focus action anymore and detection rate is a lot better

NejcZdovc:
I will squash commits before merging

app/index.js
```diff
@@ -188,9 +188,23 @@ app.on('ready', () => {
         appActions.newWindow()
       }
     } else {
-      loadedPerWindowImmutableState.forEach((wndState) => {
-        appActions.newWindow(undefined, undefined, wndState)
-      })
+      loadedPerWindowImmutableState
```
bsclifton: Maybe as a follow up, it would be good to pull out the sorting into something which can easily be unit tested.  Maybe a new method `determineZindex` or similar (maybe in `app/commmon/state/windowState`)

then the below would be like:
```
const sortedList = windowState.determineZindex(loadedPerWindowImmutableStore)
sortedList.forEach((wndState) => {
  appActions.newWindow(undefined, undefined, wndState)
})
```

js/stores/windowStore.js
```diff
@@ -788,8 +790,17 @@ const doAction = (action) => {
         break
       }
     case windowConstants.WINDOW_ON_WINDOW_UPDATE:
-      windowState = windowState.set('windowInfo', action.windowValue)
-      break
+    case appConstants.APP_WINDOW_READY:
```
bsclifton: If you did a follow up, this would be fairly easy to add a unit test for too 😄  Basically checking that `focusTime` gets added to `windowInfo`

bsclifton:
I'll hold off on merge if you wanted to squash / address feedback and maybe @bbondy can re-review 😄

NejcZdovc:
@cezaraugusto rebased and squashed

app/index.js
bbondy: ++!

js/stores/windowStore.js
```diff
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* global performance */
```
bbondy: you can just use window.performance and avoid this.

cezaraugusto:
master 843d9b3
0.20.x https://github.com/brave/browser-laptop/commit/8e370de6b1e438731e2864dce27ba6ff577d10aa

PR #10605, opened 2017-08-20
Adds state as param to currentWindow.js

### LINKED ISSUE #10457: currentWindow.js should accept appState as param
`currentWindow.js` should accept appState as a param and default to `appStoreRenderer.state` if state is not provided. Now we use this functions mostly in `mergeProps`.

NejcZdovc:
Submitter Checklist:

Resolves #10457

Auditors: @bsclifton

Test Plan:
- covered by automated tests

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/navigation/buttons/windowCaptionButtons.js
+++ b/app/renderer/components/navigation/buttons/windowCaptionButtons.js
@@ -13,7 +13,7 @@ const windowActions = require('../../../../../js/actions/windowActions')
 
 // Utils
 const locale = require('../../../../../js/l10n')
-const {getCurrentWindowId, isMaximized, isFullScreen} = require('../../../currentWindow')
+const {getCurrentWindowId} = require('../../../currentWindow')
 const cx = require('../../../../../js/lib/classSet')
 
 class WindowCaptionButtons extends ImmutableComponent {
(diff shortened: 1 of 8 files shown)


app/renderer/components/navigation/buttons/windowCaptionButtons.js
```diff
@@ -60,7 +61,7 @@ class WindowCaptionButtons extends ImmutableComponent {
     const props = { tabIndex: -1 }
 
     return <div className={cx({
-      fullscreen: this.props.windowMaximized,
+      fullscreen: this.props.windowMaximizedFullScreen,
```
bridiver: maximized and fullscreen are only the same thing on macos. We should specify one or the other

NejcZdovc: I preserved the same functionality, I didn't change anything. So it's working the same way it was before. The only thing that I changed is naming, because we had/have both things in it

bridiver: I'm only referring to the naming. We have maximized and fullscreen in one property? That's a bit confusing imo

NejcZdovc: Yeah this is how we have it now :smiley: 

https://github.com/brave/browser-laptop/blob/6fe9849cd3aea3faddf379b96fddb7abdecdd247/app/renderer/components/main/main.js#L533

NejcZdovc: what do you propose? split it and use it separately?

bridiver: the existing logic seems wrong to me because fullscreen and maximized should not be treated the same way

NejcZdovc: Current logic is working (no reports about any problems), so I would suggest that if it's not working we should open new issue for it. If we just want to refactor it we should open refactor issue. I think that this PR is not a right place to fix this, because we didn't introduced this naming in this PR

bbondy: @NejcZdovc that seems fair but pls open an issue to track it.

NejcZdovc: issue created https://github.com/brave/browser-laptop/issues/10989

PR #10607, opened 2017-08-21
Adds progress indicator to the import process

### LINKED ISSUE #9476: Importing bookmarks should show progress
Importing bookmarks doesn't show a status or progress bar.
- When folks have a lot of bookmarks, the browser will appear to be doing nothing until it finishes (see Import bookmark causes Brave to hang momentarily before showing successful import #6699)
- the message box is tied to either bookmarks manager OR the tab that owned the modal

needs mockup

cc: @alexwykoff @bradleyrichter

NejcZdovc:
Submitter Checklist:

Resolves #9476

Auditors:

Test Plan:
- try to import data from firefox and check if your toggles are respected
- try to import big file (check qa resources for 5k file) and check if you see new text notifying you about import status
image

Reviewer Checklist:

Tests


app/importer.js
```diff
@@ -279,8 +281,12 @@ importer.on('import-success', (e) => {
       appActions.changeSetting(settings.SHOW_BOOKMARKS_TOOLBAR, true)
     }
   }
+  setImmediate(() => {
+    windowActions.setImportBrowserDataDetail()
```
NejcZdovc: This is code is not called, because `windowActions` is not propagated to the windows. We could be using appState and introduce new object `ui` and fill this things into it. We could clear it at the end. This would be a first step of migrating window state ui into app state ui. @bsclifton wdyt?

bsclifton: Good call- I think this will have to be AppState actually.  We can could even show badges or progress, similar to that which was added with https://github.com/brave/browser-laptop/pull/9864

NejcZdovc: now we have appAction for import :smiley:

NejcZdovc:
this PR is blocked until we decide if we want to move windows from array to object, so that we can use simple `state.getIn(['windows', id])`

NejcZdovc:
blocked by #11151

bsclifton:
Closing as this is stale:
- it's been open for ~3 months without solution
- it needs rebase
- it's blocked on https://github.com/brave/browser-laptop/issues/11151 which hasn't been solved yet

PR #10628, opened 2017-08-22
Improve installation instructions for raw x64 binaries

TheAssassin:
Just improved the docs, I guess I don't need the check list for code changes.


bbondy:
Thanks!

PR #10652, opened 2017-08-24
Return 0% when download totalBytes is 0 or undefined

diracdeltas:
Fix https://github.com/brave/browser-laptop/issues/10264

Submitter Checklist:

Test Plan:
Download a large file and hit 'pause' immediately. It should not show Infinity% downloaded.

Reviewer Checklist:

Tests

### DIFF
--- a/js/state/downloadUtil.js
+++ b/js/state/downloadUtil.js
@@ -56,8 +56,15 @@ const getL10nId = (download) => {
   return ''
 }
 
-const getPercentageComplete = (download) =>
-  Math.ceil(download.get('receivedBytes') / download.get('totalBytes') * 100) + '%'
+const getPercentageComplete = (download) => {
+  const totalBytes = download.get('totalBytes')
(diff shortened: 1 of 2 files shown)


bbondy:
++

bbondy:
Actually I think we didn't used to show anything at all. Will this show 0% always for downloads that don't specify a content length in the reply?

diracdeltas:
@bbondy i think it would have shown NaN% or Infinity% previously

bbondy:
I mean originally we didn't show anything at all. There is a case where there is no known % because we don't know the content length being downloaded. For example if that happens with a 3GB file we shouldn't show 0% the whole time it is downloading.  Do you know if that will happen?

diracdeltas:
@bbondy pretty sure that is unaffected by this PR. looks like that check happens in app/renderer/components/download/downloadItem.js using totalBytes

Issue #6242, opened 2016-12-15
Improve Drag and Drop for Tabs


cezaraugusto:
We have React-DnD implemented as a dependency, and already used on new tab page.

React-DnD is a great library and since it's already implemented,  could be useful for all features that require drag-n-drop. In this case, tabs.

## Specific tasks
- [ ] Animation should be nice and smooth. Sliding instead of popping (originally reported with https://github.com/brave/browser-laptop/issues/8291, https://github.com/brave/browser-laptop/issues/8288)
- [x] Drop zone needs to be clear to user (see https://github.com/brave/browser-laptop/issues/6863 for more info)
- [ ] moving mouse up and down (Y axis) should not move the tab (or the ghost preview) up or down (until the threshold for tear off is hit)
- [ ] tear off on Y axis should be about 15 pixels before it does a tear off (see https://github.com/brave/browser-laptop/issues/9211)
- [ ] dragging the tab moves the actual tab (not a ghosted preview of it)

## Related
- Extend drag-n-drop functionality between URL, tabs, bookmarks, tiles, etc. #8067
- Tab.less conversion #11272

cezaraugusto:
Setting as backlog until we can deep dive on it

Liunkae:
Specific issue: #6033

lifeinchords:
It would be great to be able to drag + drop between windows, to group tabs by topic, ie:
These 10 have to do with a vacation
These 15 tabs in this window are for this project
etc

bsclifton:
Assigning over to @cezaraugusto who did a great job with the New Tab page (which uses the mentioned React-DnD library) :smile: 

@cezaraugusto let's check this one out AFTER you finish with these two issues:
https://github.com/brave/browser-laptop/pull/5918 (tabs responsiveness)
https://github.com/brave/browser-laptop/issues/5979

bsclifton:
@lifeinchords good call- we have an issue specifically tracking that if you wanted to subscribe. Check it out here: https://github.com/brave/browser-laptop/issues/4402

bsclifton:
cc: @bradleyrichter per our recent discussions

One of the behaviors he wanted to address was accidental tab tear off. We may choose to restrict the Y axis (have to drag more than 20px above or below the tab, etc).

@bradleyrichter can you help me break this task into several tangible tasks? Like
- dragging tab should move it over to the next one once it's dragged 1/3 of the way into another tab
- tab should be shown sliding BEHIND the tab next to it
- etc

[...]

PR #10686, opened 2017-08-28
[wip] Tab sorting with drag and drop moves tabs with animation and creates space

petemill:
**This is a work in progress!**

Addresses #6242 (Improve Drag and Drop for Tabs)

requires: https://github.com/brave/browser-laptop/pull/11530

State so far:
brave-tab-sort-smooth

I put some time investigating a more physical experience for dragging tabs around. Some of the aims:
1. Immediate physical movement during tab sorting in a tab-page of a window
2. Immediate space-making when dragging a tab to a new window
3. nicer detach-to-new-window experience:
   - higher threshold
   - more visual preview

What I have so far, achieves 1. and, almost perfectly, 2. I can certainly polish it up and make the inter-window experience quite nice, and increase the threshold before detaching a tab to a new window for 3., but first wanted some feedback before progressing both on the direction of the UX and the implementation.

The main difficulty that doesn't seem possible within the current architecture is being able to have *both* of the following:
- Dragging the tabs to re-sort constrained to an axis (x-axis currently)
- After dragging in the opposite axis past a threshold (e.g. 50px), some kind of preview image is attached to the mouse cursor and able to be dragged *outside of* the window

This is because:
- The only way to get a drag-image attached to the mouse cursor that will work outside of the renderer window is to use html `draggable` (as is currently done)
- An html `draggable` cannot be constrained to an axis
- An html `draggable` cannot have its drag-image changed after the `dragstart` event has fired (it can have *no* image, but has to stay that way)

Furhtermore, an html `draggable` drag cannot be fired from javascript (e.g. whilst the mouse is already dragging), so we cannot start with our own tab-sorting `mousedown` and `mousemove` handler and then start a proper cross-window drag event when the mouse moves past a certain threshold outside the tab area.

Since we have to use html `draggable` in order to get the drag data being sent outside the window, or to a new window, we are constrained by the above.

Some brief thoughts on ways this could be solved if we do want the perfect constrained-axis dragging:

 1. Do it from the muon / chromium side - provide an API to attach a tab's dragdata and a window preview to the mouse, that the renderer javascript could call whilst a mousedown and mousemove are in progress, or perhaps even more specific APIs.
 2. See if we can do it through IPC purely in this layer of the app - as the mouse is dragged away, actually create a new window with the new tab before the mouse is released, and update the position of the window as the mouse moves. When a tab bar in another window receives the mouse whilst the other window is still reporting that the mouse is down and the tab selected, start showing the tab being inserted, whilst the main process closes the new window. If the process is interrupted, the tab goes back to the original position.

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/tabs/tab.js
+++ b/app/renderer/components/tabs/tab.js
@@ -93,9 +93,9 @@ class Tab extends React.Component {
 
   get isDragging () {
     const sourceDragData = dnd.getInterBraveDragData()
-    return sourceDragData &&
+    return sourceDragData && this.props.dragData &&
       sourceDragData.get('tabId') === this.props.tabId &&
-      sourceDragData.get('draggingOverWindowId') === getCurrentWindowId()
+      this.props.dragData.get('windowId') === getCurrentWindowId()
   }
 
   get isDraggingOverSelf () {
@@ -128,7 +128,7 @@ class Tab extends React.Component {
       // this is added back to original size when onDrag event is happening
       this.tabSentinel.style.width = 0
 
-      dnd.onDragStart(dragTypes.TAB, this.frame, e)
+      dnd.onDragStart(dragTypes.TAB, this.frame.set('displayIndex', this.props.displayIndex), e)
       // cancel tab preview while dragging. see #10103
       windowActions.setTabHoverState(this.props.frameKey, false, false)
     }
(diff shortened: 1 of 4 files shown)


petemill:
Just looking more in to this. In order to get the more perfect version where sorting the tabs is axis-constrained and sorted in realtime, and then the tab is visually immediately removed past a certain threshold, then there are two options:
1. (Similar to firefox) - immediately set the dragImage to the frame preview, and monitor the drag move
2. (Similar to Chrome and Safari) - sort the tabs when drag starts, and set no dragImage on the cursor, when the mouse moves outside the tab area, set a drag image (Safari) or create a new window attached to the mouse (Chrome) that stays that way (and becomes a new window) or disappears when dragging to another existing window.

The simplest wold be 1., the closest to what we have now, and could possibly work within the renderer process/context, 2. would require either creating a new BrowserWindow in realtime as the mouse is moving (also from electron) or manipulating the dragevent from electron (similar to [electron-drag-drop](https://github.com/electron-utils/electron-drag-drop))

cezaraugusto:
hi @petemill closing this by now please re-open when you have a window. Thanks!

petemill:
Replaced by #11720

PR #10690, opened 2017-08-28
Reduce actions call when doing previews

### LINKED ISSUE #10689: onTabMouseMove is called too many times
Window action `onTabMouseMove` is called too many times (it's not debounced) when you move mouse inside the tab. Issue is set to the same milestone as #9887

NejcZdovc:
Submitter Checklist:

Resolves #10689

Auditors:

Test Plan: described in https://github.com/brave/browser-laptop/pull/9887#issue-240866594

Reviewer Checklist:

Tests


NejcZdovc:
Sure not a problem. 

Test was to open two tabs. One with loaded tab and another with new tab. Then I just moved a mouse around, no clicking.

before:
screen shot 2017-08-30 at 07 28 38

after:
image

Problem with the current implementation is that we are calling action every single time and we only debounce it in the store. Even though that we are not changing the state we are still calling the whole dispatcher cycle (app and window).

cc @bridiver for his thoughts

cezaraugusto:
you have a point but I'm not sure if we want to move to that direction so deferring review to @bridiver. Initial PR I did was rejected with the same pattern as you did here so I'm not the right person to review.

app/renderer/components/tabs/tab.js
```diff
@@ -137,19 +140,26 @@ class Tab extends React.Component {
...
     // if user is in previewMode, which is defined by mouse move
+    clearTimeout(this.mouseTimeout)
     windowActions.setTabHoverState(this.props.frameKey, true, this.props.previewMode)
   }
 
   onMouseMove () {
     // dispatch a message to the store so it can delay
     // and preview the tab based on mouse idle time
-    windowActions.onTabMouseMove(this.props.frameKey)
+    clearTimeout(this.mouseTimeout)
```
bridiver: would `this.onMouseMove = debounce(this.onMouseMove.bind(this), this.mouseTimeout)` not work in the constructor?

js/stores/windowStore.js
```diff
@@ -398,21 +397,8 @@ const doAction = (action) => {
         }
         break
       }
-    case windowConstants.WINDOW_TAB_MOUSE_MOVE:
-      {
-        // previewMode is only triggered if mouse is idle over a tab
-        // for a given amount of time based on timing defined in prefs->tabs
-        // we use actions here because that is the only way to delay updating the state
-        clearTimeout(mouseTimeout)
```
bridiver: we specifically wanted to keep the timeout in the action handler vs putting it in the component

NejcZdovc: can you please elaborate a little bit more. I don't see a real benefit and we are reducing performance when calling action without state change. Because mouse move is happening on a tab and this functionality is limited to this component, I don't see why we can't have it inside the component.

app/renderer/components/tabs/tab.js
NejcZdovc: I am not sure how to clearTimeout with debouce function? If we can use clear then this can be used yeah

js/stores/windowStore.js
bridiver: the benefit is that it keeps the hover logic out of the components. The action shouldn't be an expensive call and is likely dwarfed by the relatively expensive `setTimeout` call. Have you measure the performance difference between the two of them?

app/renderer/components/tabs/tab.js
cezaraugusto: I think we should not use debounce here. Debouncing leads to bad preview behavior when `previewMode` is enabled after consecutive previews

js/stores/windowStore.js
bridiver: I'm not 100% opposed to this change, but I'd like to see some hard perf numbers first

NejcZdovc: @bridiver here are some numbers https://github.com/brave/browser-laptop/pull/10690#issuecomment-325886207

bsclifton:
@cezaraugusto @NejcZdovc @bridiver are there clear next steps for this issue? I believe @petemill may have input on this (see https://github.com/brave/browser-laptop/issues/11515 for more info).  It sounds like we should setup a meeting and come to a resolution

@NejcZdovc as the PR owner, would you mind reaching out over Slack? If there's not a quick resolution, let's setup a call 😄

js/stores/windowStore.js
petemill: My 2c is that it's actually cleaner (as well as more performant) to put this logic in the component - the UI component is deciding when the right UI actions have been performed by the user and the action should be sent, not the store.

cezaraugusto:
the method @NejcZdovc is proposing is the same I did when this change came up but changed again after @bridiver feedback. I think the main blocker here is to decide what components actually should know. The way we did in this means that components are dumb and all the logic is handled by the store -- which I agree with. The profile is not clear for me about the perf gains.

anyway I'm ok either decision we take just wish that after the final decision we properly document/announce the steps we are going from now i.e. define where logic should be put in to avoid further discussions/refactors.

petemill:
I don't think we should be so purist as to declare 'all the logic' should be in the store. A line needs to be drawn somewhere, otherwise we'll end up sending every mousemove and click to the store for it to interpret the action based on coordinates, which is obviously too far 😀

The main question is what is more maintainable as well as more performant, and to me that is to keep UI-specific logic, e.g:
- when is a click considered a click (for a button)?
- when should an interaction animation occur? and;
- what is the timing for a hover to trigger an event (like in this component).

The best react / redux architectures are the ones which make the right decision for _each_ type of component, as there are no a blanket set of rules that should apply to each project. It's easier to get the perfect mix of readability, maintainability, performance, and speed of development this way. @cezaraugusto shall we create an issue and work on some suggested docs update to that effect?
Related list of links and discussions: https://redux.js.org/docs/faq/OrganizingState.html#organizing-state-only-redux-state

In fact, I would argue that the store is **definitely** not the place for this logic, since we're not even talking about state, but side-effects which goes against the functional flux philosophy: this 'reducer' was not dealing with state at all, but running a timeout which is then dispatching a separate action to the store. The more of that type of architecture we can remove, the cleaner and more performant the code will be. If anything, this logic should be part of the action creation process, but in the UI component is fine for something so specific IMO.
Related: https://redux.js.org/docs/faq/Actions.html#actions-side-effects

js/stores/windowStore.js
bridiver: I disagree that it's cleaner. First, we're back to an imperative action and those have been the source of endless problems. The component is also now deciding when the state should change, not when the UI action has been performed. The UI action is a mouse move. The state change for hover state may or may not occur depending based on that mouse move and conceivably something else might care about that mouse move without the timeout.

app/renderer/components/tabs/tab.js
```diff
@@ -153,19 +156,26 @@ class Tab extends React.Component {
...
+    this.mouseTimeout = setTimeout(
+      () => {
+        windowActions.setTabHoverState(this.props.frameKey, true, true)
```
bridiver: no matter what we do with this in terms of the timeout, I don't think we should go back to an imperative action. onmouseenter/leave should be changed at some point as well

bsclifton: agreed- should have a name like windowActions.onTabHoverChanged

bsclifton: Because it would take some work to either a) create an action like this or b) modify the existing uses of the action which could use a better name, I think that work should be deferred until the code behind setTabHoverState is touched.  Trying to capture this use case and the existing use-cases with one name would be tricky. The long term solution would be to move this logic to reducers

bsclifton:
master b9b49249e00dd2bd838fe5f48d0b44097a5ab9a2
0.21.x 30edcdd353356e3d2075cd8723a01258144bdac2
0.20.x 9df9566141d06d08f1c3c84e6b363f03c3170d1a

PR #10904, opened 2017-09-11
apply brave UA whitelist to first-party subresources

### LINKED ISSUE #10800: Brave UA whitelist only applies to top level page loads
## Test plan

https://github.com/brave/browser-laptop/pull/10904#issue-256869633

----

**- Did you search for similar issues before submitting this one?**

Yes

**- Describe the issue you encountered:**

Brave's UA whitelist only applies to top level page loads and not sub resources. I believe it's this line in the code:

https://github.com/brave/browser-laptop/blob/master/js/data/siteHacks.js#L46

**- Platform (Win7, 8, 10? macOS? Linux distro?):**

macOS 10.12.16 x64 - Reproduced
Win 10 x64 - Reproduced

**- Brave Version (revision SHA):**

Brave: 0.18.26 
rev: 67f256891907248a20563451d8ed455493bc4cc2 
Muon: 4.3.15 
libchromiumcontent: 60.0.3112.101 

**- Steps to reproduce:**

1. load adobe.com within Brave
2. open the developer tools via "option + command + I"
3. select the "Network" tab and reload adobe.com via "command + r"
4. select on the first resource and you'll notice the correct `Brave Chrome/60.0.3112.101` UA
5. select another resource that's underneath the first one and is from `Host:www.adobe.com`

You'll notice that the incorrect UA is being used.
    
**- Actual result:**

Only the first resource (top level) is correctly using the UA.

**- Expected result:**

The correct UA should be applied to every resource that's from the whitelist domain.

**- Will the steps above reproduce in a fresh profile? If not what other info can be added?**

Yes

**- Is this an issue in the currently released version?**

#10614 was supposed to fix this issue, however it only partially fixed the problem.

**- Can this issue be consistently reproduced?**

Yes

**- Screenshot if needed:**

Correct Example:

good

Incorrect Example:

bad

**- Any related issues:**

#10614

diracdeltas:
fix #10800

Test Plan:
load adobe.com within Brave
open the developer tools via "option + command + I"
select the "Network" tab and reload adobe.com via "command + r"
select on the first resource and you'll notice the correct Brave Chrome/60.0.3112.101 UA
select another resource that's underneath the first one and is from Host:www.adobe.com
the Brave UA should be there

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/siteHacks.js
+++ b/app/siteHacks.js
@@ -17,19 +17,20 @@ module.exports.init = () => {
   }
 
   Filtering.registerBeforeSendHeadersFilteringCB((details, isPrivate) => {
-    if (details.resourceType !== 'mainFrame') {
+    const mainFrameUrl = Filtering.getMainFrameUrl(details)
+    if (!mainFrameUrl) {
       return {
         resourceName
       }
     }
 
-    // This filter only applies to top-level requests, so details.url == mainFrameUrl
-    let domain = urlParse(details.url).hostname.split('.').slice(-2).join('.')
-    let hack = siteHacks[domain]
+    const domain = urlParse(mainFrameUrl).hostname.split('.').slice(-2).join('.')
+    const hack = siteHacks[domain]
     let customCookie
     let requestHeaders
     let cancel
-    if (hack && hack.onBeforeSendHeaders) {
+    if (hack && hack.onBeforeSendHeaders &&
+      domain === urlParse(details.url).hostname.split('.').slice(-2).join('.')) {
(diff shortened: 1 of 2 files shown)


NejcZdovc:
removing `sprint/1` label as this was not part of sprint planing. If we want to have this in this sprint please add this label on the issue as well. cc @zentagonist

diracdeltas:
@bsclifton IIUC provisional headers are shown before the filtering is applied, so that seems fine

PR #10911, opened 2017-09-12
Adds history limit into the historySites

### LINKED ISSUE #10895: History limit in the state

With site split PR I accidentally removed history limit in the state, so right now we are saving all history sites in the state without any limitations.

NejcZdovc:
## Submitter Checklist:

Resolves #10895

Auditors:

## Test Plan:
covered with automated tests

## Reviewer Checklist:

Tests


NejcZdovc:
@bbondy we use to do this inside `appStore` where we had direct access to `store` variable. But I am not sure how we can debounce it in the reducer where we need to return state from it.

This is the old implementation: https://github.com/brave/browser-laptop/blob/0.18.x/js/stores/appStore.js#L348

bbondy:
@NejcZdovc you can just debounce an action that will limit it on demand.  When the action actually goes through to the reducer it can be done sync at a lower frequency.

NejcZdovc:
@bbondy but if we debounce action, then history will not be update for a minute, which means that if you visit a site and go to `about:history` right after that you will not see it there and we will have the same problem that we have for top sites

bbondy:
The debounce only has to apply for enforcing the fuzzy limit, not for doing the add.

bbondy:
I was thinking you can have an action to clean up and wrap that in a debounced function that you call.  That call calls the action, and when the action hits the store it does the trimming.

bsclifton:
@bbondy can you please re-review when you have a chance? 😄

NejcZdovc:
this is not ready yet, we need to add debounce

bsclifton:
master a24496c14dbe6d5c0320ef947a7a26380adf8f69
0.21.x b36d1eb2c807f91bd7428c3b441341ffce1828d7
0.20.x f49ba730c4b8c66bbaa62c47235583e588f51da0

Issue #1353, opened 2016-04-13
whitelist script signatures


bridiver:
I recently went to adops.com and the site was basically unusable because they are on the blocking list. I don't really see any reason why we should be blocking jquery and other scripts that are not ad related. Not sure if a whitelist is the answer, but maybe consider the referrer url (when not in an iframe) because these are 1st-party scripts in this context.

diracdeltas:
this is a bug - TP/AB should not be blocking first party resources in the first place

diracdeltas:
need @bbondy input. i would have expected `adblock.matches(details.url, mapFilterType[details.resourceType], firstPartyUrl.host)` to return false when `details.url` is first-party to `firstPartyUrl.host`. apparently this is not the case. should this be fixed in the ABP module or is it a check we need to add to `app/adBlock.js`?

bbondy:
I thought we already had this check but apparently not.  We should just do an extra check to make sure the `details.url` host is different from the `firstPartyUrl.host` in app/adBlock.js.

bbondy:
re-opening, I think a rule like this will fix: `||adops.com$~third-party`

bbondy:
mostly fixed by the filtering rule committed above, but some things are still getting blocked that shouldn't (logo). I think it's a bug in the ad block library and could be a more generalized problem so blocking on this since the first party filtering change first lands in 0.20.x too.

bbondy:
Never mind, I just tested with combined data file and it's working. I was testing before with custom adblock rule exception which runs  in an independent instance

cezaraugusto:
heads-up I added a basic test plan please add more checks in case mine is incomplete. Thanks

PR #11005, opened 2017-09-19
Allow adblock checking within the same origin

### LINKED ISSUE #11004: Allow adblock checking within the same origin
Currently we only do same origin adblock checks for a couple websites.  We should be doing it for all sites for the best blocking results.

bbondy:
Fix #11004

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/ads/adBlockUtil.js
+++ b/app/browser/ads/adBlockUtil.js
@@ -2,10 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const {siteHacks} = require('../../../js/data/siteHacks')
 const {FilterOptions} = require('ad-block')
-const isThirdPartyHost = require('../isThirdPartyHost')
-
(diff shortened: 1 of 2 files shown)


bbondy:
pls hold on reviewing, just adding some tests in another commit first.

bbondy:
OK good to go.

bbondy:
Mentioned on Slack but this isn't meant to review all filtering list rules.  Some won't work like the example you gave but it works the same as uBlock and other adblockers.  We should fix cases like this, if we care about them in our maintained lists on github.com/brave/adblock-lists

diracdeltas:
@bbondy noted. in #1353 i asked whether this was an issue that should be fixed in the adblock module and you mentioned (https://github.com/brave/browser-laptop/issues/1353#issuecomment-209698848) that we should add a check for first-partiness in app/adblock.js. so this appeared to be a regression

bbondy:
Yep but things change in a year and a half. 

We don't have the same adblock library we had back then, we now have lists we maintain which we didn't then, and also we were consciously trying to not block first party things then. I'm going to re-open that one and fix with an adblock definition update.

Issue #10288, opened 2017-08-03
Prevent further tracking by blocking additional methods


snyderp:
- Did you search for similar issues before submitting this one?
  Yap

- Describe the issue you encountered:
   There are a large number of methods that are overwhelmingly (>95%) used for tracking on the web, but which are currently not blocked out with shields up.  Many of these would make good candidates for blocking

- Platform (Win7, 8, 10? macOS? Linux distro?):
   all

- Brave Version (revision SHA):
   00cc603

- Steps to reproduce:
    1. Try calling a tracking related method, like `SVGTextContentElement.prototype.getComputedTextLength()`
    
- Actual result:
   The method works as the standard defines

- Expected result:
   Access to it to be blocked in shields up mode

- Will the steps above reproduce in a fresh profile? If not what other info can be added?
   Yep

- Is this an issue in the currently released version?
   Yep

- Can this issue be consistently reproduced?
  Yep

The below is a list of methods that we found to have a 95% or more reduction in use when installing Ghostery, and which are used on >= 100 sites in the Alexa 10k at the time of measurement (~1.5 years ago…).  Some methods that are heavily related to tracking are 
* `SVGTextContentElement.prototype.getComputedTextLength`
* `CanvasRenderingContext2D.prototype.quadraticCurveTo`
* `MessagePort.prototype.start`
* `Document.prototype.elementFromPoint`

There are many other compelling "options" on the list  

feature_name | num_sites_using | pct_reduction_when_blocking
-- | -- | --
CanvasRenderingContext2D.prototype.scale | 583 | 0.9503
HTMLDocument.prototype.captureEvents | 121 | 0.9504
WebSocket.prototype.close | 281 | 0.9537
Element.prototype.scrollIntoView | 108 | 0.9537
Element.prototype.getClientRects | 1064 | 0.9539
window.atob | 4950 | 0.9541
URL.createObjectURL | 1670 | 0.9545
HTMLInputElement.prototype.checkValidity | 3894 | 0.9545
CSSStyleDeclaration.prototype.removeProperty | 478 | 0.9561
CanvasRenderingContext2D.prototype.bezierCurveTo | 185 | 0.9568
performance.onresourcetimingbufferfull | 210 | 0.9571
window.postMessage | 4762 | 0.9591
window.onunload | 2505 | 0.9593
window.btoa | 4146 | 0.9595
document.writeln | 1928 | 0.9595
HTMLDocument.prototype.writeln | 1928 | 0.9595
[...]

PR #11055, opened 2017-09-20
Block additional fp-related methods

snyderp:
Fixes brave/browser-laptop#10288

This change would block the following 5 methods (presented below, with brief motivation for each).  Numbers for feature use / tracking use are taken from https://www.cs.uic.edu/%7Epsnyder/static/papers/Browser_Feature_Usage_on_the_Modern_Web.pdf

`CanvasRenderingContext2D.prototype.isPointInPath`
- Is used in popular [live, popular fingerprinting code](https://github.com/Valve/fingerprintjs2/blob/master/fingerprint2.js#L673)
- Is infrequently used on the web (was observed on only 166 sites in the Alexa 10k)
- Is frequently blocked by anti-tracking tools (in the presence of Ghostery, it is only seen on 28 sites in the Alexa 10k, suggesting its used for tracking 83% of the time) 

`WebGLRenderingContext.prototype.getUniformLocation` and `WebGLRenderingContext.prototype.getAttribLocation`
- Are used in popular live, popular fingerprinting code, ([here](https://github.com/Valve/fingerprintjs2/blob/master/fingerprint2.js#L760) and [here](https://github.com/Valve/fingerprintjs2/blob/master/fingerprint2.js#L759), for example)
- Are infrequently used on the web (was observed on only 255 and 250 sites in the Alexa 10k, respectivly)
- Are frequently blocked by anti-tracking tools (in the presence of Ghostery, it is only seen on 44 and sites in the Alexa 10k, suggesting its used for tracking 82.75% and 82.4% of the time)
- Non-obvious use case (e.x.: its not obvious to see why you'd need to query these parameters out of the context, if you'd already set them)

`SVGPathElement.prototype.getTotalLength`
- Anecdotally Used in ways similar to canvas finger printing (font enumeration, getting subtle differences in rendering between platforms)
- Very infrequently used on the web (observed on 140 of the Alexa 10k)
- Very frequently associated with tracking (use goes down to only 2 sites, or a 98.57% reduction, in the presence of Ghostery)

`SVGTextContentElement.prototype.getComputedTextLength`
- Anecdotally Used in ways similar to canvas finger printing (font enumeration, getting subtle differences in rendering between platforms)
- Extremely associated with tracking (use goes from 1003 sites in the Alexa 10k, to 1, or a 99.9% reduction, in the presence of Ghostery)


luixxiul:
@snyderp @diracdeltas were those commits merged somewhere?

luixxiul:
nvm, I found the PR: https://github.com/brave/browser-laptop/pull/11140

PR #11084, opened 2017-09-21
remove unittest from pre-push hook

### LINKED ISSUE #11081: remove unittest from pre-push hook
i find that the unittest pre-push hook is more annoying than useful. ex:

1. unittests are sometimes broken on master (such as right now) for reasons unrelated to my change, so i end up having to use --no-verify.
2. pre-push hooks aren't run when merging via Github's UI so ultimately there is no guarantee that the unittests are passing
3. unittests take a long time to run and should be checked after-push in Travis anyway

diracdeltas:
fix #11081

### DIFF
--- a/tools/lib/ignoredPaths.js
+++ b/tools/lib/ignoredPaths.js
@@ -3,7 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 module.exports = [
-  'pre-comit',
+  'pre-commit',
   'pre-push',
   'test/',
   'docs/',
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "flow": "flow; test $? -eq 0 -o $? -eq 2",
     "lint": "standard",
     "postinstall": "npm run download-sync-client && webpack",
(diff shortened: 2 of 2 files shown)


tools/lib/ignoredPaths.js
```diff
@@ -3,7 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 module.exports = [
-  'pre-comit',
+  'pre-commit',
```
luixxiul: I've been not sure if this was a typo: https://github.com/brave/browser-laptop/pull/9690#discussion_r123872847

diracdeltas: cc @NejcZdovc on whether this is a typo

diracdeltas:
@bsclifton that's a reasonable option

diracdeltas:
@bsclifton done!

tools/lib/ignoredPaths.js
bbondy: These are sub-dirs of node_modules mostly so yes it's a typo. Thx.

PR #11103, opened 2017-09-23
always return copy of parse url and remove option for query as object

### LINKED ISSUE #11101: urlParse returns inconsistent types for `.query`
There are really no steps to replicate the problem because it was partially addressed here https://github.com/brave/browser-laptop/commit/d142b243150684da669d86c4d1e83cc5d20769a5

bridiver:
fix #11101

Submitter Checklist:

Test Plan:
As mentioned in the ticket there really isn't a good way to QA this

Reviewer Checklist:

Tests

### DIFF
--- a/app/common/urlParse.js
+++ b/app/common/urlParse.js
@@ -16,14 +16,13 @@ try {
 
 let cachedUrlParse = new LRUCache(config.cache.urlParse)
 
-module.exports = (url, ...args) => {
+module.exports = (url) => {
   let parsedUrl = cachedUrlParse.get(url)
-  if (parsedUrl) {
-    // make a copy so we don't alter the cached object with any changes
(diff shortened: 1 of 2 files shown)


bridiver:
this is also for legacy node url.parse and muon.url.parse compat because muon always returns a string

NejcZdovc:
Yeah we have this scenario in `ledger.js` `urlParse(location, true)`

bridiver:
@NejcZdovc but the query string is never used so it has no effect

bbondy:
This was also needed for 0.21.x and 0.20.x


0.20.x: 498ce5ce8600ffe4c342329c70384596a98eda70
0.21.x: 6e5b80b51b6eb16e470ea677962d7e0a87dbc191

PR #11104, opened 2017-09-23
Muon issue 230

bridiver:
Changes for https://github.com/brave/muon/pull/279


darkdh:
we might also need to change user-data-dir to user-data-dir-name in package.json

Issue #11142, opened 2017-09-25
extension icons do not appear at top right of app


kevinlawler:
### Description
Extensions like Pocket or Pinterest do not have their icons appear in the top right of the app. This is due to the wrong protocol (`file://`) being caught by CSP restrictions.

### Steps to Reproduce

Build and load browser. Turn on Pocket extension

**Actual result:**
![screen shot 2017-09-25 at 2 29 12 pm](https://user-images.githubusercontent.com/376238/30831864-f0d23772-a1fd-11e7-944c-6b1104d9ecf8.png)

Note vertical grey bar `|`.

**Expected result:**
Pocket icon should appear.

**Reproduces how often:**
100%

### Brave Version

Brave: 0.21.0 
V8: 6.1.534.39 
rev: 25130655e6717f9e3b6f77d6dfed94e990a09960 
Muon: 4.6.3 
OS Release: 16.7.0 
Update Channel: Release 
OS Architecture: x64 
OS Platform: macOS 
Node.js: 7.9.0 
Brave Sync: v1.3.5 
libchromiumcontent: 61.0.3163.98

**Reproducible on current live release:**
No. Not a problem on the live build.

### Additional Information
None

bsclifton:
Fixed with https://github.com/brave/browser-laptop/pull/11143

PR #11143, opened 2017-09-25
extension icons should appear

kevinlawler:
changes `file://` to `chrome://brave` for extension icon loading

Issue brave/browser-laptop#11142

Auditors:
@bridiver @diracdeltas @darkdh @jonathansampson

### DIFF
--- a/js/lib/appUrlUtil.js
+++ b/js/lib/appUrlUtil.js
@@ -24,6 +24,13 @@ module.exports.fileUrl = (filePath) => {
   return encodeURI('file://' + fileUrlPath)
 }
 
+module.exports.chromeUrl = (filePath = '') => {
+  filePath = module.exports.fileUrl(filePath)
+  filePath = filePath.replace('file://', 'chrome://brave')
+
+  return filePath
+}
+
 /**
  * Gets the URL of a page hosted by the braveExtension or torrentExtension
  * Returns 'chrome-extension://<...>'
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -3,7 +3,7 @@ const contextMenus = require('./browser/extensions/contextMenus')
 const extensionActions = require('./common/actions/extensionActions')
 const config = require('../js/constants/config')
 const appConfig = require('../js/constants/appConfig')
-const {fileUrl} = require('../js/lib/appUrlUtil')
+const {chromeUrl} = require('../js/lib/appUrlUtil')
 const {getExtensionsPath, getBraveExtUrl, getBraveExtIndexHTML} = require('../js/lib/appUrlUtil')
 const {getSetting} = require('../js/settings')
 const settings = require('../js/constants/settings')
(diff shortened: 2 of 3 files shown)


js/lib/appUrlUtil.js
```diff
@@ -24,6 +24,13 @@ module.exports.fileUrl = (filePath) => {
   return encodeURI('file://' + fileUrlPath)
 }
 
+module.exports.chromeUrl = (filePath = '') => {
```
diracdeltas: would be good to have some unit tests for this

js/lib/appUrlUtil.js
```diff
@@ -24,6 +24,13 @@ module.exports.fileUrl = (filePath) => {
...
+  filePath = module.exports.fileUrl(filePath)
+  filePath = filePath.replace('file://', 'chrome://brave')
```
diracdeltas: where is the `chrome://brave` to `file://` mapping implemented in muon? it seems like a potential security hole since it can be used to bypass CSP restrictions; we should make sure `chrome://brave` loads are blocked in untrusted contexts.

diracdeltas: also i'm curious what CSP violations are being triggered? `app/extensions/brave/index.html` explicitly allows `img-src file:` but not `chrome:` so i'm confused why the original issue happens

kevinlawler: I played with the CSP file but couldn't get the violation to reveal itself either. It might be some other security restriction. The error from the console is

    [75566:775:0922/145254.257520:ERROR:CONSOLE(0)] "Not allowed to load local resource:
        file:///Users/kevin/Desktop/brave/browser-laptop-bootstrap/src/browser-laptop/app/extensions/metalmash/images/icon-19.png", 
        source: chrome://brave/Users/kevin/Desktop/brave/browser-laptop-bootstrap/src/browser-laptop/app/extensions/brave/index-dev.html (0)

and it disappears when switching to `chrome://brave`. Subsequently the image loads.

js/lib/appUrlUtil.js
kevinlawler: added in 1bc13ba

js/lib/appUrlUtil.js
bridiver: @diracdeltas chrome://brave can't be loaded from web pages `Not allowed to load local resource: chrome://brave/usr/local/brave/browser-laptop-bootstrap/README.md`. When I originally tested them it also wasn't possible to load them by entering the url manually, but I just tested again and it is now possible because we added support for `chrome://*` urls in general so we should probably block them from the urlbar

bridiver: Also, the csp can't allow something that is prohibited by default (access to file resources) which is why the file:// urls don't work

cezaraugusto:
what's the current status of this?

js/lib/appUrlUtil.js
bridiver: can we use a relative path instead and make it filesystem agnostic?

bsclifton:
Submitted security review after seeing the above text:
https://github.com/brave/internal/issues/167

js/lib/appUrlUtil.js
diracdeltas: I think it's fine if chrome://brave  URLs can be loaded manually from the urlbar since that is also possible for file URLs. just want to make sure there is nothing they can do that files URLs can't do, other than be loaded in the chrome:// context.

bsclifton:
master cc39571d40a761f3a31021dc584191f228cba783
0.21.x 1890583d8ccb25443ad7551be062a98b495d548d
0.20.x 494cf4ece7b639ca40a66542f85e8fda8f3d0b90

Issue #11161, opened 2017-09-27
siteCache Enhancement


darkdh:
1. Support clearing method triggered by clearing browserHistory
2. Fix invalid favicon only entry generation
3. Migration of invalid favicon entry and trailing slash siteCache

darkdh:
closed by #11162

PR #11174, opened 2017-09-27
Manually apply changes from siteCache Enhancement PR to master

bsclifton:
See https://github.com/brave/browser-laptop/pull/11162 for more info

Fixes https://github.com/brave/browser-laptop/issues/11161 in master (already fixed in 0.19.x and 0.20.x)

Auditors: @darkdh, @NejcZdovc, @ayumi 

Submitter Checklist:

Test Plan:
`npm run unittest`

Reviewer Checklist:

Tests


app/common/state/clearDataState.js
```diff
@@ -0,0 +1,24 @@
...
+const { makeImmutable, isMap } = require('./immutableUtil')
+
+const validateState = function (state) {
+  state = makeImmutable(state)
+  assert.ok(isMap(state), 'state must be an Immutable.Map')
+  return state
+}
+
+const api = {
+  getClearDefaults: (state) => {
```
bsclifton: this could also be used to refactor these areas
https://github.com/brave/browser-laptop/blob/54915c9655e2efd0a44a40fe43f2da679500add7/app/browser/reducers/historyReducer.js#L44
https://github.com/brave/browser-laptop/blob/54915c9655e2efd0a44a40fe43f2da679500add7/app/browser/menu.js#L718
https://github.com/brave/browser-laptop/blob/54915c9655e2efd0a44a40fe43f2da679500add7/app/ledger.js#L184
https://github.com/brave/browser-laptop/blob/54915c9655e2efd0a44a40fe43f2da679500add7/js/stores/appStore.js#L433

bsclifton: I did actually convert them but historyReducer had a failing test so I backed out.  I felt it's better to leave those alone since they're outside the scope of this PR.

test/unit/app/common/state/bookmarksStateTest.js
```diff
@@ -71,6 +71,59 @@ describe('bookmarkState unit test', function () {
       assert.deepEqual(bookmarksState.getBookmarkOrder(stateWithData), stateWithData.getIn(['cache', 'bookmarkOrder']))
     })
   })
+
+  describe('updateFavicon', function () {
```
bsclifton: This section mostly copied from https://github.com/brave/browser-laptop/blob/0213d8f30eedf909690c8876fa9f31205bcf0ff5/test/unit/state/siteUtilTest.js#L1068

@NejcZdovc there might be more tests from the old siteUtil that we can pull over (similar to this). For example, `moveSite`

NejcZdovc:
`bookmarkLocationCache` is not the same as browser history. In `bookmarkLocationCache` we are saving bookmarks, but not via bookmark key but based on location. So let's say that you bookmark `clifton.io`, in this cache we would have `clifton.io: [ { key: `https://clifton.io|0|0`}]`. Clearing this would mean that we lose caching data.

test/unit/app/common/state/bookmarksStateTest.js
NejcZdovc: I plan to cover all this functions in #10310, it will be unblocked soon

app/common/state/bookmarksState.js
```diff
@@ -187,7 +187,9 @@ const bookmarksState = {
     }
 
     siteKeys.forEach((siteKey) => {
-      state = state.setIn([STATE_SITES.BOOKMARKS, siteKey, 'favicon'], favicon)
+      if (state.getIn([STATE_SITES.BOOKMARKS, siteKey])) {
+        state = state.setIn([STATE_SITES.BOOKMARKS, siteKey, 'favicon'], favicon)
```
bsclifton: this change in particular was causing @darkdh to have duplicate entries... one regular and one with only a favicon entry

app/common/state/bookmarksState.js
```diff
@@ -187,7 +187,9 @@ const bookmarksState = {
     }
 
     siteKeys.forEach((siteKey) => {
-      state = state.setIn([STATE_SITES.BOOKMARKS, siteKey, 'favicon'], favicon)
+      if (state.getIn([STATE_SITES.BOOKMARKS, siteKey])) {
```
NejcZdovc: If we only checking here if exists I would change it to `state.hasIn([STATE_SITES.BOOKMARKS, siteKey])`

Issue #11256, opened 2017-10-03
ledger notification should appear above the tab area


diracdeltas:
ledger notifications are appearing in the tab area instead of above it. this is not ideal because:
1. they are browser notifications, not tab-specific notifications. i.e., if you change tabs, the notification still appears.
2. they can be spoofed by an attacker webpage if they are in the tab area

![screen shot 2017-10-03 at 5 49 53 pm](https://user-images.githubusercontent.com/549654/31140318-962df27c-a863-11e7-984c-90eacac61c6c.png)

expected behavior: browser notifications that are not tab-specific should appear above the tab strip, like the update notification for example:
![screen shot 2017-10-03 at 5 53 38 pm](https://user-images.githubusercontent.com/549654/31140434-e0c91d98-a863-11e7-8631-df165a84f83a.png)

 
Brave: 0.19.27 
rev: e7f3a69f10bb7688841947efe56c302ed2f8d1dd 
Muon: 4.4.25 
libchromiumcontent: 61.0.3163.100 
V8: 6.1.534.41 
Node.js: 7.9.0 
Update Channel: Beta 
OS Platform: macOS 
OS Release: 15.6.0 
OS Architecture: x64

cc @cezaraugusto @NejcZdovc

diracdeltas:
cc @bradleyrichter

bradleyrichter:
@diracdeltas yes...absolutely.

see also: https://github.com/brave/browser-laptop/pull/11265

diracdeltas:
moving the milestone to 0.19.x based on what @luixxiul set for https://github.com/brave/browser-laptop/issues/8934

bsclifton:
@diracdeltas @bradleyrichter is this a blocker for 0.19.x / the BAT Mercury launch?

bradleyrichter:
@bsclifton Can't we just define it correctly to use the same alert bar as the update notification?

Issue #8934, opened 2017-05-18
wrong alert bar used for app-level alerts:


bradleyrichter:
![image](https://cloud.githubusercontent.com/assets/13509546/26216318/7b482daa-3bb8-11e7-8650-e33a77a3dc4f.png)
The original alert bar should be used. (the one which is located above the tabs bar, and does not include the orange bottom border.

luixxiul:
adding the milestone just in case.

diracdeltas:
This isn't a problem for all app alerts - for instance the update alert bar appears in the right place. 

https://github.com/brave/browser-laptop/issues/11256 is probably a duplicate of this

bsclifton:
Closing this issue as a duplicate of https://github.com/brave/browser-laptop/issues/11256 (the issue you created @diracdeltas has much better info / requirements 😄 )

PR #11265, opened 2017-10-03
Ledger notifications refactor and add new BTC=>BAT notification

bsclifton:
screen shot 2017-10-03 at 4 48 39 pm

Fixes https://github.com/brave/browser-laptop/issues/11021

Presents this one-time "Converted BTC to BAT" alert:
 - if payments are enabled
 - user has a positive balance
 - this is an existing profile (new profiles will have firstRunTimestamp matching btcToBatTimestamp)
 - notification has not already been shown yet

Includes a "learn more" link pointing to:
https://brave.com/faq-payments/#brave-payments

Has a possible security risk- the deep link allows URL hash to modify state (limited to boolean state variables on about:preferences)

Auditors: @NejcZdovc, @diracdeltas

## Submitter Checklist:

## Automated Test Plan:
`npm run unittest -- --grep="ledger api unit tests"`
`npm run unittest -- --grep="ledgerReducer unit tests"`
`npm run unittest -- --grep="Preferences component"`

## Reviewer Checklist:

## Merge Checklist:
- [x] Merge to master
- [ ] rebase to 0.20.x (taking care to make sure setting names match)
- [ ] cherry-pick the 0.20.x rebase to 0.19.x
- [ ] list commit hashes here


js/about/preferences.js
```diff
@@ -759,13 +766,42 @@ class AboutPreferences extends React.Component {
...
+      newState[params] = true
+    }
+    this.setState(newState)
   }
 
+  /**
+   * Parses a query string like:
+   * about:preferences#payments?ledgerBackupOverlayVisible
+   * and returns the part:
+   * `about:preferences#payments`
```
diracdeltas: This comment doesn't seem right. Previously this would return 'payments' not 'about:preferences#payments'.

bsclifton: Good catch; fixed 😄

app/browser/api/ledger.js
```diff
@@ -104,13 +102,264 @@ const fileTypes = {
...
+    }
+  },
+  onInterval: (state) => {
+    if (!getSetting(settings.PAYMENTS_NOTIFICATIONS)) {
+      return
+    }
+    if (getSetting(settings.PAYMENTS_ENABLED)) {
+      notifications.showEnabledNotifications(state)
+    } else {
+      notifications.showDisabledNotifications(state)
```
bbondy: before we showed this notification only when `!getSetting(settings.PAYMENTS_NOTIFICATIONS)` is that change intentional?

bsclifton: logic shouldn't have changed at all (just moved). If PAYMENT_NOTIFICATIONS is set, none of the notifications should be shown

app/browser/api/ledger.js
```diff
@@ -104,13 +102,264 @@ const fileTypes = {
...
+            windowId: activeWindow.id
+          })
+        }
+        break
+
+      default:
+        return
+    }
+
+    appActions.hideNotification(message)
```
bbondy: maybe the above default: should be just break so this always runs?

app/browser/api/ledger.js
bbondy: I guess I'm thinking about getSetting(settings.PAYMENTS_NOTIFICATIONS) is false but getSetting(settings.PAYMENTS_ENABLED is true, before it would show.

app/browser/api/ledger.js
bsclifton: I considered that... but since other messages may be coming through (since this fires on `ipc.on(messages.NOTIFICATION_RESPONSE`) I didn't want to change the behavior

app/browser/api/ledger.js
bsclifton: Great catch- I rolled back to the original behavior (so that there is no change from previous code)

bradleyrichter:
@bsclifton This alert bar location looks wrong. The screen shot above shows the per-tab location and this should be at the app level.

luixxiul:
https://github.com/brave/browser-laptop/issues/11256

luixxiul:
also: https://github.com/brave/browser-laptop/issues/8934

bsclifton:
master https://github.com/brave/browser-laptop/commit/9285507f67b07a2537b2e1581c4c30e3bf2b3748
0.20.x https://github.com/brave/browser-laptop/commit/8f2fc5cefbd9d8c3c24391d66bfeb64c6e471440
0.19.x https://github.com/brave/browser-laptop/commit/e84899a1390d4c5e2bc4def1e09272105a5e3c62

PR #11423, opened 2017-10-10
Remove capital B from brave in snap instructions

Torild:
Installing doesn't work using a capital B


luixxiul:
Hello, that bug will be going to be fixed with https://github.com/brave/browser-laptop/pull/11408, so please let me close the issue in favor of that. 

Please keep up the nice work :-)

PR #11469, opened 2017-10-11
Do not re-create stylesheet rules every time tabs are rendered

### LINKED ISSUE #11465: (minor) Tab performance - aphrodite stylesheets created during render
### Description

Some of the components used to render tabs are calling Stylesheet.create on every render() call. This was introduced in f9b2a2c543a9f514782c7c689bbe544422d13c9a which was the tab component module split (#8347, #8348). This can create a lot of unnecessary css rules added to the document, and maybe even unnecessary paints.

Affected components:
tab.js
favIcon.js
tabIcon.js

### Brave Version
master - introduced in f9b2a2c543a9f514782c7c689bbe544422d13c9a
**Reproducible on current live release:** Affected code not yet deployed to master

petemill:
Helps with (a very small amount of) performance - instead use css variables to pass state variables to css. But every little thing helps when working on tab re-ordering.

Fix #11465

Submitter Checklist:

Test Plan:
Since this is a small refactor, suggested test plans from previous implementations welcome. There should be 0 visual regressions from master. Affected components:
- Tab theme color
- Tab text / filter when theme color is light or dark
- Favicon display
- General background / text / hover colors when there is and isn't a favicon, should match master

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/tabs/content/favIcon.js
+++ b/app/renderer/components/tabs/content/favIcon.js
@@ -52,45 +52,40 @@ class Favicon extends React.Component {
       return null
     }
 
-    const iconStyles = StyleSheet.create({
-      icon__loading_color: {
-        filter: this.props.tabIconColor === 'white'
-          ? filter.makeWhite
-          : 'none'
-      },
-      icon__favicon: {
-        backgroundImage: `url(${this.props.favicon})`,
-        filter: this.props.tabIconColor === 'white'
-          ? filter.whiteShadow
-          : 'none'
-      },
-      icon__default_sizeAndColor: {
-        WebkitMaskSize: this.props.showIconAtReducedSize ? '10px' : '12px',
-        backgroundColor: this.props.tabIconColor === 'white'
-          ? color.white100
-          : color.mediumGray
-      }
-    })
+    const themeLight = this.props.tabIconColor === 'white'
+    const instanceStyles = { }
+    if (this.props.favicon) {
+      instanceStyles['--faviconsrc'] = `url(${this.props.favicon})`
+    }
 
     return <TabIcon
       data-test-favicon={this.props.favicon}
       data-test-id={this.testingIcon}
       className={css(
         styles.icon,
-        this.props.favicon && iconStyles.icon__favicon,
+        this.props.favicon && styles.icon_favicon,
+        this.props.favicon && themeLight && styles.icon_faviconLight,
         !this.props.isPinned && this.props.showIconWithLessMargin && styles.icon_lessMargin,
         !this.props.isPinned && this.props.showIconAtReducedSize && styles.icon_reducedSize
       )}
+      style={instanceStyles}
       symbol={
         this.props.tabLoading
           ? (
             // no loading icon if there's no room for the icon
             !this.props.showIconAtReducedSize &&
-            css(styles.icon__loading, iconStyles.icon__loading_color)
+            css(
+              styles.icon__loading,
+              themeLight && styles.icon__loading_colorLight
+            )
           )
           : (
             !this.props.favicon &&
-            css(styles.icon__default, iconStyles.icon__default_sizeAndColor)
+            css(
+              styles.icon__defaultIcon,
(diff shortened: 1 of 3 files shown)


app/renderer/components/tabs/content/tabIcon.js
```diff
@@ -43,14 +30,16 @@ class TabIcon extends ImmutableComponent {
...
       {...altProps}
     >
       {
         this.props.symbol
           ? <span
             className={cx({
               [this.props.symbol]: true,
-              [css(styles.tabIcon)]: true
+              [css(styles.tabIcon)]: true,
+              [css(styles.tabIcon_hasSymbol)]: this.props.symbolContent
```
luixxiul: Would you mind combining these two lines into `[css(styles.tabIcon, this.props.symbolContent && styles.tabIcon_hasSymbol)]: true`, since `cx` basically does what this document describes: https://github.com/Khan/aphrodite#dont-do-this?

> Why does it matter? Although the second one will produce a valid class name, it cannot guarantee that the bar styles will override the foo ones.

app/renderer/components/tabs/content/favIcon.js
```diff
@@ -52,45 +52,40 @@ class Favicon extends React.Component {
...
+    }
 
     return <TabIcon
       data-test-favicon={this.props.favicon}
       data-test-id={this.testingIcon}
       className={css(
         styles.icon,
-        this.props.favicon && iconStyles.icon__favicon,
+        this.props.favicon && styles.icon_favicon,
+        this.props.favicon && themeLight && styles.icon_favicon__light,
```
luixxiul: Should it not be `styles.icon_favicon_light`?

app/renderer/components/tabs/content/tabIcon.js
petemill: Thanks for the spot - I had just read that "don't do this" section today and had the same thought

app/renderer/components/tabs/content/favIcon.js
petemill: @luixxiul yes, my mind getting confused with the different characters for bem that we're using! I've just updated all the names and think that they're conforming now so would be interested in what you think of the latest - thanks!

luixxiul: maybe we'd create a bit more complicated example on https://github.com/brave/browser-laptop/blob/master/docs/style.md#defining-our-blocks-elements-and-modifiers ;-)

app/renderer/components/tabs/content/favIcon.js
```diff
@@ -143,9 +146,22 @@ const styles = StyleSheet.create({
...
-    WebkitMaskImage: `url(${defaultIconSvg})`
+    WebkitMaskImage: `url(${defaultIconSvg})`,
+    WebkitMaskSize: '12px',
+    backgroundColor: color.mediumGray
+  },
+
+  icon__defaultIcon_reducedSize: {
+    WebkitMaskSize: '10px'
+  },
+  icon__defaultIcon_colorLight: {
```
luixxiul: would you please add a blank line above this? https://github.com/brave/browser-laptop/blob/master/docs/style.md#always-add-space-between-styles-objects

app/renderer/components/tabs/content/favIcon.js
luixxiul: The other names look good to me :-)

app/renderer/components/tabs/content/favIcon.js
petemill: thanks - done

luixxiul:
> ++ nice work. Would be cool if at some point you could reference the usage of native variables instead of creating new styles w/ Aphrodite in a wiki/docs. cc @luixxiul for better insight

+1 from me 😄

petemill:
Moved to milestone 0.22.x due to requiring it for a clean merge of #12565 and the same / similar test plan for that version.

petemill:
0.22.x 5a9bb4745757dc7be49b20c91d55e97dcae601bc

Issue #11492, opened 2017-10-12
QR codes not being displayed when creating a new wallet


kjozwiak:
### Description

When creating a new wallet, you'll notice that the QR codes are not being displayed under the "Add Fund" manager.

### Steps to Reproduce

  1. launch the latest brave (`0.19.48` was being used in this instance)
  2. create a new wallet under `about:preferences#payments`
  3. select "Add Funds" and click on any cryptocurrency that's listed in the manager

**Actual result:**

<img width="723" alt="screen shot 2017-10-12 at 1 16 59 pm" src="https://user-images.githubusercontent.com/2602313/31510195-002aa91e-af52-11e7-8f82-d6279c662f3a.png">

QR code isn't being displayed.

**Expected result:**

QR code should be visible as another way of adding funds into the wallet.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100% reproducible using the above STR.

### Brave Version

**about:brave info:**

> Brave: 0.19.48 
> rev: de939f6dcd1647f43de544b22dfdd103585b4ec2 
> Muon: 4.4.28 
> libchromiumcontent: 61.0.3163.100 

**Reproducible on current live release:**

No, part of BAT which is being introduced in `0.19.x`

bsclifton:
I think this error in the console is related:
`Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>. See p > ... > p.`

bsclifton:
I believe I found the issue

This call should be generating the code- it's never getting called:
https://github.com/brave/browser-laptop/blob/a314b4e1c83cfbcc6b958bdc7ce6d00837c7cb4c/app/browser/api/ledger.js#L1735

I believe this was broken here:
https://github.com/brave/browser-laptop/commit/c96500235aa9d1f431b67edfae875b1d56bad528

Because the unconfirmed amount is 0 (which matches the state), it returns without making the call.  Looking into a fix...

kjozwiak:
Also checked the following under Ubuntu 17.04 x64:

* ensure that the QR codes are appearing when creating a new wallet with a clean profile under `0.19.52`
* ensure that the QR codes are appearing when upgrading from `0.19.36` -> '0.19.52'
* esnure that the QR codes are appearing once a wallet as been recovered under '0.19.52'

PR #11496, opened 2017-10-13
Fix QR codes not displaying for new wallets

bsclifton:
Fix unconfirmed funds check which prevented code from setting payment data

Fixes https://github.com/brave/browser-laptop/issues/11492

Auditors: @NejcZdovc

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/api/ledger.js
+++ b/app/browser/api/ledger.js
@@ -1717,16 +1717,11 @@ const onWalletProperties = (state, body) => {
 
   // unconfirmed amount
   const unconfirmed = parseFloat(body.get('unconfirmed'))
-  if (unconfirmed > 0) {
-    if (ledgerState.getInfoProp(state, 'unconfirmed') === unconfirmed) {
-      return state
-    }
-
+  if (unconfirmed >= 0) {
     state = ledgerState.setInfoProp(state, 'unconfirmed', unconfirmed)
     if (clientOptions.verboseP) {
       console.log('\ngetBalance refreshes ledger info: ' + ledgerState.getInfoProp(state, 'unconfirmed'))
     }
-    return state
   }
 
   if (clientOptions.verboseP) {
--- a/app/browser/api/ledger.js
+++ b/app/browser/api/ledger.js
@@ -1717,7 +1717,7 @@ const onWalletProperties = (state, body) => {
 
   // unconfirmed amount
   const unconfirmed = parseFloat(body.get('unconfirmed'))
-  if (unconfirmed >= 0) {
+  if (unconfirmed > 0) {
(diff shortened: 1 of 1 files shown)


bsclifton:
Problem was introduced with https://github.com/brave/browser-laptop/commit/c96500235aa9d1f431b67edfae875b1d56bad528#diff-3cdfdb124f718787d584e9e2b82fcedcR1720 via https://github.com/brave/browser-laptop/pull/11457

When reviewing, we need to make sure this issue is still solved:
https://github.com/brave/browser-laptop/issues/11449

app/browser/api/ledger.js
```diff
@@ -1717,7 +1717,7 @@ const onWalletProperties = (state, body) => {
 
   // unconfirmed amount
   const unconfirmed = parseFloat(body.get('unconfirmed'))
-  if (unconfirmed >= 0) {
+  if (unconfirmed > 0) {
```
NejcZdovc: I think this change is not correct, because you could have wallet that have unconfirmed balance, let's say 10$. Then you import wallet from the backup, which have unconfirmed balance 0. New value will not be saved.

This `return` in the if block is more or less from the old code. What I would do is to change it to

```
const unconfirmed = parseFloat(body.get('unconfirmed'))
if (unconfirmed >= 0) {
    state = ledgerState.setInfoProp(state, 'unconfirmed', unconfirmed)
    if (clientOptions.verboseP) {
      console.log('\ngetBalance refreshes ledger info: ' + ledgerState.getInfoProp(state, 'unconfirmed'))
    }
}
```

bbondy: want to just PR an alt fix @NejcZdovc ?

NejcZdovc: will push to this PR

NejcZdovc:
@bbondy @bsclifton pushed a fix that removes return statements. Please check if everything is ok.

bsclifton:
- QR code loads properly with this
- recovering wallet should also work great since flow is unchanged

bsclifton:
master https://github.com/brave/browser-laptop/commit/0541557ca770e27516ee23ea0a825a138a365046
0.21.x https://github.com/brave/browser-laptop/commit/4403ea43f655492b733af711f7b66ee1391aee5c
0.20.x https://github.com/brave/browser-laptop/commit/4d5b42294fe7c4f4a38d3823ac8d008e03042498
0.19.x https://github.com/brave/browser-laptop/commit/9bed2cf0bad2f6d02e02d5f1c4736bc0d1ce6cee

Issue #11476, opened 2017-10-12
Tracking: Improve webpack build time/size


cezaraugusto:
* ~disallow `uglifyJS` compression feature #11549~
* ~allow parallel minification https://github.com/brave/browser-laptop/pull/12129~
* ~cache `babel-loader` #11474~
* better babel config (needs investigation)
* better code split (separate about pages from app entry)

petemill:
For the images, would it be more ideal to compress in source code since there wouldn't be any reason to have higher quality inside git, and then compress at compile-time on every build? I've often found that images used for different purposes have different compression vs quality concerns, and each use in a PR could be evaluated from a design point of view. For SVGs, when adding a new one, I always use https://jakearchibald.github.io/svgomg/

cezaraugusto:
that's nice, please jump in here or make a new issue/PR to track image optimization! 

Plus, we make use of few FA icons and likely not use new ones in the near future so it would be nice to extract only the FA SVGs we use and put them in the tool you mentioned, or just run standalone `imagemin` for all images and push the optimized versions. What do you think about that?

luixxiul:
FYI: there is another tool https://github.com/svg/svgo

petemill:
@luixxiul 👌 - the tool I linked to is just a manual GUI for SVGO. It would be interesting to try a pass on all existing SVGs with that, and then put a policy in place for future assets. We would have to test the output for quality and any over-optimization, especially if we're using CSS to style SVG elements that may have been stripped out 🖌

PR #11550, opened 2017-10-16
reduce webpack build time by ~50s

### LINKED ISSUE #11549: disallow uglifyJS compression feature
We could save a lot of time (~50s, for a dev build) by just disallowing `uglify` compression feature, given "no whitespace" and symbol mangling accounts for most of the size reduction and are already built-in for `uglify`.

cezaraugusto:
Auditors: @bsclifton, @evq, @bbondy
close #11549
addresses #11476

We use most of UglifyJS `compress` options (at least we use all which are `on` by default), which makes no real difference in build size and ends up requiring more time to build. This PR removes them all.

Test Plan:
1. Before checking out this PR, run `CHANNEL=dev npm run build-package`
2. Pay attention to webpack time's, mine is described below
3. Pay attention to build size
4. Check out this PR
5. Go to step 2 and 3 again
6. You should see some improvements (mine was ~50s)
7. Build size should not be affected in a meaningful way (for me prior size was 374.4mb and now is 375.2mb)

Tests varied a bit but ranged around 45s-55s total time (all bundles together)

**without this change**

| bundle         | main bundle | devtools bundle | webtorrent bundle |
|--------|-------------|-----------------|--------------------|
| time     | 45415ms     | 49132ms            | 14508ms                 |

**after this change (first commit)**

| bundle         | main bundle | devtools bundle | webtorrent bundle |
|--------|-------------|-----------------|--------------------|
| time     | 25491ms     | 27002ms           | 12271ms                 |

## Edit

Turns out that letting uglify run in parallel + skipping already minified libs we did another **bold** perf gain

**after this change (second commit)**

| bundle         | main bundle | devtools bundle | webtorrent bundle |
|--------|-------------|-----------------|--------------------|
| time     | 15261ms     | 9533ms           | 8480ms                 |

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -106,9 +106,7 @@ function production () {  // eslint-disable-line
   if (env !== 'test') {
     prod.plugins.push(new UglifyJsPlugin({
       uglifyOptions: {
-        compress: {
-          warnings: false
-        },
+        compress: false,
         mangle: {
           reserved: ['module', 'exports', 'require']
         }
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -105,8 +105,11 @@ function production () {  // eslint-disable-line
   prod.plugins.push(new webpack.optimize.OccurrenceOrderPlugin(true))
   if (env !== 'test') {
     prod.plugins.push(new UglifyJsPlugin({
+      parallel: true,
       uglifyOptions: {
         compress: false,
+        // skip pre-minified libs
+        exclude: [/\.min\.js$/gi],
         mangle: {
           reserved: ['module', 'exports', 'require']
         }


evq:
It makes sense that compression negligibly affects the overall build size, I thought we had it enabled for a possible speedup on load. thinking on this though, it's not like we're moving things over the network - so it's probably slowing things down there as well to decompress

cezaraugusto:
@evq sorry I don't get what you mean. Which scenario should I test to ensure your concern is addressed?

bbondy:
I'm a bit concerned that this will reduce perf for a benefit in build time.

bsclifton:
@cezaraugusto I think @evq is saying that this should increase performance for loading the JavaScript, since it will no longer need to do decompression before processing

Is there any impact to startup time? (after installing a packaged build).  The time spent building a package doesn't matter too much, since we automate that w/ Jenkins (the time savings there wouldn't win much).

bsclifton:
Closing for now- once we can address the items Bondy has mentioned, maybe let's re-open and we can re-review 😄

bsclifton:
Heads up: this PR sped up build time by ~45 seconds on my machine while keeping minification 😄 
https://github.com/brave/browser-laptop/pull/12129

cezaraugusto:
ya `parallel: true` is a great helper

Issue #8901, opened 2017-05-17
should the update password notification bar be specific to tab?


luixxiul:
**Describe the issue you encountered:** should the update password notification bar be specific to tab? Currently the bar remains displayed after switching tabs.

![clipboard01](https://cloud.githubusercontent.com/assets/3362943/26138782/8c82d096-3b06-11e7-82c0-3ae81f761310.gif)

- Brave Version (revision SHA): 0.15.304

- Steps to reproduce:
    1. Update your password on a site
    2. Switch tabs

- Extra QA steps:
    1.
    2.
    3.

- Any related issues:

diracdeltas:
It should probably be tab-specific, but it shouldn't be origin-specific. We ran into problems where the origin of the site after login was different from the origin of the login page, so the password prompt would unexpectedly disappear.

diracdeltas:
This isn't different from the previous behavior, so I don't think it needs to be in 0.15.300

PR #11625, opened 2017-10-23
make greeting notifications global

### LINKED ISSUE #11256: ledger notification should appear above the tab area

cezaraugusto:
fix #11256

Manual Test Plan:
* all ledger notifications should appear in the safe UI area
* other notifications should work the same

Test Plan:

```
npm run test -- --grep="notificationBarState test"
```

### DIFF
--- a/app/common/state/notificationBarState.js
+++ b/app/common/state/notificationBarState.js
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const assert = require('assert')
+const Immutable = require('immutable')
+const {makeImmutable, isMap} = require('./immutableUtil')
+
+const api = {
+  validateState: function (state) {
+    state = makeImmutable(state)
+    assert.ok(isMap(state), 'state must be an Immutable.Map')
+    return state
+  }
+}
+
+const notificationBarState = {
+  /**
+   * Gets an immutable list of notifications
+   * @param {Map} appState - The app state object
+   * @return {List} - immutable list of notifications
+   */
+  getNotifications: (state) => {
+    state = api.validateState(state)
+    return state.get('notifications', Immutable.List())
+  },
+
+  /**
+   * Gets an immutable list of ledger notifications
+   * @param {Map} appState - The app state object
+   * @return {List} - immutable list of ledger notifications
+   */
+  getLedgerNotifications: (state) => {
+    const notifications = notificationBarState.getNotifications(state)
+    return notifications.filter(item => item.get('from') === 'ledger')
+  }
+}
+
+module.exports = notificationBarState
--- a/app/renderer/components/main/notificationBar.js
+++ b/app/renderer/components/main/notificationBar.js
@@ -13,6 +13,7 @@ const NotificationItem = require('./notificationItem')
 // Utils
 const {getOrigin} = require('../../../../js/lib/urlutil')
 const frameStateUtil = require('../../../../js/state/frameStateUtil')
+const notificationBarState = require('../../../common/state/notificationBarState')
 
 // Styles
 const commonStyles = require('../styles/commonStyles')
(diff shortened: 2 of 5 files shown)


app/renderer/components/main/notificationBar.js
```diff
@@ -49,6 +55,32 @@ class NotificationBar extends React.Component {
   }
 }
 
+// TODO maybe this should be merged in <NotificationBar /> and defined
+// per conditional prop such as isGlobal={conditional}
+class BraveNotificationBar extends React.Component {
```
NejcZdovc: Please move this into a separate file

app/common/state/notificationBarState.js
```diff
@@ -0,0 +1,85 @@
...
+   * origin match the current frame origin
+   */
+  isNotificationSameOrigin: (state, item) => {
+    state = api.validateState(state)
+    const activeFrame = getActiveFrame(state.get('currentWindow')) || Immutable.Map()
+
+    if (!item.has('frameOrigin')) {
+      return false
+    }
+    return getOrigin(activeFrame.get('location')) === item.get('frameOrigin')
```
diracdeltas: i think you should be comparing `getOrigin(activeFrame.get('location'))` to the frameOrigin, not the location directly

app/common/state/notificationBarState.js
```diff
@@ -0,0 +1,85 @@
...
+   */
+  getActiveTabNotifications: (state) => {
+    const notifications = api.getNotifications(state)
+    return notifications.filter(item => api.isNotificationSameOrigin(state, item))
+  },
+
+  /**
+   * Checks whether or not the notification should be shown in a per-tab basis
+   * @param {Map} appState - The app state object
+   * @param {Boolean} - whether or not the notification should be shown per tab
```
diracdeltas: this should be `@return` not `@param`

app/common/state/notificationBarState.js
```diff
@@ -0,0 +1,85 @@
...
+   * @param {Map} appState - The app state object
+   * @return {List} - list of the current active tab's notification
+   */
+  getActiveTabNotifications: (state) => {
+    const notifications = api.getNotifications(state)
+    return notifications.filter(item => api.isNotificationSameOrigin(state, item))
+  },
+
+  /**
+   * Checks whether or not the notification should be shown in a per-tab basis
```
diracdeltas: i think this should say `Checks whether or not there are per-tab notifications for the active tab`

app/renderer/components/main/notificationBar.js
```diff
@@ -26,19 +27,24 @@ class NotificationBar extends React.Component {
     const notifications = state.get('notifications', Immutable.List())
 
     const props = {}
+    // TODO: Make use of getActiveTabNotifications API instead
     props.activeNotifications = notifications
       .filter((item) => {
         return item.get('frameOrigin')
           ? activeOrigin === item.get('frameOrigin')
-          : true
+          : !item.get('greeting') // greetings indicate a global notification
```
diracdeltas: this may break password saving notifications. those do *not* have a frameOrigin (see https://github.com/brave/browser-laptop/issues/8901#issuecomment-301991642), but they are also not greeting messages.

cezaraugusto: ya but if they do not have greetings the condition will evaluate to `true` and it would be visible as-is now. You can assert by running `npm run test -- --grep="shows notification for login form"`. Let me know if I missed something

cezaraugusto: ya but if no `greetings` then it will evaluate to `true` as well. Anyway, your call is correct and the check I was doing for per tab notifications were wrong and I updated. Thanks for checking

You can assert now it works by running npm run test -- --grep="shows notification for login form". Let me know if I missed something

app/renderer/components/main/notificationBar.js
cezaraugusto: this is very similar to `NotificationBar` and plan is to make it as the comment and merge both components together, that's why it's living there. I just avoided making a big change in this PR to do so

app/common/state/notificationBarState.js
cezaraugusto: not sure if I get it, it's checking against frameOrigin rn

diracdeltas: nvm, i misread this. sorry

app/common/state/notificationBarState.js
```diff
@@ -0,0 +1,85 @@
...
+   * @param {Map} appState - The app state object
+   * @return {List} - list of all global notifications
+   */
+  getGlobalNotifications: (state) => {
+    const notifications = api.getNotifications(state)
+    return notifications.filter(item => item.has('greeting'))
+  },
+
+  /**
+   * Get an immutable list of global notifications
```
diracdeltas: minor - this comment is not right

app/common/state/notificationBarState.js
```diff
@@ -0,0 +1,85 @@
...
+   * @param {Map} appState - The app state object
+   * @return {List} - list of the current active tab's notification
+   */
+  getActiveTabNotifications: (state) => {
+    const notifications = api.getNotifications(state)
+    return notifications.filter(item => api.isNotificationSameOrigin(state, item))
+  },
+
+  /**
+   * Checks whether or not there are per-tab notifications for the active tab
```
diracdeltas: should this comment be updated? now it just checks if global notifications is empty

diracdeltas:
Bug: The per-tab notifications do not appear until the global notification has been dismissed. In the first screenshot, I went to a site and entered a password, then I went to a site that uses geolocation. However, because there is a global notification, neither of the tab notifications appeared. In the second screenshot, I dismissed the global notification and then the tab notifications appeared.

screen shot 2017-10-27 at 6 55 50 pm

after dismissing the global notification:
screen shot 2017-10-27 at 6 56 09 pm

cezaraugusto:
ok this is bigger than I thought. we need to create an `id` for each kind of notification so we can properly filter whether or not a notification should show as global or per tab. I'm making a list.

given RC1 is today I'm moving it to the next milestone

cezaraugusto:
setting this as WIP I'm going to need to cover all notification cases

see https://github.com/brave/browser-laptop/wiki/User-Notifications for a list

cezaraugusto:
cc @diracdeltas ready for re-review

cezaraugusto:
master 3ddb0a4
0.21.x 76b693c24
0.20.x e0f276e5d
0.19.x 3f70619cf

PR #11642, opened 2017-10-23
Fix automated download bar tests

petemill:
### Functional fixes:
 - Do not show 'save location' dialog when restarting a cancelled download

### Test fixes:
 - Fix automated download bar tests by preventing showing save location dialog for all file downloads

Auditors: @brave/automated-tests-team

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -569,9 +569,7 @@ function registerForDownloadListener (session) {
       webContents.forceClose()
     }
 
-    if (getSetting(settings.DOWNLOAD_ALWAYS_ASK)) {
-      item.setPrompt(true)
-    }
+    item.setPrompt(getSetting(settings.DOWNLOAD_ALWAYS_ASK) || false)
 
     const downloadId = item.getGuid()
     item.on('updated', function (e, st) {
--- a/test/misc-components/downloadItemTest.js
+++ b/test/misc-components/downloadItemTest.js
@@ -31,6 +31,8 @@ describe('downloadItem test', function () {
     yield setup(this.app.client)
 
     yield this.app.client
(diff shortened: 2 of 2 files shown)


app/filtering.js
```diff
@@ -569,9 +569,7 @@ function registerForDownloadListener (session) {
       webContents.forceClose()
     }
 
-    if (getSetting(settings.DOWNLOAD_ALWAYS_ASK)) {
-      item.setPrompt(true)
-    }
+    item.setPrompt(getSetting(settings.DOWNLOAD_ALWAYS_ASK) || false)
```
petemill: Normal download, via `.url()` (and I assume content-disposition), was correctly not showing the prompt. But when choosing 're-download' it calls `webContents.downloadUrl` which I assume has a default to show the prompt, and it was not explicitly being set here to `false`.

NejcZdovc:
master e6225e0a6
0.21 c37a22cdd5a9103243ccfa11cdae417a52421d46
0.20 68cf29dac2b04d7f4259fd0e195ebf84a7a9067b
0.19 9c33f43d7dbfdde460ce89344a11eefe8bda2546

test/misc-components/downloadItemTest.js
```diff
@@ -31,6 +31,8 @@ describe('downloadItem test', function () {
     yield setup(this.app.client)
 
     yield this.app.client
+      .changeSetting('general.download-always-ask', false)
```
NejcZdovc: Only later I notice that this can be variable, so I change it for 0.21 and 0.22

https://github.com/brave/browser-laptop/blob/e6225e0a6acf67eaf40dc362f989d17010cc4cc2/test/misc-components/downloadItemTest.js#L88-L89

PR #11655, opened 2017-10-24
Amazon SQS and SNS support

SergeyZhukovsky:
The fix for the sync library. Added support of SQS, SNS notifications.

### DIFF
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -227,7 +227,7 @@ let generateSyncManifest = () => {
     'default-src': '\'self\'',
     'form-action': '\'none\''
   }
-  const connectSources = ['\'self\'', appConfig.sync.serverUrl, appConfig.sync.s3Url]
+  const connectSources = ['\'self\'', appConfig.sync.serverUrl, appConfig.sync.s3Url, appConfig.sync.snsUrl, appConfig.sync.sqsUrl]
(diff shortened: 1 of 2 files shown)


luixxiul:
Is this no-qa-needed?

ayumi:
@luixxiul it's currently not manually verifiable, as the required code is in the https://github.com/brave/sync staging branch. after we do a new sync release with SNS+SQS support, we should QA sync within browser-laptop.

bsclifton:
@SergeyZhukovsky @ayumi does there need to be an update to the sync library before this can be merged?  What milestone do you all think it should be in?

SergeyZhukovsky:
@bsclifton we need that to be able to test sync functionality. I would tell we need it as soon as possible. Everything is merged on sync library side.

bsclifton:
@ayumi is this safe to merge before sync is released? If so, maybe we can merge it and tentatively put it (and the issue, if there is one) in the 0.22.x milestone 😄

ayumi:
@bsclifton yes, this is safe to merge

kjozwiak:
Removing the `QA/test-plan-required` tag as per https://github.com/brave/browser-laptop/pull/11655#issuecomment-340216208.

PR #11720, opened 2017-10-30
Tabs transitions

petemill:
*This is a work in progress, but very much ready for both design and engineering feedback*

# Main features

### Dragging tabs to sort
- Dragging the tab moves the tab to follow the mouse, fixed to the axis of the tab bar
- As the edge of a tab passes the middle of another tab, it swaps positions with that tab

### Smooth transitions for tabs
- Opening a tab
   - Animate Y from 50% to 100%
   - Duration 710ms
- Closing a tab
    - Animate Y from 100% to 0
    - Duration 710ms
- Re-ordering a tab
  - Slight staggered delay of 30ms (if moving multiple positions, from a single continuous drag)
  - Animation duration of 710ms
  - Ease Out

### Detaching a tab from a window
- Threshold of 44px on Y axis when the tab hasn't yet been dragged past another tab horizontally
- Threshold increases to 90px on Y axis when the tab has been dragged past another tab horizontally
- Threshold *also* always requires mouse has been outside that threshold for 300ms (meaning the mouse can drag very far away and back within 300ms and it wil not trigger a detach)
- Detaches whilst drag is still happening
- New window follows mouse until mouse is let go, ESC is pressed, or tab is dragged to another window's tab bar for attachment

### Attaching a tab to a window
- Dragging a tab in a window with only one tab, moves the whole window
- When it approaches the tab bar of another window, it is attached to that window

# Issues

### Done
- [x] Tabs don't animate 'appear' when switching between pages
- [x] Tab content which has own animation (favicon, title, audio icon, etc) don't restart animations when tab is moved
- [x] Moving animation does not restart when tab is dragged back and forth, before previous animation completes
- [x] Ensure 60fps, at least in original window where drag originates

#### Major
- If the user **starts** tab dragging from a single-tab window and then attaches the tab to another window, the dragging will stop and the user will need to start dragging (with a new click) the tab in its new location again in order to re-sort the tab in the new window. This is because the event is controlled from the originating window, which gets destroyed at that point. So this isn't an issue if the drag originates in a window with multiple tabs - you can attach, sort, detach from as many windows as you can in one single movement if that's the case. This can be fixed by having the window hanging around until the drag is ended.
- Tab pages. The idea is that the user should be able to drag to an arrow and the tab moves to the previous / next page. This might feel jarring perhaps, because if you drag to the 'previous page' button and the tab moves to the previous page - making the currently displayed page change - the mouse will still be over the 'previous page' button, which would maybe cause an expectation that the page would keep moving to the previous one until it ends up at the first one... It could just be that mousing over the previous / next button only moves one page until a mouse-out and mouse-in occurs again. We need to solve this otherwise there'll be no way to change tab pages for a tab unless using a keyboard shortcut.

#### Medium
- Speed of window creation. Could create a buffer window when any drag operation starts, so it's rendered hopefully by the time a tab is detached, or just wait for the window to be fully rendered before showing, avoiding flash-of-white (separate issue?)
- Some transparency on the top bar of a detached window might be a good visual aid to know the drag is still happening.
- Linux/Ubuntu - not getting mousemove outside window unlike macOS and Windows, so detach not working so well

#### Minor
- Fix tab borders
- Tab does not retain its size from the source window when attached to another window. The main use-case here is that a user may drag a wide tab into a window with many tabs, and then the tab becomes narrower. It may feel as if you are dragging the space next to the tab. Functionality still works though. It would be possible to make the tab retain its width, but a few css and physics calculations would need to be re-worked. I left this for a future issue so as to not risk breaking tab styles.
- Disable navigation bar mouseout / mouseenter handlers so that url bar (and others) does not flash between states
- Cancellation. When you press escape, the tab should go back to it's original position / window. The plumbing is set up for this, so should be relatively low effort
- Consider not allowing portion of windows to be dragged outside entire screen bounds during a tab-drag, since some platforms don't allow it anyway
- Tab attachment works even to a window that is completely hidden behind other windows. I have made that window go to the forefront and become focused, but I haven't yet found a reliable way to detect which windows are visible and not obscured.
- Pinned tabs - offset the position of the non-pinned tab dragging when there are pinned tabs present.
- You can no longer drag a tab from non-pinned to pinned status or the opposite. This is similar to how dragging in other browsers work. Plus it doesn't work if there are not already pinned tabs already. And since pinned tabs are multi-window, it could break the connection between moving a physical single-tab and what ends up happening. If this is important, I'll possibly have to re-work quite a bit.
- Remove `process.stdout.write` debug logging

#### Arthitecture
- Create a HOC for Tab to separate the dragging activation onDragStart handler, the event handlers that are then attached, and the comms with the other potential windows. Or do all that in the windowStore, since it has access to the window object.
- Also consider a wrapper or HOC for responding to the events and props that are part of responding to being dragged (movement, X and Y detection for changes).
- Helper module for the common window / pointer calculations

# Engineering notes

- Uses the FLIP technique for checking to see where a component's DOM node has been rendered, comparing it to where it was previously rendered (when props last changed), setting the node's position back to the last position and animating it to the new position. https://aerotwist.com/blog/flip-your-animations/

- The component `ListWithTransitions` is used to perform these animations on component enter / leave / move. It has nothing to do with dragging, and everything to do with responding to re-ordered tabs or new tabs or closing tabs, and animating those states. I've tried to keep it agnostic as to what is being rendered, so it can be used for other things in the future (e.g. bookmarks).

- The existing `Tab` component can do two things:
  1. Initiate a tab dragging via the normal DOM onDragStart event.
     - Dispatches an event informing which tab is being dragged
     - Sends messages to other windows when the mouse moves (see the code comments for why)
  2. Respond to the state instructing a specific tab to be 'dragged'
     - Listens for mousemove in order to move the tab with the mouse, dispatch actions to re-order the tabs, or detach the tab from the window
 
  Both these things can actually happen in separate components and windows during a drag operation, since a drag may start inside a component in one window, then move the tab to another window, where a different component and window will need to know to be `props.isDragging = true`, and listen for the mouse position.

# QA

TODO


app/renderer/components/tabs/tab.js
```diff
@@ -333,20 +538,17 @@ class Tab extends React.Component {
...
           this.props.showAudioTopBorder && styles.tabArea__tab_audioTopBorder,
-
           // Windows specific style (color)
           isWindows && styles.tabArea__tab_forWindows,
-
           // Set background-color and color to active tab and private tab
           this.props.isActive && styles.tabArea__tab_active,
           this.props.isPrivateTab && styles.tabArea__tab_private,
           (this.props.isPrivateTab && this.props.isActive) && styles.tabArea__tab_private_active,
-
```
luixxiul: nit: let's keep the blank lines above to make those comments meaningful :-)

petemill: Whilst it highlights this specific area, it may harm the ability to scan of the rest of the render structure because typically blank lines should be used for component separation, and not usually in the middle of a component definition. I would suggest that the comments themselves act as padding between the lines of code, to highlight their importance or grouping. But this was your code so I’ll add the spacing back before I’m done. Thanks

luixxiul: The reason why I thought the blanks lines are important is not only related with appearance but also   with style cascading; the sorting order in css() is so critical that reordering them without being sure how they are ordered would easily cause visual regressions (some of them would be critical) if the conditions are specified vaguely, while reordering components would basically not cause that kind of issue.

Ideally blanks lines inside css() would be removed, if the naming in BEM is properly done and makes it quite sure how the lines is aligned and why they are done so.

How does that sound? 😄

bsclifton:
Updated PR and all issues to be in 0.22.x milestone

bsclifton:
Closing - @petemill evaluated if we could pull this in and unfortunately it would take 4-6 days to complete (and we'd get this for free with Brave Core)

Issue #11399, opened 2017-10-09
Browser crashes with "Cannot read property 'get' of undefined"


rafid059:
### Description

I upgraded brave to the latest release on Fedora 24 as described here: https://github.com/brave/browser-laptop/blob/master/docs/linuxInstall.md#fedora-x86_64

After the update:

* there was no 'new tab' already opened after starting the browser (img3)
* The 'plus' sign doesn't open any new tabs. (img3)
* The address bar is invisible too.
* The help dropdown opens but that's it. Selecting any of the options like 'settings' or 'downloads' doesn't open any tabs. (img2)
* Tried the keyboard shortcuts to open new tabs or the downloads tab, but that doesn't work either.
* Bookmarks toolbar is there, and the bookmark folder dropdown works too BUT selecting a bookmarked page doesn't open any tab. (img1)

I uninstalled and reinstalled the latest release (which is 0.18.36-1) from the repo and it's the same. Also, `brave --help` cmd crashes and opens the browser instead with all of the issues mentioned above. Here's the following error that the cmd spits out:

```
$ brave --help
Error: unrecognized flag --icu_case_mapping
Try --help for options
Crash reporting enabled
Could not access master password; aborting
An uncaught exception occurred in the main process Uncaught Exception:
TypeError: Cannot read property 'get' of undefined
    at migrate (/usr/share/brave/resources/app.asar/app/browser/reducers/spellCheckerReducer.js:13:13)
    at spellCheckerReducer (/usr/share/brave/resources/app.asar/app/browser/reducers/spellCheckerReducer.js:35:15)
    at reducers.reduce (/usr/share/brave/resources/app.asar/js/stores/appStore.js:382:24)
    at Array.reduce (<anonymous>)
    at applyReducers (/usr/share/brave/resources/app.asar/js/stores/appStore.js:380:68)
    at handleAppAction (/usr/share/brave/resources/app.asar/js/stores/appStore.js:424:14)
    at callbacks.forEach (/usr/share/brave/resources/app.asar/js/dispatcher/appDispatcher.js:97:7)
    at Array.forEach (<anonymous>)
    at AppDispatcher.dispatchToOwnRegisteredCallbacks (/usr/share/brave/resources/app.asar/js/dispatcher/appDispatcher.js:96:20)
    at AppDispatcher.dispatchInternal (/usr/share/brave/resources/app.asar/js/dispatcher/appDispatcher.js:122:10)
Waiting 60 seconds for process to load
[...]

PR #11723, opened 2017-10-30
Ensure state is truthy before acting on it (preventing crash)

bsclifton:
Fixes https://github.com/brave/browser-laptop/issues/11399

Auditors: @NejcZdovc, @bbondy

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests


bbondy:
This should never be null, it makes me think there is some other error happening earlier which fails here. But if that root thing is not fixed it'll just fail somewhere else after this error.

bsclifton:
@bbondy that's a good point- that call is only made once... so likely there's a problem loading the session store

PR #11727, opened 2017-10-31
Adds media publishers support

### LINKED ISSUE #11851: Adds media publishers support

NejcZdovc:
Fix #11851 

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests


app/browser/api/ledger.js
```diff
@@ -44,6 +44,7 @@ const request = require('../../../js/lib/request')
 const ledgerUtil = require('../../common/lib/ledgerUtil')
 const tabState = require('../../common/state/tabState')
 const pageDataUtil = require('../../common/lib/pageDataUtil')
+const ledgerVideoCahce = require('../../common/cache/ledgerVideoCache')
```
bsclifton: s/ledgerVideoCahce/ledgerVideoCache

app/browser/api/ledger.js
```diff
@@ -559,9 +561,14 @@ const getPublisherData = (result, scorekeeper) => {
     pinPercentage: result.pinPercentage,
     weight: result.pinPercentage
   }
-  // HACK: Protocol is sometimes blank here, so default to http:// so we can
-  // still generate publisherURL.
-  data.publisherURL = (result.protocol || 'http:') + '//' + result.publisherKey
+
+  data.publisherURL = result.publisherURL || (result.protocol || 'http:') + '//' + result.publisherKey
```
bsclifton: would be nice to use parens to make this clear-  I think you're meaning it to be:
`result.publisherURL || ((result.protocol || 'http:') + '//' + result.publisherKey)`

app/browser/api/ledger.js

app/browser/api/ledger.js

bsclifton:
while testing, I got this crash:
```
AssertionError: state must contain ledgerVideos as Immutable.Map
    at validateState (/Users/clifton/Documents/browser-laptop/app/common/cache/ledgerVideoCache.js:12:10)
    at Object.getDataByVideoId (/Users/clifton/Documents/browser-laptop/app/common/cache/ledgerVideoCache.js:17:11)
    at Object.onMediaRequest (/Users/clifton/Documents/browser-laptop/app/browser/api/ledger.js:2495:34)
    at ledgerReducer (/Users/clifton/Documents/browser-laptop/app/browser/reducers/ledgerReducer.js:410:27)
    at reducers.reduce (/Users/clifton/Documents/browser-laptop/js/stores/appStore.js:169:24)
    at Array.reduce (<anonymous>)
    at applyReducers (/Users/clifton/Documents/browser-laptop/js/stores/appStore.js:167:68)
    at handleAppAction (/Users/clifton/Documents/browser-laptop/js/stores/appStore.js:231:16)
    at callbacks.forEach (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:107:7)
    at Array.forEach (<anonymous>)
    at AppDispatcher.dispatchToOwnRegisteredCallbacks (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:106:20)
    at AppDispatcher.dispatchInternal (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:132:10)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
    at processImmediate [as _immediateCallback] (timers.js:617:5)
```

NejcZdovc:
@bsclifton this is new thing that I added for tests, apparently I broke something :smile:

bsclifton:
Interestingly enough, when that error happens (and likely many other events we're processing), the video stutters

NejcZdovc:
@bsclifton ok tests are fixed and functionality should be working again. Please re-test

bsclifton:
Happy path is working great! This is awesome
screen shot 2017-11-06 at 12 41 10 pm

bsclifton:
OK noticed a few bad ones here- it seems that:
- pinning channels doesn't work
- enabling / disabling channels affects all channels at once
- deleting channels (trash can icon) deletes all the channels (and youtube.com)

weird-behavior
weird-behavior2

app/common/constants/ledgerMediaProviders.js
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const mapValuesByKeys = require('../../../js/lib/functional').mapValuesByKeys
```
bsclifton: I believe there is a performance impact by using this... Going forward, I think it might be better to just assign a value directly.  For example:
```js
const providers = {
  YOUTUBE: 'youtube.com',
  TWITCH: 'twitch.tv'
  //... etc ...
}
```

bsclifton: cc: @bbondy

app/common/lib/ledgerUtil.js
```diff
@@ -183,16 +186,136 @@ const stickyP = (state, publisherKey) => {
...
+    time += parseFloat(endTime[i]) - parseFloat(startTime[i])
+  }
+
+  // we get seconds back, so we need to convert it into ms
+  time = time * 1000
+
+  return parseInt(time)
+}
+
+const isMediaProvider = (url) => {
```
bsclifton: Would this be better named as `getMediaProvider`?

app/common/lib/ledgerUtil.js
```diff
@@ -183,16 +186,136 @@ const stickyP = (state, publisherKey) => {
...
+    case ledgerMediaProviders.YOUTUBE: {
+      duration = getYoutubeDuration(data)
+      break
+    }
+  }
+
+  return duration
+}
+
+const getYoutubeDuration = (data) => {
```
bsclifton: nit: to respect the brand name, it should be YouTube (ex: `getYouTubeDuration`)

bsclifton:
If I play a few videos at once, the time for each is recorded (even if it's not the active tab).  I think this is OK, however the ones which are not active have their view count incremented each time the heartbeat comes in

app/common/constants/ledgerMediaProviders.js
NejcZdovc: I used what we have now for other constants, but I can change it

app/sessionStore.js
```diff
@@ -926,7 +926,8 @@ module.exports.defaultAppState = () => {
     },
     cache: {
       bookmarkLocation: undefined,
-      bookmarkOrder: {}
+      bookmarkOrder: {},
+      ledgerVideos: {}
```
NejcZdovc: We need to add deep merge for state migration. Suggestion by @bsclifton https://github.com/KyleAMathews/deepmerge

app/common/lib/ledgerUtil.js
NejcZdovc: fixed

app/common/lib/ledgerUtil.js
NejcZdovc: fixed

app/common/constants/ledgerMediaProviders.js
NejcZdovc: replaced with string

NejcZdovc:
@bsclifton visits in background tabs fixed as well

app/sessionStore.js
NejcZdovc: will add it after PR will be confirmed

bsclifton:
I think I need to reset my profile... but testing again, I'm getting this:
```
AssertionError: state must contain ledgerVideos as Immutable.Map
    at validateState (/Users/clifton/Documents/browser-laptop/app/common/cache/ledgerVideoCache.js:12:10)
    at Object.getDataByVideoId (/Users/clifton/Documents/browser-laptop/app/common/cache/ledgerVideoCache.js:17:11)
    at Object.onMediaRequest (/Users/clifton/Documents/browser-laptop/app/browser/api/ledger.js:2493:34)
    at ledgerReducer (/Users/clifton/Documents/browser-laptop/app/browser/reducers/ledgerReducer.js:409:27)
    at reducers.reduce (/Users/clifton/Documents/browser-laptop/js/stores/appStore.js:169:24)
    at Array.reduce (<anonymous>)
    at applyReducers (/Users/clifton/Documents/browser-laptop/js/stores/appStore.js:167:68)
    at handleAppAction (/Users/clifton/Documents/browser-laptop/js/stores/appStore.js:231:16)
    at callbacks.forEach (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:107:7)
    at Array.forEach (<anonymous>)
    at AppDispatcher.dispatchToOwnRegisteredCallbacks (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:106:20)
    at AppDispatcher.dispatchInternal (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:132:10)
    at async.cargo (/Users/clifton/Documents/browser-laptop/js/dispatcher/appDispatcher.js:213:19)
    at process (/Users/clifton/Documents/browser-laptop/node_modules/async/dist/async.js:2317:17)
    at Immediate.<anonymous> (/Users/clifton/Documents/browser-laptop/node_modules/async/dist/async.js:66:16)
    at runCallback (timers.js:672:20)
```

I'm not sure if there's a graceful way to handle this... when it happens, it crashes the browser hard (you have to force quit).  We don't have a global exception handler unfortunately for when the assert raises the error.  Can you wrap w/ a try/catch?

NejcZdovc:
@bsclifton please check this comment https://github.com/brave/browser-laptop/pull/11727#discussion_r149584403. So we will fix state after this PR is approved and before merged. Because this is not a problem of this PR, but we still want to fix it as part of this PR

srirambv:
Seeing errors mentioned in https://github.com/brave/browser-laptop/pull/11727#issuecomment-342736381 on a new profile. Also seeing a lot of break in video every time that error pops up on the console. Build on 145f454

srirambv:
Unable to switch to other tabs after getting this error @bsclifton do you also see the problem on your profile?

bbondy:
Added issue #11851 to track this (release notes and QA get done via issue tracking per OS, so it's important to have an associated issue). Thanks.

NejcZdovc:
@srirambv 2db6073 should fix problem with state

diracdeltas:
This should not be merged without security review.

app/browser/api/ledger.js
```diff
@@ -569,9 +571,14 @@ const getPublisherData = (result, scorekeeper) => {
...
-  // HACK: Protocol is sometimes blank here, so default to http:// so we can
-  // still generate publisherURL.
-  data.publisherURL = (result.protocol || 'http:') + '//' + result.publisherKey
+
+  data.publisherURL = result.publisherURL || ((result.protocol || 'http:') + '//' + result.publisherKey)
+
+  // media publisher
+  if (result.faviconName) {
+    const on = locale.translation('on').toLowerCase()
+    data.siteName = `${result.faviconName} ${on} ${result.providerName}`
```
diracdeltas: [minor] this might cause the wrong word order in some languages. i think we support l10n parameters to solve this problem, but maybe not in the `locale` module

app/browser/api/ledger.js
```diff
@@ -569,9 +571,14 @@ const getPublisherData = (result, scorekeeper) => {
     pinPercentage: result.pinPercentage,
     weight: result.pinPercentage
   }
-  // HACK: Protocol is sometimes blank here, so default to http:// so we can
-  // still generate publisherURL.
-  data.publisherURL = (result.protocol || 'http:') + '//' + result.publisherKey
+
+  data.publisherURL = result.publisherURL || ((result.protocol || 'http:') + '//' + result.publisherKey)
```
diracdeltas: should probably use `https` instead of `http` by default since that is what https://github.com/brave-intl/bat-publisher/blob/master/index.js#L100 does

app/browser/api/ledger.js
```diff
@@ -853,7 +860,7 @@ const excludeP = (publisherKey, callback) => {
       return done(err, result)
     }
 
-    let props = ledgerPublisher.getPublisherProps('https://' + publisherKey)
+    let props = ledgerPublisher.getPublisherProps(publisherKey)
```
diracdeltas: where is `publisherKey` ultimately generated? just making sure it has some kind of input validation

app/browser/api/ledger.js
```diff
@@ -1440,21 +1449,12 @@ const cacheRuleSet = (state, ruleset) => {
...
 
-      ctx = underscore.mapObject(ctx, function (value) {
-        if (!underscore.isFunction(value)) return value
-      })
-      ctx.URL = location
-      ctx.SLD = tldjs.getDomain(ctx.host)
-      ctx.RLD = tldjs.getSubdomain(ctx.host)
-      ctx.QLD = ctx.RLD ? underscore.last(ctx.RLD.split('.')) : ''
+      if (publisher.publisherURL) ctx.URL = publisher.publisherURL
+      if (!ctx.URL) ctx.URL = (publisher.get('protocol') || 'http:') + '//' + publisherKey
```
diracdeltas: same as above for 'https' vs 'http'

app/browser/api/ledger.js
```diff
@@ -2457,6 +2458,117 @@ const transitionWalletToBat = () => {
...
+    return module.exports.saveVisit(state, cache.get('publisher'), duration, revisited)
+  }
+
+  const options = underscore.extend({roundtrip: roundtrip}, clientOptions)
+  const mediaProps = {
+    mediaId,
+    providerName: type
+  }
+
+  ledgerPublisher.getMedia.getPublisherFromMediaProps(mediaProps, options, (error, response) => {
```
diracdeltas: 1. bat-publisher should throw an error if the media URL protocol is not "https:"
2. it appears `bat-publisher` is making a call to a youtube.com page from the browser main process, scraping and parsing the page HTML, and extracting page data from it. from a security perspective, it would be much better to do this in a page content script that runs when the user visits a youtube page (ex: how we currently extract favicons from publisher pages).

package.json
```diff
@@ -88,12 +88,13 @@
...
     "bat-balance": "^1.0.3",
-    "bat-client": "^1.0.15",
-    "bat-publisher": "^1.0.3",
+    "bat-client": "^1.2.2",
+    "bat-publisher": "^1.2.0",
     "bignumber.js": "^4.0.4",
     "bloodhound-js": "brave/bloodhound",
     "clipboard-copy": "^1.0.0",
     "compare-versions": "^3.0.1",
+    "deepmerge": "^2.0.1",
```
diracdeltas: i don't see anything obviously wrong with deepmerge but fwiw we use lodash.merge in sync. i wonder if @ayumi knows why?

ayumi: no particular reason. lodash seemed well maintained.

app/browser/api/ledger.js
NejcZdovc: we get it from bat-publisher library https://github.com/brave/browser-laptop/pull/11727/files#diff-3cdfdb124f718787d584e9e2b82fcedcR2505, with `ledgerPublisher.getMedia.getPublisherFromMediaProps`

app/browser/api/ledger.js
NejcZdovc: cc @mrose17

app/browser/api/ledger.js

app/browser/api/ledger.js

app/browser/api/ledger.js

package.json
NejcZdovc: we can switch to loadash as well

diracdeltas: either way is fine

diracdeltas:
There's a security issue in https://github.com/brave-intl/bat-publisher/blob/master/getMedia.js#L55 
```
if (!schemes.length) return (mediaURL.indexOf(rule.domain) !== -1) ...
```

say that rule.domain is youtube.com, then `youtube.com.evil.com` would match for it

app/browser/api/ledger.js
diracdeltas: in https://github.com/brave-intl/bat-publisher/blob/d896a93465df57b3cca227b90530f0b40044d5fc/getMedia.js#L91, where does author_url come from?

NejcZdovc: In https://github.com/brave-intl/bat-publisher/blob/d896a93465df57b3cca227b90530f0b40044d5fc/getMedia.js#L77 we call publisher oembed, which is defined here https://github.com/brave-intl/bat-publisher/blob/d896a93465df57b3cca227b90530f0b40044d5fc/media/providers.json#L7 (for youtube is `https://www.youtube.com/oembed?url=https://www.youtube.com/watch?v=kLiLOkzLetE`) and we get back something like this
```
{"html":"\u003ciframe width=\"480\" height=\"270\" src=\"https:\/\/www.youtube.com\/embed\/kLiLOkzLetE?feature=oembed\" frameborder=\"0\" gesture=\"media\" allowfullscreen\u003e\u003c\/iframe\u003e","title":"This Is Brave","thumbnail_height":360,"version":"1.0","provider_name":"YouTube","thumbnail_width":480,"width":480,"author_name":"Brave","author_url":"https:\/\/www.youtube.com\/channel\/UCFNTTISby1c_H-rm5Ww5rZg","height":270,"thumbnail_url":"https:\/\/i.ytimg.com\/vi\/kLiLOkzLetE\/hqdefault.jpg","provider_url":"https:\/\/www.youtube.com\/","type":"video"}
``` which contain `author_url`

Comment (diracdeltas) [reply app/browser/api/ledger.js]: i think https://github.com/brave-intl/bat-publisher/blob/d896a93465df57b3cca227b90530f0b40044d5fc/getMedia.js#L91 needs to validate that author_url is a valid URL. this is easy because if not then `parts` will be null

Comment (diracdeltas) [reply app/browser/api/ledger.js]: https://github.com/brave-intl/bat-publisher/blob/d896a93465df57b3cca227b90530f0b40044d5fc/index.js#L101 should also return immediately if `props` is null

File: app/browser/api/ledger.js [hunk 13]
```diff
@@ -107,6 +110,38 @@ underscore.keys(fileTypes).forEach((fileType) => {
 })
 signatureMax = Math.ceil(signatureMax * 1.5)
 
+if (ipc) {
```
NejcZdovc: Accidentally removed in #11882

app/browser/api/ledger.js
mrose17: these last two are fixed in `bat-publisher@1.2.1` and `bat-client@1.4.2`

app/browser/api/ledger.js
mrose17: 1. `bat-publisher` is being passed a URL that the browser is playing. whether it is `http?` is immaterial. the goal is to look at the URL string and figure out where to look for the oembed data.

2. i think that this is a worthy future, but not essential at this point. it is certainly good to have routines that parse data run in a separate context. it's not clear to me why that is "more necessary" in this case. we can open up an issue for this, and when we do, it would be worthwhile to see what other cases might exist!

app/browser/api/ledger.js
NejcZdovc: versions updated in this PR

app/browser/api/ledger.js
evq: @mrose17 I think comment 1 is still unaddressed? I'd also strongly prefer we address 2 before moving on. The follow up issue seems to be to address the general case, but no reason for us to not adhere to the correct practices now.

bsclifton: @mrose17 I'm going to hold off on merge until @evq 's concern can be addressed (and we have a sign off in https://github.com/brave/internal/issues/141). Let us know if we need to update libs after that is addressed

mrose17: 1. is addressed. we aren't opening a URL, we are parsing it in order to find an oembed provider. if the browser decides to play it and then ask for information, then the module will provide the information.  however, in the interest of moving things along, i have pushed `bat-publisher@1.2.3` that makes this change.

2. it is not a simple thing to make this change (it has to be made in the browser). nor am i convinced that it makes sense to require this code be held to a standard higher than any other code in the browser. as i noted in my comment.

everything has a cost. i suppose we can address 1, because it's a small change, even if i do view it as unnecessary. in addressing 2, it has a large cost (i already researched it with @NejcZdovc) and addresses a theoretical concern. there is significant cost to that we will not be able to merge this upcoming week. i don't have an issue with the browser folks reviewing the issue that @diracdeltas created after this merge, and having the same standard applied to the entire browser. i most certainly have an issue having this new standard being applied in this context and in this timeline.

bsclifton:
master fafa483e8ff3b984cf7c88c0eb1fd7f912112654
0.21.x 466cc7f82f6bb9ab383f642a9bee7e495662d7a6
0.20.x 41e432723d96da4636a79bb66bddd673febabf73
0.19.x @NejcZdovc I need your help merging this one 😄

NejcZdovc:
@bsclifton merged in 0.19 2d55627b9

Issue #11582, opened 2017-10-17
reduce size of session-store-1 by removing 'data:' URLs


diracdeltas:
## Test plan

https://github.com/brave/browser-laptop/pull/11754#issue-270469950

----

My production session-store-1 file is now 23MB. `egrep -o '"data:(.+?)"'` shows that nearly 80% of this (18MB) of this is `data:` URLs, mostly favicons it seems. We should investigate if the favicons can be moved to external files or replaced with links to the favicon URL.

bridiver:
see https://github.com/brave/browser-laptop/commit/11a2bc4849eade05eec5ba603be8298684aee64e#diff-70ca3257d79251708167c6ce9bf1cd81
You may have some leftover in your session store, but there shouldn't be any new ones being created

diracdeltas:
i think most/all of the data URLs are from external sites which use data: URLs as their favicon URL

NumDeP:
@diracdeltas I made a bloody inference to this here #10975 and you completely blanked me. I've been waiting around like a nervous hooker.

bsclifton:
@NumDeP she ( @diracdeltas) wasn't involved with that issue, I was.  There are thousands of issues in our backlog, so even though I was involved I had to re-read the details

I don't see any reference to data URL/URIs in that issue, so I wasn't ignoring you

Let's keep things civil here- we want to help 😄

NumDeP:
Sorry. It's just that from all the crashes I've been hoarding all the S-S-1's locally and going through the process of going through unnecessary data.

bsclifton:
@NumDeP I understand your frustration- there are a lot of small details that we're constantly evaluating and trying to assess the priority for (if you use GitHub's `Watch` feature for this repo for a few days, you'll get a sense of how many requests we get).

If you have a suggestion for what we can remove, that is appreciated.  Please do know that harassment of any kind (including comments or language which make a contributor uncomfortable) is not welcomed

We keep a fairly up to date record of what goes into the session store here:
https://github.com/brave/browser-laptop/blob/master/docs/state.md

If you find data that you think can be removed, please open an issue.  If you're not comfortable with making a PR, we should be able to help make the change (or we could also give it some visibility by tagging it as `good-first-bug` or similar).

diracdeltas:
i have a fix for this, it seems to help for performance but would be good to test against a benchmark

diracdeltas:
added ledger label because all the data favicons i had were from the ledger

[...]

PR #11754, opened 2017-11-01
Move ledger data:* favicons to external files

diracdeltas:
This moves data:*  favicon URLs in the ledger appState to external image files during session store cleanup. Reduces the size of my session-store-1 from ~23 MB to 5 MB.

Fix https://github.com/brave/browser-laptop/issues/11582

Test plan:
1. Copy session-store-1 from a Brave profile that has had payments enabled for a while to your brave-development appData directory.
2. Start the browser and go to about:preferences#payments. The synopsis table should appear fine.
3. Close the browser.
4. In your brave-development appData directory, you should now see a ledger-favicons subdirectory which contains a lot of image files. You should also notice that session-store-1 is now much smaller.
5. Repeat step 2.

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -37,6 +37,7 @@ const {defaultSiteSettingsList} = require('../js/data/siteSettingsList')
 const filtering = require('./filtering')
 const autofill = require('./autofill')
 const {navigatableTypes} = require('../js/lib/appUrlUtil')
+const {isDataUrl, parseFaviconDataUrl} = require('../js/lib/urlUtil')
 const Channel = require('./channel')
 const BuildConfig = require('./buildConfig')
 const {isImmutable, isMap, makeImmutable, deleteImmutablePaths} = require('./common/state/immutableUtil')
@@ -55,8 +56,8 @@ const getTempStoragePath = (filename) => {
     : path.join(process.env.HOME, '.brave-test-session-store-' + filename + '-' + epochTimestamp)
 }
 
-const getStoragePath = () => {
-  return path.join(app.getPath('userData'), sessionStorageName)
+const getStoragePath = (filename = sessionStorageName) => {
+  return path.join(app.getPath('userData'), filename)
 }
 /**
  * Saves the specified immutable browser state to storage.
@@ -421,7 +422,7 @@ module.exports.cleanAppData = (immutableData, isShutdown) => {
     })
   }
 
-  // Leader cleanup
+  // Ledger cleanup
   if (immutableData.has('pageData')) {
     immutableData = immutableData.delete('pageData')
   }
@@ -442,6 +443,21 @@ module.exports.cleanAppData = (immutableData, isShutdown) => {
     })
   }
 
+  try {
+    // Prune data: favicons by moving them to external files
+    const basePath = getStoragePath('ledger-favicons')
+    if (immutableData.get('createdFaviconDirectory') !== true) {
+      const fs = require('fs')
+      if (!fs.existsSync(basePath)) {
+        fs.mkdirSync(basePath)
+      }
+      immutableData = immutableData.set('createdFaviconDirectory', true)
+    }
+    immutableData = cleanFavicons(basePath, immutableData)
+  } catch (e) {
+    console.error('cleanAppData: error cleaning up data: urls', e)
+  }
+
   return immutableData
 }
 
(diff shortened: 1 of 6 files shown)


NejcZdovc:
@diracdeltas this is really cool. Maybe one suggestion. Now we have synopsis and about synopsis, but synopsis is parent of about synopsis. So what is in about synopsis is for sure in synopsis. If I understand this code correctly we are create all favicons twice. My suggestion would be to generate favicons for synopsis and then just assign this values to about synopsis as well. wdyt?

NejcZdovc:
In near future we would like to remove this duplicate entries, so that you would only have one synopsis.

app/sessionStore.js
```diff
@@ -447,6 +459,42 @@ module.exports.cleanSessionDataOnShutdown = () => {
...
+            return
+          }
+          const faviconPath = path.join(basePath,
+            typeof index === 'number'
+              ? `favicon-${index}-${Date.now()}.${parsed.ext}`
+              : `favicon-${index}.${parsed.ext}`
+          )
+          fs.writeFile(faviconPath, parsed.data, 'base64', (err) => {
+            if (err) {
+              console.error(`Error writing file: ${faviconPath} ${err}`)
```
bridiver: are we ok with potentially losing these?

app/sessionStore.js
```diff
@@ -428,6 +429,17 @@ module.exports.cleanAppData = (immutableData, isShutdown) => {
     immutableData = immutableData.deleteIn(['ledger', 'locations'])
   }
 
+  try {
+    const basePath = getStoragePath('ledger-favicons')
+    const fs = require('fs')
+    if (!fs.existsSync(basePath)) {
```
bridiver: should use the async versions of these to avoid blocking the UI thread

app/sessionStore.js
diracdeltas: iirc the ledger tries to get the favicon whenever the user visits the page, so it would be re-fetched on the next visit

app/sessionStore.js
diracdeltas: it needs to be synchnrous because this directory must exist before `cleanFavicons` is called, and `cleanFavicons` returns immutableData synchronously. I could add a migration flag that indicates whether this directory was created, which would avoid having to do the synchronous mkdir in future runs.

app/sessionStore.js
NejcZdovc: @diracdeltas that's not the case. We only get favicon if there is not one

diracdeltas: i could add an img onerror in the synopsis table that sets the faviconURL to `'' / null` if the file does not exist on disk - would that cause the favicon to be re-fetched?

NejcZdovc: yes if you set it to null it will work. More info here: https://github.com/brave/browser-laptop/blob/master/app/browser/api/ledger.js#L1146

bsclifton:
For review status, @bridiver @NejcZdovc do you approve / not approve?  @diracdeltas was any action needed on your part after the above?

diracdeltas:
@bsclifton there's some follow-up enhancements which i'll try to do early this week

bsclifton:
Moving to 0.20.x (as the next 0.19.x release is reserved for Chromium 63)

app/sessionStore.js
```diff
@@ -447,6 +459,42 @@ module.exports.cleanSessionDataOnShutdown = () => {
...
+            immutableData = immutableData.setIn(
+              synopsisPath.concat([index, 'faviconURL']), '')
+            return
+          }
+          const faviconPath = path.join(basePath,
+            typeof index === 'number'
+              ? `favicon-${index}-${Date.now()}.${parsed.ext}`
+              : `favicon-${index}.${parsed.ext}`
+          )
+          fs.writeFile(faviconPath, parsed.data, 'base64', (err) => {
```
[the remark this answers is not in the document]
diracdeltas: fs.writeFile was easier because it supported the base64 encoding parameter

diracdeltas:
@NejcZdovc i addressed all review comments. however, i thought of something which is that the `ledger-favicon/` directory should be deleted whenever the user chooses to delete payments data. do you remember if that's doable via the 'Clear browsing data...' menu item or something else in the UI?

NejcZdovc:
we have issue up for this option https://github.com/brave/browser-laptop/issues/8537, so I think that we should finally implement it :smiley: will take a look and probably add it

diracdeltas:
ok, if there's currently no way to clear ledger data from the session store in the UI, i don't think this PR needs to address that. ready for re-review.

diracdeltas:
master: 4e7995dd468f7e6987a0ceca3938e9d0bc6fdfbe
0.21.x: ca4bc9af0ace764bc4562e8b0f2c3b79f50cfc32
0.20.x: 0b45c879ad08d9dc70bb7eb041665d0c20a2576b

petemill:
Do we have any sample large profiles with this data in, so we can assess impact and clear this change from any perf regressions we're seeing? @diracdeltas @NejcZdovc

Issue #11757, opened 2017-11-01
Unit tests fail with Error: global leak detected: @@any-promise/REGISTRATION


bsclifton:
First occurrence of the error that I know about was via travis-ci when ran against https://github.com/brave/browser-laptop/pull/11700

I'm now getting the error locally when running:
`npm run unittest -- --grep="ledger api unit tests"`

PR #11763, opened 2017-11-02
Set transition back to false when not needed to avoid being stuck in limbo

bsclifton:
This commit also fixes a leak that the tests had (which was due to tests running and including bat-publisher).

Fixes https://github.com/brave/browser-laptop/issues/11566
Fixes https://github.com/brave/browser-laptop/issues/11757

Auditors: @NejcZdovc, @bbondy

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/api/ledger.js
+++ b/app/browser/api/ledger.js
@@ -2367,8 +2367,8 @@ const yoDawg = (stateState) => {
   return stateState
 }
 
-const checkBtcBatMigrated = (state, status) => {
-  if (!status) {
+const checkBtcBatMigrated = (state, paymentsEnabled) => {
+  if (!paymentsEnabled) {
     return state
   }
 
@@ -2378,6 +2378,8 @@ const checkBtcBatMigrated = (state, status) => {
   if (!isNewInstall && !hasUpgradedWallet) {
     state = migrationState.setTransitionStatus(state, true)
     module.exports.transitionWalletToBat()
+  } else {
+    state = migrationState.setTransitionStatus(state, false)
   }
 
   return state
@@ -2513,6 +2515,7 @@ const getMethods = () => {
     privateMethods = {
       enable,
       addVisit,
+      checkBtcBatMigrated,
       clearVisitsByPublisher: function () {
         visitsByPublisher = {}
       },
--- a/test/unit/app/browser/api/ledgerTest.js
+++ b/test/unit/app/browser/api/ledgerTest.js
@@ -9,6 +9,8 @@ const sinon = require('sinon')
 const mockery = require('mockery')
 const settings = require('../../../../../js/constants/settings')
 const appActions = require('../../../../../js/actions/appActions')
+const migrationState = require('../../../../../app/common/state/migrationState')
+const batPublisher = require('bat-publisher')
 
 const defaultAppState = Immutable.fromJS({
   ledger: {},
@@ -135,6 +137,16 @@ describe('ledger api unit tests', function () {
     ledgerClient.returns(lc)
     mockery.registerMock('bat-client', ledgerClient)
 
+    // ledger publisher stubbing
+    const lp = {
+      ruleset: [],
+      getPublisherProps: function (publisher) {
+        return null
+      },
+      Synopsis: batPublisher.Synopsis
+    }
+    mockery.registerMock('bat-publisher', lp)
+
     // once everything is stubbed, load the ledger
     ledgerApi = require('../../../../../app/browser/api/ledger')
   })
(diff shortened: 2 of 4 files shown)


app/browser/api/ledger.js
```diff
@@ -2362,8 +2362,8 @@ const yoDawg = (stateState) => {
   return stateState
 }
 
-const checkBtcBatMigrated = (state, status) => {
-  if (!status) {
+const checkBtcBatMigrated = (state, paymentsEnabled) => {
+  if (!paymentsEnabled) {
```
bsclifton: renamed the variable to be more meaningful (and also match other places in the file)

app/browser/api/ledger.js
```diff
@@ -2373,6 +2373,8 @@ const checkBtcBatMigrated = (state, status) => {
   if (!isNewInstall && !hasUpgradedWallet) {
     state = migrationState.setTransitionStatus(state, true)
     module.exports.transitionWalletToBat()
+  } else {
+    state = migrationState.setTransitionStatus(state, false)
```
bsclifton: This is the big bug fix- if it's an existing install and hasn't upgraded yet, it needs to start (or in the case where you quit, restart) the transition process.  Otherwise, it should hide the transition because that is no longer valid

app/renderer/components/preferences/payment/enabledContent.js
```diff
@@ -170,7 +171,7 @@ class EnabledContent extends ImmutableComponent {
   render () {
     const ledgerData = this.props.ledgerData
     const walletStatusText = walletStatus(ledgerData)
-    const inTransition = ledgerData.getIn(['migration', 'btc2BatTransitionPending']) === true
+    const inTransition = migrationState.inTransition(ledgerData)
```
bsclifton: simple cleanup to put in helper method

bsclifton: HOWEVER... notice that it says `migration` instead of `migrations`.  This was a bug I caught with the unit tests. I'm not sure how this ever returned true?

test/unit/app/browser/api/ledgerTest.js
```diff
@@ -135,6 +139,16 @@ describe('ledger api unit tests', function () {
...
 
+    // ledger publisher stubbing
+    const lp = {
+      ruleset: [],
+      getPublisherProps: function (publisher) {
+        return null
+      },
+      Synopsis: batPublisher.Synopsis
+    }
+    mockery.registerMock('bat-publisher', lp)
```
bsclifton: This is basically what fixes the tests. Without `bat-publisher` being mocked, I was getting the leak error on my machine

bsclifton:
Marking as work-in-progress... the tests are having another issue now

bsclifton:
removing WIP label- I confirmed this does fix the unit tests 😄 👍 
https://travis-ci.org/brave/browser-laptop/jobs/296108034

app/browser/api/ledger.js
NejcZdovc: Just tested out PR and transition is not triggered for me when I upgrade from 0.18 wallet.

app/renderer/components/preferences/payment/enabledContent.js
NejcZdovc: This is failing for me and only see disabled screen on payments when upgrading from 0.18.x. This is the error:

```
(BLESSED_EXTENSION context for mnojpmjdmbbfmejpflffifhffcmidifd) http://localhost:8080/gen/aboutPages.entry.js:17915: Uncaught AssertionError: state must contain an Immutable.Map of migrations{AssertionError: state must contain an Immutable.Map of migrations
    at validateState (http://localhost:8080/gen/aboutPages.entry.js:173611:10)
    at Object.inTransition (http://localhost:8080/gen/aboutPages.entry.js:173626:13)
    at EnabledContent.render (http://localhost:8080/gen/aboutPages.entry.js:102839:41)
    at http://localhost:8080/gen/aboutPages.entry.js:78075:21
    at measureLifeCyclePerf (http://localhost:8080/gen/aboutPages.entry.js:77355:12)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (http://localhost:8080/gen/aboutPages.entry.js:78074:25)
    at ReactCompositeComponentWrapper._renderValidatedComponent (http://localhost:8080/gen/aboutPages.entry.js:78101:32)
    at ReactCompositeComponentWrapper.performInitialMount (http://localhost:8080/gen/aboutPages.entry.js:77641:30)
    at ReactCompositeComponentWrapper.mountComponent (http://localhost:8080/gen/aboutPages.entry.js:77537:21)
    at Object.mountComponent (http://localhost:8080/gen/aboutPages.entry.js:25250:35)}
```

NejcZdovc: This is working because this is not `state` but `ledgerData` which is generated here https://github.com/brave/browser-laptop/blob/f52a10989bb9c632c2fbbfecb2ec538fa28cfc7c/app/browser/tabs.js#L243. We can't use state helper here

bsclifton: OK I can fix this- we are OK to use the state helper because the path is identical, but it might be sloppy to do so (in case things change). The problem in our case is the validation that is done, which throws the error.

I'll change it back to the hardcoded path since the state helper shouldn't be used to evaluate ledgerData (even if the path matches)

bsclifton: wait- nevermind, I see the problem... fixing

bsclifton:
@NejcZdovc fixed, ready for re-review

NejcZdovc:
master 88974f25391199a9f3a2ca98524d021b62a565ce
0.21 e81dfcefe
0.20 5af92b45b
0.19 54fecec18

PR #11764, opened 2017-11-02
Hide windows until they are either rendered or a timeout expires

### LINKED ISSUE #8121: Add temporary notice to sync prefs page regarding mobile sync, + Beta label
## Test plan

https://github.com/brave/browser-laptop/pull/8131#issue-220415095

----

image

string text:

Note: Sync for iOS and Android is still being completed and will be released soon.

(Beta label can be copied from Brave Payments label.)

### LINKED ISSUE #8128: New windows are just white for a bit.
## Test plan

https://github.com/brave/browser-laptop/pull/11764#issue-270588807

----

yes
Opening new windows show chromeless completely white before the window frame etc are loaded.
This gives me a feeling of beeing not solid software and reminds me of 'badly packaged electron webapps'. I thought this was just due to me using webpack watch mode but i was shocked to see this is the behaviour in the release version too.

mac os 

Brave	0.14.2
rev	10539c2 and before

- Steps to reproduce:
    1. start brave
    2. open new window
    3. stare at a white, completely chromeless window for fractions of a second
    
- Actual result:
 i see a white square
- Expected result:
i see a browser window

yes
yes
yes

- Screenshot if needed:
bildschirmfoto 2017-04-08 um 14 18 04

#3025

petemill:
Fix #8128

Waits for React to do its initial render of components, before firing an event to the browser process, which will show the window.

If the event is not received within 5 seconds (hard-coded, but in _config.js_), the window is shown (and will probably be a white screen).

Also:
- Separates parsing of APP_NEW_WINDOW action properties, and window creation / visibility logic.
- Moves window creation logic to windows API.
- Creates tests for existing APP_NEW_WINDOW action handler logic in windowsReducer
- Creates tests for new windows.createWindow logic
- Fixes test for downloadsReducer with not restoring a stub required by windowsReducer test

One complication is with macOS fullscreen windows for a couple reasons. 
1. macOS (or muon) will ignore the `show` property and open the window immediately if the window is fullscreen. This code counteracts that by removing the fullscreen property and then setting it once the window is ready (or the timeout happens)
2. When starting the app with many windows in the state to be opened, some of which are fullscreen and some of which are non-fullscreen on the same displays as a fullscreen window(s), the non fullscreen windows will open in the fullscreen space, which can be annoying for a user. I've gone between just having those windows show immediately and some more complex logic than what I have right now which attempted to open all non-fullscreen windows before opening fullscreen windows. The complex logic did not work because at that point, macOS would not put more than 1 window fullscreen per display even if the state was instructing it to. master currently supports that, and I didn't want to break it so right now is a balance (using a setTimeout of 100ms for fullscreen windows) between being able to open many fullscreen windows on app startup and preventing combined window spaces in most circumstances.

Thoughts for the future:
- Set a default (non-react) html content which displays that there has been an error, and provides tips for how to proceed.
- Remove the white background and put a gray one in (BrowserWindow has a background-color property)
- If fading in the windows is desired on some platforms where it is normal, we may need to [bring back muon's transparency property in BrowserWindow](https://github.com/brave/muon/issues/373), or use window handles.

### Automated
`npm run unittest -- --grep "window API unit tests"`
`npm run unittest -- --grep "windowsReducer"`

### Manual
- Make sure windows open at app startup
 - Make sure windows loaded from state at app startup remember their locations and fullscreen status
- macOS / linux: when in a fullscreen window, ensure new windows opened are also fullscreen
- Try to have a window that errors (put `throw new Error()` in `entry.js`, it should open after 5 seconds
- Ideally if you have a profile that takes more than 5 seconds to open a window, you could observe that the window opens in a not-so-ready state after 5 seconds, but I have not been able to produce such a state.

Submitter Checklist:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/reducers/windowsReducer.js
+++ b/app/browser/reducers/windowsReducer.js
@@ -23,14 +23,10 @@ const electron = require('electron')
 const BrowserWindow = electron.BrowserWindow
 const firstDefinedValue = require('../../../js/lib/functional').firstDefinedValue
 const appConfig = require('../../../js/constants/appConfig')
-const messages = require('../../../js/constants/messages')
-const appUrlUtil = require('../../../js/lib/appUrlUtil')
 const settings = require('../../../js/constants/settings')
 const getSetting = require('../../../js/settings').getSetting
-const {zoomLevel} = require('../../common/constants/toolbarUserInterfaceScale')
+
 const platformUtil = require('../../common/lib/platformUtil')
-const {initWindowCacheState} = require('../../sessionStoreShutdown')
-const appDispatcher = require('../../../js/dispatcher/appDispatcher')
 
 const isDarwin = platformUtil.isDarwin()
 const isWindows = platformUtil.isWindows()
@@ -43,7 +39,32 @@ function isModal (browserOpts) {
 
 const navbarHeight = () => {
   // TODO there has to be a better way to get this or at least add a test
+  // TODO try creating a window and measuring the difference between window and content area
+  // and updating this number with that value once the first window is created
   return 75
+}
+
+function clearFramesFromWindowState (windowState) {
+  return windowState
+    .set('frames', Immutable.List())
+    .set('tabs', Immutable.List())
+}
+
+/**
+ * Determine the frame(s) to be loaded in a new window
+ * based on user preferences
+ */
+function getFramesForNewWindow () {
+  const startupSetting = getSetting(settings.STARTUP_MODE)
+  const homepageSetting = getSetting(settings.HOMEPAGE)
+  if (startupSetting === 'homePage' && homepageSetting) {
+    return homepageSetting
+      .split('|')
+      .map((homepage) => ({
+        location: homepage
+      }))
+  }
+  return [ { } ]
 }
 
 /**
@@ -51,9 +72,10 @@ const navbarHeight = () => {
  */
 const setWindowDimensions = (browserOpts, defaults, immutableWindowState) => {
   assert(isImmutable(immutableWindowState))
-  if (immutableWindowState.getIn(['windowInfo'])) {
-    browserOpts.width = firstDefinedValue(browserOpts.width, immutableWindowState.getIn(['windowInfo', 'width']))
-    browserOpts.height = firstDefinedValue(browserOpts.height, immutableWindowState.getIn(['windowInfo', 'height']))
+  const windowInfoState = immutableWindowState.get('windowInfo')
+  if (windowInfoState) {
+    browserOpts.width = firstDefinedValue(browserOpts.width, windowInfoState.get('width'))
+    browserOpts.height = firstDefinedValue(browserOpts.height, windowInfoState.get('windowInfo'))
   } else {
     browserOpts.width = firstDefinedValue(browserOpts.width, browserOpts.innerWidth, defaults.width)
     // height and innerHeight are the frame webview size
@@ -108,7 +130,6 @@ const setMaximized = (state, browserOpts, immutableWindowState) => {
 
 function windowDefaults (state) {
   return {
-    show: false,
     width: state.getIn(['defaultWindowParams', 'width']) || state.get('defaultWindowWidth'),
     height: state.getIn(['defaultWindowParams', 'height']) || state.get('defaultWindowHeight'),
     x: state.getIn(['defaultWindowParams', 'x']) || undefined,
@@ -146,10 +167,10 @@ function setDefaultWindowSize (state) {
   return state
 }
 
-const createWindow = (state, action) => {
+const handleCreateWindowAction = (state, action) => {
   const frameOpts = (action.get('frameOpts') || Immutable.Map()).toJS()
   let browserOpts = (action.get('browserOpts') || Immutable.Map()).toJS()
-  const immutableWindowState = action.get('restoredState') || Immutable.Map()
+  let immutableWindowState = action.get('restoredState') || Immutable.Map()
   state = setDefaultWindowSize(state)
   const defaults = windowDefaults(state)
   const isMaximized = setMaximized(state, browserOpts, immutableWindowState)
@@ -160,21 +181,30 @@ const createWindow = (state, action) => {
   delete browserOpts.left
   delete browserOpts.top
 
+  // decide which bounds to restrict new window to
   const screen = electron.screen
+  // use primary display by default
   let primaryDisplay = screen.getPrimaryDisplay()
-  const parentWindowKey = browserOpts.parentWindowKey
-  if (browserOpts.x != null && browserOpts.y != null) {
-    const matchingDisplay = screen.getDisplayMatching(browserOpts)
+  // can override with provided x, y coords
+  if (browserOpts.x != null && browserOpts.y != null && browserOpts.width != null && browserOpts.height != null) {
+    const matchingDisplay = screen.getDisplayMatching({
+      x: browserOpts.x,
+      y: browserOpts.y,
+      width: browserOpts.width,
+      height: browserOpts.height
+    })
     if (matchingDisplay != null) {
       primaryDisplay = matchingDisplay
     }
   }
-
+  // always override with parent window if present
+  const parentWindowKey = browserOpts.parentWindowKey
   const parentWindow = parentWindowKey
     ? BrowserWindow.fromId(parentWindowKey)
     : BrowserWindow.getFocusedWindow()
   const bounds = parentWindow ? parentWindow.getBounds() : primaryDisplay.bounds
 
+  // decide which screen to position on
   // position on screen should be relative to focused window
   // or the primary display if there is no focused window
   const display = screen.getDisplayNearestPoint(bounds)
@@ -229,7 +259,8 @@ const createWindow = (state, action) => {
     autoHideMenuBar: autoHideMenuBarSetting,
     title: appConfig.name,
     webPreferences: defaults.webPreferences,
-    frame: !isWindows
+    frame: !isWindows,
+    disposition: frameOpts.disposition
   }
 
   if (process.platform === 'linux') {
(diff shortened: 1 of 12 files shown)


NejcZdovc:
@petemill did you maybe define wrong fix issue number?

petemill:
Yeah @NejcZdovc thanks, edited. Will edit commit message before merge.

app/browser/reducers/windowsReducer.js
```diff
@@ -35,6 +35,8 @@ const appDispatcher = require('../../../js/dispatcher/appDispatcher')
 const isDarwin = platformUtil.isDarwin()
 const isWindows = platformUtil.isWindows()
 
+const TIMEOUT_WINDOW_SHOW_MS = 5000
```
bsclifton: I think this is fine here, but for future reference you may consider `js/constants/config.js` 😄

app/browser/reducers/windowsReducer.js
```diff
@@ -245,7 +246,52 @@ const createWindow = (state, action) => {
...
 
   setImmediate(() => {
-    const win = new BrowserWindow(Object.assign(windowProps, browserOpts, {disposition: frameOpts.disposition}))
+    // normally macOS will open immediately-created windows from fullscreen
+    // parent windows as fullscreen
+    // but since we are showing the window async, we will set the window
+    // fullscreen once it is ready to be shown
+    // (browserOpts.fullscreen may already be set when loading from saved state,
+    // so this just sets it for other scenarios)
+    if (isDarwin && parentWindow && parentWindow.isFullScreen()) {
```
bsclifton: this logic (all of what is in this setImmediate) might be better to pull out to a discrete function.  This would help with unit testing and could help with readability. Something like `getCreateWindowArgs` (passing what's needed). Then whatever is returned is directly passed to the call `new BrowserWindow`

bsclifton: Unit tests would then be a great way to capture the behavior (of setting fullscreen to false) and the tests themselves would act as documentation (if someone makes a change which breaks the test).  Could we get some unit tests for this?

app/browser/reducers/windowsReducer.js
petemill: moved to _config.js_

app/browser/reducers/windowsReducer.js
petemill: I have split all the window creation, showing, deferred showing, and timeout showing logic to a different place - _app/browser/windows.js_ which seemed an appropriate place which deals with calling functions on a muon BrowserWindow (and children). It's also the same place that the whenRendered action is handled, so that nicely puts all the logic in one place, including the temporary but ugly win.__blah properties - so they are at least encapsulated in one module.

I created tests for this new logic in _windowsTest.js_ handling:
- showing the window immediately
- showing the window if the timout expires
- showing the window on whenRendered firing for the window
- ensuring the window is fullscreen or maximized on each of those scenarios 

I also created tests for most of the existing action handler logic for APP_NEW_WINDOW in _windowReducerTest.js_, of which there were no existing tests, including tests for:
- creating the right tabs in the new window
- window position and size
- maximizing the window

test/unit/app/browser/reducers/windowsReducerTest.js
```diff
@@ -20,36 +21,192 @@ describe('windowsReducer unit test', function () {
...
+
+    it('creates a window, with a single specified frame', function () {
+      const action = {
+        actionType: appConstants.APP_NEW_WINDOW,
+        frameOpts: sampleFrame1
+      }
+      windowsReducer(state, action)
+      fakeTimers.tick(0)
+      // ensure the window api was asked to create the frame
+      const actualCreateWindowFrameArg = spy.args[0][3]
```
bsclifton: this is an interesting way to check the params- any reason you'd do this over something like `assert(spy.withArgs(args_here).calledOnce)`?

app/browser/reducers/windowsReducer.js
```diff
@@ -325,11 +309,14 @@ const windowsReducer = (state, action, immutableAction) => {
       state = windows.init(state, action)
       break
     case appConstants.APP_NEW_WINDOW:
-      state = createWindow(state, action)
+      state = handleCreateWindowAction(state, action)
       break
     case appConstants.APP_WINDOW_READY:
       windows.windowReady(action.get('windowId'))
       break
+    case appConstants.APP_WINDOW_RENDERED:
```
bsclifton: Great having an event for this! 😄

app/browser/windows.js
```diff
@@ -111,6 +120,26 @@ const updatePinnedTabs = (win) => {
   }
 }
 
+function showDeferredShowWindow (win) {
```
bsclifton: much better encapsulation this go-around 😄 👍

test/unit/app/browser/reducers/windowsReducerTest.js
petemill: @bsclifton In this case, sure that more expressive function could be used. I was just being consistent with the other tests here, which need to inspect the windowOptions object argument for property values, and not necessarily enforce every property on the object argument. If there's a nicer way to do that vs being tied to the argument order I'd love to learn!

petemill: @bsclifton I updated the couple of tests I could using that syntax. It looks nicer, thanks. Only downside is the error messages aren't as expressive as using the chai.assert functions.

petemill:
Just noticed that multiple homepage tabs, as specified in https://community.brave.com/t/how-to-set-up-multiple-home-pages/9998 is not working in master (or this branch). I'll see if it's an easy fix we can bundle in this, or create the issue

bsclifton:
Found a problem- I grabbed latest (saw there was a rebase / force push after I last looked at code) and tried it out.  The sizes are not being saved for the windows ☹️  Here's a clip: basically the height for the window on the right is not saved
not-saving-size

app/browser/reducers/windowsReducer.js
```diff
@@ -43,17 +39,43 @@ function isModal (browserOpts) {
...
  */
 const setWindowDimensions = (browserOpts, defaults, immutableWindowState) => {
   assert(isImmutable(immutableWindowState))
-  if (immutableWindowState.getIn(['windowInfo'])) {
-    browserOpts.width = firstDefinedValue(browserOpts.width, immutableWindowState.getIn(['windowInfo', 'width']))
-    browserOpts.height = firstDefinedValue(browserOpts.height, immutableWindowState.getIn(['windowInfo', 'height']))
+  const windowInfoState = immutableWindowState.get('windowInfo')
+  if (windowInfoState) {
+    browserOpts.width = firstDefinedValue(browserOpts.width, windowInfoState.get('width'))
+    browserOpts.height = firstDefinedValue(browserOpts.height, windowInfoState.get('windowInfo'))
```
bsclifton: Here's the bug causing the height issue 🐛  👀 !!

petemill: oh good spot @bsclifton, I remember deleting that code 🤦‍♂️

app/browser/windows.js
```diff
@@ -330,6 +372,108 @@ const api = {
...
+      // window should be shown already
+      // manual maximize
+      if (maximized) {
+        win.maximize()
+      }
+      // NOTE: we don't need to fullscreen manually since it's specified in options
+      // passed to BrowserWindow constructor
+    }
+    // let store know there's a new window
+    // so it can subscrive to state updates
```
bsclifton: s/subscrive/subscribe

test/unit/lib/fakeWindow.js
```diff
@@ -3,17 +3,65 @@
...
-  constructor (id) {
-    super()
-    this.id = id
-    this.webContents = Object.assign(new EventEmitter())
-    this.webContents.send = this.webContents.emit
-  }
-  getId () {
-    return this.id
+// cannot be a class since sinon has
+// trouble stubbing the constructtor for a class
```
bsclifton: s/constructtor/constructor

test/unit/lib/fakeElectron.js
```diff
@@ -100,7 +88,16 @@ const fakeElectron = {
...
+  screen: {
+    getDisplayMatching: () => new FakeElectronDisplay(),
+    getPrimaryDisplay: () => new FakeElectronDisplay(),
+    getDisplayNearestPoint: () => new FakeElectronDisplay(),
+    getAllDisplays: () => [new FakeElectronDisplay()],
+    getCursorScreenPoint: () => ({ x: 200, y: 200 })
   }
 }
 
+// function getFakeDisplay
```
bsclifton: can this be removed?

bsclifton:
master 01e76241e53bbf4627e11aaee87411d5f40df9f4
0.21.x 57c20b682b6801a8b6342a368609210dbe11d565
0.20.x 0b27319147c8809e216a2f6e2feb555e480b5b58
0.19.x need some help merging this one, @petemill 😄

petemill:
@bsclifton I've got a successful merge on 0.19.x (pushed to [origin/0.19.x-hide-white-windows](https://github.com/brave/browser-laptop/commits/0.19.x-hide-white-windows)) - complication was that all the window creation logic had been moved and edited from appStore.js to windowReducer.js previous to this in #9963 and #10458, both of which were only merged to 0.20.x.  This could be a risk, but we've actually introduced more unit tests here, so up to you. If we do keep the 0.19.x merge, I would say incorporate the test plan from those PRs, but they both don't have one, so this actually would merge those fixes and create a test plan for them 🤷‍♂️ 😄 

I had to introduce all windowsReducerTests and windowsTests - and all the ones we created pass. However, there were two that were created before this PR that do not pass (and I removed them):
1. _windowsReducerTest.js_/`APP_WINDOW_UPDATED`/`updateDefault is true` is not passing because saving window dimensions is not working very well in 0.19.x. On the current 0.19.x, your [screencap above](https://github.com/brave/browser-laptop/pull/11764#issuecomment-346251228) doesn't work, i.e. the app only remembers one of the window positions, and layers all windows at that position, on top of eachother. This PR does not fix that, since it already had been fixed in master (in #10458). I removed that test.
2. _windowsTest.js_/`updatePinnedTabs` these tests were added as part of the #10531 fix, that is not in 0.19.x, so I removed them.

bsclifton:
Based on the above (and in the interest of reducing complexity), let's push this to 0.20.x 👍 

Thanks for taking the time to evaluate the merge 😄

PR #11818, opened 2017-11-07
Avoid 'flashes of white' when tabs open or close

### LINKED ISSUE #11813: White flash after tab open or close
### Test plan
See https://github.com/brave/browser-laptop/pull/11818

### Description
When opening a tab, there is a flash of the background of the `<webview>` component, which is explicitly set to white in window.less.
When closing a tab, there is a flash of the background of the whole window (or at least `div.tabContainer`) as the active frame is removed, but the next active frame is not set active yet.

The opening tab scenario can be solved by setting the `<webview>` background to not be white in certain cases, or not showing it until the background of the incoming page is ready (perhaps, more risky).

the closing tab scenario can be solved by either:
1. Setting next active tab before closing the tab, and then waiting Xms to actually close the tab (tested it and works OK, but the browser process feels the wrong place to do it for a rendering issue), or;
2. Keeping the frame (and `<webview>`) around until the newly active frame has a chance to be shown (tested and works OK, but an arbitrary number of ms)

### Steps to Reproduce
Set new tab page to Dashboard.
Open and close many tabs.

**Actual result:**
Observe white flashes

**Expected result:**
No white flashes between opening or closing tabs

**Reproduces how often:**
100%

### Brave Version

**about:brave info:**
master

Yes

### Additional Information
PR incoming

petemill:
Remove white flash for new tab pages which use about:newtab.
Sets the frame background to the same background as the newtab page, for a smooth transition between frame creation and the tab beginning to load content.
Keeps a closed frame around until the next frame is ready, to avoid flash when closing tabs.

Fix #11813
Addresses #5309

When opening a tab, there is a flash of the background of the <webview> component, which is explicitly set to white in window.less.
This is avoided by setting the background to the 'new tab dashboard' background color (only when that is the Url for the frame), and by hiding the frame if it's about:blank and for the first 150ms since when a frame is first created, it starts as about:blank and then sets its location to the specified Url.

When closing a tab, there is a flash of the background of the whole window (or at least div.tabContainer) as the active frame is removed, but the next active frame is not set active yet. This is avoided by keeping the `Element` for the frame around for 100ms, with a low z-index, so that the new / next active tab will show on top when it is ready.

Closing tabs - Before:
https://www.dropbox.com/s/aon11cov3nqcau2/p11813-close-tabs-before.mov?dl=0

Closing tabs - After:
https://www.dropbox.com/s/0x1n8msobxssihh/p11813-close-tabs-after.mov?dl=0

## Test Plan:
- When creating a new tab there is no flash between the frames (~difficult to see until #11817 lands since all new frames have white background, and the window has a white background~).
- When closing a tab with a loaded previous and next tab, there is no white flash between leaving the old tab and showing the next tab.

## Submitter Checklist:

## Security
Introduces new package: https://github.com/reactjs/react-transition-group

Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/frame/frame.js
+++ b/app/renderer/components/frame/frame.js
@@ -859,12 +859,15 @@ class Frame extends React.Component {
 
     const props = {}
     // used in renderer
+    props.transitionState = ownProps.transitionState
     props.partition = frameStateUtil.getPartition(frame)
     props.isFullScreen = frame.get('isFullScreen')
     props.isPreview = frame.get('key') === currentWindow.get('previewFrameKey')
     props.isActive = frameStateUtil.isFrameKeyActive(currentWindow, frame.get('key'))
     props.showFullScreenWarning = frame.get('showFullScreenWarning')
     props.location = location
+    props.isDefaultNewTabLocation = location === 'about:newtab'
+    props.isBlankLocation = location === 'about:blank'
     props.tabId = tabId
     props.showMessageBox = tabMessageBoxState.hasMessageBoxDetail(state, tabId)
     props.isFocused = isFocused(state)
@@ -906,7 +909,17 @@ class Frame extends React.Component {
     return props
   }
 
+  getTransitionStateClassName (stateName) {
+    // handle missing data
+    if (!stateName) {
+      return null
+    }
+    // convert Transition element state string to a more consistent css classname
+    return `is${stateName[0].toUpperCase()}${stateName.slice(1)}`
+  }
+
   render () {
+    const transitionClassName = this.getTransitionStateClassName(this.props.transitionState)
     return <div
       data-partition={this.props.partition}
       data-test-id='frameWrapper'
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -118,8 +118,9 @@
     "react-dnd": "^2.1.4",
     "react-dnd-html5-backend": "^2.1.2",
     "react-dom": "^15.5.4",
-    "snazzy": "^7.0.0",
     "react-select": "^0.9.1",
+    "react-transition-group": "^2.2.1",
```
diracdeltas: approved in https://github.com/brave/internal/issues/140#event-1334755820

petemill:
I've rebased this branch and specified a test plan. Would appreciate a review and merge to master so I can land open a PR for some other features based on this.

app/renderer/components/frame/frame.js
```diff
@@ -906,16 +909,29 @@ class Frame extends React.Component {
     return props
   }
 
+  getTransitionStateClassName (stateName) {
+    if (!stateName) {
+      return null
+    } else {
```
bsclifton: nit: no else needed here

bsclifton:
Although we default to dashboard, users can pick others (including an empty tab). I noticed if I picked empty tab, the tab stays as the gray color
screen shot 2017-11-17 at 12 45 38 am

There might be a flash of gray when opening a new tab when set to search engine / home page, but I can't tell for sure (my eyes are horrible! it's too fast)

bsclifton:
master 93d7a771066c60bb56bcf0a07825861d447d2d88
0.21.x 225c213a77e940286f3a0f2b10287a05daec79da
0.20.x b5e7e43c13e8f27d018ea0d1610331790ce3d89b
0.19.x working though merge conflicts...

PR #11820, opened 2017-11-07
ensure first frame key is always valid

### LINKED ISSUE #11208: Pinning throws error
### Description

When you pin second tab you get error.
### Steps to Reproduce

  1. Load one page in the first tab
  2. Load second page in the second tab
  3. Try to pin second page

**Actual result:**

image

image

**Expected result:**
No error
**Reproduces how often:** [What percentage of the time does it reproduce?]
100%
### Brave Version

**about:brave info:**

 
Brave: 0.21.0 
V8: 6.1.534.41 
rev: c5a903cc2672577a41fdf8ad070aec245ecaa367 
Muon: 4.4.25 
OS Release: 16.7.0 
Update Channel: Release 
OS Architecture: x64 
OS Platform: macOS 
Node.js: 7.9.0 
Brave Sync: v1.3.5 
libchromiumcontent: 61.0.3163.100

cezaraugusto:
Auditors: @nejczdovc
fix #11208

Manual Test Plan:

* Open 2 tabs
* Pin the last
* Should throw no error

Test Plan:
npm run test -- --grep="isFirstFrameKeyIntabPage"

### DIFF
--- a/js/state/frameStateUtil.js
+++ b/js/state/frameStateUtil.js
@@ -480,7 +480,7 @@ const isFirstFrameKeyInTabPage = (state, frameKey) => {
   const firstFrame = unpinnedTabs
     .slice(startingFrameIndex, startingFrameIndex + tabsPerTabPage).first()
 
-  return firstFrame.get('key') === frameKey
+  return firstFrame && firstFrame.get('key') === frameKey
 }
 
 const getTabPageIndex = (state) => {
--- a/test/unit/state/frameStateUtilTest.js
+++ b/test/unit/state/frameStateUtilTest.js
@@ -444,4 +444,52 @@ describe('frameStateUtil', function () {
       assert.equal(result, true)
     })
   })
+
+  describe('isFirstFrameKeyInTabPage', function () {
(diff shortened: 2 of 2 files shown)


js/state/frameStateUtil.js
```diff
@@ -480,7 +480,7 @@ const isFirstFrameKeyInTabPage = (state, frameKey) => {
   const firstFrame = unpinnedTabs
     .slice(startingFrameIndex, startingFrameIndex + tabsPerTabPage).first()
 
-  return firstFrame.get('key') === frameKey
+  return !!firstFrame && firstFrame.get('key') === frameKey
```
NejcZdovc: do we need to have `!!` if I understand correctly `firstFrame` is (if exists) map, so I think that we should be ok without double negation

cezaraugusto: ya it's was me forcing coercion. I tried without it w/ tests and worked same way so I made your request

cezaraugusto:
master bf1c2d9
0.21.x 88b225f36
0.20.x 10747ad36

PR #11862, opened 2017-11-09
Prevent firing updatePinnedTabs repeatedly for the same state/window combination

### LINKED ISSUE #11861: Pinning tabs with multiple windows open results in duplicates
### Description
Pinning tabs often causes those tabs to be pinned multiple times, especially if there is more than 1 window open. This is likely because `updatePinnedTabs` fires cyclically, i.e:
	1. A tab is pinned in windowA
	2. updatePinnedTabs is called to ensure all other windows get the pinned tabs
	3. windowB has the tab pinned, which causes…
	4. updatePinnedTabs to be called
        5. State is not necessarily up to date, so windowB or windowC gets the tab pinned again

### Steps to Reproduce

  1. Open window1
  2. Open window2
  3. Open a tab in window1
  4. Pin a tab in window1

This can then cause some weird errors when a window's tab state is out of sync with what it should have.

**Actual result:**
  - The tab is pinned twice in window2 and once in window1

**Expected result:**
 - The tab is pinned once in window2 and once in window1

**Reproduces how often:**

### Brave Version

**about:brave info:**
master (0.22)

No

petemill:
Fix #11861 (and #12377)

Submitter Checklist:

Test Plan:

### Steps to Reproduce

  1. Open window1
  2. Open window2
  3. Open a tab in window1
  4. Pin a tab in window1

This can then cause some weird errors when a window's tab state is out of sync with what it should have.

**Expected result:**
 - The tab is pinned once in window2 and once in window1

**Previous actual result:**
  - The tab is pinned twice in window2 and once in window1

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/windows.js
+++ b/app/browser/windows.js
@@ -5,6 +5,7 @@
 const electron = require('electron')
 const Immutable = require('immutable')
 const appActions = require('../../js/actions/appActions')
+const appStore = require('../../js/stores/appStore')
 const appUrlUtil = require('../../js/lib/appUrlUtil')
 const {getLocationIfPDF} = require('../../js/lib/urlutil')
 const debounce = require('../../js/lib/debounce')
@@ -29,6 +30,7 @@ const {app, BrowserWindow, ipcMain} = electron
 
 // TODO(bridiver) - set window uuid
 let currentWindows = {}
+const windowPinnedTabStateMemoize = new WeakMap()
 
 const getWindowState = (win) => {
   if (win.isFullScreen()) {
@@ -83,7 +85,7 @@ const siteMatchesTab = (site, tab) => {
   return matchesLocation && matchesPartition
 }
 
-const updatePinnedTabs = (win) => {
+const updatePinnedTabs = (win, appState) => {
   // don't continue if window won't need pinned tabs updated
   if (
     !win ||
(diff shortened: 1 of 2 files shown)


app/browser/windows.js
```diff
@@ -72,19 +75,26 @@ const siteMatchesTab = (site, tab) => {
   return matchesLocation && matchesPartition
 }
 
-const updatePinnedTabs = (win) => {
+const updatePinnedTabs = (win, appState) => {
```
NejcZdovc: not sure why we are passing state in, which we get in the same way as we did before?

petemill: Seems like a better practice and more consistent with the way other functions are structured. Plus it avoids calling AppStore.getstate multiple times in a row for multiple windows (though yes I know it is a very cheap function) and it feels cleaner and easier to test. Plus I’m not sure at all why the function had the AppStore require inside the function before...?? Was it to stop the module from being evaluated in case it wasn’t ever required in the process (I can’t imagine that happening)

app/browser/windows.js
```diff
@@ -92,15 +94,22 @@ const updatePinnedTabs = (win) => {
...
-  let pinnedWindowTabs = getPinnedTabsByWindowId(state, windowId)
+  const statePinnedSites = pinnedSitesState.getSites(appState)
+  // no need to continue if we've already processed this state for this window
+  if (windowPinnedTabStateMemoize.get(win) === statePinnedSites) {
+    return
+  }
+  // cache that this state has been updated for this window,
+  // so we do not repeat the operation until
+  // this specific part of the state has changed
+  // See
```
bsclifton: looks like your comment just drops off here 😛

bsclifton:
master 5d78c1ca8ed98314986a2b73b7e4c6fc0bade2a2
0.21.x 83958edc8371e3bf1e60a14b0a41e1b8b8c7d6d7
0.20.x a718cb866bb0fad97f11cf343a864f3732c5950a

Issue #6944, opened 2017-01-31
Misuse of Math.random() to generate random integers


dimitri-xyz:
In     
https://github.com/brave/browser-laptop/blob/57e69652217e69c937407f1f776b370f75dbd38f/app/extensions/brave/content/scripts/adInsertion.js#L74-L75

That is:
```javascript
var segments = ['IAB2', 'IAB17', 'IAB14', 'IAB21', 'IAB20']
var segment = segments[Math.floor(Math.random() * 4)]
```
The last value 'IAB20' is *excluded* from the range of random() and will never be chosen.

Here's another bug:
https://github.com/brave/browser-laptop/blob/57e69652217e69c937407f1f776b370f75dbd38f/tools/lib/transactionHelpers.js#L48

```javascript
 const count = Math.round(Math.random() * 100)
```
In this case, both 0 and 100 are part of the results, but only occur half as frequently as the other numbers. To see why this happens consider:

```javascript
 const zeroOneTwo = Math.round(Math.random() * 2)
```
all numbers from 0 to 0.49999 will return 0
all numbers from 0.5 to 1.4999 will return 1
all numbers from 1.5 to 1.9999 will return 2

So 1 occurs with probbility 1/2 and 0 and 2 are returned with probability 1/4 each.
This is very different from the expected uniform distribution (1/3,1/3,1/3).

The problem here is the flooring or rounding of floating point random numbers to obtain integers.
I think the "proper" way to obtain random integers in a given range is to use `randomInt` from `random-lib`

*There are many more instances of this in the code. Just do a search for random() to find them.*

bsclifton:
We do use random-lib in our ledger code; it wouldn't be hard to rework instances to call that instead

[Docs on Math.random() explaining the exclusivity issue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) shown in the first example (4 never being possible because 1 is excluded from Math.random results)

Liunkae:
Which instances are worth fixing? https://github.com/brave/browser-laptop/search?l=JavaScript&q=random&type=Code&utf8=%E2%9C%93
Here is an example for the new tab page backgrounds: ([js/about/newtab.js](https://github.com/brave/browser-laptop/blob/master/js/about/newtab.js))
````js
@@ -19,7 +19,7 @@ const urlutils = require('../lib/urlutil')
 const siteTags = require('../constants/siteTags')
 const config = require('../constants/config')
[...]

PR #11874, opened 2017-11-09
Adds UGP support

### LINKED ISSUE #11100: Implement BAT matching funds promotion (UGP - user growth pool)
~With the launch of BAT Mercury, we will be creating an incentive program for users to add funds to their Brave wallet.~

The process will work as follows: Any user of the Brave desktop browser who updates to the BAT Mercury release will have at least their first $5 matched by Brave (from Brave's BAT User Growth Pool) when transferred to the Brave wallet. 

There are unfortunately still some unknowns, such as for what duration the promotion will run; will it be offered to a given user more than once, and if so at what frequency.

Regardless of the resolution of those outstanding questions, it is a requirement that the promotion be able to be switched off by setting a server-side flag. That is, if we find that the promotion is getting away from us for any reason, we will be able to discontinue it immediately, without requiring that users update their browsers.

This requires that the browser check with the server to see if the promotion has been disabled (likely when the Payments screen in Preferences is opened) and if it has been disabled, the elements of the UI that correspond to the promotion need to be disabled. This includes (in current UI designs) at least one dialog and one string. 

_**Sub-tasks**_
- [ ] Make BAT matching promotion configurable #11869
- [ ] Text on Payments screen should be retrieved from server side #11870
- [ ] Payments screen should have a "Claim free tokens" button #11871
- [ ] Show notification for BAT Matching - users with payments disabled #11867
- [ ] Show notification for BAT Matching - users with empty wallets #11868 
- [ ] Show notification for BAT Matching - users with non-empty wallets (TBD)

NejcZdovc:
## Submitter Checklist:

Resolves #11100 

## Test plan
- [x] Plan A:
- clean profile
- check if you see notification about promotion

- [x] Plan B:
- clean profile
- enable payments
- check if first notification is hidden and second one is displayed

- [x] Plan C:
- clean profile
- click on both buttons on the first notification (maybe later, should display first notification after 24h again; second button should open payments page)

- [x] Plan D:
- clean profile
- enable payments
- close preference page
- click on both buttons on the second notification (maybe later, should display second notification after 24h again; second button should open payments page)

- [x] Plan E:
- disable internet connection
- clean profile
- check if payments page have alternative text

- [x] Plan F:
- clean profile
- check if payments page have custom promotion text

- [x] Plan G:
- clean profile
- enable payments 
- check if claim button is there
- click on it
- you would get free BAT's and text below `Add Funds`
- claim button and notifications should be hidden

## Reviewer Checklist:

Tests


app/browser/api/ledger.js
```diff
@@ -2343,6 +2360,51 @@ const onMediaPublisher = (state, mediaKey, response, duration, revisited) => {
...
+    appActions.saveLedgerPromotion(result)
+  })
+}
+
+const claimPromotion = (state, promotionId) => {
+  if (!client) {
+    return
+  }
+
+  client.setPromotion(promotionId, (err, result) => {
```
diracdeltas: getPromotion and setPromotion both use retryTrip. need to be careful that the retry algorithm doesn't trigger whatever we end up using for rate-limiting and protection against DoS (cloudflare?) on the grant server

cc @evq

diracdeltas:
how random do the calls to randomInt have to be? they won't generate perfect random distributions due to #6944

app/browser/api/ledgerNotifications.js
```diff
@@ -307,8 +342,48 @@ const showBraveWalletUpdated = () => {
...
+  if (notification.isEmpty()) {
+    return
+  }
+
+  appActions.hideNotification(notification.get('message'))
+}
+
 if (ipc) {
-  ipc.on(messages.NOTIFICATION_RESPONSE, (e, message, buttonIndex) => {
+  ipc.on(messages.NOTIFICATION_RESPONSE, (e, message, buttonIndex, checkbox, index, buttonActionId) => {
```
diracdeltas: is this supposed to do something with `checkbox` and `index`?

app/renderer/components/preferences/payment/disabledContent.js
```diff
@@ -17,27 +20,63 @@ const uphold = require('../../../../extensions/brave/img/ledger/uphold_logo_smal
...
+    }
+
+    const text = this.props.ledgerData.getIn(['promotion', 'panel', 'optInMarkup'])
+    const claimed = this.props.ledgerData.has('claimedTimestamp')
+
+    if (!text || !claimed) {
+      return
+    }
+
+    this.text = <div dangerouslySetInnerHTML={{ __html: text }} />
```
diracdeltas: we should avoid using dangerouslySetInnerHTML. could `optInMarkup` be modified to return text and templating data instead of raw HTML?

app/renderer/components/preferences/payment/enabledContent.js
```diff
@@ -167,6 +207,39 @@ class EnabledContent extends ImmutableComponent {
...
+
+    return <div className={cx({[css(styles.enabledContent__grant)]: true, 'enabledContent__grant': true})}>
+      <div
+        className={cx({
+          [css(styles.enabledContent__grant_close)]: true,
+          'fa fa-close': true
+        })}
+        onClick={this.closeClick}
+      />
+      <p className={css(styles.enabledContent__grant_title)} dangerouslySetInnerHTML={{ __html: successText }} />
```
diracdeltas: same as above for `dangerouslySetInnerHTML`

diracdeltas:
marking as blocked on https://github.com/brave/internal/issues/147

docs/state.md
```diff
@@ -238,6 +238,83 @@ AppStore
         publisher: string // url of the publisher in question
       }
     },
+    promotion: {
```
diracdeltas: need to check that if a user modifies/deletes this, they aren't able to re-trigger the promotion and claim funds into the same wallet

evq: this is enforced by ledger and grant servers. they both enforce the max of one grant allocated from a given promotionId per wallet.

app/browser/api/ledgerNotifications.js
NejcZdovc: no we don't need this two params, but we need the last one `buttonActionId`, so I needed to add them

NejcZdovc:
@diracdeltas `dangerouslySetInnerHTML` removed

app/browser/api/ledger.js
```diff
@@ -1114,6 +1114,10 @@ const enable = (state, paymentsEnabled) => {
     }, 15 * ledgerUtil.milliseconds.second)
   } else if (paymentsEnabled) {
     // on toggle
+    if (!togglePromotionTimeoutId) {
```
bsclifton: Is this correct? shouldn't it be clearing when it's non-zero?

NejcZdovc: no it's not. Fixed

bsclifton:
master 01477552d5268880c7abee24fad69a91e59b7a2b
0.21.x in progress
0.20.x TODO
0.19.x TODO

bsclifton:
@NejcZdovc can you please double check my 0.19.x manual merge? I manually tested and it looks great- much better than previous attempt (0.19.98). I put in some extra time to make sure it's perfect

PR #11916, opened 2017-11-12
[WIP] comment out APP_TAB_ATTACHED and APP_TAB_WILL_ATTACH

ayumi:
~20ms / tab is spent in these reducer actions when starting up.
Removing this logic doesn't seem to cause any new test failures.

Auditors:

Test Plan:

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/reducers/tabsReducer.js
+++ b/app/browser/reducers/tabsReducer.js
@@ -73,16 +73,18 @@ const tabsReducer = (state, action, immutableAction) => {
       state = tabState.maybeCreateTab(state, action)
       break
     case appConstants.APP_TAB_ATTACHED:
-      state = tabs.updateTabsStateForAttachedTab(state, action.get('tabId'))
+      // XXX Commented out for SCIENCE - Ayumi
+      // state = tabs.updateTabsStateForAttachedTab(state, action.get('tabId'))
       break
     case appConstants.APP_TAB_WILL_ATTACH: {
-      const tabId = action.get('tabId')
-      const tabValue = tabState.getByTabId(state, tabId)
-      if (!tabValue) {
-        break
-      }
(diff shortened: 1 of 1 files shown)


ayumi:
@petemill @bridiver is it safe to comment these out

petemill:
@ayumi What seems to be happening here is that on WILL_ATTACH, the old window is being synced with the tab object, and on DID_ATTACH, the new window is being synced with the muon tab object, via `updateTabsStateForWindow`. It's difficult to tell immediately (for me) if this is necessary because the `tab` object coming from muon is a bit of a black box in the context of browser-laptop, in that it's not clear what could change and require state sync, or if it's still a thing.

At the moment it may not be doing anything but, as it happen, I'm using APP_TAB_ATTACHED in the [continuous tab dragging work](https://github.com/brave/browser-laptop/pull/11720) I'm doing at the moment, to know when the tab being dragged has been attached to the window, and the drag operation can continue... I think it might be worthwhile considering batching these though so that opening hundreds of tabs does not thrash the stores. I have experimented with that recently.

But if someone (@bridiver perhaps?) can tell us we don't need to perform `updateTabsStateForWindow` then let's at least remove the store reducer action handlers and I can add what I need to back in for the branch I'm working on. The other issue with removing the store action calls is that the muon tab events are not documented anywhere AFAIK, so the code in **app/browser/tabs.js** is the only place we can discover that there are `will-attach` and `did-attach` events for the `tab` object.

ayumi:
closing – i think doing diligence on removing these events seems to be effort; and @petemill's related work like https://github.com/brave/browser-laptop/pull/11968 will help with the many tabs slow launch problem.

Issue #11945, opened 2017-11-14
if the PIA proxy service is blocked, fallback to non-proxy server


mrose17:
some users operate in an environment in which the PIA service is blocked. experimentally, such environments return 403.

mrose17:
cf., https://github.com/brave/browser-laptop/tree/issue-11945

bsclifton:
I feel confident that this is well covered by automated tests

Issue #11936, opened 2017-11-13
Transition is still occurring after update to 0.19.88 for some users


LaurenWags:
### Description
Some users are still reporting seeing the transition after updating to 0.19.88
https://twitter.com/stshank/status/929846370835378183

### Steps to Reproduce
TBD - further information needed.

**Actual result:**
Transition is still occuring

**Expected result:**
When not needed, transition should not occur.

**Reproduces how often:**
unknown

### Brave Version

**about:brave info:**
0.19.88

**Reproducible on current live release:**
Yes

### Additional Information
cc @kjozwiak

BrendanEich:
We need a timeout to stop the endless spinner in a way that the user can grok, and take better action in response to.

bsclifton:
OK- after some investigation, I've found at least one fatal issue which we need to handle

A different user shared logs and ran into this scenario:
- they were a long time user (not a new user)
- they upgraded to this new release, which has the transition to BAT
- in the code, they qualified to transition (and call was made)
  - transition was not done because ledger client marked itself as busy
  - ledger client does not retry

The ledger client was stuck in a busy situation for this user because their workplace blocked the "privateinternetaccess.com" used for reconciling. We contact this service to get timestamps for when reconciliation needs to happen, so it's not specific to the transition

There are two actions I'm going to try and take to resolve this issue:
1. Update the transition code to retry after a delay (15 mins, for example)
2. Update the Payments screen to show fatal errors which are preventing the ledger from working (corrupt file, can't reach server, etc)

PR #11950, opened 2017-11-14
Hardening for ledger errors

bsclifton:
Fixes https://github.com/brave/browser-laptop/issues/11936
Fixes https://github.com/brave/browser-laptop/issues/11945

Changes contain two fixes:
- for any calls: if a 403 is received back and proxy is enabled, ledger will disable proxy and try again
- for `transitionWalletToBat`: if busyP returns a truthy value, `transitionWalletToBat` will be retried up to 3 times before quitting.

## Submitter Checklist:

## Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/api/ledger.js
+++ b/app/browser/api/ledger.js
@@ -1262,7 +1262,7 @@ const roundtrip = (params, options, callback) => {
     if (err) return callback(err, response)
 
     if (Math.floor(response.statusCode / 100) !== 2) {
-      if ((params.useProxy) && (response.statusCode === 403)) {
+      if (params.useProxy && response.statusCode === 403) {
         params.useProxy = false
         return roundtrip(params, options, callback)
       }
@@ -2094,6 +2094,8 @@ const getNewClient = () => {
   return newClient
 }
 
+let busyRetryCount = 0
+
 const transitionWalletToBat = () => {
   let newPaymentId, result
 
@@ -2164,7 +2166,13 @@ const transitionWalletToBat = () => {
   }
 
   if (client.busyP()) {
-    console.log('ledger client is currently busy; transition will be retried on next launch')
+    if (++busyRetryCount > 3) {
+      console.log('ledger client is currently busy; transition will be retried on next launch')
+      return
+    }
+    const delayTime = random.randomInt({ min: miliseconds.minute, max: 10 * miliseconds.minute })
+    console.log('ledger client is currently busy; transition will be retried shortly (this was attempt ' + busyRetryCount + ')')
+    setTimeout(() => transitionWalletToBat(), delayTime)
     return
   }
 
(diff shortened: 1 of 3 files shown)


app/browser/api/ledger.js
```diff
@@ -2160,7 +2167,13 @@ const transitionWalletToBat = () => {
   }
 
   if (client.busyP()) {
-    console.log('ledger client is currently busy; transition will be retried on next launch')
+    if (++busyRetryCount > 3) {
+      console.log('ledger client is currently busy; transition will be retried on next launch')
+      return
+    }
+    const delayTime = random.randomInt({ min: 1, max: 500 })
```
NejcZdovc: I would add higher values, because now we are setting timeout between 1ms and 500ms

test/unit/app/browser/api/ledgerTest.js
```diff
@@ -421,8 +424,14 @@ describe('ledger api unit tests', function () {
   })
 
   describe('transitionWalletToBat', function () {
+    let fakeClock
+
+    before(function () {
+      fakeClock = sinon.useFakeTimers()
```
bsclifton: This is needed because code paths hitting setTimeout() will cause unit tests to crash (seemingly randomly) and in some cases hang.  When the fake timer is destroyed, all methods registered with setTimeout are cleaned up.

app/browser/api/ledger.js
bsclifton: Fixed 👍

app/browser/api/ledger.js
```diff
@@ -2160,7 +2167,13 @@ const transitionWalletToBat = () => {
   }
 
   if (client.busyP()) {
-    console.log('ledger client is currently busy; transition will be retried on next launch')
+    if (++busyRetryCount > 3) {
+      console.log('ledger client is currently busy; transition will be retried on next launch')
```
NejcZdovc: should we hide transition overlay when busy fails for the third time?

bsclifton: I don't know if it makes sense to hide it without providing a work-around. We don't want to encourage using the old servers indefinitely- I think it would be better to update the transition screen to provide a community link about stuck transitions.

cc: @alexwykoff

alexwykoff:
Is 403 the only case or is it also common for firewalls to redirect to a 'this site is blocked...' with a 200?

bsclifton:
@alexwykoff good question- I'm not sure

bsclifton:
**_UPDATE_**: I confirmed Websense and Barracuda content filters should be using HTTP status code 403 by default for blocked content

bsclifton:
master c369861a75a996956c7f2f4d2ed73eead004de46
0.21.x 82ef6f17e662cc2f1d77d76f6531831f4fee6fed
0.20.x aa9149663c858abd0eb38cd67d3ab1c2d394c73f
0.19.x a949f744ed4e289887b0a110dd98669aa0b354b3

PR #11962, opened 2017-11-15
cache-require-paths

ayumi:
Add [cache-require-paths](https://github.com/bahmutov/cache-require-paths).

BL version | 1st start | Subsequent starts
-- | -- | --
master 01477552d5268880c7abee24fad69a91e59b7a2b | 770 ms | 770 ms
feature/require-faster 9ffc7eed3de2bc63264b1a27dc6172672b314611 | 790 ms | 700 ms

The first start is slightly slower (maybe due to writing the cache) but subsequent starts are faster.

### Test plan:

1. Try packaged build on MacOS, Linux and Windows.
2. Try dev on MacOS, Linux and Windows.

#### Measuring perf
1. Log run time of app/index.js from 0 to perWindowStateLoaded ready = true
  a. To L10 add:
```js
const tStart = process.hrtime()
```
  b. To L361 after `ready = true` add:
```js
const tTime = process.hrtime(tStart)
console.log(`index.js window ready: ${tTime[0] * 1e3 + tTime[1] * 1e-6} ms`)
```
2. Start brave and record the run time.
3. Repeat

Note: when switching branches delete node_modules and reinstall to be sure the npm dedupe changes got applied

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ public/built
 .DS-Store
 temp.VisualElementsManifest.xml
 app/extensions/gen/
+.cache-require-paths.json
 
 # electron-packager output dirs
 *-darwin-x64
--- a/app/index.js
+++ b/app/index.js
@@ -8,6 +8,7 @@
 let ready = false
 
 require('v8-compile-cache')
+require('cache-require-paths')
 
 const CrashHerald = require('./crash-herald')
 const telemetry = require('./telemetry')
--- a/package-lock.json
+++ b/package-lock.json
@@ -2666,6 +2666,11 @@
       "resolved": "https://registry.npmjs.org/bytes/-/bytes-3.0.0.tgz",
       "integrity": "sha1-0ygVQE1olpn4Wk6k+odV3ROpYEg="
     },
+    "cache-require-paths": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/cache-require-paths/-/cache-require-paths-0.3.0.tgz",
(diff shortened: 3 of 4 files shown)


ayumi:
@bsclifton thanks for the reminder! I rebased and tried it out again on Linux (dev and built package)

bsclifton:
Security review created https://github.com/brave/internal/issues/161

Will wait for results before merging

diracdeltas:
i wonder if we will run into cache invalidation issues? https://github.com/bahmutov/cache-require-paths/issues/6

diracdeltas:
sorry, accidental close

bsclifton:
@ayumi do you know if the cache is invalidated on each run? If not, we could encounter the problem @diracdeltas linked to when running from source (we wouldn't run into it for packaged builds though)

ayumi:
@diracdeltas @bsclifton That's a good catch. The only cache invalidation happens when the requiring a cached path fails, in that case it resolves from scratch: https://github.com/bahmutov/cache-require-paths/blob/master/index.js#L40

So it sounds like this may affect dev builds, however wouldn't affect packaged builds.
Would it be sufficient to add a post `npm install` hook to delete the cache file?

bsclifton:
@ayumi I think that would be perfect actually 😄 👍

bsclifton:
pushing back a few releases - @ayumi we can re-review once the hook is done (per notes above) 😄

bsclifton:
Closing PR for now- let's reopen if we have the time to revisit 😄

Issue #5419, opened 2016-11-05
Set max number of tabs loaded at the same time


luixxiul:
**Describe the issue you encountered:** Sometimes I notice the browser opens so many tabs at the same time by running script that the browser becomes quite slow because they eat up the resource on my machine.

**Expected behavior:** Set max number of tabs which should be loaded at the same time. The more tabs than the number should be loaded successively. Ideally the number should be decided dynamically based on the usage stats or telemetry for example, but the magic number would be acceptable as a starting point. 

- Brave Version: 0.12.7

- Any related issues:

PR #11968, opened 2017-11-15
Prevent rendering a frame's <webview> when it is created in an unloaded state

petemill:
Preventing load of the `<webview>` of frames/tabs that are not to be loaded yet eliminates the 5+ minute lockup / wait for windows to be created with hundreds of tabs.

All the tabs from persisted state are created immediately on window creation after startup. Whilst the rapid fire of store actions causes some delay (which can also be improved, separately), rendering a `<webview>` for every frame, attaching it to a tab and hooking in to its events causes exponential growth in the time taken to load frames simultaneously.

This PR makes sure the `<webview>`, and its associated events, are not created until the associated tabs are made active, or previewed.

It's possible that we can achieve the same results by rendering the `<webview>` and handling / forwarding less events, or batching them, but this isolation at least shows where the bulk of the time is spent.

Issues with doing this:
- [ ] can't reorder never-loaded tabs (probably because we don't have an accurate tab id since we're not getting the 'tab-id-changed' event until the `<webview>` is created)
  - move the handling of some events that deal with this data from the renderer `<webview>` to browser tab API. Do this as minimally as possible to avoid regressions and get this PR out ASAP

## Results:
6 windows, 150 unique tabs in each window
Before - 5+ mins of 200% cpu and increasing memory usage [﻿video](https://www.dropbox.com/s/lfnvbggcsztong8/6windows-150tabs-each-master.mov?dl=0)
After - 22 seconds [video](https://www.dropbox.com/s/clqkr18krm95hgp/6windows-150tabs-each-noframerender.mov?dl=0)

## Submitter Checklist:

## Test Plan:
 - Generate a state with many windows and unique tabs
    e.g.
    ```
    node generate-session-store.js --windows 6 --tabs 150 --out ~/Library/Application\ Support/brave-development/session-store-1 --unloadedtabs true
    ```
 - start the browser with the profile location of the generated profile
 - All windows and active tabs render in a considerably lower time than master
 - Repeat with blank, manually generated, profile, i.e. opening tabs in Brave, then restarting the browser.

## Reviewer Checklist:

### DIFF
--- a/app/renderer/components/frame/frame.js
+++ b/app/renderer/components/frame/frame.js
@@ -101,7 +101,7 @@ class Frame extends React.Component {
   }
 
   shouldCreateWebview () {
-    return !this.webview
+    return !this.webview && (!this.props.unloaded || this.props.isActive || this.props.isPreview)
   }
 
   allowRunningWidevinePlugin () {
@@ -190,7 +190,7 @@ class Frame extends React.Component {
   }
 
   componentDidMount () {
-    this.updateWebview(this.onPropsChanged)
+    this.updateWebview(() => this.onPropsChanged())
     if (this.props.activeShortcut) {
       this.handleShortcut()
     }
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -160,6 +160,7 @@ const newFrame = (state, frameOpts) => {
     const tabId = frameOpts.tabId
     const frame = frameStateUtil.getFrameByTabId(state, tabId)
     state = frameStateUtil.updateTabPageIndex(state, tabId)
+    state = frameStateUtil.setActiveFrameKey(state, nextKey)
     if (!active || !frame) {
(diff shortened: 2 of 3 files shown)


luixxiul:
@petemill do you think this resolves #5419? 😄

ayumi:
locally this saved ~1000ms on the 50 tab test (`'Performance startup type a URL and navigate 50 tabs'`)!

bridiver:
does this create any additional delay on first activation? Should we start a low-priority queue to load them in the background? I think we should only load them in the background if there is a noticeable difference.

petemill:
Yes but I think it’s not a fair test until some of the events that fire are
easy to move have been moved. If it’s still perceptible then we could
prioritise based on relationship to current active tab (e.g. next and
previous tabs)

bsclifton:
Closing as stale

PR #12105, opened 2017-11-27
memoize state reading

### LINKED ISSUE #11515: component mergeProps are unnecessarily running
### Description
Our architecture of a redux-style state and state -> props component connecting is interesting, and helpful in order to read from the state from any level of the component tree. However, it can be wasteful in that each component's `mergeProps` function is being called every single time any property anywhere in the state changes. Mostly these functions are not doing anything computationally expensive and just reading from the state, so it seems fine (although they occasionally are from what I've seen). However, these very frequent Immutable object searches can add up to cause some jank, and a lot of them are doing exactly the same lookups, repeating them multiple times every time any state changes. More interestingly, we don't seem to be taking advantage of the main feature of the ImmutableJS objects which allows for cheap equality comparison at all levels to know if any descendent properties have been modified anywhere underneath that level, rather than more expensive deep comparisons that we are doing at the moment.

Other react / redux projects I've worked on, especially ones even approaching this size, have benefitted from having their mergeProps-like functions only run when a specific portion of the state tree that the specific UI component depends on is changed.

I did some quick measurement to try to see how much time was being 'wasted'. I isolated invocations of mergeProps which do not result in any change props, measured how long that took, and then reported how much time over a 1 second period that the js thread has spent in those functions. This is the report with about 20 tabs open and just moving the mouse over various UI components:
image

I don't know how much of this time we can recover - perhaps a lot of it - but it's perhaps worth investigating.

## Memoize
Using a library like [reselect](https://github.com/reactjs/reselect), each component instance can create memoized functions that explicitly define their dependent values / trees in the state. Their `mergeProps` function will only run if those are not === their previous ImmutableJS  Map / List value. The [reselect + immutablejs pattern](https://github.com/reactjs/reselect#q-how-do-i-use-reselect-with-immutablejs) is common

## Split containers / components (or smart vs dump components)
Have some high level containers connect to redux store and have a mergeProps, then they pass props directly to pure child components which are not connected to the redux store (and have no mergeProps). This would reduce the occurrence of the problem.

Of course, a mixture of both could be best if we think this is an issue. Probably put some kind of structure in place to help define what should be a container and what a component. 

Related immutablejs / redux / selector topics:
- http://redux.js.org/docs/recipes/ComputingDerivedData.html
- http://redux.js.org/docs/recipes/UsingImmutableJS.html
- https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0
- http://www.thegreatcodeadventure.com/the-react-plus-redux-container-pattern/

Would appreciate any feedback from @bsclifton @NejcZdovc or @cezaraugusto as to your opinion of the issue and the suggestions... 😄

petemill:
Fix #11515 

The [reactjs/reselect library](https://github.com/reactjs/reselec) is a common method to memoize the result of functions that read or compute data based on state. It creates functions that are only re-computed when the relevant state changes. It becomes useful when:
 1. A state tree is large
 2. When portions of the state tree are modified often, and not all of the state tree is modified on every action dispatch
3. When **immutablejs** is used to represent state, and therefore partial tree equality comparison oldstate.get('portion1') === newstate.get('portion2') can be used very effectively.

Problems this PR aims to solve
- Time spent recalculating react/redux component instance `mergeProps` functions
  - A high number of components exist that subscribe to state changes
  - A high number of actions are dispatched that change certain relevant state properties
  - A high number of time is spent re-computing state changes where dependent state properties have not changed
  - Many components run the same expensive state-parsing functions within the same render-cycle but re-run those functions nevertheless.

This PR adds the following features:
- Common state accessors that are not dynamic, and should only be re-evaluated when a portion of the state tree changes are turned in to static memoized functions (e.g. in _xxxState.js_)
  - [x] Identify and address some main ones
  - [ ] Create Issues for refactoring some others
- Component instance-specific state accessors that are computed from dynamic properties, e.g. the state of a specific tab, frame, etc, are memoized per-instance use-case
  - [ ] Allow mergeProps to return a function or instance-specific memoized selector
  - [ ] Consider [re-reselect](https://github.com/toomuchdesign/re-reselect) to cache common dynamic selectors (e.g. frames or tabs)
- [ ] Some components are passed data as ownProps, when a child's state is very closely tied to its immediate parent, instead of calculating the same props again, and will benefit from its parent's mergeProp memoization
- [x] Enhances equality comparison for appState / windowState by only merging the two when neccessary

This PR results in the following metric improvements:
...

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/common/state/notificationBarState.js
+++ b/app/common/state/notificationBarState.js
@@ -2,17 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const assert = require('assert')
 const Immutable = require('immutable')
-const {makeImmutable, isMap} = require('./immutableUtil')
+const { createSelector } = require('reselect')
+const frameStateUtil = require('../../../js/state/frameStateUtil')
+const { getOrigin } = require('../../../js/lib/urlutil')
+const { generateWindowStateSelector } = require('./windowState')
 
-const api = {
-  validateState: function (state) {
-    state = makeImmutable(state)
-    assert.ok(isMap(state), 'state must be an Immutable.Map')
-    return state
-  }
-}
+// the portion of the appStore state that concerns notifications
+const selectNotificationState = state => state.get('notifications', Immutable.List())
 
 const notificationBarState = {
   /**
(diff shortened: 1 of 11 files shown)


alexwykoff:
closing in favor of other priorities

PR #12109, opened 2017-11-27
Fix loading of remote PDF files

ltilve:
Remote PDF files were downloaded instead of showed in browser
with pdfJS. Brave rewrites the URL for those resources so that
the pdfJS extension can show it. The regression appeared after
ba8e3f52a94 which was fixing the loading of local PDF files.

The problem was the condition used to filter those PDF files
which was excluding remote PDF files (only local ones where
properly shown).


ltilve:
We are finding that the fix might not be totally correct, as that code was already working in C62 and it seems that it was the switch to PlzNavigate what broke it.

svillar:
Yeah, that code broke with PlzNavigate as we are not getting the LOAD_URL_REQUEST anymore.

Issue #4651, opened 2016-10-10
Brave cannot load PDFs with JS PDF reader in session tabs


pureooze:
**Did you search for similar issues before submitting this one?**
Yes.

**Describe the issue you encountered:**
PDF files do not load using the JS pdf reader when using a session tab. I just get a blank tab when I attempt to load a PDF in a session tab.

**It works fine in non-session and private tabs.**

**Expected behavior:**
PDF should load.
- Platform (Win7, 8, 10? macOS? Linux distro?):
  Linux, OpenSUSE 42.1
- Brave Version:
  0.12.4
- Steps to reproduce:
  1. Open a new session tab
  2. Visit any link that should load a PDF
- Screenshot if needed:
  Session tab (PDF does not load):
  ![image](https://cloud.githubusercontent.com/assets/2337524/19244283/9e62e4a0-8ee9-11e6-8f3b-5ec3171ba8a1.png)

Regular tab (PDF loads):
![image](https://cloud.githubusercontent.com/assets/2337524/19244295/aa16152e-8ee9-11e6-97a7-160e664a54fd.png)

diracdeltas:
confirmed on mac that pdfjs works in private and regular tabs but not session tabs

kjozwiak:
This issue is reproducible in Windows as well. When loading PDF's via session tabs, you'll get the following error within the tab:

> Your file was not found
> 
> It may have been moved or deleted.
> ERR_FILE_NOT_FOUND

* Platform: Win 10 x64 Build: 10.0.15063
* Brave version: 0.18.14, 0.18.20

STR:

- open a new `regular tab` and load https://bitcoin.org/bitcoin.pdf (no issue)
- open a new `private tab` and load https://bitcoin.org/bitcoin.pdf (no issue)
- open a new `session tab` and load https://bitcoin.org/bitcoin.pdf (ERR_FILE_NOT_FOUND)

**Example:**

![pdfsession](https://user-images.githubusercontent.com/2602313/29151082-d1bd787c-7d4c-11e7-92b2-8e680d9160a4.gif)

Issue #2715, opened 2016-07-25
PDFJS returns CORS error when PDF is loaded directly


diracdeltas:
611f61707bfdf6087c62c965db48339d55454a74 works around the fact that a PDF loaded directly in Brave via webview loadUrl causes a CORS error. reloading always fixes the error for me. seems like some kind of race condition with how the origin is being set.
![screen shot 2016-07-25 at 6 20 05 pm](https://cloud.githubusercontent.com/assets/549654/17120923/b660000c-529d-11e6-827e-07ad931421f5.png)

Rob--W:
Why is the commit called "Fix PDFJS MIME type detection"? I tried hard, but couldn't relate that commit description to the code that had been committed.

I didn't look at Brave's source code, but I know that in Chrome there is an issue with granting host permissions (=setting up the origin whitelist in the renderer for the given frame) when an extension is being loaded in a `data:-URL`. For instance, when PDF.js is embedded as follows: `data:text/html,<iframe src="chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/https://bitcoin.org/bitcoin.pdf">`, then the same error as in your screenshot is shown. This is not PDF.js-specific, any extension suffers from it.

diracdeltas:
@Rob--W originally, to work around this CORS issue, I added a shortcut in Brave to bypass PDF.JS's `onHeadersReceived` PDF mime type detection: 00a70f52302406ef862042970605b7dbc5fd7462. This turned out to be what was causing a bunch of 'Malformed PDF' bugs that we were seeing, including #2654. 

The CORS issue is still a mystery, but it seems to always go away when the viewer URL is reloaded, so that's the workaround for now.

diracdeltas:
it's worth noting the CORS error never appears when a PDF is navigated to, such as by clicking a link in the page. it seems to only fire when <webview>.loadURL (http://electron.atom.io/docs/api/web-view-tag/) is called. calling <webview>.reload() fixes it.

diracdeltas:
the `webview.reload()` workaround no longer works as of https://github.com/brave/electron/commit/21383015964b4a693ad5b29327a1629aea

diracdeltas:
workaround fixed in https://github.com/brave/electron/commit/7e61ac2b79b6e5b04cd4a80f5fb231d4dd77d2ee

PR #12138, opened 2017-11-29
move pdf.js loading logic into browser-laptop

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/4651
probably fixes https://github.com/brave/browser-laptop/issues/2715 but ideally the workaround in brave/pdf.js should be removed too
may also fix other PDF loading errors

## Submitter Checklist:

## Test Plan:
1. go to any page that contains PDF links such as https://letsencrypt.org/repository/
2. try to open the PDF links in a new session tab. it should work.

## Reviewer Checklist:

Tests

### DIFF
--- a/app/extensions.js
+++ b/app/extensions.js
@@ -3,7 +3,6 @@ const contextMenus = require('./browser/extensions/contextMenus')
 const extensionActions = require('./common/actions/extensionActions')
 const config = require('../js/constants/config')
 const appConfig = require('../js/constants/appConfig')
-const messages = require('../js/constants/messages')
 const {fileUrl} = require('../js/lib/appUrlUtil')
 const {getExtensionsPath, getBraveExtUrl, getBraveExtIndexHTML} = require('../js/lib/appUrlUtil')
 const {getSetting} = require('../js/settings')
@@ -17,7 +16,7 @@ const fs = require('fs')
 const path = require('path')
 const l10n = require('../js/l10n')
 const {bravifyText} = require('./renderer/lib/extensionsUtil')
-const {ipcMain, componentUpdater, session} = require('electron')
+const {componentUpdater, session} = require('electron')
 
 // Takes Content Security Policy flags, for example { 'default-src': '*' }
 // Returns a CSP string, for example 'default-src: *;'
@@ -394,10 +393,6 @@ module.exports.init = () => {
     }
   })
 
-  ipcMain.on(messages.LOAD_URL_REQUESTED, (e, tabId, url) => {
-    appActions.loadURLRequested(tabId, url)
-  })
-
   process.on('extension-load-error', (error) => {
     console.error(error)
   })
@@ -449,8 +444,26 @@ module.exports.init = () => {
   }
 
   let loadExtension = (extensionId, extensionPath, manifest = {}, manifestLocation = 'unpacked') => {
-    if (extensionId === config.PDFJSExtensionId) {
+    const isPDF = extensionId === config.PDFJSExtensionId
+    if (isPDF) {
       manifestLocation = 'component'
+      // Override the manifest. TODO: Update manifest in pdf.js itself after enough
+      // clients have updated to Brave 0.20.x+
+      manifest = {
+        name: 'PDF Viewer',
+        manifest_version: 2,
+        version: '1.9.457',
+        description: 'Uses HTML5 to display PDF files directly in the browser.',
+        icons: {
+          '128': 'icon128.png',
+          '48': 'icon48.png',
+          '16': 'icon16.png'
+        },
+        permissions: ['storage', '<all_urls>'],
+        content_security_policy: "script-src 'self'; object-src 'self'",
+        incognito: 'split',
+        key: config.PDFJSExtensionPublicKey
+      }
     }
     if (!extensionInfo.isLoaded(extensionId) && !extensionInfo.isLoading(extensionId)) {
       extensionInfo.setState(extensionId, extensionStates.LOADING)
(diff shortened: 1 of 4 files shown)


app/pdfJS.js
```diff
@@ -5,23 +5,102 @@
...
 const UrlUtil = require('../js/lib/urlutil')
 const Filtering = require('./filtering')
-const config = require('../js/constants/config')
 const appActions = require('../js/actions/appActions')
 const getSetting = require('../js/settings').getSetting
 const settings = require('../js/constants/settings')
 
-const pdfjsBaseUrl = `chrome-extension://${config.PDFJSExtensionId}/`
-const viewerBaseUrl = `${pdfjsBaseUrl}content/web/viewer.html`
+const getViewerUrl = UrlUtil.getPDFViewerUrl
```
diracdeltas: most of the changes below are just copied from brave/pdf.js pdfhandler.js and extension-router.js. Those scripts used to run as background scripts in the pdf.js extension but this PR deprecates them.

app/extensions.js
```diff
@@ -394,10 +393,6 @@ module.exports.init = () => {
     }
   })
 
-  ipcMain.on(messages.LOAD_URL_REQUESTED, (e, tabId, url) => {
```
bsclifton: Curious why this was removed? What is/was emitting this?

app/pdfJS.js
```diff
@@ -5,23 +5,102 @@
...
+
+/**
+ * Check if the request is a PDF file.
+ * @param {Object} details First argument of the webRequest.onHeadersReceived
+ *                         event. The properties "responseHeaders" and "url"
+ *                         are read.
+ * @return {boolean} True if the resource is a PDF file.
+ */
+function isPdfFile (details) {
```
bsclifton: To match other places, this (and the function below) might be named with the same casing (ex: `isPDFFile`). If the functions aren't being exported, they could be changed to `isFile` and `isDownloadable` since this file is already named `pdfJS.js`

app/pdfJS.js
```diff
@@ -5,23 +5,102 @@
...
+  var header = details.responseHeaders['Content-Type']
```
bsclifton: nitpick: potentially falsey values could be weeded out to prevent null exception when called. For example:
`var header = details && details.responseHeaders && details.responseHeaders['Content-Type']`

app/extensions.js
```diff
@@ -449,8 +444,26 @@ module.exports.init = () => {
...
+    const isPDF = extensionId === config.PDFJSExtensionId
+    if (isPDF) {
       manifestLocation = 'component'
+      // Override the manifest. TODO: Update manifest in pdf.js itself after enough
+      // clients have updated to Brave 0.20.x+
+      manifest = {
+        name: 'PDF Viewer',
+        manifest_version: 2,
+        version: '1.9.457',
+        description: 'Uses HTML5 to display PDF files directly in the browser.',
```
bsclifton: By having this here, do we lose any translation that was being done? Should this be loading using `locale.translation()`?

app/extensions.js
diracdeltas: @bsclifton Good question! This is actually being emitted in PDF.JS: https://github.com/brave/pdf.js/blob/master/extensions/chromium/pdfHandler.js#L136. However I moved all of that logic into this PR [1], so it no longer needs to be handled in browser-laptop with this PR.

[1] Technically all of this logic is still in the PDF.JS extension since we can't update browser-laptop and PDF.JS at the same time. However the manifest override in line 452 means that once someone updates to this version of browser-laptop, the duplicated code in PDF.JS will no longer run.

app/extensions.js
diracdeltas: @bsclifton I don't think translation would be lost. It was also a literal string in the old manifest: https://github.com/brave/pdf.js/blob/master/extensions/chromium/manifest.json#L5. AFAIK passing a manifest object to the extension-loading function is the same as reading a manifest.json file, so if it was translated before, it would be translated now.

diracdeltas:
@bsclifton ready for re-approval!

app/extensions.js
bsclifton: Ah- perfect 😄  Thanks for the explanation!

bsclifton:
master d1a05f022454c20d1d581396a615a6c22c766291
0.21.x 349a1f04d885c8a2f347a5b0c97e0e888578f596
0.20.x 26abd9966826ef65fba72eb5b7a7de3dbfec526e

test/unit/lib/urlutilTest.js
```diff
@@ -208,6 +208,16 @@ describe('urlutil', function () {
     })
   })
 
+  describe('getPDFViewerUrl', function () {
+    const baseUrl = 'chrome-extension://jdbefljfgobbmcidnmpjamcbhnbphjnb/content/web/viewer.html?file='
+    it('regular url', function () {
+      assert.equal(urlUtil.getPDFViewerUrl('http://example.com'), baseUrl + 'http%3A%2F%2Fexample.com')
+    })
+    it('file url', function () {
+      assert.equal(urlUtil.getPDFViewerUrl('file:///Users/yan/some files/test.pdf'), baseUrl + 'file%3A%2F%2F%2FUsers%2Fyan%2Fsome%20files%2Ftest.pdf')
```
pureooze: Possibly a silly question, but will these work on other machines? Seems to me the path is hardcoded.

bsclifton: @pureooze since the function just builds the path (using the provided input), the actual input itself isn't important 😄  The important part to note is that the output is prefixed with `baseUrl` (and the parts afterword are URL encoded)

Issue #11603, opened 2017-10-19
improve browser resource loading


cezaraugusto:
This is somehow related to #3025 but split as it's for resource loading only

* make use of resource hints (ref: https://www.w3.org/TR/resource-hints/)
* Pre-load common assets (preload/prefetch)
* Cache resources? (needs investigation)
* Audit flow/reflow/repaint (tracked alone in #10027)
* audit use cases for `async`/`defer`
* better code split for webpack (tracked in #11476)
* inline SVG
* Compress images (ref https://github.com/brave/browser-laptop/issues/11476#issuecomment-336950891)
* remove unneeded fonts (mostly FA)
* async module loading? (needs investigation)
* decide when to use inline data: URIs and when to load via separate file (completed with #12225)

PR #12225, opened 2017-12-08
Tighten webpack rules for which assets to inline as data URIs

petemill:
Do not inline our included font files, since there are many small files, which fell under the 10kb limit and were being sent to everyone despite different languages / character ranges covered by each file.
Do inline other small SVG / asset files that fall under the 10kb limit, since they are presumed to be always used by the client for most places they are require'd

Relevant docs: https://github.com/webpack-contrib/url-loader

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -36,14 +36,22 @@ function config () {
           test: /\.css$/,
           loader: 'style-loader!css-loader?-minimize'
         },
-        // Loads font files for Font Awesome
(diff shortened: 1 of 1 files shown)


cezaraugusto:
btw could you reference this in https://github.com/brave/browser-laptop/issues/11603?

PR #12247, opened 2017-12-11
Store promo code in session file

aekeus:
This is required so that requests to the updater will always send the
  initial ref code instead of the ref code from the current browser
  binary.

Auditors: @clifton, @bbondy


bsclifton:
master 65cb44a24a10e260ec5151376d2818f9fa2c45fc
0.21.x 8083d472d7fe020c0055353639d318b1a6e603a5
0.20.x 4e825c26f631bbc8dd70ef4b64f95e3af25ceba2
0.19.x 83f5a1db4a66a93e3345ec4664f8ba2bc7b77bcc

PR #12455, opened 2018-01-01
Gray out label when disabled

### LINKED ISSUE #7634: On disable switch make the text grayed out
## Test plan
See https://github.com/brave/browser-laptop/pull/12455

## Original issue description

- Steps to reproduce:
    1.
    2.
    3.
    
- Actual result:

- Expected result:

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues:

user512:
Fix #7634

## Submitter Checklist:

## Test Plan:
go to about:preferences#plugins without installing Pepper Flash and observe label is gray out.
screen shot 2018-01-01 at 2 27 36 pm

## Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/components/common/switchControl.js
+++ b/app/renderer/components/common/switchControl.js
@@ -131,6 +131,7 @@ class SwitchControl extends ImmutableComponent {
               styles.switchControl__text__label,
               styles.switchControl__text__label_right,
               this.props.small && styles.switchControl__text__label_small,
+              this.props.disabled && styles.switchControl__text_disabled,
               this.props.customStyleTextRight
             )}
               data-l10n-id={this.props.rightl10nId}


app/renderer/components/common/switchControl.js
```diff
@@ -131,7 +131,8 @@ class SwitchControl extends ImmutableComponent {
               styles.switchControl__text__label,
               styles.switchControl__text__label_right,
               this.props.small && styles.switchControl__text__label_small,
-              this.props.customStyleTextRight
+              this.props.customStyleTextRight,
+              this.props.disabled && styles.switchControl__text_disabled
```
luixxiul: @user512 would you mind moving `this.props.customStyleTextRight` under this line as it is expected to override the inherited properties?

app/renderer/components/common/switchControl.js
```diff
@@ -131,6 +131,7 @@ class SwitchControl extends ImmutableComponent {
               styles.switchControl__text__label,
               styles.switchControl__text__label_right,
               this.props.small && styles.switchControl__text__label_small,
+              this.props.disabled && styles.switchControl__text_disabled,
```
user512: @luixxiul, thank you for the feedback.
I am not very familiar with styling in this app or styling in React, would you mind telling me how does order affect styling? I tried reading https://github.com/brave/browser-laptop/blob/master/docs/style.md but can't find how order affect styling.

Would be great if you can suggested some reading about styling in React, thanks 😄

cezaraugusto: hey @user512. The order affects styles the same way as CSS cascading works. The last rule overrides the previous so in your changes, the disabled interface would override other styles, which is the correct and expected behavior. Nice work!

cezaraugusto: maybe this could be the last conditional but I think it's ok as-is. Thanks!

srirambv:
@bsclifton @user512 Is this fix only for Flash switch or all switch in general when off?

bsclifton:
@srirambv this should be for all places where it's used (in a disabled state). The test plan captures the only known occurrence... but if you set the `disabled` field on any future ones, it would work like this too.

srirambv:
@bsclifton none of the switches which are turned off shows gray out. The flash text does it when there is no flash installed on the machine. Otherwise all other text just looks normal.

bsclifton:
@srirambv per the test plan, it should only be grayed out when Flash is not installed on the machine (disabled). Because you can't turn something on if it's not installed 😛 So you just verified this is working as expected 😄 👍

srirambv:
That works but the original issue was intended to show grayed out text when the switch was turned off. So i was expecting to see all text in gray which were turned off. Basically like this https://github.com/brave/browser-laptop/issues/7634#issuecomment-285217620

PR #12664, opened 2018-01-15
logic for dropping a bookmark into toolbar is now inside the store

cezaraugusto:
Test Plan:

```
npm run test -- --grep="APP_ON_DROP_BOOKMARK"
```

1. Should be able to drag from urlbar helper icon (lock/unlock etc) to the bookmarks toolbar
2. Should be able to drag from urlbar helper icon (lock/unlock etc) to a folder in the bookmarks toolbar
3. Should be able to drop a bookmark into folder
4. Should able to remove bookmark from a folder and put into another folder/toolbar
4. Should be able to move (reorder) bookmarks with dnd


app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -50,83 +49,27 @@ class BookmarksToolbar extends React.Component {
...
+          droppedOn.selectedRef.state.isFolder,
+          droppedOn.isDroppedOn,
+          isRightSide
+        )
       }
-      return
+      dnd.onDragEnd()
     }
-
-    const droppedOn = getClosestFromPos(e.clientX, undefined)
```
cezaraugusto: iirc this logic was used to allow links from the webview to be made as bookmarks. Given this doesn't work anymore code was removed. Commit that made this removal is 86c1ac42ff6ef5aa026cbfa3eae5556d44141479 in case we need it back.

per git blame cc @bbondy @NejcZdovc if there's other something I missed

bsclifton:
Given that you removed the `release-blocking` tag from the issue (since the root problem was solved by other issues), I'd like to propose pushing this to 0.21.x

bsclifton:
moving to 0.21.x

cezaraugusto:
@petemill thanks for checking. Since it wasn't working in the main window I didn't check multiple windows. I'll take care of this case.

re https://github.com/brave/browser-laptop/pull/12664#pullrequestreview-89950080 I guess it's a tradeoff between a small perf regression in exchange for the ability to make it testable, as this has regressed a few times. I didn't profile tho but IMO it's a reasonable exchange. Open to suggestions btw.

bsclifton:
Moving to 0.22.x

bsclifton:
Closing as stale. Sorry we never got to this, @cezaraugusto ☹️

PR #12685, opened 2018-01-17
Safe browsing api

jumde:
Test Instructions:

- Open Brave and wait ~2-3 mins (needed to load data files for the first time)
- Verify that data files have been loaded by navigating to ` %DATA_DIR%/Safe Browsing` and check if some `.store` files are present.

Safe Browsing URLs:

1. Open:
- https://testsafebrowsing.appspot.com/s/phishing.html
- https://testsafebrowsing.appspot.com/s/malware.html
- https://testsafebrowsing.appspot.com/s/malware_in_iframe.html
- https://testsafebrowsing.appspot.com/s/unwanted.html
- https://testsafebrowsing.appspot.com/s/image_small.html
- https://testsafebrowsing.appspot.com/s/bad_css.html
- https://testsafebrowsing.appspot.com/s/bad_javascript.html

Verify if appropriate error pages are displayed when these URLs are opened.

Safe Browsing Downloads:

1. Open:
- https://testsafebrowsing.appspot.com/s/content.exe
- https://testsafebrowsing.appspot.com/s/badrep.exe
- https://testsafebrowsing.appspot.com/s/unknown.exe
- https://testsafebrowsing.appspot.com/s/pua.exe

Verify the downloads are blocked.

Enable/Disable Safe Browsing for all sites:

1. Navigate to "about:preferences" and Shields > Select `Basic Safe Browsing`
2. Navigate to any of the above URLs to confirm Google safe browsing is disabled.
3. Navigate to `freehacks.us` to verify `Basic Safe Browsing` still works.
4. Navigate to "about:preferences" and Shields > Select `Disable Safe Browsing`
5. Navigate to any of the `testsafebrowsing` URLs to verify Google Safe Browsing is disabled
6. Navigate to `freehacks.us` to verify `Basic Safe Browsing` works.

cc: @luixxiul - Test instructions updated. Let me know if you have any questions.


app/extensions/brave/locales/en-GB/downloads.properties
```diff
@@ -1,6 +1,7 @@
 downloads=Downloads
 noDownloads=There are no downloads
 
+downloadSafeBrowsingBlocked=Blocked by Safe Browsing
```
diracdeltas: fyi you only need to update these in en-US

js/constants/downloadStates.js
```diff
@@ -13,7 +13,8 @@ const downloadStates = {
   COMPLETED: _,
   CANCELLED: _,
   INTERRUPTED: _,
-  UNAUTHORIZED: _
+  UNAUTHORIZED: _,
+  SAFEBROWSINGBLOCKED: _
```
diracdeltas: please change this to `SAFE_BROWSING_BLOCKED`

app/extensions/brave/locales/en-US/bravery.properties
```diff
@@ -25,7 +25,12 @@ httpsEverywhere=HTTPS Everywhere
 noScript=Block Scripts
 noScriptPref=Block Scripts (will break many sites)
 safeBrowsing=Block Phishing / Malware
+safeBrowsingControl=Block Phishing / Malware *
+advancedSafeBrowsing=Advanced Safe Browsing
+advancedSafeBrowsingInfo=* Advanced safe browsing supplements Brave's phishing and malware protection with Google Safe Browsing.
```
diracdeltas: cc @flamsmark for thoughts on wording

js/state/contentSettings.js
```diff
@@ -355,7 +355,8 @@ const updateUserPrefs = (appState, appConfig, isPrivate = false) => {
   const braveryDefaults = siteSettings.braveryDefaults(appState, appConfig)
   const defaultUserPrefs = getDefaultUserPrefContentSettings(braveryDefaults, appState, appConfig)
 
-  return { 'content_settings': getSettingsFromSiteSettings(defaultUserPrefs, appState, appConfig, isPrivate) }
+  return { 'content_settings': getSettingsFromSiteSettings(defaultUserPrefs, appState, appConfig, isPrivate)
+  }
```
diracdeltas: unneeded formatting change?

js/constants/appConfig.js
```diff
@@ -78,6 +79,9 @@ module.exports = {
     msBetweenRechecks: 1000 * 60 * 60 * 2, // 2 hours
     enabled: true
   },
+  safeBrowsingAll: {
+    enabled: true
```
diracdeltas: does this correspond to enabling both brave and google safebrowsing? shouldn't it default to false?

js/state/syncUtil.js
```diff
@@ -43,7 +43,7 @@ module.exports.siteSettingDefaults = {
   shieldsUp: true,
   adControl: 1,
   cookieControl: 0,
-  safeBrowsing: true,
+  safeBrowsingControl: 0,
```
diracdeltas: this will require changes to the brave/sync library and cross-platform sync implementations. i think we should avoid changes to sync for now, since other platforms don't support Google safebrowsing yet.

diracdeltas: FYI the sync changes would be needed in https://github.com/brave/sync/blob/staging/lib/api.proto#L68

js/state/siteSettings.js
```diff
@@ -68,6 +70,8 @@ module.exports.activeSettings = (siteSettings, appState, appConfig) => {
     })()
   })
 
+  setUserPref('safebrowsing.enabled', settings['safeBrowsing'] && settings['safeBrowsingAll'])
```
diracdeltas: i think this should be called in the reducers for when settings/siteSettings changed, instead of every time that `activeSettings` is called.

see  `userPrefsUpdateTrigger` in `js/state/contentSettings.js` for an example

diracdeltas:
please add any new properties or property changes to `docs/state.md`

it would be great to add tests for the new setting. see `test/bravery-components/safebrowsingTest.js`

js/constants/appConfig.js
jumde: safeBrowsingAll - Google + Brave Safe Browsing
safeBrowsing - Brave Safe Browsing

I was thinking `safeBrowsingAll` will be the `safeBrowsingControl` by default.

diracdeltas: i thought we decided safeBrowsing (without Google) would be the default, but cc @flamsmark

app/filtering.js
```diff
@@ -548,6 +548,10 @@ function registerPermissionHandler (session, partition) {
 function updateDownloadState (win, downloadId, item, state) {
   updateElectronDownloadItem(win, downloadId, item, state)
 
+  if (win) {
+    win.webContents.send(messages.SHOW_DOWNLOADS_TOOLBAR)
```
bridiver: why was this added?

app/extensions/brave/locales/en-US/bravery.properties
flamsmark: This wording is okay, but I don't think it really teases out the differences.

`````
+safeBrowsingControl=Block Phishing and Malware *
+advancedSafeBrowsing=Protection plus Google Safe Browsing 
+advancedSafeBrowsingInfo=* You can supplement Brave's phishing and malware protection with Google's Safe Browsing service.
+basicSafeBrowsing=Phishing and malware protection
+disableSafeBrowsing=No phishing and malware protection
`````

File: js/constants/appConfig.js (hunk 5 above)
Comment (flamsmark) [reply js/constants/appConfig.js]: I think we should use both by default. If it's not the default, I can't imagine how many people would turn it on. And this is a protection which most benefits those least likely to wrangle their settings.

Comment (diracdeltas) [reply js/constants/appConfig.js]: alright defaulting to both sgtm

File: app/browser/reducers/siteSettingsReducer.js [hunk 9]
```diff
@@ -79,6 +81,33 @@ const siteSettingsReducer = (state, action, immutableAction) => {
...
+          (parsedUrl.protocol === 'https:' || parsedUrl.protocol === 'http:')) {
+          let ruleKey = `https?://${parsedUrl.host}`
+          let currentSettings = currentSiteSetting.get(ruleKey)
+          if (currentSettings !== undefined) {
+            siteSafeBrowsing = currentSettings.get('safeBrowsing')
+          }
+        }
+        siteSafeBrowsing = siteSafeBrowsing === undefined ? true : siteSafeBrowsing
+
+        setUserPref('safebrowsing.enabled',
```
bridiver: You can't turn a global setting on/off per tab. I don't think it makes sense or is even possible in the current implementation to have site settings for safe browsing. With this implementation it could turn off for a tab loading in the background because the foreground tab has it disabled

js/state/siteSettings.js
```diff
@@ -81,6 +87,18 @@ module.exports.activeSettings = (siteSettings, appState, appConfig) => {
     return defaults.adControl
   })()
 
+  settings.safeBrowsingControl = (() => {
+    if (settings.shieldsUp === false) {
```
bridiver: @diracdeltas should safe browsing turn off with shields? Safe browsing seems independent of that to me.

jumde: Updated!

diracdeltas:
Currently `about:safebrowsing` > `Advanced` says: 

> You can access this site by clicking the Brave icon in the top right corner and temporarily disabling 'Block Phishing / Malware'. This is not recommended!

If we are taking safebrowsing out of shields, then this should be changed to instruct the user to disable safebrowsing globally in `about:preferences#shields`

js/state/syncUtil.js
jumde: https://github.com/brave/sync/pull/198

app/extensions/brave/locales/en-US/app.properties
```diff
@@ -207,9 +207,9 @@ runInsecureContentWarning=This page is trying to load scripts from insecure sour
 runtimeUpdatePending=A runtime update which requires the application to be restarted is available.
 safebrowsingError=Dangerous Site Blocked
 safebrowsingErrorAdvanced=Advanced
-safebrowsingErrorBypass=You can access this site by clicking the Brave icon in the top right corner and temporarily disabling 'Block Phishing / Malware'. This is not recommended!
+safebrowsingErrorBypass=You can access this site by navigating to 'about:preferences#shields' and disabling 'Block Phishing/Malware'. This is not recommended!
 safebrowsingErrorHideAdvanced=Hide advanced
-safebrowsingErrorInfo=Brave uses crowdsourced blacklists from Spam404 and Disconnect.me to block dangerous domains. If you believe this domain should not be blocked, Please report the site at https://community.brave.com/c/site-blocking
+safebrowsingErrorInfo=Brave uses crowdsourced blacklists from Google Safe Browsing, Spam404 and Disconnect.me to block dangerous domains. If you believe this domain should not be blocked, Please report the site at https://community.brave.com/c/site-blocking
```
diracdeltas: Does this text show up if Google safebrowsing is enabled, or only if our own safebrowsing implementation is enabled?

js/state/siteSettings.js
```diff
@@ -81,6 +87,15 @@ module.exports.activeSettings = (siteSettings, appState, appConfig) => {
     return defaults.adControl
   })()
 
+  settings.safeBrowsingControl = (() => {
```
diracdeltas: do we need this since safebrowsing is no longer a site setting?

js/state/syncUtil.js
diracdeltas: per comment on the sync PR, i think we should just stub `safeBrowsing` to be `undefined` for sync.

app/extensions/brave/locales/en-US/app.properties
jumde: This text will only show up if `Brave Safe Browsing` is enabled.

alexwykoff:
@jumde @bridiver could you update on the status of this? any outstanding issues?

bsclifton:
Closing as I believe we're no longer implementing this in Muon

Issue #12673, opened 2018-01-17
add the ability to change the default monthly budget via server side


kjozwiak:
### Description

As per our discussions earlier today, we'll need a way to change the default monthly budget values via server side rather than client side. Currently, whenever we need to change the default monthly budget, we require new builds which is expensive in terms of QA/Dev resources. Having the ability to change the default monthly budget via the server will save us a lot of time.

@mrose17 I believe you mentioned that you'll add some information for @NejcZdovc so he could get started on this ASAP so we can hit the timeline of releasing `0.19.x Hotfix 13` on Frirday.

### Steps to Reproduce

N/A (test cases will be added when feature is complete and ready for QA)

**Actual result:**

Changing the default monthly budget requires code changes on the client which requires new builds.

**Expected result:**

Changes will be pushed via server side and won't require new builds.

**Reproduces how often:**

N/A

### Brave Version

**about:brave info:**

> Brave: 0.19.136 - Reproducible 
> rev: 1659471cf1b1016568f63a6ab3e35b604d7e3db3 
> Muon: 4.7.1 
> libchromiumcontent: 64.0.3282.85 

**Reproducible on current live release:**

Yes, the issue/limitation also exists under `0.19.134`.

> Brave: 0.19.134 - Reproducible 
> rev: 316cfa522ffe6a558fe9f2f3f299c0c602dfb691 
> Muon: 4.5.38 
> libchromiumcontent: 63.0.3239.132 

### Additional Information

CCing @mrose17 @evq @davidtemkin  @NejcZdovc @bsclifton

PR #12686, opened 2018-01-17
Make it easier to debug the ledger when using a balance server

mrose17:
set BALANCE_SERVER_URL

Part of a fix for https://github.com/brave/browser-laptop/issues/12673


package.json
```diff
@@ -25,6 +25,7 @@
     "electron-rebuild": "electron-rebuild",
     "flow": "flow; test $? -eq 0 -o $? -eq 2",
     "lint": "standard --verbose | snazzy",
+    "lint-standard": "standard",
```
NejcZdovc: why do we need another standard linter command?

mrose17: because the `lint-stanadrd` command generates output so that it can be parsed by programs that will position the editor to each error. when you do the `--verbose | snazzy` thing, you lose that.

NejcZdovc: ok for me it's working with normal lint as well

mrose17: you probably don't use emacs

NejcZdovc:
@mrose17 please create issue for this PR and what should be the target for it. I am guessing we are needing this for dynamic amounts right? so hotfix 13?

mrose17:
@NejcZdovc - done.

package.json
NejcZdovc: no, using vsc 😃

NejcZdovc:
master 5f7756f
0.21 e442d8eae
0.20 bb9b3b48e
0.19  2758bdaae

PR #12795, opened 2018-01-23
Mitigating CVE-2018-1000006

jumde:
https://electronjs.org/blog/protocol-handler-fix
resolves  https://github.com/brave/browser-laptop/issues/12797

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -25,6 +25,7 @@ const UpdateStatus = require('../js/constants/updateStatus')
 const settings = require('../js/constants/settings')
 const siteTags = require('../js/constants/siteTags')
 const downloadStates = require('../js/constants/downloadStates')
+const {defaultProtocols} = require('../js/constants/appConfig')
 
 // State
 const tabState = require('./common/state/tabState')
@@ -874,8 +875,10 @@ const protocolHandlerMigration = (immutableData) => {
     if (platformUtil.isWindows() &&
       immutableData.getIn(['settings', settings.IS_DEFAULT_BROWSER])) {
       // Update the protocol handler to be safe
-      app.setAsDefaultProtocolClient('http', ['--'])
-      app.setAsDefaultProtocolClient('https', ['--'])
+      defaultProtocols.forEach((protocol) => {
+        app.removeAsDefaultProtocolClient(protocol)
+        app.setAsDefaultProtocolClient(protocol, undefined, ['--'])
+      })
     }
     immutableData = immutableData.set('defaultProtocolMigration', true)
   }
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -868,8 +868,23 @@ const fingerprintingProtectionMigration = (immutableData) => {
   return immutableData
 }
 
+const protocolHandlerMigration = (immutableData) => {
+  // Fix https://github.com/brave/browser-laptop/issues/12797
+  if (!immutableData.get('defaultProtocolMigration')) {
+    if (platformUtil.isWindows() &&
+      immutableData.getIn(['settings', settings.IS_DEFAULT_BROWSER])) {
+      // Update the protocol handler to be safe
+      app.setAsDefaultProtocolClient('http', ['--'])
+      app.setAsDefaultProtocolClient('https', ['--'])
+    }
+    immutableData = immutableData.set('defaultProtocolMigration', true)
+  }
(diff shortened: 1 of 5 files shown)


js/stores/appStore.js
```diff
@@ -496,9 +496,10 @@ const handleAppAction = (action) => {
         if (platformUtil.isLinux()) {
           const Channel = require('../../app/channel')
           const desktopName = Channel.getLinuxDesktopName()
+          var args = [desktopName, "--"];
           for (const p of defaultProtocols) {
-            app.setAsDefaultProtocolClient(p, desktopName)
-            app.setAsDefaultProtocolClient('', desktopName)
+            app.setAsDefaultProtocolClient(p, args)
```
darkdh: it should be `setAsDefaultProtocolClient(p, desktopName, ["--"])`, desktopName is not part of args

darkdh: my bad, Linux implementation only reads desktop name so no need to add extra args
the args that needs -- is https://github.com/brave/browser-laptop/pull/12795/files#diff-23ca389e2bcb77191b5a9c10900eb3a3R507

darkdh: https://github.com/brave/muon/blob/80442e401dd3f78d55f78daf2f5f2d8505970296/atom/browser/browser_linux.cc#L92

diracdeltas: yeah this is not needed for linux afaict

js/stores/appStore.js
```diff
@@ -496,14 +496,15 @@ const handleAppAction = (action) => {
...
+            app.setAsDefaultProtocolClient('', args)
           }
           isDefaultBrowser = app.isDefaultProtocolClient('', desktopName)
         } else {
           for (const p of defaultProtocols) {
-            app.setAsDefaultProtocolClient(p)
+            app.setAsDefaultProtocolClient(p, ["--"])
```
diracdeltas: i think you need single quotes for the linter to pass

package.json
```diff
@@ -218,7 +218,6 @@
     "plugins": [
       "flowtype"
     ],
-    "parser": "babel-eslint",
```
diracdeltas: was this not supposed to be removed?

diracdeltas:
updated PR to do the protocol handler migration on first launch

js/stores/appStore.js
```diff
@@ -503,7 +503,7 @@ const handleAppAction = (action) => {
           isDefaultBrowser = app.isDefaultProtocolClient('', desktopName)
         } else {
           for (const p of defaultProtocols) {
-            app.setAsDefaultProtocolClient(p)
+            app.setAsDefaultProtocolClient(p, ['--'])
```
diracdeltas: doesn't the 2nd argument (`path`) need to be set? https://electronjs.org/docs/api/app#appsetasdefaultprotocolclientprotocol-path-args-macos-windows

darkdh: it uses default value `Defaults to process.execPath`

darkdh: we should add `['--']` just for Windows just like `protocolHandlerMigration` did

app/sessionStore.js
```diff
@@ -868,8 +869,25 @@ const fingerprintingProtectionMigration = (immutableData) => {
...
 
+const protocolHandlerMigration = (immutableData) => {
+  // Fix https://github.com/brave/browser-laptop/issues/12797
+  if (!immutableData.get('defaultProtocolMigration')) {
+    if (platformUtil.isWindows() &&
+      immutableData.getIn(['settings', settings.IS_DEFAULT_BROWSER])) {
+      // Update the protocol handler to be safe
+      defaultProtocols.forEach((protocol) => {
+        app.removeAsDefaultProtocolClient(protocol)
+        app.setAsDefaultProtocolClient(protocol, undefined, ['--'])
```
darkdh: function signature here is not right `app.setAsDefaultProtocolClient(protocol, [path, args])`
https://github.com/brave/muon/blob/80442e401dd3f78d55f78daf2f5f2d8505970296/atom/browser/browser.h#L84

darkdh:
https://github.com/brave/browser-laptop/blob/59159138e141e32c26959fc2276d5822f3ffa282/res/dev/braveDefaults.nsi#L66
has already appended `--` when we do packaging
but we still need muon changes to address the issue

Issue #12632, opened 2018-01-12
Privacy issue: Browser connects to third party hosts in the background.


ghost:
### Description
Brave browser shows non privacy respecting behavior disregarding user settings.

### Steps to Reproduce

Downloading Brave for the first time, never used before.

1. `wget https://laptop-updates.brave.com/latest/linux64 -O- | tar xj`
2. Turn off any additional network services such as ntpd and similar (to prevent "poluting" the tcpdump monitoring with external data)
3. Reboot and login to Plasma desktop (openSUSE Leap 42.3). Do not start any applications.
4. In a console run:
`# tcpdump -i eth1 -l > /tmp/tcpdump.log & tail -f /tmp/tcpdump.log`

**Actual result:**

Lots of communication not explicitly initiated by the user with hosts like:
```
amazonaws.com
herokudns.com
analytics.brave.com
mercury.basicattentiontoken.org
qwant.com
twitter.com
yandex.com
ocsp.digicert.com
phicdn.net
fastly.net
```

**Expected result:**
No packets should be sent without user initiating that.

**Reproduces how often:**
Always

### Brave Version
```
Brave: 0.19.131 
rev: da2f34787e51eb44190a7a77249e100fb2368306 
Muon: 4.5.36 
libchromiumcontent: 63.0.3239.132 
V8: 6.3.292.49 
Node.js: 7.9.0 
Update Channel: Release 
OS Platform: Linux 
OS Release: 4.4.104-39-default 
OS Architecture: x64
```

### Additional Information

Further testing:

5. Changing settings in an attempt to harden privacy:

* about:preferences#general

Brave starts with: a new tab
A new tab shows: Blank page
My home page is: about:blank
Enable Spell Check: OFF

* about:preferences#search

Search engine: DDG.
Use DuckDuckGo by default for search in Private Tabs: ON

* about:preferences#security

Password Manager: Don't manage my passwords
Enable Autofill: OFF
Autoplay Media: Always deny (was 'Always allow')
Send a 'Do Not Track' header with browsing requests *: ON
Enhance security by loading each site in its own process (experimental) *: ON
Autoplay Media(Clear all): cleared

* about:preferences#extensions

Disable all (PDF Viewer and Torrent Viewer are enabled by default). There are 8 extensions total, none of them seems possible to uninstall.

* about:preferences#shields

Cookie Control: Block all cookies
Block Scripts (will break many sites): ON
Fingerprinting Protection (may break some sites): ON

* about:adblock

[...]

PR #12798, opened 2018-01-23
Adds referral code read into browser

### LINKED ISSUE #12789: Read promo code from profile directory and set AppState
- There should be an argument passed during first run (on Windows there is; not sure about macOS after it's changed to PKG format). 
- During this first run, we should check for the presence of the `promoCode` file inside the user's profile directory (ex: `~/Library/Application Support/brave/promoCode`)
- Once file is detected and if code is valid, we need to set App State:
`state.setIn(['updates', 'promoCode'], promoCode)`

### LINKED ISSUE #12790: Contact promo API on first install and obtain download_id
**NOTE: https://github.com/brave/browser-laptop/issues/12789 must be completed first**
**NOTE: https://github.com/brave/browser-laptop/issues/12788 must be completed first**

During first run after install, the promo API needs to be contacted.

Example input/output from endpoint:
```
Input {
  "api_key": "key",
  "referral_code": "1234",
  "platform": "winx64"
}

Output {
  "download_id": "45f84b3d-6162-4565-ab44-dbed30f3ba6e",
  "referral_code": "1234"
}
```

The `download_id` returned should be stored in AppState

NejcZdovc:
Submitter Checklist:

Resolves #12789
Resolves #12790

Test Plan:
1) clean profile
2) create empty folder `brave-development` and create file `promoCode` with `123456` inside
3) start browser `LEDGER_ENVIRONMENT=staging npm run start`

Reviewer Checklist:

Tests


app/browser/api/ledger.js
```diff
@@ -81,6 +83,7 @@ let verifiedTimeoutId = false
 let v2RulesetDB
 const v2RulesetPath = 'ledger-rulesV2.leveldb'
 const statePath = 'ledger-state.json'
+const referralCodePath = 'promoCode/code'
```
petemill: Please update to simply 'promoCode', or even better use the function for reading first-run promoCode from the module I included in #12920

NejcZdovc: will change to your read when #12920 is merged

diracdeltas:
please request a security review for this using the usual process when it's ready, thanks

app/browser/api/ledger.js

app/browser/api/ledger.js
```diff
@@ -1852,7 +1868,63 @@ const onCallback = (state, result, delayTime) => {
...
+
+  if (code.length > 2) {
+    module.exports.roundtrip({
+      server: referralServer,
+      method: 'PUT',
+      path: '/promo/initialize/nonua',
+      payload: {
+        api_key: referralAPI,
+        referral_code: code,
+        platform: 'winx64' // TODO add correct platform
```
bbondy: ^ seems blocking

app/browser/api/ledger.js
```diff
@@ -1875,6 +1947,19 @@ const initialize = (state, paymentsEnabled) => {
     }
   }
 
+  if (!ledgerState.getInfoProp(state, 'referralTimestamp')) {
+    fs = require('fs')
```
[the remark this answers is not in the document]
NejcZdovc: fs is defined locally, so not sure if there is any scenario where fs would exists?

app/browser/api/ledger.js

NejcZdovc:
don't merge before #12920 is merged

app/browser/api/ledger.js
```diff
@@ -95,6 +99,27 @@ const clientOptions = {
...
+  'linux': 'linux'
+}
+
+let platform = platforms[process.platform]
+
+if (process.platform === 'win32') {
+  platform = platforms[process.platform + process.arch]
+}
+
+let referralServer = 'https://laptop-updates-staging.herokuapp.com'
```
diracdeltas: Privacy note: We are trying to not add any more 3rd party requests in Brave (https://github.com/brave/browser-laptop/issues/12632). For production, please make sure this endpoint is a brave.com subdomain, assigning a CNAME if needed.

cc @aekeus

app/browser/api/ledger.js
```diff
@@ -95,6 +99,27 @@ const clientOptions = {
...
+let referralAPI = 'key'
+
+if (clientOptions.environment === 'production') {
+  referralServer = 'https://laptop-updates.brave.com'
```
diracdeltas: ++, i think this addresses the comment above

app/browser/api/ledger.js
```diff
@@ -1852,7 +1881,63 @@ const onCallback = (state, result, delayTime) => {
...
+    return
+  }
+
+  if (!code) {
+    return
+  }
+
+  code = code.trim()
+
+  if (code.length > 2) {
```
diracdeltas: i thought the code was always going to be 6 characters? or can it be less?

app/browser/api/ledger.js
```diff
@@ -2627,6 +2725,46 @@ const onPublisherTimestamp = (state, oldTimestamp, newTimestamp) => {
...
+  if (body && body.finalized) {
+    appActions.onReferralActivity()
+    return
+  }
+
+  if (clientOptions.verboseP) {
+    console.log('Referral is still not ready, please wait at least 30days')
+  }
+
+  updater.checkForUpdate(false, true)
```
diracdeltas: minor: `activityRoundTrip` and `checkReferralActivity` would be clearer if they only handled things related to referral activity instead of also checking for browser updates. instead of returning after `appActions.checkReferralActivity()` in `updater.checkForUpdate`, you could always check for the browser update so that these functions here don't need to check for browser updates.

app/updater.js
```diff
@@ -212,8 +214,21 @@ var downloadHandler = (err, metadata) => {
...
-  const updateStatus = AppStore.getState().getIn(['updates', 'status'])
+exports.checkForUpdate = (verbose, skipReferral = false) => {
+  const state = AppStore.getState()
+
+  // check for referral 30 days
+  if (!skipReferral && !ledgerState.getInfoProp(state, 'referralTimestamp')) {
+    const installTime = state.get('firstRunTimestamp')
+    const month = ledgerUtil.milliseconds.day * 30
+
+    if (installTime + month < new Date().getTime()) {
```
diracdeltas: this function will check for referral activity every 30 days. i thought it was only supposed to check referral activity once after 30 days and then never check again.

docs/state.md
```diff
@@ -217,8 +217,9 @@ AppStore
         EUR: number,
         USD: number
       },
-      reconcileFrequency: number // duration between each reconciliation in days
+      reconcileFrequency: number, // duration between each reconciliation in days
       reconcileStamp: number,  // timestamp for the next reconcilation
+      referralTimestamp: number,
```
diracdeltas: please document this

docs/state.md
```diff
@@ -600,6 +601,7 @@ AppStore
       name: string, // name of the update
       notes: string // release notes for the active update
     },
+    downloadId: string,
```
diracdeltas: please document this

app/browser/api/ledger.js
```diff
@@ -1875,6 +1960,19 @@ const initialize = (state, paymentsEnabled) => {
     }
   }
 
+  if (!ledgerState.getInfoProp(state, 'referralTimestamp')) {
+    fs = require('fs')
+    const codePath = pathName(referralCodePath)
+
+    fs.access(codePath, fs.FF_OK, (err) => {
+      if (err) {
```
diracdeltas: this should set a flag in session state so that the referralCodePath check doesn't run every time a user starts Brave if we already know that they don't have a referral code. i guess you could just set `referralTimestamp` to a special value like 'none'.

app/browser/api/ledger.js
```diff
@@ -1875,6 +1960,19 @@ const initialize = (state, paymentsEnabled) => {
     }
   }
 
+  if (!ledgerState.getInfoProp(state, 'referralTimestamp')) {
```
diracdeltas: shouldn't this check if `downloadId` exists instead of `referralTimestamp`? IIUC, `referralTimestamp` is only set when the activity check is called, after 30 days, whereas `downloadId` is set as soon as the referral is initialized.

app/browser/api/ledger.js

app/browser/api/ledger.js
NejcZdovc: we want to reduce amount of calls, so we only want to call update server when referral fails.

app/updater.js
NejcZdovc: server can be in a different zone or not aligned with the browsers, so we need to check until referral is successful. We have `referralTimestamp` above which will not trigger this checks anymore as soon as we get green light from the server. We can add amount of attempts after which we will just stop checking?

docs/state.md
NejcZdovc: done

docs/state.md
NejcZdovc: done

app/browser/api/ledger.js
NejcZdovc: you are correct, fixed

app/browser/api/ledger.js
NejcZdovc: we have check above `if (!updateState.getUpdateProp(state, 'downloadId')) {`

diracdeltas: @NejcZdovc but isn't `!updateState.getUpdateProp(state, 'downloadId')` always true for a user who didn't install from a referral? i mean that this code shouldn't keep running on every startup for those users.

app/updater.js
diracdeltas: Where is `referralTimestamp` set to a false value after getting a green light from the server?

NejcZdovc: we don't set it to false, we actually set value (timestamp) to it in the ledger reducer

app/browser/api/ledger.js
NejcZdovc: that's correct, good catch

NejcZdovc: But how would we know that we don't need to check for referral code?

NejcZdovc: this file check is the only indicator that we have to know if install is from referral or not

NejcZdovc: on thing that I can do is to set referral flag to false after we fail the first time. Because this should be only checked once that should be fine. wdyt?

NejcZdovc: @diracdeltas

NejcZdovc: ok did implement it b93829db3d5eaf7b0803625f225b87f5ac2b4748

NejcZdovc: Implemented logic: we only check for that file once. And if we get error when preforming `fs.access` we will not check again

app/browser/api/ledger.js
```diff
@@ -1852,7 +1881,63 @@ const onCallback = (state, result, delayTime) => {
...
+    return
+  }
+
+  if (!code) {
+    return
+  }
+
+  code = code.trim()
+
+  if (code.length > 0) {
```
diracdeltas: not a big deal but why was this changed back to `> 0` from `=== 6`?

NejcZdovc: because server is currently expecting code 1234 for testing, I change it to this

petemill:
0.22.x f5eefc227
0.21.x b03ab808c
0.20.x 111b59b35

PR #12806, opened 2018-01-23
fix topsites never returning after exclusion

### LINKED ISSUE #10411: Removing top site tiles is buggy
# Test Plan
    1. Visit few sites
    2. Open the dashboard page in a new tab.
    3. Click the "X" button to remove any page tile.
    4. The tile should be removed immediately.
# 
Yes.

When I try to remove top site tiles on the dashboard page, the tile is not removed immediately. It sometimes gets removed later randomly, when I navigate to another website, open a new tab, etc. Sometimes it does not get removed even after trying multiple things. Refreshing the dashboard page does not help.

macOS 10.12.x

0.18.21 (263b6d5). I've had this problem for several major versions.

- Steps to reproduce:
    1. Visit any site.
    2. Open the dashboard page in a new tab.
    3. Click the "X" button to remove any page tile.
    
- Actual result:
The tile is not removed immediately.

- Expected result:
The tile should be removed immediately.

Not sure.

I think I've had the problem throughout 0.18.x, so yes.

Almost all the time.

- Extra QA steps:
    1. Visit any site.
    2. Open the dashboard page in a new tab.
    3. Click the "X" button to remove any page tile.
    4. Make sure the tile is removed immediately.

- Screenshot if needed:
None.

- Any related issues:
None.

cezaraugusto:
cc @LaurenWags 

fix #10411

STR:

1. Exclude a top site
2. With 5 tiles only, visit a new website
3. Website tile should be shown with a total of 6 tiles in newtab page

Please note that we debounce all newtab page data so the sixth tile will be shown only after a new site visit, even if you have other sites on the list.

### DIFF
--- a/app/browser/api/topSites.js
+++ b/app/browser/api/topSites.js
@@ -28,6 +28,10 @@ const isPinned = (state, siteKey) => {
   })
 }
 
+const isIgnored = (state, siteKey) => {
+  return aboutNewTabState.getIgnoredTopSites(state).includes(siteKey)
+}
+
 const sortCountDescending = (left, right) => {
   const leftCount = left.get('count', 0)
   const rightCount = right.get('count', 0)
@@ -85,6 +89,7 @@ const getTopSiteData = () => {
   let sites = historyState.getSites(state)
     .filter((site, key) => !isSourceAboutUrl(site.get('location')) &&
       !isPinned(state, key) &&
+      !isIgnored(state, key) &&
       (minCountOfTopSites === undefined || (site.get('count') || 0) >= minCountOfTopSites) &&
       (minAccessOfTopSites === undefined || (site.get('lastAccessedTime') || 0) >= minAccessOfTopSites)
     )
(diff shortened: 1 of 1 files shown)


app/browser/api/topSites.js
```diff
@@ -85,6 +89,7 @@ const getTopSiteData = () => {
   let sites = historyState.getSites(state)
     .filter((site, key) => !isSourceAboutUrl(site.get('location')) &&
       !isPinned(state, key) &&
+      !isIgnored(state, site.get('key')) &&
```
NejcZdovc: can we use just `key` here same as for `isPinned`?

cezaraugusto: ok done

bsclifton:
master 4e6cbdd25e987345d7d2848c64a053c0124cb20c
0.21.x 57abcd3af3434cd7a82033f6918ed2b6264432a2
0.20.x 08680411ed75754e1562a79523937bb115733dec

PR #12832, opened 2018-01-24
Add an option for "no delay" in tab preview timing

### LINKED ISSUE #8860: Make tab preview based on mouse idle time

cezaraugusto:
closes #8860

this is a follow-up of #9887 and adds a _none_ option for tab preview delay.

Test Plan:

1. Go to pref->tabs
2. Confirm tab preview defaults to "short" 
3. Change tab preview timing to "none"
4. tab previews should be instant


app/renderer/components/tabs/tabs.js
```diff
@@ -195,6 +196,7 @@ class Tabs extends React.Component {
                 key={'tab-' + frameKey}
                 ref={(node) => this.tabRefs.push(node)}
                 frameKey={frameKey}
+                tabsPerTabPage={this.props.tabsPerTabPage}
```
petemill: Wondering why a new prop has been added here, but no change to the Tab component to utilize it. Is this fixing a different bug?

cezaraugusto: oh my bad this was for another pr thanks for spotting

cezaraugusto:
@petemill ya I like _instant_ and also suggested _immediate_ but I guess @bradleyrichter is searching for a better phrase for the select title.

re timing I'm ok changing to 40 btw but can't really spot the difference. Below 30ms I can repro the unwanted flash so I'd go 40-50ms

Jacalz:
It is getting late in Sweden so I will  veck this one out early tomorrow 🙂

I was thinking of names and maby it could be called “Classic” it something because it is the old style 😅

js/constants/appConfig.js
```diff
@@ -153,7 +153,7 @@ module.exports = {
     'tabs.tabs-per-page': 20,
     'tabs.close-action': 'parent',
     'tabs.show-tab-previews': true,
-    'tabs.preview-timing': 1000,
+    'tabs.preview-timing': tabPreviewTiming.SHORT,
```
bsclifton: nice!! Love this 😄

petemill:
master 16740dc
0.21.x 9ef9571f84a391671757c61111fc511841f2d3ba 4b91dfe15be436c9ec55ba81cc966245bcdf4953
0.20.x 9ef9571f84a391671757c61111fc511841f2d3ba 4b91dfe15be436c9ec55ba81cc966245bcdf4953

Jacalz:
Can’t get the browser to build because I reinstalled my computer 🖥Going to need to wait for a new beta build from @clifton 🙂

PR #12963, opened 2018-01-31
Prevent cmd-w or ctrl-w closing a pinned tab.

### LINKED ISSUE #1563: pinned tabs should be ignored in tab-closure sequence using control+W
current behavior works like chrome and is fine for pinned tabs that only exist in one window at a time.

The prefered behavior should mimic Safari on Mac which also has pinned tabs on every browser window. Those pins are ignored so they are not cycled when closing the window with Command+W.

NHDaly:
Change the `CmdOrCtrl+w` accelerator to check whether a tab is pinned before sending `tabCloseRequested`, and do nothing otherwise.

This will allow closing a pinned tab via any other method to continue to work normally.

Fixes #1563.

Submitter Checklist:

Test Plan:
Still unsure!

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/menu.js
+++ b/app/browser/menu.js
@@ -85,7 +85,12 @@ const createFileSubmenu = () => {
       label: locale.translation('closeTab'),
       accelerator: 'CmdOrCtrl+W',
       click: function () {
-        appActions.tabCloseRequested(tabState.TAB_ID_ACTIVE)
+        const tabId = tabState.TAB_ID_ACTIVE
+        // Do not allow closing a pinned tab.
+        const isPinned = tabState.isTabPinned(state, tabId)
+        if (!isPinned) {
+          appActions.tabCloseRequested(tabId)
+        }
       }
     }, {
       // This should be disabled when no windows are active.


NHDaly:
I never got a respone on the original issue (#1563), which is why i haven't moved forward with any more of these required steps. Maybe i'd be better off asking the questions here:

--------------------

My questions:
1. Is this basically the right fix? Is this even a thing people want, and if so, did I implement it correctly? (It's at least a thing that *I* want. Mistakenly closing pinned tabs is hugely frustrating for me.)

2. Implementation question: 
"Oh, actually, after looking at it more, I think maybe this would prevent closing a pinned tab from the file menu as well... i don't think that's what we want, is it? Perhaps there's a way to see how the menuItem was invoked (ie via an accelerator or through clicking)?"
--> Does anyone know how i can fix that?

app/browser/menu.js
```diff
@@ -83,7 +83,12 @@ const createFileSubmenu = () => {
       label: locale.translation('closeTab'),
       accelerator: 'CmdOrCtrl+W',
       click: function () {
-        appActions.tabCloseRequested(tabState.TAB_ID_ACTIVE)
+        const tabId = tabState.TAB_ID_ACTIVE
+        // Do not allow closing a pinned tab.
+        const isPinned = tabState.isTabPinned(state, tabId)
+        if (!isPinned) {
+          appActions.tabCloseRequested(tabId)
+        }
```
NHDaly: This might also prevent closing through the File menu. I don't think that's desired.

is there a way to see if this was invoked via the keyboard accelerator or via the file menu?

bsclifton:
@NHDaly apologies that we never got around to reviewing this ☹️

PR #12968, opened 2018-02-01
clear old pinned items stored in wrong position

### LINKED ISSUE #12941: top site tiles not being populated when updating profile from 0.19.147
## Test plan

https://github.com/brave/browser-laptop/pull/12968#issue-293390058

----

### Description
I have a profile that I was using with version 0.19.147 which encountered issue https://github.com/brave/browser-laptop/issues/12027 (to the degree that I only have one (default, pinned) top site tile). When this profile was updated to 0.20.27 and I began to browse, I noticed my top site tiles were not being populated.

### Steps to Reproduce

  1. Have a profile from 0.19.x that is missing top site tiles due to https://github.com/brave/browser-laptop/issues/12027
  2. Update profile to 0.20.27.
  3. Open a new tab, note you still have the one (default, pinned) top site tile.
  4. Browse normally (visit some sites in new tabs, etc). 
  5. Open another new tab.

**Actual result:**
No new top site tiles added, only the default, pinned top site tile is displayed:
screen shot 2018-01-30 at 6 14 57 pm

**Expected result:**
Additional top site tiles should appear based on your browsing.

**Reproduces how often:**
Easily

### Brave Version

**about:brave info:**
Brave | 0.20.27
V8 | 6.4.388.40
rev | 76102c5
Muon | 4.7.9
OS Release | 16.7.0
Update Channel | Release
OS Architecture | x64
OS Platform | macOS
Node.js | 7.9.0
Brave Sync | v1.4.2
libchromiumcontent | 64.0.3282.119

n/a

### Additional Information
Missed when testing https://github.com/brave/browser-laptop/issues/12027

cezaraugusto:
fix #12941 

TL;DR: The bug is caused because pinned top sites were stored in a wrong position -- instead of own index they were pushed to the first index, causing duplicates. Since duplicated keys are hidden you ended up having only one tile.

The fix cleans pinned top sites from the previous version.

Test Plan:

1. checkout 0.19.x
2. Visit some websites, restart browser a couple times until some tiles start disappearing, you'll reach a point where only twitter is shown
3. Switch to 0.20.x  
4. Top sites should be populated with sites you visited in step 1

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -847,6 +847,16 @@ module.exports.runPreMigrations = (data) => {
       if (data.cache) {
         delete data.cache.bookmarkLocation
       }
+
+      // pinned top sites were stored in the wrong position in 0.19.x
+      // allowing duplicated items. See #12941
+      // in this case eliminate pinned items so they can be properly
+      // populated in their own indexes
+      if (data.about.newtab.pinnedTopSites) {
+        // Empty array is currently set to include default pinned sites
+        // which we avoid given the user already have a profile
+        data.about.newtab.pinnedTopSites = [null]
+      }
     }
   }
 


app/sessionStore.js
```diff
@@ -847,6 +847,16 @@ module.exports.runPreMigrations = (data) => {
...
       }
+
+      // pinned top sites were stored in the wrong position in 0.19.x
+      // allowing duplicated items. See #12941
+      // in this case eliminate pinned items so they can be properly
+      // populated in their own indexes
+      if (data.about.newtab.pinnedTopSites) {
+        // Empty array is currently set to include default pinned sites
+        // which we avoid given the user already have a profile
+        data.about.newtab.pinnedTopSites = [null]
```
petemill: How come we're setting this to an array with one `null` item inside and not an empty array with `[]` @cezaraugusto ?

cezaraugusto: see https://github.com/brave/browser-laptop/blob/master/app/common/state/aboutNewTabState.js#L41 an empty array would populate top sites with default pinned site, which should only happen on an empty profile

cezaraugusto: could be done with something like `Array.from({length: 18}, _ => null)` as pinned top sites are set as 18 null spaces but conversion is covered already in both newTab and newTabState

petemill:
0.22.x 940436347
0.21.x 5bc903fba
0.20.x 9f7fe12ef

PR #12977, opened 2018-02-01
separate rules for whether a tab can be created discarded and whether it can be autoDiscardable

### LINKED ISSUE #12976: Tabs are not loading discarded, even when requested to
### Description
https://github.com/brave/browser-laptop/issues/12916 was mitigating a bug with discarding tabs (automatically) which had already been loaded, and causing them to not be able to be switched to (#10673).

However, it also unintentionally results in startup tabs not being created 'unloaded' (or 'discarded'), meaning they all load simultaneously. This does not need to be the case as tabs during startup are not affected by #10673.

### Steps to Reproduce

  1. Create many tabs (regular http / https sites)
  2. Quit Brave
  3. Load Brave again

**Actual result:**
Tabs are all loaded

**Expected result:**
Only active tab is loaded

**Reproduces how often:**
100%

### Brave Version

**about:brave info:**
0.20

petemill:
Fix #12976

Submitter Checklist:

Test Plan:

Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -12,7 +12,7 @@ const tabState = require('../common/state/tabState')
 const {app, BrowserWindow, extensions, session, ipcMain} = require('electron')
 const {makeImmutable, makeJS} = require('../common/state/immutableUtil')
 const {getTargetAboutUrl, getSourceAboutUrl, isSourceAboutUrl, newFrameUrl, isTargetAboutUrl, isIntermediateAboutPage, isTargetMagnetUrl, getSourceMagnetUrl} = require('../../js/lib/appUrlUtil')
-const {isURL, getUrlFromInput, toPDFJSLocation, getDefaultFaviconUrl, getLocationIfPDF} = require('../../js/lib/urlutil')
+const {isURL, getUrlFromInput, toPDFJSLocation, getDefaultFaviconUrl, isHttpOrHttps, getLocationIfPDF} = require('../../js/lib/urlutil')
 const {isSessionPartition} = require('../../js/state/frameStateUtil')
 const {getOrigin} = require('../../js/lib/urlutil')
 const settingsStore = require('../../js/settings')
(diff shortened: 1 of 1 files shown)


app/browser/tabs.js
```diff
@@ -912,8 +912,16 @@ const api = {
...
-      if (!isAutoDiscardable(createProperties)) {
+
+      // tabs are allowed to be initially discarded (unloaded) if they are regular tabs
+      // i.e not about: pages, and not pinned
+      // but the tab creation request can determine that ultimately by passing in
+      // `discarded: true`
+      if (createProperties.pinned || !isHttpOrHttps(createProperties.url)) {
         createProperties.discarded = false
+      }
+      // similarly, autoDiscardable will happen for regular tabs (not about: tabs)
```
bridiver: should have a note that this is temporarily disabled and link to the issue

petemill: it's still in the separate function, perhaps I should just put that logic inline here too since there's only 1 usage and it's 1 line. thoughts?

bridiver: That seems fine. This is the problem with these github reviews sometimes because important code is often not visible. The original issue with `isAutoDiscardable` was missed for a similar reason

petemill: @bridiver I inlined it for now. Normally I like the functional / semantic language of bundling that logic out, but it was causing confusion in this case.

app/browser/tabs.js
```diff
@@ -112,12 +112,6 @@ const needsPartitionAssigned = (createProperties) => {
     createProperties.partition
 }
 
-const isAutoDiscardable = (createProperties) => {
-  // Temporarily disabled due to bad handling for auto discarded tabs
```
bridiver: I know we can get it from the history and I should have brought this up with the original PR, but I think we should keep this method and preserve the original logic, but comment it out for now and replace with `false`

petemill:
Addressed @bridiver feedback and made a separation in defined logic between wheter a tab is 'auto-discardable' and whether a tab can be 'lazy-loaded'. Also moved logic inline to `tabs.create:` fn since this module is very large and is maybe more readable to put this all in one place to avoid anymore future confusion.

petemill:
master 8314998
0.22.x 82877c228823cfa85516bf93ce8152eeb7585dd4
0.21.x 9dc1f00cc6fe59486f13a322e934a47d67652f76
0.20.x d53e4381a0dd9d62565ba07192c3f893104b226a

PR #12998, opened 2018-02-02
Usermodel

NejcZdovc:
**Known issues**

- [ ] don't display notification if there is no window opened
- [ ] notification on windows is still visible only for 5s (time is not take in to account)
- [ ] notification is not opened if window is minimized/in the background


js/constants/appConfig.js
```diff
@@ -180,6 +180,8 @@ module.exports = {
     'sync.type.history': false,
     'sync.type.siteSetting': true,
     'general.downloads.default-save-path': null,
+    // BATads
+    'BATads.enabled': true, // need to default it to off for merge into master
```
NejcZdovc: switch to false, before merge!!

locklin: OK, will do. 
I'm assuming the purpose of this PR is to ultimately merge back into main branch?
I apologize for my ignorance; most of my career I'm out working on my own repo, which eventually gets used by clients.

NejcZdovc: Yes purpose of this PR is to track progress, add comments for others and prepare code for the review. When review is done and code (PR) is approved we will merge it to the appropriate version.

mandar-brave:
Bug#1: On clicking "view" button from an ad, Brave opened three tabs to the landing page URL
image

mandar-brave:
bug#2 - saw ad; clicked view. Ad went away but tab did not open (the ad was related to a brave promotion - got it after seeing a brave site

mandar-brave:
bug#3 opened 5 tabs on 3rd ad (when clicked)
image

mandar-brave:
Bug# Clicked 4th ad and opened 6 windows (there is a clear bug re: ads clicking - increased clicking is opening more tabs) -- will stop filing for now

lukemulks:
Windows 10 x64 bit
Clicked on a notification and 3 tabs launched with destination URL. 

Pretty sure I only clicked once on the notification, but will attempt to repro.

diracdeltas:
reminder that this will definitely need a security review; please follow the usual process

NejcZdovc:
@diracdeltas will do

app/browser/api/userModel.js
```diff
@@ -115,11 +115,11 @@ function randomKey (dictionary) {
   return keys[keys.length * Math.random() << 0]
 }
 
-const goAheadAndShowTheAd = (windowId, categoryName, notificationText, notificationUrl) => {
+const goAheadAndShowTheAd = (windowId, advertiser, notificationText, notificationUrl) => {
```
mrose17: shouldn't we just call this `notificationTitle` ?

mrose17: i think from the browser's perspective: it needs a title, text, and url for the notification.

it doesn't really matter if the title was a category yesterday, an advertiser today, or something else tomorrow..

.nsprc
```diff
@@ -6,6 +6,7 @@
     "https://nodesecurity.io/advisories/534",
     "https://nodesecurity.io/advisories/535",
     "https://nodesecurity.io/advisories/550",
-    "https://nodesecurity.io/advisories/532"
+    "https://nodesecurity.io/advisories/532",
+    "https://nodesecurity.io/advisories/146"
```
NejcZdovc: We need to remove this one before merging into the master

package.json
```diff
@@ -93,10 +94,12 @@
     "bat-publisher": "^2.0.15",
     "bignumber.js": "^4.0.4",
     "bloodhound-js": "brave/bloodhound",
+    "brave-node-notifier": "brave-intl/brave-node-notifier",
     "clipboard-copy": "^1.0.0",
     "compare-versions": "^3.0.1",
     "date-fns": "^1.29.0",
     "deepmerge": "^2.0.1",
+    "detect-ssid": "^1.0.3",
```
diracdeltas: why is this needed?

package.json
```diff
@@ -130,6 +133,7 @@
     "string.prototype.startswith": "^0.2.0",
     "tablesort": "5.0.2",
     "tldjs": "^2.3.1",
+    "truncate-url": "^1.0.0",
```
diracdeltas: appears not to be used

app/browser/api/userModel.js
```diff
@@ -0,0 +1,325 @@
...
+  return keys[keys.length * Math.random() << 0]
+}
+
+const goAheadAndShowTheAd = (windowId, notificationTitle, notificationText, notificationUrl) => {
+  appActions.nativeNotificationCreate(
+    windowId,
+    {
+      title: notificationTitle,
+      message: notificationText,
+      icon: path.join(__dirname, '../../../img/BAT_icon.png'),
```
diracdeltas: are you sure this works in packaged builds?

package.json
diracdeltas: to clarify, i see that it's needed to set the current SSID, but i'm wondering why this is a product requirement.

mrose17: the original "ask" from the folks at research was to report the SSID, the IP address and the geolocation for each event. i changed this so that the browser would detect the SSID internally, and allow the user to optionally select a tag from a static list to go along with it. that tag would be reported instead of the three things in the original ask.

the list of tags has entries like "home", "work", "transit", etc., which are both generic and user-selected. i suspect that this is much less a privacy risk than the data requested in the original "ask"...

diracdeltas: @mrose17 thanks, i will check in with them. is this just for a test group of users, or will it be shipped out in Brave stable?

mrose17: my understanding is that it is only for test groups. when differential privacy is added, perhaps it will return; however, i have no data on that.

app/browser/api/userModel.js
```diff
@@ -71,10 +71,16 @@ const generateAdReportingEvent = (state, eventType, action) => {
     case 'load':
       {
         const tabValue = action.get('tabValue')
-        map.tabId = tabValue.get('tabId')
-        map.tabType = 'general'
-        map.tabUrl = tabValue.get('url')
-        map.tabClassification = lastSingleClassification
+        const tabUrl = tabValue.get('url')
+
+        if (!tabUrl.startsWith('http://')) {
```
mrose17: @kevinlawler - either http: or https:

app/adBlock.js
```diff
@@ -47,7 +49,8 @@ const startAdBlocking = (adblock, resourceName, shouldCheckMainFrame) => {
     const firstPartyUrl = urlParse(mainFrameUrl)
     const url = urlParse(details.url)
     const cancel = shouldDoAdBlockCheck(details.resourceType, firstPartyUrl, url, shouldCheckMainFrame) &&
-      adblock.matches(details.url, mapFilterType[details.resourceType], firstPartyUrl.host)
+      adblock.matches(details.url, mapFilterType[details.resourceType], firstPartyUrl.host) &&
+      !braveExemptHosts.includes(url.hostname)
```
mrose17: thank you!

.nsprc
diracdeltas: @NejcZdovc i thought this was going to be maintained as a separate branch, not merged into master?

app/browser/api/userModel.js
mrose17: i believe so, we are seeing it work now in the current RC under private test.

.nsprc
mrose17: it still is in the `usermodel` branch, which we regularly rebase from master to stay current with "the latest and greatest", however, afaik, none of the `usermodel` branch has gone into master.

app/browser/api/userModel.js
```diff
@@ -0,0 +1,752 @@
...
+
+let testingP = (process.env.NODE_ENV === 'test') || (process.env.LEDGER_VERBOSE === 'true')
+const oneDay = (testingP ? 600 : 86400) * 1000
+const oneHour = (testingP ? 25 : 3600) * 1000
+const hackStagingOn = true
+const roundTripOptions = {
+  debugP: process.env.LEDGER_DEBUG === 'true',
+  loggingP: process.env.LEDGER_LOGGING === 'true',
+  verboseP: process.env.LEDGER_VERBOSE === 'true',
+  server: url.parse('https://' + (hackStagingOn || testingP ? 'collector-staging.brave.com' : 'collector.brave.com'))
```
diracdeltas: Use urlParse instead of `url.parse`

mrose17: thanks!

app/browser/api/userModel.js
```diff
@@ -0,0 +1,752 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+const um = require('@brave-intl/bat-usermodel')
+const elph = require('@brave-intl/bat-elph')
+const braveNotifier = require('brave-node-notifier')
+const path = require('path')
+const getSSID = require('detect-ssid')
+const underscore = require('underscore')
+const url = require('url')
```
diracdeltas: replace this with `const urlFormat = require('url').format` so that url.parse is not callable (which is insecure and should be always replaced with our urlParse module)

mrose17: ok.

js/state/contentSettings.js
```diff
@@ -118,6 +118,10 @@ const getDefaultUserPrefContentSettings = (braveryDefaults, appSettings, appConf
       setting: getSetting(settings.METAMASK_PROMPT_DISMISSED) || getSetting(settings.METAMASK_ENABLED) ? 'block' : 'allow',
       primaryPattern: '*'
     }],
+    BATads: [{
+      setting: 'allow',
```
diracdeltas: this needs to be set to 'block' if the user hasn't opted into ad trials

app/browser/reducers/userModelReducer.js
```diff
@@ -46,16 +46,16 @@ const userModelReducer = (state, action, immutableAction) => {
         const stat = changeInfo && changeInfo.get('status')
         const complete = stat === 'complete'
 
-        if (complete) {
-          state = userModel.generateAdReportingEvent(state, 'load', action)
-        }
-
         const tabValue = action.get('tabValue')
+        if ((tabValue) && (tabValue.get('incongnito') === true)) {
+          appActions.onUserModelLog('Private Tab', { url: tabValue.get('url') })
```
NejcZdovc: We should not log private tab url's

app/browser/reducers/userModelReducer.js
```diff
@@ -76,10 +76,16 @@ const userModelReducer = (state, action, immutableAction) => {
...
         const tabId = action.get('tabId')
-        const tab = tabState.getByTabId(state, tabId)
-        if (tab == null) break
+        const tabValue = tabState.getByTabId(state, tabId)
 
-        const url = tab.get('url')
+        if (tabValue == null) break
+
+        if (tabValue.get('incognito') === true) {
+          appActions.onUserModelLog('Private Tab', { url: tabValue.get('url') })
```
NejcZdovc: Same here

app/browser/reducers/userModelReducer.js
```diff
@@ -99,13 +105,19 @@ const userModelReducer = (state, action, immutableAction) => {
...
       {
         const tabId = action.get('tabId')
-        const tab = tabState.getByTabId(state, tabId)
-        if (tab == null) break
+        const tabValue = tabState.getByTabId(state, tabId)
+
+        if (tabValue == null) break
+
+        if (tabValue.get('incognito') === true) {
+          appActions.onUserModelLog('Private Tab', { url: tabValue.get('url') })
```
NejcZdovc: Same here

js/state/contentSettings.js
NejcZdovc: fixed

app/common/state/userModelState.js
```diff
@@ -149,6 +149,18 @@ const userModelState = {
     return respectsHourLimit && respectsDayLimit
   },
 
+  elphAppendLetter: (state, letter) => {
+    state = validateState(state)
+    let tmp = state.getIn(['userModel', 'elphstring'])
```
NejcZdovc: Please add all this state fields into `state.md` as well

bsclifton:
Closing PR as I don't believe there's any reason we need to keep it open. Let's re-submit once it's ready

NejcZdovc:
We use this PR to track commits and comments about the code. This helps us with rebases as well

PR #13110, opened 2018-02-12
Extension data no longer removed in clear cookies…

ryanml:
… action. Called method is now cookie specific.

Fixes Issue #13098

## Submitter Checklist:

## Test Plan:

Test plan specified in issue:

1. Configure Brave to clear all site cookies on close in about:preferences#security.
2. Enable and sign-into the LastPass extension
3. Restart Brave
4. Lastpass should not prompt for credentials

## Reviewer Checklist:

Tests

Previously, there was one very general method being called when "clear all site cookies" was enabled called clearStorageData(). This deleted: appcache, cookies, filesystem, indexdb, local storage,shadercache, websql, and serviceworkers. This was way too aggressive. 

Per https://github.com/brave/muon/blob/master/atom/browser/api/atom_api_session.cc#L84

We can see that specific storage types can be specified, the only one we should need to touch for this option is "cookies"

In the filtering module, specific clearCache() and clearHistory() methods exist. I created an pseudo alias of the clearStorageData() method called clearCookies() that is called instead. bound to the clearStorageData() call is `{'storages': ['cookies']}` which specifies the cookies storage type.

See: https://github.com/electron/electron/blob/master/docs/api/session.md#sesclearstoragedataoptions-callback

### DIFF
--- a/app/renderer/components/main/clearBrowsingDataPanel.js
+++ b/app/renderer/components/main/clearBrowsingDataPanel.js
@@ -48,11 +48,9 @@ class ClearBrowsingDataPanel extends React.Component {
     this.onHide()
 
     if (
-      (this.props.browserHistory &&
-      this.props.allSiteCookiesNoLocalStorage &&
-      this.props.cachedImagesAndFiles) ||
-      this.props.allSiteCookies
-
+      this.props.browserHistory &&
+      (this.props.allSiteCookiesNoLocalStorage || this.props.allSiteCookies) &&
+      this.props.cachedImagesAndFiles
     ) {
       ipc.send(messages.PREFS_RESTART)
     }


ryanml:
cc: @srirambv @jonathansampson @bsclifton

bsclifton:
@ryanml quick nit-pick on the PR titles; could we keep the title as what is being fixed? and then (in the body), we can share what it fixes (so that it's easy to click + follow link to issue) 😄

Thanks for submitting this, you've been doing a ton of great patches lately 😄 👍

ryanml:
@bsclifton no problem :) I'll keep this in mind going forward.

My pleasure! This is a fantastic project and it's fun to work on.

jonathansampson:
I'm a little torn on the solution here; while it addresses the issue, I'm not sure that it addresses the original intent of the feature. Brave often says "cookies" when it makes _client data_. For instance, enabling _3rd Party Cookies_ in the Site Shields actually enables both cookies and localStorage. I think the intent here is to give the user an option to remove all client data when Brave shuts down.

@diracdeltas, can you share your thoughts here?

ryanml:
@jonathansampson makes sense given the definition of cookies here. Is this something that makes more sense to solve via redefinition or maybe easier yet, the addition of another data selection when a user is prompted to clear data? (Ex: All Stored Data, which would tie to the original clearStorageData function. 

cc: @bsclifton @diracdeltas

jonathansampson:
@ryanml I think your PR here is good, but reveals our need for an additional option. Or, this reveals our need to be able to distinguish between site-data and extension-data, and give the user control over one or both at shutdown.

ryanml:
@jonathansampson Got it. Sounds like this PR should probably be tabled until a few others give their thoughts on it as well. I like your approach that the user should be able to distinguish between the two in the controls, and it would have my vote if it comes down to it. Thanks!

ryanml:
@diracdeltas thanks for your review :) I think the changes you requested are in line with what @jonathansampson mentioned with distinguishing those two controls by adding an additional one.

While waiting to see what @bridiver thinks, I'm going to pre-implement those changes and push them to this branch if their thoughts are in line with the solutions in this thread

ryanml:
I only have one concern, and I'm sure it can be addressed with some historical knowledge with how a settings change has been handled in the past. 

I can see why it is beneficial for a user who has "enable clear cookies on shutdown" enabled to continue to have it enabled and garner the old expected behavior of all storage data cleared.

Would it look a little odd in the codebase if we keep the settings constants the same but switch out the labels? I imagine it would cause a little confusion. Again, if there is an existing migration plan in place for something like this my concern may be invalid :) cc: @jonathansampson @diracdeltas

diracdeltas:
@ryanml you're right that the naming in the codebase would be confusing since the setting is currently called `shutdown.clear-all-site-cookies`. i would keep that as-is for simplicity and just call the new method ``shutdown.clear-cookies-no-localstorage` or something, with a comment explaining it.

ryanml:
@diracdeltas Sounds good and thanks for the clarification :)

ryanml:
Given the consensus above about having two distinguishing settings for clearing cookies and all local storage, I've decided to push my changes to this PR as they should provide a better basis for further review.

Summary of the changes in [my commit](https://github.com/brave/browser-laptop/pull/13110/commits/f84f00827cf4296d2247d6f621abbb5454e2d90f):

1. The previous setting of 'allSiteCookies' has been redefined as a "Clear all stored site data" option. This is for the purpose or retaining the original behavior of the setting, so as to not disturb users' previous settings upon migration. This ties back in the original filtering method of clearStorageData
2. An additional setting 'allSiteCookiesNoLocalStorage' has been created, which is now labeled "Clear cookies". This ties in to the new filtering method "clearCookies" and does only such.
(Both of the settings retain their respective shutdown actions in security preferences.

There was one change in logic that made sense for me to make [here:](https://github.com/brave/browser-laptop/pull/13110/commits/f84f00827cf4296d2247d6f621abbb5454e2d90f#diff-61bcd61c5d3195f216c4fdf636fff62fL49)

Previously, the criteria for triggering a browser restart were that the browser history, cached images and files, and clear cookies switches were enabled for clearing history. 

As the 'allSiteCookies' property now clears all local storage, it makes sense that if that is checked at all, a restart should be prompted. The original behavior of browser history, cached images and files, and clear cookies being enabled triggering a browser restart is changed. 

I am sure there will be some feedback on the new naming (I took diracdeltas suggestion pretty literally). I think these changes will provide a good basis for further modifications and review going forward :)

cc: @jonathansampson @diracdeltas @bsclifton

app/renderer/components/main/clearBrowsingDataPanel.js
```diff
@@ -47,9 +48,11 @@ class ClearBrowsingDataPanel extends React.Component {
     this.onHide()
 
     if (
-      this.props.allSiteCookies &&
-      this.props.browserHistory &&
-      this.props.cachedImagesAndFiles
+      (this.props.browserHistory &&
+      this.props.allSiteCookiesNoLocalStorage &&
+      this.props.cachedImagesAndFiles) ||
+      this.props.allSiteCookies
```
diracdeltas: this should be 

```
this.props.browserHistory &&
      (this.props.allSiteCookiesNoLocalStorage || this.props.allSiteCookies) &&
      this.props.cachedImagesAndFiles
```

ryanml: Per this request I've pushed a change 👍

ryanml:
@diracdeltas is everything else with this round looking pretty okay to you? I think that unless labels and naming needs to be adjusted we are getting pretty close :)

diracdeltas:
sorry for the delay. this approach looks good to me!

ryanml:
Thanks @diracdeltas ! :D

ryanml:
See #13490

PR #13120, opened 2018-02-13
Fix tab auto-discarding ability

### LINKED ISSUE #1796: feature suggestion: tab unloading from memory when not in use
from #470 

> Problem is Chromium is a memory eater so it makes sense to have an option which makes tabs only loaded into memory when actually clicking on them, or the opposite when a tab is longer time not in use that it is unloaded out of memory until being reactivated again with clicking on it.

### LINKED ISSUE #13210: Tab content should be discarded when memory usage is too high
## Description
Muon / Chromium has a feature whereby a tab's contents will be disposed and the memory freed when under memory pressure.

This was previously enabled, but disabled in #12916 due to #10673.

However, performance is hindered by not having tabs discarded, so we should do two things:
1. Fix tab discarding
   a. The ability to switch to and away from discarded tabs
   b. The ability to detach previously-discarded tabs
2. Re-enable automatic discarding

This issue tracks both the re-enabling of the feature and the fixing of the underlying issues which led to the feature being disabled.

## Prerequisites

- [x] TabState is not updated for the new discarded WebContents. This happens in `add-new-contents` event which isn't received for discarded replacement WebContents. Therefore if the window has 2 tabs, one being discarded, the state reports 1 tab and the detach will not be allowed.
**Status**: Issue https://github.com/brave/muon/issues/487. Muon tab event `tab-replaced-at` implemented in https://github.com/brave/muon/pull/504

- [x] When a discarded tab is detached, it seems to be destroyed immediately, and the 'did-detach' event never fires. Therefore the handler which adds the frame to the new window never fires. Even if the handler is forced to run, the tab in the new window is dead, since it has been destroyed.
**Status**: Issue https://github.com/brave/muon/issues/496. Fix in https://github.com/brave/muon/pull/504

- [ ] When a discarded tab is detached, attached to a new window, and loaded, it does not remember its scroll position. The new webview has a src element (unlike when detaching and attaching non-discarded tabs) even though browser-laptop is not setting it, but does use the existing tabId.

- [ ] New muon version with above changes integrated with browser-laptop

## Steps to Reproduce
The problems with tab discarding can be viewed with the new Debug menu option **'Allow manual tab discarding'**
**Please turn off the preference for tab previews whilst testing!**

### Tab discarding works

  1. Open some tabs with websites (not about:... pages)
  2. Visit each tab to ensure it is loaded
  3. Right-click an inactive tab and choose **'Discard'**
  4. Switch to the discarded tab

#### Actual result:
Cannot switch to tab

#### Expected result:
Tab is made active and tab's contents are loaded again

### Tab detaching works

  1. Open some tabs with websites (not about:... pages)
  2. Visit each tab to ensure it is loaded
  3. Right-click an inactive tab and choose **'Discard'**
  6. Right-click the same tab and choose detach (be careful not to hover over tab and cause a tab-preview as this will load the tab contents)

#### Actual result:
Tab either disappears or attaches to a new window but is blank and not loadable / useable.

#### Expected result:
Tab detaches to a new window with its contents intact.

### Tab detaching after reloading after discarding works 😮 

  1. Open some tabs with websites (not about:... pages)
  2. Visit each tab to ensure it is loaded
  3. Right-click an inactive tab and choose **'Discard'**
  4. Switch to the discarded tab to load its contents
  5. Switch away from the discarded tab
  6. Right-click the same tab and choose detach

#### Actual result:
Tab either disappears or attaches to a new window but is blank and not loadable / useable.

#### Expected result:
Tab detaches to a new window with its contents intact.

More test conditions not neccessarily connected to this issue, but there because of some changed code, are specified in the PR #13120 

## Brave Version

**about:brave info:**
0.21.x source

No, since automatic discarding is disabled

petemill:
Fix #13210 _(13120 / 13210 - what are the chances...)_
Fix #1796
~Depends on #13118 (new Debug menu for manually discarding a tab)~

**~Depends on new muon version with brave/muon#504~**

## Blocking issues
As described in #13210:
 - ~https://github.com/brave/muon/issues/487~
 - ~https://github.com/brave/muon/issues/496~

## What this fixes
Ensure that tab/WebContents objects are added to memory as soon as they are created.

Use muon's new `tab-replaced-at` event (instead of <webview>.tab-id-changed) to ensure state gets updated correctly when a tab is discarded, as well as communicating temporary contents replacement to a window's frame (such as when detaching a tab to a new window, the tab in the old window gets a temporary WebContents that then gets destroyed).

In the case of tabs which get discarded, they are replaced with a fresh, clean WebContents. We were not receiving this new object, so when the renderer asked for this new object's tabId to be made active, the cache did not have that object.

We must also handle updating references to the old TabId to the new TabId in state, in the module which remembers opener tab IDs and the module which handles historical tab 'active' history for a window.

The guestInstanceId for a tab also changes when a discarded tab is reloaded, so make sure we pass that new value in the frame options when a discarded tab is detached to a new window.

Adds more relevant log entries under the `--debug-tab-events` flag.

## Testing
Verification steps specified on #13210

And in addition, some modified tests from https://github.com/brave/browser-laptop/pull/12193, since now a tab ID can change, and we must update history for which tabs were active with the new tab ID.
Here those modified tests are:

**_NOTE: on Windows, you should disable tab preview_**

#### Parent tab
- Preferences: default (When closing an active tab, select its parent tab)
- Visit a site
- Cmd-click 5 links to open 'child' tabs from that site
- Switch active focus to the 3rd tab
- Right-click the 1st tab and choose Discard
- Close the 3rd (current) tab
- Expected: Tab active focus switches to the 1st tab (parent)

#### Parent tab 2
- Preferences: default (When closing an active tab, select its parent tab)
- Visit a site
- Cmd-click 5 links to open 'child' tabs from that site
- Right-click the 3rd tab and choose Discard
- Switch active focus to the 3rd tab, so it loads
- Close the 3rd (current) tab
- Expected: Tab active focus switches to the 1st tab (parent)

#### Next tab
- Select the last tab
- Right-click the previously-active 1st tab (new tab page) and choose Discard
- Switch active focus back to that 1st tab, so it loads
- Right-click -> Discard **all** the other tabs (except the current active tab)
- Close the (active) new tab page
- Expected: Tab active focus switches to the tab that was 2nd, and now is 1st
- Preferences: When closing an active tab Select the next tab
- Open the new tab page
- Cmd-click all the top sites icons to open 'child' tabs
- Switch active focus to the 3rd tab
- Close the 3rd (current) tab
- Expected: Tab active focus switches to the tab that was 4th, and is now 3rd

#### Last active tab
- Preferences: When closing an active tab Select the last viewed tab
- Open at least 5 tabs
- Make the 1st tab active
- Make the 3rd tab active
- Make the last tab active
- Right-click -> Discard all the tabs (except the current active tab)
- Close the currently-active tab (the last tab)
- Expected: Tab active focus switches to the 3rd tab
- Close the currently-active tab (the 3rd tab)
- Expected: Tab active focus switches to the 1st tab
- Close the currently-active tab (the 1st tab)
- Expected: Tab active focus switches to the 2nd tab

#### Parent tab
- Preferences: detault (When closing an active tab, select its parent tab)
- Open the new tab page
- Cmd-click all the top sites icons to open 'child' tabs
- Switch active focus to the 3rd tab
- Right-click / drag -> Detach the 3rd (current) tab
- Expected: Tab active focus switches to the 1st tab (new tab page)

#### Next tab
- Right-click / drag -> Detach the (active) new tab page
- Expected: Tab active focus switches to the tab that was 2nd, and now is 1st
- Preferences: When closing an active tab Select the next tab
- Open the new tab page
- Cmd-click all the top sites icons to open 'child' tabs
- Switch active focus to the 3rd tab
- Right-click / drag -> Detach the 3rd (current) tab
- Expected: Tab active focus switches to the tab that was 4th, and is now 3rd

#### Last active tab
- Preferences: When closing an active tab Select the last viewed tab
- Open at least 5 tabs
- Make the 1st tab active
- Make the 3rd tab active
- Make the last tab active
- Right-click / drag -> Detach the currently-active tab (the last tab)
- Expected: Tab active focus switches to the 3rd tab
- Right-click / drag -> Detach the currently-active tab (the 3rd tab)
- Expected: Tab active focus switches to the 1st tab
- Right-click / drag -> Detach the currently-active tab (the 1st tab)
- Expected: Tab active focus switches to the 2nd tab

## For reference:
- When a Tab is discarded, it is given a new Tab ID. It will use that new Tab ID when contents are then loaded in the future. The old Tab ID is not used anymore.
- When a Tab is discarded, its blank replacement contents will cause `app.on('web-contents-created')` to be fired.
- We will never receive a `process.on('add-new-contents')` event for the tab's new contents / tabId. So we rely on muon's new `tab-replaced-at` event which fires on the original tab.


NejcZdovc:
#13118 was merged, so removing PR blocked label

petemill:
This is ready for review but not merge until brave/muon#504 is merged, released and referenced in whichever branch this PR should end up in (currently 0.21.x).

app/browser/tabs.js
```diff
@@ -609,6 +619,29 @@ const api = {
...
+          } else {
+            console.log(`Tab [${tabId}] permanently changed to tabId ${newTabId}. Updating state references...`)
+          }
+        }
+
+        // update state
+        appActions.tabReplaced(tabId, getTabValue(newTabId), getTabValue(tabId).get('windowId'), !isPlaceholder)
+        if (!isPlaceholder) {
+          // update in-memory caches
+          webContentsCache.tabIdChanged(tabId, newTabId)
```
bridiver: I think this could introduce a race condition because the webContentsCache and activeTabHistory will be temporarily out-of-sync with the app state. Can we make them reducers instead that update on the tabReplaced action?

petemill: `activeTabHistory` and `webContentsCache[].openerTabId` (openerTabId is the only prop changed here) are only called in response to tab events coming from muon, not app state actions. Is that enough to satisfy your concern @bridiver? My thinking was to change them immediately to actually avoid a race condition, in case we're getting the tab 'detached' / 'destroyed' event quicker than the store can update those values - and we need the values updated then so that the next active TabId can be correctly chosen. Or is your concern that the `tabReplaced` action may rely on that data?

Either way, I should test that detaching a tab correctly sets the appropriate parent tab on detach, because there may be an issue with the 'temporary' contents id, in which case even if `isPlaceholder === true` we should call `webContentsCache.tabIdChanged` and `activeTabHistory.tabIdChanged` since that is window-specific and the info is needed on `tab.on('will-destroy')` which is fired very soon after `tab-id-changed`.

petemill: ^ detaching is fine, and we shouldn't change the activehistory / openerTabId tabId for placeholder replacements since `tabs.moveTo` directly calls `tabs.getNextActiveTabId` passing in the moved tabId (and not the temporary replacement tabId) - so detaching and setting the appropriate next tab active in the departing window, depending on preference, works nicely.

petemill:
I added some more test cases

petemill:
Looks like the 'Parent Tab 2' discard test isn't working, I'm looking in to it

petemill:
Thanks for the approval @bsclifton. Looking forward to getting it merged and out on beta release to test auto-discarding in the wild.

bsclifton:
master f5db19e3a1a544195b24714ddddf57602dc8bf3f
0.22.x 73e52acb3b0cd99d409de79752fc8dec28b20288
0.21.x @petemill can you help me merge this? getting a conflict

PR #12438, opened 2017-12-29
Create a Buffer Window in order to pre-render and open new windows faster

petemill:
A buffer window is created at startup, and any time a previous buffer window is detached in order to be utilized as a real window.
This type of window is not shown and has no tabs (pinned or non-pinned) until it is shown.
State is not persisted for Buffer Windows until they are utilized as real windows.

A new command-line flag is introduced: '--debug-window-events' which shows logging around Buffer Window creation and utilization, as well as all other window events. No additional logging should be present without this flag.

I had built some of this functionality for #11720 (Tabs transitions) in order to decrease the time to detach a tab to a new window. At the same time, a request came through Asana for the performance initiative to do this same thing for every window. This PR is the result of adapting the code from #11720 for all (compatible) window creation.

Fix #12437

#### Window creation
- Brave starts with 1 window
- A new window can be opened
   - The window opens instantly without any wait

#### Window + frame creation
- Open a tab that has links on it
- Right-click any link and choose 'Open in New Window'
   - The window opens instantly, and the tab with the link's url is opened

#### Window persistance
 - Open at least 2 windows, with at least 2 tabs in each
 - Quit Brave
 - Open Brave again
     - All windows and tabs are re-opened
     - No extra windows are re-opened

#### Pinned tabs
 - Open 2 windows
 - Pin a tab in window 1
    - Tab is pinned in window 1 and window 2
 - Open a new window
     - Window opens instantly
     - new window (3) has the pinned tab

#### Window focus (especially for Windows OS due to brave/muon#424)
- Create a new window
    - The window opens instantly
    - The window gets focus instantly
    - The window does not lose focus, and the urlbar can be typed in without any further action
Especially on Windows: ensure that always when a new window is created, that the new window gets and keeps focus (in the url bar for example)

#### Detach tab to new window
- Create 1 window with mulitple tabs
- Detach a tab (right-click --> Detach, or drag the tab away from Window)
    - Ensure the tab's new window was created from a buffer window
    - Ensure the window is positioned correctly by the mouse cursor (if dragged out), as per previous version of Brave.

#### Popup window
- Open two windows with two different sites in each window
- Pin 1 site in 1 window
   - Site will be pinned in Window 1 and Window 2
- Visit http://www.popuptest.com/goodpopups.html
- Click "Good Popup #1"
     - Popup should open in a new window
     - Popup window should be buffer window
     - Popup window should be sized small (according to the popup request)
     - Popup window **should not have any pinned tabs**
- Pin 1 site in window 2
[...]

PR #13161, opened 2018-02-16
windows restore in the correct order to ensure correct focused window

### LINKED ISSUE #13158: Focused window does not have focus when restarting
note: I noticed this, and made a fix, when testing Buffer Windows in #12438, but it is a regression from 0.20.x or before. Assigning to same milestone as buffer windows, but can be moved if the fix shouldn't land in that milestone.

### Description
This is caused by two reasons:
- An incorrect redux action name is preventing 'focusTime' from getting saved to windowState accurately
- We now open windows asynchronously, so window's could render and be shown, at different times and not in the order the windows were asked to be created

### Test plan / Steps to Reproduce

  1. Open several windows with tabs
  2. Place the windows at different coordinates with different sizes
  2. Focus the windows in a specific order
  2. Quit Brave
  3. Start Brave

**Actual result:**
- The last-focused window does not always have focus.
- The stacking order is not remembered.

**Expected result:**
The windows are layered in the same stacking order as they were previously. The last window to have focus, has focus after restart

petemill:
Do not merge before #12438 since it's best to test this fix with those changes too as that PR changes a lot of window logic. This PR should only show 1 commit to merge.

Fix #13158 

Note that this test plan may never be 100% robust, but this gets us closer.
There will be additional cases that cannot reproduce success (i.e. if a window is macOS fullscreen), but the code to 'fix' those scenarios would likely introduce new issues and/or not be maintainable, for arguably small benefit.

## Submitter Checklist:

## Test Plan:
On #13158 

## Reviewer Checklist:

Tests

### DIFF
--- a/js/entry.js
+++ b/js/entry.js
@@ -77,6 +77,9 @@ ipc.on(messages.INITIALIZE_WINDOW, (e, mem) => {
   const windowValue = message.windowValue
 
   currentWindow.setWindowId(windowValue.id)
+  if (process.env.NODE_ENV === 'development') {
+    console.debug(`This Window's ID is:`, windowValue.id)
+  }
   const newState = Immutable.fromJS(message.windowState) || windowStore.getState()
 
   appStoreRenderer.state = Immutable.fromJS(message.appState)
--- a/js/stores/windowStore.js
+++ b/js/stores/windowStore.js
@@ -2,8 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/* global performance */
-
 const appDispatcher = require('../dispatcher/appDispatcher')
 const EventEmitter = require('events').EventEmitter
 const appActions = require('../actions/appActions')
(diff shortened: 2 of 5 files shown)


js/entry.js
```diff
@@ -77,6 +77,9 @@ ipc.on(messages.INITIALIZE_WINDOW, (e, mem) => {
   const windowValue = message.windowValue
 
   currentWindow.setWindowId(windowValue.id)
+  if (process.env.NODE_ENV === 'development') {
+    console.debug(`This Window's ID is:`, windowValue.id)
```
petemill: Just a little something in each window's console.log to aid window debugging (when in dev mode, and set to show debug messages).

js/stores/windowStore.js
```diff
@@ -752,21 +750,18 @@ const doAction = (action) => {
         }
         break
       }
-    case windowConstants.WINDOW_ON_WINDOW_UPDATE:
```
petemill: This constant does not exist

js/stores/windowStore.js
```diff
@@ -752,21 +750,18 @@ const doAction = (action) => {
...
     case appConstants.APP_WINDOW_READY:
-      {
-        const oldInfo = windowState.get('windowInfo', Immutable.Map())
-        let windowValue = makeImmutable(action.windowValue)
-
-        if (windowValue.get('focused')) {
-          windowValue = windowValue.set('focusTime', performance.timing.navigationStart + performance.now())
```
petemill: This should be set whenever window is updated (not only when it's ready) as we want to save which window **last** had focus. Also we can use `new Date().getTime()` as milliseconds are perfectly fine, we don't need to resolve to a microsecond for this.

bsclifton:
master 1f4058e5646fa78523b5668ceac05fd55176cd52
0.22.x 7fd3c16bd2810e34c9d1c4931e65b57511e7a97c
0.21.x 9d1d278efded01dbbee0ee5df962cdec942c53a8

PR #13197, opened 2018-02-20
sync v2

### LINKED ISSUE #9254: Remove sync devices feature - Sync'd device manager
**Describe the issue you encountered:** Feature to remove devices synchronized with the same passphrase should be added.

- Brave Version (revision SHA): 0.15.314

- Steps to reproduce:
    1. Sync more than 2 profiles with the same sync passphrase
    
- Actual result: You cannot remove sync'd profile.

- Expected result: It would be nice if it would be possible to 
  1. Have a sync'd devices manager
  2. With the maneger, select whether you would just remove the profile, or delete it once the browser is opened and connected to the sync server (anti-theft). To prevent the bad guy from clearing the other profiles on the stolen laptio, that feature should be enabled with either passcode or 2fa.

- Extra QA steps:
    1.
    2.
    3.

- Any related issues:

### LINKED ISSUE #12269: sync v2: Add first-use wizard dialogs
**Overview:**

The wizard guides the user step by step through an optimal path based on their choice of device type to sync with. 

Additionally, imagery at each step (if applicable) adds clarity to the task and reduces time for understanding.

Most wizard pages will allow back/forward navigation. 

If the wizard is closed BEFORE the user adds the first device, they will return to the first-use state. If they close the wizard AFTER a device is added, they will be taken to the device list and options view.

Wizard Screens:

First-run landing page for sync (to be detailed in a different issue)
image

Start a New Sync Chain: Wizard Screen 1: Choose a device type
image

Start a New Sync Chain: Wizard Screen 2a: Scan the code
image

Start a New Sync Chain: Wizard Screen 2b: Enter the code (viewer)
image

Start a New Sync Chain: Wizard Screen 3: Device list (loading devices)
image

Start a New Sync Chain: Wizard Screen 4: Device list (view or add devices)
image

Enter a sync code: Wizard screen 1
image

Enter a sync code: Match check failure
image

image

### LINKED ISSUE #12356: sync V2: Add remove function for any single device in sync chain
As a user, I need to be able to remove a device from my device list in the case that that device was lost or otherwise no longer accessible to me.

**Feature:**
Any device in the sync chain will have the ability to remove any other device. When a device is removed, a new file is posted to the server and the previous data file is deleted. The new data file holds a record of the deleted device. 

See desktop UI:
image

**Approach:**
This record is read by other devices in the sync chain so that they can update the status of the deleted device in their local lists. 

If the device itself has been deleted from the chain, it will read it's own deletion file and display a dialog. "The sync chain is no longer available to this device."

**Dialogs:**
If the active device (this computer) is removed, a warning dialog will appear explaining the result.

image

If a different device is removed, a different warning dialog appears to confirm the result. 

image

Notes:
If the active device is removed, the UI will reset to the first-time-use wizard.

The delete button from the ledger UI control should be used in this device list.

The device list styles should match other list controls.

cezaraugusto:
Closes #12269
Closes #9254
Closes #12356

Test Plan:

* Run `npm start` and `npm run start2` so you can sync between devices
* You should be able to copy passphrase content using the clipboard button
* You should be able to Sync two devices by typing the correct passphrase
* Passphrase should have a counter and it should match the number of words (current default 16)
* Once you sync both devices, they should appear in the postSetup content (inside sortableTable)

> **Note for reviewers**: There's a bug reported in #13258 where you can't sync devices again once you reset sync. That will be tracked separately so in case you're testing multiple options, ensure to clear your profile after every reset you want.

Screenshots **UPDATED**:

screen shot 2018-04-04 at 12 00 34 am
screen shot 2018-04-04 at 12 00 30 am
screen shot 2018-04-04 at 12 00 18 am
screen shot 2018-04-04 at 12 00 10 am
screen shot 2018-04-03 at 11 56 19 pm
screen shot 2018-04-03 at 11 56 14 pm
screen shot 2018-04-03 at 11 56 10 pm
screen shot 2018-04-03 at 11 56 06 pm


app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -347,10 +363,15 @@ syncResetMessageOtherDevices=If you've synced other devices, they will continue
 syncResetMessageWhat=Resetting Sync clears data stored on the Sync server and resets this device's Sync settings.
 syncResetMessageWhatNot=You will keep any bookmarks, history and other browsing data currently on this device.
 syncRetryButton=Try again
+syncScan=Scan the code
+syncWelcome1=Brave Sync allows you to sync bookmarks, tabs, history and other data privately between your Brave Browsers on your varios devices.
```
diracdeltas: various

app/sessionStore.js
```diff
@@ -1015,7 +1015,8 @@ module.exports.defaultAppState = () => {
       lastFetchTimestamp: 0,
       objectsById: {},
       pendingRecords: {},
-      lastConfirmedRecordTimestamp: 0
+      lastConfirmedRecordTimestamp: 0,
+      setupCompleted: false
```
diracdeltas: please add this to docs/state.md

package.json
```diff
@@ -241,10 +242,7 @@
     "template": "COMMIT_TEMPLATE"
   },
   "pre-push": {
-    "run": [
-      "lint",
-      "pre-push-tests"
-    ]
```
diracdeltas: should this have been removed? (btw you can delete the commit hooks in `.git/hooks` instead so it doesn't change any files that are checked into git)

diracdeltas:
is this blocked on https://github.com/brave/browser-laptop/issues/13313?

app/renderer/components/preferences/syncTab.js
```diff
@@ -428,76 +750,93 @@ class SyncTab extends ImmutableComponent {
...
         this.setupSyncProfile(true)
+        appActions.syncSetupCompleted(true)
         return
       }
     }
     window.alert('Invalid input code; please try again or create a new profile.')
   }
 
   render () {
+    console.log(JSON.stringify(this.props.syncData.get('devices')))
```
diracdeltas: extra console.log

diracdeltas:
UI comments:

It doesn't seem to update the word count (stays at 1 no matter what i type)
screen shot 2018-02-26 at 11 17 21 am

If I click 'Start new sync chain' > 'Computer', copy the code, then click 'Done', it still shows the pre-sync preferences screen. Is that expected? It only shows the post-sync screen once another device is synced.

On the post-sync preferences screen, if I click 'Sync a new device..., 'Computer', then click the 'X' button to close the window, it goes back to showing the pre-sync screen. This seems like a bug.
screen shot 2018-02-26 at 11 20 58 am

app/renderer/components/preferences/syncTab.js
cezaraugusto: oh noes! fixed

package.json
cezaraugusto: ya shouldn't be removed. added back

app/sessionStore.js
cezaraugusto: done 👍

cezaraugusto:
ref https://github.com/brave/browser-laptop/pull/13197#issuecomment-368618473

> It doesn't seem to update the word count (stays at 1 no matter what i type)

I'll fix it. I only tested with copy/paste, which works but should work with any user input

> If I click 'Start new sync chain' > 'Computer', copy the code, then click 'Done', it still shows the pre-sync preferences screen. Is that expected? It only shows the post-sync screen once another device is synced.

I'm not sure if I followed right but you can't hit `done` until you go to the next screen (devices list). If you hit the X (close) button you're expected to go to the main screen. The only way to activate Sync is to hit `done` on the last screen. (there are changes coming on this after Brad feedback).

> On the post-sync preferences screen, if I click 'Sync a new device..., 'Computer', then click the 'X' button to close the window, it goes back to showing the pre-sync screen. This seems like a bug.

Ya that's a bug. Addressed in 260ba3e52df2ce1adca4f30034c890bfad7e14bc

cezaraugusto:
removing reviewers and setting as `work in progress` to apply new UI changes per Brad request

bradleyrichter:
@diracdeltas We have made some changes that eliminate the ability to start a sync group (chain) but not add any devices. 

Do you have an important use case where this is desired? From the user feedback and testing we received, it led to confusion.

bsclifton:
Going to start reviewing this and giving feedback 😄 Please note that I'm building off the `sync-v2` branch. In order to get it working, I had to delete the `package-lock.json` file and redo npm install (worked great after that)

bsclifton:
Splash should have the image marked with `user-select: none;` so that it's not selectable via click/drag
user-select

bsclifton:
Splash screen doesn't quite match spec.  

***Current screen***
splash-impl

***Spec***
splash-spec

1. `Brave Sync beta` logo should be inside the image
2. Buttons should be stacked vertically

bsclifton:
Screens showing the code need some tweaks

_**Current screen**_
screen shot 2018-04-02 at 11 14 10 pm

_**Spec**_
screen shot 2018-04-02 at 11 15 50 pm

_**What's different**_
1. `On your computer, go to:` should be `Go to:`
2. `Settings > Sync and click "Enter a sync chain code".` should be `Brave Settings > Sync > Enter a Sync Chain Code`
3. `Then enter this code` text should be deleted (since title already says `enter this code`) and textbox with code words should be moved up

bsclifton:
Enter a sync code screen

**_Current screen_**
screen shot 2018-04-02 at 11 34 18 pm

**_Spec_**
screen shot 2018-04-02 at 11 34 50 pm

**_What's different_**
1. Word count needs to be left aligned
2. Needs clipboard functionality for copy
3. Title `Enter a sync chain code` should be `Enter a sync code`
4. `.` should be removed after `Go to Brave Settings > Sync > Display sync code`
5. Spec shows misspelled / non-matching words as orange- is this do-able?
6. If I put an invalid code and click `Confirm sync code`, I'm shown an alert saying `Invalid input code; please try again or create a new profile.`. Text should match spec: `Invalid sync code. Please try again.`
7. If I put an invalid code and click `Confirm sync code`, the alert is shown. After clicking `OK`, the `Enter a sync code` screen is closed (instead of just hiding the alert and letting me find/fix the broken words)

bsclifton:
I was easily able to join a sync chain 👍 Page needs some tweaking though
_**Current screen**_
screen shot 2018-04-02 at 11 46 30 pm

**_Spec_**
screen shot 2018-04-02 at 11 46 09 pm

**_What's different_**
1. `Add device` and `View sync code` buttons need to be aligned to right of table which shows devices
2. Top portion does not match spec.
  - Need to remove the `Sync this device toggle` / Device name
  - Brave Sync (beta) should be in Black with `beta` being orange
  - Description and tooltip need to be removed
3. The title for what to sync should incorporate the device name. In my example, `Sync Data` should be `Data to sync from Mac Laptop`
4. Table needs adjustment
  - `Device name` should be `Device Name`
  - `Last active` column should actually show `Added on` (if that data is available) and should only show date (not time)
  - Trash can under `Remove` needs center alignment
5. Leave sync chain should be a secondary button

bsclifton:
Once you have sync setup and you click `View sync code`, this screen comes up (which needs some minor tweaks)
screen shot 2018-04-02 at 11 59 23 pm

1. Top of QR code should align with top of text box with words
2. `QR code` title should be center aligned above the actual QR code
3. `Word Code` title needs left padding

bsclifton:
Clicking `Leave Sync Chain` does not match the spec
**_Current Screen_**
screen shot 2018-04-03 at 12 06 51 am

**_Spec_**
screen shot 2018-04-03 at 12 07 49 am

**_What's different_**
1. The icon is missing
2. `Reset Sync` should be `Remove "<device name>" from this sync chain?`
3. Text inside needs to match
4. `Reset sync` button needs to be `Remove`
5. When removing device, a confirmation is shown which I didn't see in the spec. Is this intended?

cezaraugusto:
thanks @bsclifton, great review. Feedback addressed. Some comments/answers:

re: 

> Spec shows misspelled / non-matching words as orange- is this do-able?

Yes but we agreed that this can be made as a follow-up. Needs integration with niceware/bit39. I also need to split this into a separated component so ledger can re-use it

> Brave Sync (beta) should be in Black with beta being orange

Colors were kept to match other pages like payments. Asked Brad and he's ok leaving it as-is

> When removing device, a confirmation is shown which I didn't see in the spec. Is this intended?

Yes, intended

> In the spec (and in this code), I didn't see anywhere I could set device name. Since I'm on a mac, all of my devices are Mac Laptop

Brad can speak better but decision was deliberated. You can differentiate between devices with `(this device)` text (there's an updated screenshot of PR's comment with a demo)

> Given my scenario of syncing two instances of Brave on the same machine:
>
>* I create sync group on instance 1
>* On instance 2, I join existing sync group (instance 1) using code words
>* Instance 2 shows the other device (instance 1)
>* Instance 1 is still stuck at the Looking for device screen:

This action takes 20s on average, which is bad but should work. Needs consensus between the library/design team if we have to either reduce the time or indicate to the user that device is being searched i.e. a countdown

bsclifton:
That's it - everything else has been resolved 😄 👍 

Awesome work on this! 😻

cezaraugusto:
re https://github.com/brave/browser-laptop/pull/13197#issuecomment-379011594

> Can we tighten this up (ex: reduce the margin above/below) to closer match the spec?

> To match the spec, can text be updated to say Go to: Brave Settings > Sync > Enter a Sync Chain Code?

Up to Brad. He asked to reduce text size to keep all in the same line (at least in English);

> Clipboard button always seems to be disabled for some reason- even if you have the proper amount of words entered

Yes that's expected. You can only copy to clipboard pre-generated content

js/stores/appStore.js
```diff
@@ -568,6 +570,9 @@ const handleAppAction = (action) => {
         if (device.name) {
           appState = appState.setIn(['sync', 'devices', deviceId, 'name'], device.name)
         }
+        if (!hasMainDevice && getSetting(settings.SYNC_DEVICE_NAME) === device.name) {
+          appState = appState.setIn(['sync', 'devices', deviceId, 'mainDevice'], true)
```
diracdeltas: mainDevice should be added to docs/state.md

js/stores/appStore.js
```diff
@@ -614,6 +619,12 @@ const handleAppAction = (action) => {
       appState = appState.setIn(['sync', 'devices'], {})
       appState = appState.setIn(['sync', 'objectsById'], {})
       break
+    case appConstants.APP_REMOVE_SYNC_DEVICE:
+      appState = appState.deleteIn(['sync', 'devices', action.deviceId])
```
diracdeltas: I don't know if this is the intended behavior, but since this does not delete the device seed, if you set up sync again after removing a device, it will re-join the old sync chain.

app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -325,41 +326,67 @@ submitFeedback=Help Center
...
 syncNewDevice3=If asks you to enter code words, type in the words below.
-syncNewDevice=Sync a new device…
+syncNewDeviceButton=Sync a new device…
+syncPhoneOrTablet=Phone/Tablet
+syncQRCode=QR Code
 syncQRImg.title=Brave sync QR code
+syncRemove=Remove
+syncRemoveDeviceModal=Remove "{{device}}" from this sync chain?
+syncRemoveDevice.title=Remove this device
+syncRemoveActiveDeviceWarning1=Local device data will remain intact on all devices. Other devices in this sync chain will remain sync'd.
```
diracdeltas: can we use `synced` instead of `sync'd` (which isn't a word)?

app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -325,41 +326,67 @@ submitFeedback=Help Center
...
+syncRemoveActiveDeviceWarning2=To join a sync chain again, choose "Enter a sync chain code".
+syncRemoveOtherDeviceWarning= Note: Removing this device from this sync chain does not clear previously sync'd data from the device.
```
diracdeltas: same as above

js/stores/appStore.js
diracdeltas: this is probably not the right behavior because the text says 'Start a new sync chain' not 'Rejoin my old sync chain'

app/renderer/components/preferences/syncTab.js
```diff
@@ -321,91 +396,440 @@ class SyncTab extends ImmutableComponent {
...
+    }
+  }
+
+  scanCodeOverlayNextAction () {
+    // close current modal
+    this.props.hideOverlay('syncScanCode')
+  }
+
+  /**
+   * QR and Passphrase Overlay
```
diracdeltas: Currently there is no text that indicates this QR code or passphrase should be kept secret. I am worried this will lead to users sharing their QR code or passphrase in support requests. They should not do this since the information is an encryption key. 

@bradleyrichter can you mock up a warning somewhere telling users to never share their code words or QR code?

diracdeltas: for reference this is what ledger says: 

screen shot 2018-04-09 at 3 34 34 pm

js/stores/appStore.js
bradleyrichter: correct. It should start new, not rejoin.

app/extensions/brave/locales/en-US/preferences.properties
bradleyrichter: I hate the way the word looks but since it is more correct, let's use "synced".

app/renderer/components/preferences/syncTab.js
bradleyrichter: mockup:
image

js/stores/appStore.js
cezaraugusto: not sure if I get it but isn't the seed related to the main device only?

app/extensions/brave/locales/en-US/preferences.properties
cezaraugusto: done in 63c57cd76e472929fefdc66b030c088b7c87f4b3

js/stores/appStore.js
cezaraugusto: done in 49113e5b3571ccaabaeeb7d6b69ef415cb137dfe

app/renderer/components/preferences/syncTab.js
cezaraugusto: done in 327d4a6da2bbdbff3ffae2a3b0f7df3138be486f

screen shot 2018-04-10 at 6 38 00 pm
screen shot 2018-04-10 at 6 29 24 pm
screen shot 2018-04-10 at 6 29 17 pm

cezaraugusto: second screenshot has the text in different position given the extra footer button eating space

js/stores/appStore.js
cezaraugusto: so that option is for other devices (other than the main device). IIRC considering the above use case, if user setup sync again after removing a device, `seed` will be cleared given that removing the main device triggers `RESET_SYNC`.

cezaraugusto:
reviews request addressed, left a question regarding `seed` but per https://github.com/brave/browser-laptop/pull/13197#pullrequestreview-110639075 that's a non-blocker.

Screenshots for passphrase warning added in https://github.com/brave/browser-laptop/pull/13197#discussion_r180578145. We're ready for a re-review

js/stores/appStore.js
diracdeltas: @cezaraugusto 

i think tese were my test steps:
1. join an existing sync chain (the one corresponding to bytes `[0, 0, 0, 0...]`, so the current device is not the main device
2. click `Leave Sync Chain`
3. click `Start a new sync chain`
4. click `Computer`
5. it showed the old sync code

however i can't seem to repro this consistently

cezaraugusto: sorry unable to repro but by clicking `Leave Sync Chain` you trigger `DELETE_SYNC_USER`which iirc clears all data in client. Maybe a response delay from the server? 

Created #13799 for further investigation

cezaraugusto:
Added notice for QR code screen per https://github.com/brave/browser-laptop/pull/13197#issuecomment-380265703:

screen shot 2018-04-11 at 11 26 40 am

Phrase is the same as other screens cc @bradleyrichter if this one needs adjustment

js/stores/appStore.js
```diff
@@ -563,6 +565,9 @@ const handleAppAction = (action) => {
         if (device.name) {
           appState = appState.setIn(['sync', 'devices', deviceId, 'name'], device.name)
         }
+        if (!hasMainDevice && getSetting(settings.SYNC_DEVICE_NAME) === device.name) {
```
diracdeltas: i think this line just sets the mainDevice to be true for any device with the same name as the current device, since `SYNC_DEVICE_NAME` is the name of the current device. is that the intended behavior?

diracdeltas: (as long as the current device doesn't know of any existing main devices)

bsclifton:
Great job on this PR - I know you've already shifted work towards Brave Core. I captured what has been done so far with https://github.com/brave/brave-browser/issues/356 for when we do chose to implement this 😄

PR #13211, opened 2018-02-20
Make sure we only forget a tab once it is destroyed.

petemill:
Fixes a potential race condition as pointed out by @bridiver in https://github.com/brave/browser-laptop/commit/b7141404c6a7dfd1fc2bc7f8127d42e968f4a0d4#r27463923 and possibly introduced with #12193

## Test Plan:
Cannot reproduce

## Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -1406,7 +1406,29 @@ const api = {
     return state
   },
   forgetTab: (tabId) => {
-    webContentsCache.cleanupWebContents(tabId)
+    const tab = webContentsCache.getWebContents(tabId)
+    if (!tab) {
+      // perhaps tab was set to be null, but other cache data still exists
+      webContentsCache.cleanupWebContents(tabId)
+      return
+    }
+    // Do not remove tab until it is destroyed, as we still need to refer to it,
+    // even though state has let us know it does not care about the tab anymore
+    // But, we do this here in case state still needs to refer to tab
+    // after it is destroyed, for a brief time.
+    if (tab.isDestroyed()) {
+      if (shouldDebugTabEvents) {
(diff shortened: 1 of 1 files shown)


petemill:
@bridiver do you have STR for how you created and destroyed multiple tabs and the tab was removed from the cache prematurely? I haven't been able to reproduce, but this may fix it.

app/browser/tabs.js
```diff
@@ -1359,7 +1359,28 @@ const api = {
     return state
   },
   forgetTab: (tabId) => {
-    webContentsCache.cleanupWebContents(tabId)
+    const tab = webContentsCache.getWebContents(tabId)
+    if (!tab) {
+      return
```
bsclifton: Should this codepath also call `webContentsCache.cleanupWebContents(tabId)`?

petemill: I suppose it is conceivable that in `webContentsCache` module, `tabData.tab` could be null, but `tabData` could be an object. I don't _think_ there is a code path where that can happen at the moment, but it wouldn't be bad to explicitly call `cleanupWebContents` here, in case that state happens in the future. wdyt?

petemill:
Need to fix timing issue between `will-destroy` and `destroyed`

petemill:
Talking to @bridiver, we would definitely like to remove `tabs.forgetTab` and simply have the cache be cleared on destroy, and probably refactor getNextActiveTabId to store all its data in appState, as long as we can get the timings right.

bsclifton:
master c0754f026d0e264ef23f7b45724ec1e4db062f46
0.23.x 5c9efab6e5ed24fdb916f03f3efed31937251978
0.22.x 1dec344629f558e2c44e234f24103cd4c95e6d8f

Issue #13214, opened 2018-02-21
[hackerone] 317931


jumde:
https://hackerone.com/reports/317931

bsclifton:
Fixed with https://github.com/brave/browser-laptop/pull/13215

LaurenWags:
@jumde could we get a test plan for this one?

kjozwiak:
@jumde @diracdeltas I'm still seeing the same behaviour with `0.23.17` bc754eb. Example:

* ebаy.com@ebаy.com --> xn--eby-7cd.com@xn--eby-7cd.com
* ebаy.com/ebаy.com --> xn--eby-7cd.xn--com/eby-7fg.com
* ebаy.com/@ebay.com --> xn--eby-7cd.com/@ebay.com

Reproduced on both `macOS 10.13.14 x64` and `Win 10 x64.` I'm not sure what the expected behaviour is but from the looks of the hackerone report, the behaviour looks the same as the initial report.

Should this be re-opened?

diracdeltas:
@jumde please add test steps to this issue

given https://github.com/brave/browser-laptop/pull/13215/files#diff-67408f7e7eb48fc100595349f7540aa9R303, seems like the expected behavior is 'ebаy.com/@ebаy.com' becomes 'xn--eby-7cd.com/@xn--eby-7cd.com'

kjozwiak:
Results after talking to @jumde:

ebаy.com@ebаy.com --> xn--eby-7cd.com@xn--eby-7cd.com
ebаy.com/ebаy.com --> xn--eby-7cd.xn--com/eby-7fg.com
ebаy.com/@ebay.com --> xn--eby-7cd.com/@xn--eby-7cd.com

Verified on `macOS 10.13.4 x64` using the following build:

* `0.23.17` bc754ebe4dd8d8f59a322660e8af1962a99753ea 
* muon: 7.1.2 
* libchromiumcontent: 67.0.3396.87

Verified on `Win 10 x64 VM` using the following build:

* `0.23.17` bc754ebe4dd8d8f59a322660e8af1962a99753ea 
* muon: 7.1.2 
* libchromiumcontent: 67.0.3396.87

Verified on `Ubuntu 18.04 x64 VM` using the following build:

* `0.23.17` bc754ebe4dd8d8f59a322660e8af1962a99753ea 
* muon: 7.1.2 
* libchromiumcontent: 67.0.3396.87

PR #13215, opened 2018-02-21
Punycode bypass with `@`

jumde:
Strings with `@` are treated as email addresses, so the punycode conversion does not happen correctly. See: https://www.npmjs.com/package/punycode#punycodetoasciiinput

https://github.com/brave/browser-laptop/issues/13214

### DIFF
--- a/js/lib/urlutil.js
+++ b/js/lib/urlutil.js
@@ -387,7 +387,9 @@ const UrlUtil = {
       parsed.hostname = punycode.toASCII(parsed.hostname)
       return urlFormat(parsed)
     } catch (e) {
-      return punycode.toASCII(url)
+      var splitUrl = url.split('@')
+      splitUrl = splitUrl.map(str => punycode.toASCII(str))
+      return splitUrl.join('@')
     }
   },
 
--- a/test/unit/lib/urlutilTest.js
+++ b/test/unit/lib/urlutilTest.js
@@ -299,6 +299,9 @@ describe('urlutil', function () {
     it('returns the punycode URL when given a valid URL', function () {
       assert.equal(urlUtil.getPunycodeUrl('http://brave:brave@ebаy.com:1234/brave#brave'), 'http://brave:brave@xn--eby-7cd.com:1234/brave#brave')
     })
+    it('returns the punycode URL when given a URL contains @', function () {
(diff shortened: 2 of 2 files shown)


js/lib/urlutil.js
```diff
@@ -387,7 +387,11 @@ const UrlUtil = {
       parsed.hostname = punycode.toASCII(parsed.hostname)
       return urlFormat(parsed)
     } catch (e) {
-      return punycode.toASCII(url)
+      var splitUrl = url.split('@')
+      for (var i = 0; i < splitUrl.length; i++) {
```
diracdeltas: minor: instead of `for ... `, you can do `splitUrl.map(...)`

test/unit/lib/urlutilTest.js
```diff
@@ -299,6 +299,9 @@ describe('urlutil', function () {
     it('returns the punycode URL when given a valid URL', function () {
       assert.equal(urlUtil.getPunycodeUrl('http://brave:brave@ebаy.com:1234/brave#brave'), 'http://brave:brave@xn--eby-7cd.com:1234/brave#brave')
     })
+    it('returns the punycode URL when given a URL contains @', function () {
+      assert.equal(urlUtil.getPunycodeUrl('ebаy.com/@brave.com'), 'xn--eby-7cd.com/@brave.com')
```
darkdh: it is better to test `ebаy.com/@ebаy.com` or something like that which makes both contains punnycode

test/unit/lib/urlutilTest.js
```diff
@@ -299,6 +299,9 @@ describe('urlutil', function () {
     it('returns the punycode URL when given a valid URL', function () {
       assert.equal(urlUtil.getPunycodeUrl('http://brave:brave@ebаy.com:1234/brave#brave'), 'http://brave:brave@xn--eby-7cd.com:1234/brave#brave')
     })
+    it('returns the punycode URL when given a URL contains @', function () {
+      assert.equal(urlUtil.getPunycodeUrl('ebаy.com/@ebаy.com'), 'xn--eby-7cd.com/@xn--eby-7cd.com')
```
diracdeltas: technically only the host name should be shown in punycode. but since this code only applies to strings which are not valid URLs, i don't think this matters much.

bsclifton:
master e162adf0717f60b8ce8a058f5168f8df63125f0c
0.22.x 809155cd84437fcebcafe276e9db48e18f50b4c7

PR #13240, opened 2018-02-21
Handle failure in creating a state diff by logging (and crash-reporting) the full path of any keys that are null

### LINKED ISSUE #13239: If state contains a [null] key, all actions will fail with little log evidence
### Description
It's possible, although undesired, to have a javascript object or Immutable Map with a key of [null]. This causes immutabediff to throw an error as it only supports strings as keys. This then further causes all subsequent state updates to not be sent to renderer windows, and a crash log to be generated upon every single action, which halts the browser momentarily.

One instance of this was caught and fixed in #13234, but there could be more, and if there are, we should log and report something more helpful, such as the exact key path where a null key was found.

petemill:
A null key path was fixed in #13234, but there may be other cases, so we should add better logging to the console and for the crash reports.

Performance should be unaffected since v8 can now optimize functions which contain a try...catch block.

Fix #13239

## Submitter Checklist:

## Test Plan:
This should cause no change in functionality, or extra log lines.
If a dev really wants to check the output in error conditions, they can force a null key path in the state, for example in appStore.js at the top of the `emitChanges` function:
`appState = appState.setIn(['historySites', null], Immutable.fromJS({ hello: 'a' }))`

## Reviewer Checklist:

Tests

### DIFF
--- a/app/common/state/immutableUtil.js
+++ b/app/common/state/immutableUtil.js
@@ -45,6 +45,22 @@ const api = {
       return defaultValue
     }
     return api.isImmutable(obj) ? obj.toJS() : obj
+  },
+
+  findNullKeyPaths (state, pathToState = []) {
+    let nullKeys = [ ]
+    if (!Immutable.Map.isMap(state) && !Immutable.List.isList(state)) {
+      return nullKeys
+    }
+    for (const key of state.keySeq()) {
+      const keyPath = [...pathToState, key]
+      if (key === null) {
+        nullKeys.push(keyPath)
+      }
+      // recursive, to find deep keys
+      nullKeys.push(...api.findNullKeyPaths(state.get(key), keyPath))
+    }
+    return nullKeys
   }
 }
 
--- a/js/stores/appStore.js
+++ b/js/stores/appStore.js
@@ -36,7 +36,7 @@ const urlUtil = require('../lib/urlutil')
 const buildConfig = require('../constants/buildConfig')
 
 // state helpers
-const {makeImmutable} = require('../../app/common/state/immutableUtil')
+const {makeImmutable, findNullKeyPaths} = require('../../app/common/state/immutableUtil')
 const basicAuthState = require('../../app/common/state/basicAuthState')
 const extensionState = require('../../app/common/state/extensionState')
 const aboutNewTabState = require('../../app/common/state/aboutNewTabState')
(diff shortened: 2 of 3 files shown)


bsclifton:
master a4a04cce5db7cc14e6321ab5dd53a3621924b9de
0.22.x dcba3feda93a22b375368242bce3d7c5d0a82268
0.21.x 8fc13f553b46ac83a1cb3613aa21cc02009970d1

js/stores/appStore.js
```diff
@@ -80,8 +80,27 @@ class AppStore extends EventEmitter {
...
-      if (!d.isEmpty()) {
+      let d
+      try {
+        d = diff(this.lastEmittedState, appState)
+      } catch (e) {
+        console.error('Error getting a diff from latest state.')
+        // one possible reason immutablediff can throw an error
+        // is due to null keys, so let's log any that we find
+        const nullKeyPaths = findNullKeyPaths(appState)
+        const error = (typeof e === 'object')
```
bridiver: I think we should be more specific about the type of error before we do a null key check. We don't want to assume that all exceptions are this specific issue.

app/common/state/immutableUtil.js
```diff
@@ -45,6 +45,22 @@ const api = {
...
+    if (!Immutable.Map.isMap(state) && !Immutable.List.isList(state)) {
+      return nullKeys
+    }
+    for (const key of state.keySeq()) {
+      const keyPath = [...pathToState, key]
+      if (key === null) {
+        nullKeys.push(keyPath)
+      }
+      // recursive, to find deep keys
+      nullKeys.push(...api.findNullKeyPaths(state.get(key), keyPath))
```
bridiver: recursion will be substantially more expensive than iteration to go through all the keys

petemill: Is it worth spending the time to get this for...of...recurse loop into a flat iteration loop considering that if this code gets run the app is in an unusable state anyway?

js/stores/appStore.js
petemill: The trouble is that immutablediff does not throw an error object, just a (dynamic) string. This is the only case I've seen immutablediff fail so far, and I did not want to go down the route of either:
- parsing the string, since it could change in a later immutablediff version
- assuming string vs Error object would mean a null-key error, since that could also change

Again, since the app will be unusable in this state, I wanted to err on the side of caution with regards to making sure we get the right error output and reported, so that we can solve the issue. But happy to take advice on this.

app/common/state/immutableUtil.js
petemill: So... I was intrigued by the challenge, and tested what I came up with for a non-recursive function. It runs ~30% slower than the recursive one https://jsperf.com/immutable-find-null-paths-recursion-vs-iteration  Perhaps someone can improve on it.

image

bsclifton:
0.20.x 2c3f9751fad5525cef0cd7a883777dc14e844e18

PR #13319, opened 2018-02-26
Add Install Error Fix

Gurran:
Even though I'm quite tech-savvy I had great trouble trying to resolve this error. Because this error WILL occur on all clean installs of Debian, Mint, and Ubuntu its critical that people easily can find a fix. 

I would usually not put such a solution directly on the install page, but I think it was necessary because of its prevalence which hurt adoption.


posix4e:
+1

PR #13429, opened 2018-03-13
increase version range for pinnedTopSites cleanup

### LINKED ISSUE #12941: top site tiles not being populated when updating profile from 0.19.147

cezaraugusto:
Different fixes were tried for this but all needed to iterate over the list to check for duplicates and ended bulking startup time.

This fix is the same take as previous with a wider range of versions. This is better for performance as it does not iterate over the list on startup/every pin interaction but instead cleans the state at update time so no user-space affected.

fix #12941

## Test Plan:

1. Hack your session-store-1 adding a few duplicated objects inside pinnedTopSites list.
2. Run Brave
3. Quit
4. Check session-store-1 file again. There should be no duplicates
5. Run Brave again. There should be no pinned top sites

## Test Plan (QA specific):

1. Install 0.19.x - have a bunch of top sites
2. Upgrade to 0.20.30 which doesn't have the fix. Verify still broken
3. Update to this release (0.22.x) and verify issue is now fixed

### DIFF
--- a/app/sessionStore.js
+++ b/app/sessionStore.js
@@ -845,11 +845,17 @@ module.exports.runPreMigrations = (data) => {
       if (data.cache) {
         delete data.cache.bookmarkLocation
       }
+    }
 
-      // pinned top sites were stored in the wrong position in 0.19.x
-      // allowing duplicated items. See #12941
-      // in this case eliminate pinned items so they can be properly
-      // populated in their own indexes
+    // pinned top sites were stored in the wrong position in 0.19.x
+    // and on some updates ranging from 0.20.x/0.21.x
+    // allowing duplicated items. See #12941
+    let pinnedTopSitesCleanup = false
+    try {
+      pinnedTopSitesCleanup = compareVersions(data.lastAppVersion, '0.22.00') < 1
+    } catch (e) {}
+
+    if (pinnedTopSitesCleanup) {
       if (data.about.newtab.pinnedTopSites) {
         // Empty array is currently set to include default pinned sites
         // which we avoid given the user already have a profile


bsclifton:
@cezaraugusto besides the dupes, this should fix folks who missed your fix (fixing their top sites).

Did we want to capture that scenario in the test plan? ex: 
1. Install 0.19.x - have a bunch of top sites
2. Upgrade to 0.20.30 which doesn't have the fix. Verify still broken
3. Update to this release and verify issue is now fixed

app/sessionStore.js
```diff
@@ -845,11 +845,17 @@ module.exports.runPreMigrations = (data) => {
...
         delete data.cache.bookmarkLocation
       }
+    }
+
+    // pinned top sites were stored in the wrong position in 0.19.x
+    // and on some updates ranging from 0.20.x/0.21.x
+    // allowing duplicated items. See #12941
+    let pinnedTopSitesCleanup = false
+    try {
+      pinnedTopSitesCleanup = compareVersions(data.lastAppVersion, '0.22.00') < 1
```
bsclifton: should we have another check for this, so it only runs once? Specifically we could capture this in the state in an area called "migrations" (which was removed with https://github.com/brave/browser-laptop/commit/ccee1223cbe565c546b6f103b661ae749bc85971#diff-b4e400e260f324745d12616a8c9c1fc6L377 since nothing else is using it).  You could basically store the timestamp this was ran on and default the value to undefined

bsclifton: Also, note to self: we'll want to update this version before we ship. We might want to create an issue to track that

cezaraugusto:
re https://github.com/brave/browser-laptop/pull/13429#issuecomment-372915466 I included your take (thanks) in the test plan. Most users did receive the fix, this PR specifically target all of them and unless someone manages to update from 0.19.x to something >0.22.00, all users should get this fix.

bsclifton:
master 9e91ffefe9e60585c81cd3dd553d4273c3e37c34
0.23.x 23e50ad013a1626b234a877aad1cbad5c896e0d7
0.22.x f1d4f722cc61cfc6cf4f332b567d2b2b3cc0983e

Issue #12223, opened 2017-12-07
mitigate HSTS fingerprinting


diracdeltas:
## Test plan
See https://github.com/brave/browser-laptop/pull/13437

## Original issue description
context: it has been reported in various places that Criteo is using HSTS supercookies (where they buy a bunch of domains and set HSTS on a different subset of domains for each user in order to uniquely identify them) for ad tracking. https://www.gothamcityresearch.com/single-post/2017/10/12/Criteo-SA-NASDAQ-CRTO-Why-We-Believe-Criteo%E2%80%99s-Undisclosed-Practices-are-Illegal-and-Harmful-to-Advertisers

possibilities:
1. double-key HSTS
2. disallow 3rd parties from setting HSTS

diracdeltas:
adding to 0.20.x, per @BrendanEich suggestion

currently the proposal seems to boil down to: don't apply HSTS on 3rd party subresource loads, unless that subresource has been visited as a first party (similar to Safari's 3rd party cookie policy)

bsclifton:
Moving to 0.21.x for now- let's move it back if this is finished before Jan 8th and doesn't require a Muon change 😄

alexwykoff:
@diracdeltas is this release blocking?

diracdeltas:
it would be really good if this were included in 0.21.x since people have been asking for it for a while. cc @jumde

bsclifton:
Depends on https://github.com/brave/muon/pull/527

@diracdeltas please note that the milestone was adjusted; once the work is complete, we can find the right place for it 😄

diracdeltas:
it now depends on https://github.com/brave/muon/pull/532 which is a simpler change. moving to 0.22.x for now

bsclifton:
Re-opening after reverting browser-laptop code with https://github.com/brave/browser-laptop/pull/13638

btlechowski:
Verified Win7 x64 v0.22.108

Steps used to verify:
1. Clear browsing history and cache
2. Go to https://jsfiddle.net/pqwdgr5x/5/
3. Bring Debugger(F12) and open Network tab
4. Tick 'Preserve log' and select 'Other'
5. Press F5 to reload

Expected result:
301 redirect

Steps used to verify:
1. Clear browsing history and cache
2. Go to https://avatars2.githubusercontent.com/u/1903815?s=40&v=4
3. Go to https://jsfiddle.net/pqwdgr5x/5/
[...]

PR #13437, opened 2018-03-13
HSTS Fingerprinting

jumde:
Fixes: https://github.com/brave/browser-laptop/issues/12223

## Test Plan:

1. Open Brave. Disable HTTPS Everywhere in preferences.
2. Open safebrowsing-proxy.brave.com (ping @jumde if the domain is not working)
3. Verify that the ninja image is loaded with a 301 redirect even if the page is loaded multiple times.
4. Clear browsing history (to avoid cached responses)
5. Load https://avatars2.githubusercontent.com/u/1903815?s=40&v=4 in the URL bar
6. Open safebrowsing-proxy.brave.com and confirm that the HTTPS upgrade happens with a 307 redirect instead of 301.

## Alternative test plan if proxy site is not working
edit session-store-1, make sure lastAppVersion is less than 0.22.x (if not, edit it)
in a new tab, visit https://avatars3.githubusercontent.com/u/7602755?v=4
now go to https://jsfiddle.net/pqwdgr5x/3/. the URLbar should show an unlocked icon.

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -367,9 +367,10 @@ function registerForHeadersReceived (session, partition) {
 
     let parsedTargetUrl = urlParse(details.url || '')
     let parsedFirstPartyUrl = urlParse(firstPartyUrl)
-    let trackingHeaders = ['Strict-Transport-Security', 'Expect-CT', 'Public-Key-Pins']
+    const trackableSecurityHeaders = ['Strict-Transport-Security', 'Expect-CT',
+      'Public-Key-Pins', 'Public-Key-Pins-Report-Only']
     if (isThirdPartyHost(parsedFirstPartyUrl.hostname, parsedTargetUrl.hostname)) {
-      trackingHeaders.forEach(function (header) {
+      trackableSecurityHeaders.forEach(function (header) {
         delete details.responseHeaders[header]
         delete details.responseHeaders[header.toLowerCase()]
       })
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -367,7 +367,7 @@ function registerForHeadersReceived (session, partition) {
 
     let trackingHeaders = ['Strict-Transport-Security', 'Expect-CT', 'Public-Key-Pins']
     if (firstPartyUrl !== details.url) {
-      trackingHeaders.forEach(function(header){
+      trackingHeaders.forEach(function (header) {
         delete details.responseHeaders[header]
         delete details.responseHeaders[header.toLowerCase()]
       })
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -860,7 +860,7 @@ module.exports.clearHSTSData = () => {
   for (let partition in registeredSessions) {
     let ses = registeredSessions[partition]
     setImmediate(() => {
-      ses.clearHSTSData()
+      ses.clearHSTSData.bind(ses)(() => {})
     })
   }
 }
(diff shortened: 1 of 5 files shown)


app/filtering.js
```diff
@@ -364,6 +364,12 @@ function registerForHeadersReceived (session, partition) {
       muonCb({ cancel: true })
       return
     }
+
+    if (firstPartyUrl !== details.url) {
```
diracdeltas: please use `isThirdPartyHost` instead on the URL hostnames since we want to check this based on the host, not the full URL

app/filtering.js
```diff
@@ -364,6 +364,17 @@ function registerForHeadersReceived (session, partition) {
       muonCb({ cancel: true })
       return
     }
+
+    let parsedTargetUrl = urlParse(details.url || '')
+    let parsedFirstPartyUrl = urlParse(firstPartyUrl)
+    let trackingHeaders = ['Strict-Transport-Security', 'Expect-CT', 'Public-Key-Pins']
```
diracdeltas: these should be `const` not `let`

diracdeltas: minor: rename `trackingHeaders` to something like `trackableSecurityHeaders` so it's more clear why we are deleting them

app/filtering.js
```diff
@@ -364,6 +364,17 @@ function registerForHeadersReceived (session, partition) {
...
+    if (isThirdPartyHost(parsedFirstPartyUrl.hostname, parsedTargetUrl.hostname)) {
+      trackingHeaders.forEach(function (header) {
+        delete details.responseHeaders[header]
+        delete details.responseHeaders[header.toLowerCase()]
```
diracdeltas: this line shouldn't be needed because `details.responseHeaders` capitalization is already normalized

app/filtering.js
diracdeltas: i think these should also include https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Public-Key-Pins-Report-Only

app/filtering.js
jumde: I checked yesterday, and I did see some headers in lowercase. `Strict-Transport-Security` and `strict-transport-security`

test/unit/app/sessionStoreTest.js
```diff
@@ -38,6 +38,7 @@ describe('sessionStore unit tests', function () {
     }
   }
   const fakeFiltering = {
+    clearHSTSData: () => {},
```
diracdeltas: it would be good to add a unit test to check that `clearHSTSData` is called if lastAppVersion is 0.21.x but not if it's 0.23.x

app/filtering.js
diracdeltas: thanks for checking - i wonder if we should also be checking `cookie` and `referer` in lowercase then

this seems potentially problematic because HTTP headers are case-insensitive according to the RFC. so someone could create a header called `sTriCT-TransPort-SecUrITY` and the browser would still process it (in theory)

diracdeltas:
lgtm. will test after https://github.com/brave/muon/pull/532 is released

diracdeltas:
Awesome work. Here is the test plan I did:
1. edit session-store-1, make sure lastAppVersion is less than 0.22.x (if not, edit it)
2. in a new tab, visit https://avatars3.githubusercontent.com/u/7602755?v=4
3. now go to https://jsfiddle.net/pqwdgr5x/3/. the URLbar should show an unlocked icon.

PR #13455, opened 2018-03-15
Setting tor path relative to browser laptop

jumde:


app/browser/tabs.js
```diff
@@ -1015,6 +1016,7 @@ const api = {
           createProperties.isolated_storage = true
           createProperties.parent_partition = ''
           createProperties.tor_proxy = 'socks5://127.0.0.1:9050'
+          createProperties.tor_path = getExtensionsPath('tor') + '/tor'
```
darkdh: it is better to use `getExtensionsPath('bin')` so that we can have different binary in it like `geth`
and also need to check platform for Windows in order to use `tor.exe`

app/browser/tabs.js
```diff
@@ -1015,6 +1016,7 @@ const api = {
           createProperties.isolated_storage = true
           createProperties.parent_partition = ''
           createProperties.tor_proxy = 'socks5://127.0.0.1:9050'
+          createProperties.tor_path = getExtensionsPath('bin') + '/tor'
```
diracdeltas: better to use `path.join` here

app/filtering.js
```diff
@@ -691,6 +692,7 @@ const initPartition = (partition) => {
     options.isolated_storage = true
     options.parent_partition = ''
     options.tor_proxy = 'socks5://127.0.0.1:9050'
+    options.tor_path = getExtensionsPath('bin') + '/tor'
```
diracdeltas: same as above

app/browser/tabs.js
darkdh: yeah, or we will get trouble on Windows

app/browser/tabs.js
```diff
@@ -1015,6 +1017,11 @@ const api = {
           createProperties.isolated_storage = true
           createProperties.parent_partition = ''
           createProperties.tor_proxy = 'socks5://127.0.0.1:9050'
+          if (process.platform === 'win32') {
+            createProperties.tor_path = path.join(getExtensionsPath('bin'), 'tor.exe').quote()
```
diracdeltas: why is `quote` needed? `String.prototype.quote` does not appear to be supported on my version of Node (on MacOS)

darkdh:
```
#if defined(OS_WIN)
// Windows-specific LaunchProcess that takes the command line as a
// string.  Useful for situations where you need to control the
// command line arguments directly, but prefer the CommandLine version
// if launching Chrome itself.
//
// The first command line argument should be the path to the process,
// and don't forget to quote it.
//
// Example (including literal quotes)
//  cmdline = "c:\windows\explorer.exe" -foo "c:\bar\"
BASE_EXPORT Process LaunchProcess(const string16& cmdline,
                                  const LaunchOptions& options);
```

diracdeltas: i see - did someone verify on windows that `string.quote()` works? it's non-standard https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/quote

Issue #13471, opened 2018-03-16
brave doesn't check URL protocol and requests external application permissions


kjozwiak:
### Test plan
See https://github.com/brave/browser-laptop/pull/13475

### Description

Brave will requests for `external application` permissions when a URL with an invalid protocol is used. We shouldn't be asking for `external application` permissions when an invalid protocol is being used.

### Steps to Reproduce

  1. launch brave and load reddit.com in a new tab
  2. within the same tab, type in abc://google.com and press 'enter'
  3. reddit.com will request for `external application` permissions

**Actual result:**

![protocalcheck](https://user-images.githubusercontent.com/2602313/37505380-fe4f035a-28ba-11e8-9df1-a6d0e273c4ef.gif)

<img width="1557" alt="screen shot 2018-03-16 at 1 12 14 am" src="https://user-images.githubusercontent.com/2602313/37505342-bcf0af58-28ba-11e8-93b7-cd6af2ed5cd3.png">

<img width="1551" alt="screen shot 2018-03-16 at 1 11 49 am" src="https://user-images.githubusercontent.com/2602313/37505344-be4c83f4-28ba-11e8-9e6d-a3effece9fab.png">

**Expected result:**

Brave shouldn't be asking for external application permissions when loading a URL with an invalid protocol.

**Reproduces how often:**

100% reproducible using the above STR.

### Brave Version

**about:brave info:**

> Brave: 0.21.658 
> V8: 6.5.254.36 
> rev: ad8bdc1d4f9c175d37dd9d4bb1b93dd0be512a1f
> libchromiumcontent: 65.0.3325.162
> Muon: 4.9.2 
> OS Release: 17.4.0 
> Update Channel: Beta 
> OS Architecture: x64 
> OS Platform: macOS 
> Node.js: 7.9.0 
> Brave Sync: v1.4.2 

**Reproducible on current live release:**

Yes, currently also reproducible under 0.21.18 580be78 which is the latest released build.

### Additional Information

CC'ing @diracdeltas @flamsmark for proper triage.

diracdeltas:
It might be hard to detect all valid external protocols since people are constantly creating new ones. However we should probably show the origin of the external site in the URL bar instead of the origin of the page that the user is currently on.

kjozwiak:
@LaurenWags @btlechowski easiest way to get a local webserver going so you can go through the test plan that was added is using `python -m SimpleHTTPServer`.

kjozwiak:
[...]

PR #13475, opened 2018-03-16
Permission host mismatch

jumde:
Fixes: https://github.com/brave/browser-laptop/issues/13471
Depends: https://github.com/brave/muon/pull/540

## Test plan
1. Open https://google.com
2. Change the address in the URL bar to sms://ggle.com. Verify the external permission message is for  sms://ggle.com and not https://google.com
3. Launch a webserver with test.html:
`<a href="sms://ggle.com">Test</a>`
4. Open http(s)://<webserver>/test.html
5. Click on the hyperlink, verify the permission is requested for `sms://ggle.com` and not `<webserver>`

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -394,7 +394,7 @@ function registerPermissionHandler (session, partition) {
   const isPrivate = module.exports.isPrivate(partition)
   // Keep track of per-site permissions granted for this session.
   let permissions = null
-  session.setPermissionRequestHandler((origin, mainFrameUrl, permissionTypes, muonCb) => {
+  session.setPermissionRequestHandler((mainFrameOrigin, requestingUrl, permissionTypes, muonCb) => {
     if (!permissions) {
       permissions = {
         media: {
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
     // TODO(bridiver) - the permission handling should be converted to an action because we should never call `appStore.getState()`
     // Check whether there is a persistent site setting for this host
     const appState = appStore.getState()
-    const isBraveOrigin = origin.startsWith(`chrome-extension://${config.braveExtensionId}/`)
-    const isPDFOrigin = origin.startsWith(`${pdfjsOrigin}/`)
+    const isBraveOrigin = mainFrameOrigin.startsWith(`chrome-extension://${config.braveExtensionId}/`)
+    const isPDFOrigin = mainFrameOrigin.startsWith(`${pdfjsOrigin}/`)
     let settings
     let tempSettings
-    if (mainFrameUrl === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
+    if (requestingUrl === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
       // lookup, display and store site settings by the origin alias
-      origin = isPDFOrigin ? 'PDF Viewer' : 'Brave Browser'
+      requestingUrl = isPDFOrigin ? 'PDF Viewer' : 'Brave Browser'
       // display on all tabs
-      mainFrameUrl = null
+      mainFrameOrigin = null
       // Lookup by exact host pattern match since 'Brave Browser' is not
       // a parseable URL
-      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), origin)
-      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), origin)
-    } else if (mainFrameUrl.startsWith('magnet:')) {
+      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), requestingUrl)
+      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), requestingUrl)
+    } else if (requestingUrl.startsWith('magnet:')) {
       // Show "Allow magnet URL to open an external application?", instead of
       // "Allow null to open an external application?"
       // This covers an edge case where you open a magnet link tab, then disable Torrent Viewer
       // and restart Brave. I don't think it needs localization. See 'Brave Browser' above.
-      origin = 'Magnet URL'
+      requestingUrl = 'Magnet URL'
     } else {
       // Strip trailing slash
-      origin = getOrigin(origin)
-      settings = siteSettings.getSiteSettingsForURL(appState.get('siteSettings'), origin)
-      tempSettings = siteSettings.getSiteSettingsForURL(appState.get('temporarySiteSettings'), origin)
+      requestingUrl = getOrigin(requestingUrl)
+      settings = siteSettings.getSiteSettingsForURL(appState.get('siteSettings'), requestingUrl)
+      tempSettings = siteSettings.getSiteSettingsForURL(appState.get('temporarySiteSettings'), requestingUrl)
     }
 
     let response = []
 
-    if (origin == null) {
+    if (requestingUrl == null) {
       response = new Array(permissionTypes.length)
       response.fill(false, 0, permissionTypes.length)
       muonCb(response)
@@ -471,7 +471,7 @@ function registerPermissionHandler (session, partition) {
         response.push(false)
       } else if (permission === 'fullscreen' &&
         // The Torrent Viewer extension is always allowed to show fullscreen media
-        origin.startsWith('chrome-extension://' + config.torrentExtensionId)) {
+        mainFrameOrigin.startsWith('chrome-extension://' + config.torrentExtensionId)) {
         response.push(true)
       } else if (permission === 'fullscreen' && alwaysAllowFullscreen) {
         // Always allow fullscreen if setting is ON
@@ -497,7 +497,7 @@ function registerPermissionHandler (session, partition) {
 
       // Display 'Brave Browser' if the origin is null; ex: when a mailto: link
       // is opened in a new tab via right-click
-      const message = locale.translation('permissionMessage').replace(/{{\s*host\s*}}/, origin || 'Brave Browser').replace(/{{\s*permission\s*}}/, permissions[permission].action)
+      const message = locale.translation('permissionMessage').replace(/{{\s*host\s*}}/, getOrigin(requestingUrl) || 'Brave Browser').replace(/{{\s*permission\s*}}/, permissions[permission].action)
 
       // If this is a duplicate, clear the previous callback and use the new one
       if (permissionCallbacks[message]) {
@@ -511,9 +511,9 @@ function registerPermissionHandler (session, partition) {
             {text: locale.translation('deny')},
             {text: locale.translation('allow')}
           ],
-          frameOrigin: getOrigin(mainFrameUrl),
+          frameOrigin: getOrigin(mainFrameOrigin),
           options: {
-            persist: !!origin,
+            persist: !!requestingUrl,
             index: i
           },
           message
@@ -530,7 +530,7 @@ function registerPermissionHandler (session, partition) {
           response[index] = result
           if (persist) {
             // remember site setting for this host
-            appActions.changeSiteSetting(origin, permission + 'Permission', result, isPrivate)
+            appActions.changeSiteSetting(requestingUrl, permission + 'Permission', result, isPrivate)
           }
           if (response.length === permissionTypes.length) {
             permissionCallbacks[message] = null
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -431,7 +431,7 @@ function registerPermissionHandler (session, partition) {
     const isPDFOrigin = mainFrameOrigin.startsWith(`${pdfjsOrigin}/`)
     let settings
     let tempSettings
-    let requestingOrigin = getOrigin(requestingUrl)
+    let requestingOrigin = getOrigin(requestingUrl) || requestingUrl
 
     if (requestingUrl === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
       // lookup, display and store site settings by the origin alias
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -433,7 +433,7 @@ function registerPermissionHandler (session, partition) {
     let tempSettings
     let requestingOrigin = getOrigin(requestingUrl)
 
-    if (requestingOrigin === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
+    if (requestingUrl === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
       // lookup, display and store site settings by the origin alias
       requestingOrigin = isPDFOrigin ? 'PDF Viewer' : 'Brave Browser'
       // display on all tabs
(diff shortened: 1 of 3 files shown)


app/filtering.js
```diff
@@ -495,9 +496,10 @@ function registerPermissionHandler (session, partition) {
         }
       }
 
+      let parsedMainFrameUrl = new URL(mainFrameUrl)
```
diracdeltas: can this use `urlParse`?

diracdeltas:
could you please add a test plan with the expected behavior?

app/filtering.js
jumde: With `urlParse`:

```
tps://oogle.com/search?q=s%3A%2F%2Fwww.lcebook.com%2F
{ hash: '',
  hostname: '',
  host: '',
  href: 'tps://oogle.com/search?q=s%3A%2F%2Fwww.lcebook.com%2F',
  path: '//oogle.com/search?q=s%3A%2F%2Fwww.lcebook.com%2F',
  pathname: '//oogle.com/search',
  port: '',
  protocol: 'tps:',
  query: 'q=s%3A%2F%2Fwww.lcebook.com%2F',
  search: '?q=s%3A%2F%2Fwww.lcebook.com%2F',
  origin: '' }
```

jumde: with `new URL`
```
URL {
  href: tps://oogle.com/search?q=s%3A%2F%2Fwww.lcebook.com%2F
  protocol: tps:
  hostname: oogle.com
  pathname: /search
  search: ?q=s%3A%2F%2Fwww.lcebook.com%2F
}
```

jumde: Fails here: `<a href="tp://ttt.com">Test Url</a>`

app/common/urlParse.js
```diff
@@ -5,15 +5,7 @@
...
-// muon.url.parse is not available in all environments (ex: unittests)
-let urlParse
-try {
-  urlParse = muon.url.parse
-} catch (e) {
-  // TODO: move to the new node URL API: https://nodejs.org/api/url.html#url_url
-  urlParse = require('url').parse
-}
-
+let urlParse = require('url').parse
```
diracdeltas: there are some security issues with Node's default URL parser - can this behavior be fixed in Muon's url parser?

jumde: yes, working on that.

app/filtering.js
```diff
@@ -511,7 +511,7 @@ function registerPermissionHandler (session, partition) {
             {text: locale.translation('deny')},
             {text: locale.translation('allow')}
           ],
-          frameOrigin: getOrigin(mainFrameUrl),
+          frameOrigin: getOrigin(origin),
```
diracdeltas: i think you want mainFrameUrl here

jumde: The variable names  need to be updated here. 

In https://github.com/brave/muon/pull/540. I updated the `origin` to be the security origin of the current frame, and the `mainFrameUrl` to the URL requesting the permission.

diracdeltas:
@jumde previously, the `openExternal` permission (as I understood it), was granted to the main frame origin, not the origin that is being opened. Ex: if you go to brave.com and click the `mailto:` link at the bottom of the page, the expected behavior is that it prompts you whether to allow brave.com to open external URLs. 

If we are going to change it to the behavior specified in your test plan, we need to also:
1. update the permission message to "Allow URL_TO_BE_OPENED to be opened in an external application?" instead of "Allow MAIN_FRAME_URL to open external applications"?
2. delete the existing openExternal permission settings, since those are no longer valid with this change

i agree that the behavior in this PR makes more sense than the existing behavior, but i also want to note that the scope of these changes is larger than https://github.com/brave/browser-laptop/issues/13471 (which can be fixed by just changing the displayed URL to be the external URL in the case where a main frame URL doesn't exist).

js/lib/urlutil.js
```diff
@@ -470,7 +470,10 @@ const UrlUtil = {
       if (parsed.protocol === 'about:') {
         return [parsed.protocol, parsed.path].join('')
       }
-      return parsed.origin.replace(/\/+$/, '')
+      if (parsed.origin.length !== 0) {
+        return parsed.origin.replace(/\/+$/, '')
+      }
+      return parsed.href.replace(/\/+$/, '')
```
diracdeltas: is this change still needed? i don't think this works on arbitrary URLs, ex:

```
muon.url.parse('sms://bing/kfdk').href.replace(/\/+$/, '')  // returns "sms://bing/kfdk"
```

app/filtering.js
```diff
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
...
+      mainFrameOrigin = getOrigin(mainFrameOrigin)
+      settings = siteSettings.getSiteSettingsForURL(appState.get('siteSettings'), mainFrameOrigin)
+      tempSettings = siteSettings.getSiteSettingsForURL(appState.get('temporarySiteSettings'), mainFrameOrigin)
     }
 
     let response = []
 
-    if (origin == null) {
+    if (requestingUrl == null) {
       response = new Array(permissionTypes.length)
```
bsclifton: This looks to be the first real change- everything else above this line is just renaming, is that correct?

jumde: Yes!

app/filtering.js
```diff
@@ -497,7 +497,7 @@ function registerPermissionHandler (session, partition) {
 
       // Display 'Brave Browser' if the origin is null; ex: when a mailto: link
       // is opened in a new tab via right-click
-      const message = locale.translation('permissionMessage').replace(/{{\s*host\s*}}/, origin || 'Brave Browser').replace(/{{\s*permission\s*}}/, permissions[permission].action)
+      const message = locale.translation('permissionMessage').replace(/{{\s*host\s*}}/, getOrigin(requestingUrl) || 'Brave Browser').replace(/{{\s*permission\s*}}/, permissions[permission].action)
```
bsclifton: Do you have an example of where the `setPermissionRequestHandler` handler gets called from? I see the definition in Muon, but am not sure when the handler is called?

bsclifton: Looking at the code in Muon...
`mainFrameOrigin` is the `current_origin`
`requestingUrl` maps to `requesting_origin`

Super glad you fixed this mismatch issue 😄

app/filtering.js
```diff
@@ -530,7 +530,7 @@ function registerPermissionHandler (session, partition) {
           response[index] = result
           if (persist) {
             // remember site setting for this host
-            appActions.changeSiteSetting(origin, permission + 'Permission', result, isPrivate)
+            appActions.changeSiteSetting(requestingUrl, permission + 'Permission', result, isPrivate)
```
diracdeltas: I'm confused why we are getting the permissions for the main frame origin in https://github.com/brave/browser-laptop/pull/13475/files#diff-8a9bac13ef6264feb2b6da7c18d86b3cR452, but here we apply permissions by `requestingUrl` instead of `mainFrameOrigin`. seems like these two lines should either both use `requestingUrl` or both use `mainFrameOrigin`

app/filtering.js
```diff
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
...
+      mainFrameOrigin = isPDFOrigin ? 'PDF Viewer' : 'Brave Browser'
       // display on all tabs
-      mainFrameUrl = null
+      requestingUrl = null
       // Lookup by exact host pattern match since 'Brave Browser' is not
       // a parseable URL
-      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), origin)
-      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), origin)
-    } else if (mainFrameUrl.startsWith('magnet:')) {
+      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), mainFrameOrigin)
```
diracdeltas: are you sure this and the other `getSiteSettingsForHostPattern` lines aren't supposed to use  `getOrigin(requestingUrl)` since that's what is actually shown in the message?

diracdeltas: https://github.com/brave/browser-laptop/pull/13475/files#diff-8a9bac13ef6264feb2b6da7c18d86b3cR500

app/filtering.js
jumde: Good catch! fixed

app/filtering.js
```diff
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
...
       // display on all tabs
-      mainFrameUrl = null
+      mainFrameOrigin = null
       // Lookup by exact host pattern match since 'Brave Browser' is not
       // a parseable URL
-      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), origin)
-      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), origin)
-    } else if (mainFrameUrl.startsWith('magnet:')) {
+      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), requestingUrl)
+      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), requestingUrl)
```
diracdeltas: you have to do `requestingUrl = getOrigin(requestingUrl)` before calling these settings since these were previously saved by origin not full URL

diracdeltas: or better yet, define `const requestingOrigin = getOrigin(requestingUrl)` in the function scope so that it's harder to confuse URL with origin

app/filtering.js
```diff
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
...
       // This covers an edge case where you open a magnet link tab, then disable Torrent Viewer
       // and restart Brave. I don't think it needs localization. See 'Brave Browser' above.
-      origin = 'Magnet URL'
+      requestingUrl = 'Magnet URL'
     } else {
       // Strip trailing slash
-      origin = getOrigin(origin)
-      settings = siteSettings.getSiteSettingsForURL(appState.get('siteSettings'), origin)
-      tempSettings = siteSettings.getSiteSettingsForURL(appState.get('temporarySiteSettings'), origin)
+      requestingUrl = getOrigin(requestingUrl)
```
diracdeltas: see https://github.com/brave/browser-laptop/pull/13475/files#r177257591

app/filtering.js
diracdeltas: can also be simplified using https://github.com/brave/browser-laptop/pull/13475/files#r177257591

app/filtering.js
diracdeltas: see https://github.com/brave/browser-laptop/pull/13475/files#r177257591 - this should apply the change by requesting origin, not full URL

js/lib/urlutil.js
diracdeltas: let's remove this diff if it's no longer needed

jumde: This is still needed, `muon.url.parse` for `muon.url.parse('sms://bing/kfdk')` returns NULL origin. The warning message if this change is not present, the warning message will read `Allow Brave Browser to open an external application?`

app/filtering.js
```diff
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
...
-    const isBraveOrigin = origin.startsWith(`chrome-extension://${config.braveExtensionId}/`)
-    const isPDFOrigin = origin.startsWith(`${pdfjsOrigin}/`)
+    const isBraveOrigin = mainFrameOrigin.startsWith(`chrome-extension://${config.braveExtensionId}/`)
+    const isPDFOrigin = mainFrameOrigin.startsWith(`${pdfjsOrigin}/`)
     let settings
     let tempSettings
-    if (mainFrameUrl === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
+    let requestingOrigin = getOrigin(requestingUrl)
+
+    if (requestingOrigin === appUrlUtil.getBraveExtIndexHTML() || isPDFOrigin || isBraveOrigin) {
```
diracdeltas: you want `requestingUrl` in this line i think

js/lib/urlutil.js
diracdeltas: @jumde what if instead of this change, you just change https://github.com/brave/browser-laptop/pull/13475/files#diff-7b5b041c814805dd8496488bc6a53d3cR469 to `parsed.origin != undefined` so that the null origin will trigger https://github.com/brave/browser-laptop/pull/13475/files#diff-7b5b041c814805dd8496488bc6a53d3cR479 ?

also please add a test case in the getOrigin unit tests

jumde: Host is also null for arbitrary URLs:

```
Url: hps://nodejs.org/en/download/
{ hash: '',
  hostname: '',
  host: '',
  href: 'hps://nodejs.org/en/download/',
  path: '//nodejs.org/en/download/',
  pathname: '//nodejs.org/en/download/',
  port: '',
  protocol: 'hps:',
  query: '',
  search: '',
  origin: '' }```

Comment (diracdeltas) [reply js/lib/urlutil.js]: @jumde i see. what if instead of changing the behavior of `getOrigin`, we just change https://github.com/brave/browser-laptop/pull/13475/files#diff-8a9bac13ef6264feb2b6da7c18d86b3cR500 such to `requestingOrigin || requestingUrl || 'Brave Browser'` so it displays correctly?

i am reluctant to change getOrigin since it's used in various places for security checks

**Comment by diracdeltas** [comment]
@jumde thanks for making the changes. PR looks good to me now :). i will test it once the muon build is out.

File: app/filtering.js [hunk 12]
```diff
@@ -471,7 +471,7 @@ function registerPermissionHandler (session, partition) {
         response.push(false)
       } else if (permission === 'fullscreen' &&
         // The Torrent Viewer extension is always allowed to show fullscreen media
-        origin.startsWith('chrome-extension://' + config.torrentExtensionId)) {
+        mainFrameOrigin.startsWith('chrome-extension://' + config.torrentExtensionId)) {
```
diracdeltas: this throws an error when you go to http://www.orimi.com/pdf-test.pdf and click the fullscreen button

should be `requestingOrigin` instead of mainFrameOrigin i think

diracdeltas: actually it shouldn't - https://github.com/brave/browser-laptop/pull/13475/files#diff-8a9bac13ef6264feb2b6da7c18d86b3cR440 set the mainFrameOrigin to null deliberately

so i think this just has to check if mainFrameOrigin is null

app/filtering.js
```diff
@@ -427,35 +427,35 @@ function registerPermissionHandler (session, partition) {
...
-      mainFrameUrl = null
+      mainFrameOrigin = null
       // Lookup by exact host pattern match since 'Brave Browser' is not
       // a parseable URL
-      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), origin)
-      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), origin)
-    } else if (mainFrameUrl.startsWith('magnet:')) {
+      settings = siteSettings.getSiteSettingsForHostPattern(appState.get('siteSettings'), requestingOrigin)
+      tempSettings = siteSettings.getSiteSettingsForHostPattern(appState.get('temporarySiteSettings'), requestingOrigin)
+    } else if (requestingOrigin.startsWith('magnet:')) {
```
diracdeltas: this `if` block` is actually not necessary anymore since it will show the magnet URL itself

bsclifton:
I am an idiot and just merged again without squashing 😛

bsclifton:
master 569450262ec1ee43dac7bdb8b3c73606fd8e9f15
0.23.x ac2b5fdc1ad6990030d2c466160037a560d01069
0.22.x cd7a8781d6fb6d0307865276de9e2ebac4fc91f0

PR #13519, opened 2018-03-20
Add the steps to install Brave on Solus since it got packaged recently

Jacalz:
Brave has now finally been packaged for Solus and it is now available in the stable repository.
This will mean that it is easy to install for users and it will see updates in the software center just like every other program on Solus. Nothing more will be needed.

This is done and ready for merging :+1:

### DIFF
--- a/docs/linuxInstall.md
+++ b/docs/linuxInstall.md
@@ -138,6 +138,13 @@ install the dependency for you:
 sudo apt-get -f install
 ```
 
+## Solus
+
+Install Brave through eopkg:
+```
+sudo eopkg install brave
+```
+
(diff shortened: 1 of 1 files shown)


Jacalz:
Would you mind reviewing this? @bsclifton

Jacalz:
Not to be rude @bsclifton but actually no, Solus is more popular than both OpenSUSE and Fedora if you check on [Distrowatch](https://distrowatch.com), I actually had some data to back me up 😅
[](https://postimg.org/image/pn35xoyez/)

Jacalz:
Not your bad, I forgot to add that to the PR, I intended to do so at first 😅

Issue #13360, opened 2018-03-01
'block 3rd party fingerprinting' sometimes appears to block 1st party fingerprinting


diracdeltas:
### Steps to Reproduce

  1. clean session store
  2. go to https://browserleaks.com/canvas and open shields
  3. note that canvas fingerprinting is blocked from browserleaks.com even though 3rd party fingerprinting block is the default setting

**Actual result:**

![screen shot 2018-03-01 at 1 59 06 pm](https://user-images.githubusercontent.com/549654/36872143-1a8281f4-1d59-11e8-8b12-eb8bad52932c.png)

**Expected result:**
there should be no fingerprinting methods blocked from the first-party origin (browserleaks.com)

### Brave Version

**about:brave info:**

master

**Reproducible on current live release:**

yes

diracdeltas:
apparently this test case uses an about:blank iframe for canvas fp, so i think the report is right - the origin should show `about:blank` instead of the main frame URL though

PR #13539, opened 2018-03-21
fallback to frame location if originating script URL is not found

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/13360

Test Plan:
clean session store
go to https://browserleaks.com/canvas and open shields
note that canvas fingerprinting is blocked from about:blank

## Submitter Checklist:

## Reviewer Checklist:

Tests

### DIFF
--- a/app/extensions/brave/content/scripts/blockCanvasFingerprinting.js
+++ b/app/extensions/brave/content/scripts/blockCanvasFingerprinting.js
@@ -144,7 +144,7 @@ if (chrome.contentSettings.canvasFingerprinting == 'block') {
   })
 
   function reportBlock (type) {
-    var script_url = getOriginatingScriptUrl()
+    var script_url = getOriginatingScriptUrl() || window.location.href
     var msg = {
       type,
       scriptUrl: stripLineAndColumnNumbers(script_url)


snyderp:
This should only happen when the injected frame is missing a src, or is some `about:*` url, no?  If so, then makes sense to attribute to the parent frame (since determining which script created the child frame isn't possible currently)

@diracdeltas If the above doesn't seem off to you, then this all makes sense / looks good to me!

diracdeltas:
@snyderp the current behavior is that the fingerprinting is attributed to the top-level frame (not the parent technically). with this PR, it will instead attribute it to the origin of the frame itself, which in the test case is `about:blank`. 

the reason for doing this is that the 3rd party FP blocker in Muon actually works by just blocking all fingerprinting in 3rd party iframes right now, and "about:blank" is considered by chromium to be 3rd party to browserleaks.com

if we showed the parent frame origin (browserleaks.com in this case) in the brave shields UI, users would be confused why we are apparently blocking a 1st party origin when the setting is set to "block 3rd party fingerprinting"

probably the correct fix is to improve the Muon behavior but this is a workaround for now

snyderp:
@diracdeltas I see, thank you for breaking it down :)

That all sounds good and makes sense, but maybe from a UX perspective, it'd be nicer (and make a bit more sense) to say something like: `"injected frame: " + window.location.href`, since `about:blank` almost looks like an error in Brave?

Im not a UX wizard, so just my 2 cents, take with a grain of salt.  But in terms of functionality, makes sense to me and sounds good!

diracdeltas:
@snyderp can you approve this? thanks

diracdeltas:
master  / 0.24.x: 40775057835e57bab3d5b3e8d3caeddbb58731a0

PR #13554, opened 2018-03-22
Display correct host when requesting external app permisssion

jumde:
Fixes: brave/browser-laptop#13471

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -495,9 +495,13 @@ function registerPermissionHandler (session, partition) {
         }
       }
 
+      let permissionUrl = origin;
+      if (permission === 'openExternal')
+        permissionUrl = mainFrameUrl
+
       // Display 'Brave Browser' if the origin is null; ex: when a mailto: link
       // is opened in a new tab via right-click
-      const message = locale.translation('permissionMessage').replace(/{{\s*host\s*}}/, mainFrameUrl || 'Brave Browser').replace(/{{\s*permission\s*}}/, permissions[permission].action)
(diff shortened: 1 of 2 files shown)


app/filtering.js
```diff
@@ -530,7 +530,7 @@ function registerPermissionHandler (session, partition) {
           response[index] = result
           if (persist) {
             // remember site setting for this host
-            appActions.changeSiteSetting(origin, permission + 'Permission', result, isPrivate)
+            appActions.changeSiteSetting(mainFrameUrl, permission + 'Permission', result, isPrivate)
```
diracdeltas: I don't think we want to change this behavior for permissions in general - ex: if you go to https://developer.mozilla.org/en-US/docs/Web/API/notification and click 'notify me', the message is `Allow https://mdn.mozillademos.org...` instead of `Allow https://developer.mozilla.org...`. we would also have to clear all saved permissions for this to work correctly. 

for https://github.com/brave/browser-laptop/issues/13471, i think the most minimal fix is to change the message to "Allow X to be opened in an external app?" for the `openExternal` permission only and set that permission by URL about to be opened instead of the current page URL

app/filtering.js
```diff
@@ -495,9 +495,14 @@ function registerPermissionHandler (session, partition) {
         }
       }
 
+      let permissionUrl = origin
+      if (permission === 'openExternal') {
+        permissionUrl = mainFrameUrl
```
diracdeltas: i think you want to set this in https://github.com/brave/browser-laptop/pull/13554/files#diff-8a9bac13ef6264feb2b6da7c18d86b3cR451 so that `settings` and `tempSettings` are the correct settings for the origin

jumde:
Closing in favor of: https://github.com/brave/browser-laptop/pull/13475

Issue #11059, opened 2017-09-21
PDF.js viewer error


luixxiul:
### Description

https://community.brave.com/t/pdf-js-viewer-error/7825?u=suguru

<blockquote>
I’ve noticed a glitch when trying to open download links via email (Airmail 3).

Scenario…
I’ve sent dropbox links to a client to download some PDF files.
Sent message links open up a brave window and fail to download.
I get an ‘Invalid or corrupt PDF file’ message when the PDF.js viewer file loads.

Only happens in Brave (0.18.36), links work fine in Safari (10.1.2). Macbook Pro running OS 10.10.5.

I don’t think the client (Windows) had an issue either but this is a recurring issue with Brave. Interrupts my workflow a bit. Hoping somebody can shed some light on it.
</blockquote>

### Steps to Reproduce

  1. Send dropbox links to a client to download some PDF files.
  2. Send message links open up a brave window and fail to download

**Actual result:**

The user claims he got an ‘Invalid or corrupt PDF file’ message when the PDF.js viewer file loaded.

**Expected result:**

It should load the PDF file.

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Brave Version

**about:brave info:**

> 0.18.36

**Reproducible on current live release:**

rhubarbwibble:
**Reproduces how often**: with every dropbox link sent to a PDF. All other file formats are fine.

tbates:
Here's another example: opens fine under safari on os x high sierra golden candidate, fails under brace on same OS

https://www.dropbox.com/s/4nfqksi0b1leh4o/Meet%20the%20Faculty%202017-18%20final.pdf?dl=0

bsclifton:
cc: @diracdeltas

danielfranca:
It seems to happen when the url is not pointing to a pdf itself, but working as a router or container for the file itself.
The same issue happens when trying to open a pdf link on github, i.e: https://github.com/datproject/docs/blob/master/papers/dat-paper.pdf

HP4k1h5:
in Brave | 0.19.48 macosx sierra
i think danielfranca is onto the issue
just tried with 
https://t.co/35e64NwVjV       
 which is for 
https://lirias.kuleuven.be/bitstream/123456789/547640/1/usenix2016-wifi.pdf
the first opens to blank page, the second displays normally

altosaar:
Another example that breaks on brave but works on chrome: https://openreview.net/pdf?id=B184E5qee

[...]

Issue #8364, opened 2017-04-17
opening pdf in new tab fails to load


diracdeltas:
## Test plan:
1. go to https://letsencrypt.org/repository/
2. right click on any of the PDF links to open in a new tab. it should work.
3. right click on any of the PDF links to open in a new private tab. it should work.

## Original issue description
- Did you search for similar issues before submitting this one?
yes 

- Platform (Win7, 8, 10? macOS? Linux distro?):
macOS

- Brave Version (revision SHA):
0.14.1 / pdfjs 1.7.428

- Steps to reproduce:
    1. right-click to open a pdf file in a new tab with pdf.js enabled
    2. the pdf does not load

cezaraugusto:
doesn't work on Ubuntu. "Save as" dialog opens while opening on the same page but not for new tab/new private tab. Reopening

bsclifton:
This works great for me on Ubuntu 14.04 using 0.15.0 RC3

Can you retest this, @cezaraugusto?

bsclifton:
Also confirmed it works great on Ubuntu 16.04

cezaraugusto:
I'm still not able to open PDF running the executable, but I did a new shot checking out RC3 tag branch and it worked. I can't say why as both are running muon 2.58.4. 

This is labeled as working on macOS and tested on two diff Ubuntu versions working good, so I'll close and keep an eye on RC4. @bsclifton thanks for tracking 🥇

PR #13587, opened 2018-03-24
 Fix #10554, #11059: remove check for .pdf in URL extensions, prefer response headers

### LINKED ISSUE #10554: URL ending with .pdf causes Brave to ignore Content-Type, then fail because document is not PDF
Yes.

_Feel free to rename the title._ :)

From community https://community.brave.com/t/url-ending-with-pdf-causes-brave-to-ignore-content-type-then-fail-because-document-is-not-pdf/6055?u=eljuno

>When a response does not declare the content to be a PDF document, a browser should not attempt to render the response as a PDF.
>
>But this is what Brave tries to do, when the URL ends with “.pdf”. Even with a response whose Content-Type is “text/html”.
>
>Of course, with a “Content-Type: text/html” document, the PDF viewer will fail.
>
>Brave should be ignoring the URL and instead use the Content-Type to decide what the content type is.
macOS

0.18.14

- Steps to reproduce:
    1.
    2.
    3.
    
- Actual result:

- Expected result:
Brave fail to load the URL

- Extra QA steps:
    1.
    2.
    3.

- Screenshot if needed:

- Any related issues:

humphd:
This fixes the issue in Brave where URLs that contain `.pdf` in a file's extension are automatically assumed to be PDF resources, and get loaded with PDF.js.  This turns out to be problematic for sites like Dropbox or GitHub, which allow user generated content to be named with a `.pdf` extension, but served as HTML (e.g., with a custom UI wrapped around it).

In debugging this, I found that the bug was introduced in https://github.com/brave/browser-laptop/pull/8366 as a partial fix for #8364.  The underlying issue has since been mostly fixed in other code, and all I needed to do was remove the temporary URL rewriting code in order to let requests get properly inspected.  I've tested with URLs and cases from the original bug, as well as current ones, and all seem to work as expected.

I've also attempted to add a regression test for the case of a `.pdf` URL being served with `text/html`. The underlying test server didn't support overriding `content-type`, so I also included a fix to allow it.  I realize you might want this done another way, or not at all.  I would have upstreamed a fix to `node-static`, but it seems mostly unmaintained.  I'm not familiar with all of your test code, so please let me know if I can improve or correct anything I've done.

Finally, I've written this up as a walkthrough for my students, who are learning open source development.  Quite a few of them are either working on Brave, or hoping to, so I wanted to give them some guidance by fixing a bug myself and writing about the process.  You can see that at https://github.com/humphd/browser-laptop/tree/good-first-experience-issue-10554#walkthrough-fixing-a-bug-in-the-brave-browser.  I'll show them this next week.


test/tab-components/frameTest.js
```diff
@@ -278,6 +279,20 @@ describe('frame tests', function () {
         .url(url)
         .waitForVisible('#viewerContainer')
     })
+
+    it('loads HTML properly despite having .pdf extension', function * () {
```
diracdeltas: thanks for adding a test! unfortunately our automated webdriver test environment is totally broken at the moment, so instead i ran the custom server and manually loaded the .pdf HTML file. confirmed that it is broken in latest release and fixed with this PR.

diracdeltas:
@humphd this needs a rebase since some of the urlutil tests were moved into a different file

humphd:
Rebased and updated urlutil test change.

humphd:
No. It's fixed and rebased. I don't have the ability to land it, so you should ask the maintaners.  I assume the Chrome rewrite work occupied them.On Jul 7, 2018 5:25 PM, "Jason B." <notifications@github.com> wrote:Is this abandoned?

diracdeltas:
@humphd it needs a rebase actually. once that's done i'll move the milestone to 0.24 and merge

humphd:
@diracdeltas, @bsclifton I've rebased again, and this should be good to go now.  Let me know if you need me to do more, or something different.

test/tab-components/frameTest.js
```diff
@@ -278,6 +279,20 @@ describe('frame tests', function () {
...
+      // Add custom headers for this request, so we can override the
+      // default mime type that will otherwise get added for a .pdf file.
+      let customHeaders = { 'Content-Type': 'text/html; charset=utf-8' }
+      let url = Brave.server.url('html_with_pdf_extension.pdf')
+      Brave.server.defineHeaders(url, customHeaders)
+
+      yield this.app.client
+        .tabByIndex(0)
+        .url(url)
+        .waitForUrl(url)
```
diracdeltas: this test needs a `.windowByUrl(Brave.browserWindowUrl)` here to switch from the page context into the Brave window context or else it fails because there's no activeTabTitle component found.

humphd: @diracdeltas OK, I've added this in https://github.com/brave/browser-laptop/pull/13587/commits/0224cd6595b5f8df80bf5018da0c87c495f9d7e9.

diracdeltas:
master: 12650d7ab08248b59f7392df22208327479427d4
0.24.x: a02742c7928d86e5d15fb11eae3105c277a729ac

thanks @humphd !!

humphd:
Excellent, thanks @diracdeltas, and thanks for the reviews.  Glad to contribute, love Brave.  Hope I can do more.

bsclifton:
Uplifted to 0.23.x with 34329d2a5a17afaaba8a88ddac85899482cc39ff

PR #13594, opened 2018-03-25
Fix 12458 v2

jayyyin:
my second attempt at fixing https://github.com/brave/browser-laptop/issues/12458
the code is a bit cleaner this time and I've managed to get the desired selection in the menu but I can't figure out how to make the dropdown menu appear, been trying to add my own windowAction to see if that would work, but came to the realization there just isn't the feature to expand the menu from outside the menubar.js object, any help here would be appreciated.


bsclifton:
Closing PR as stale- apologies that we haven't had a chance to review and help you solve the remaining work ☹️ Unfortunately, the Alt + menu navigation code is extremely custom and is not easy to change. This will be fixed as we move to Brave Core

PR #13641, opened 2018-03-28
Set the tor socks port and data directory.

riastradh-brave:
Set the tor socks port and data directory for muon to use.  For use with https://github.com/brave/muon/pull/553 once that is merged or obviated by equivalent changes.

### DIFF
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -41,7 +41,6 @@ const {getWindow} = require('./windows')
 const activeTabHistory = require('./activeTabHistory')
 const path = require('path')
 const {getTorSocksProxy} = require('../channel')
-const fs = require('fs')
 
 let adBlockRegions
 let currentPartitionNumber = 0
@@ -1025,12 +1024,6 @@ const api = {
           createProperties.isolated_storage = true
           createProperties.parent_partition = ''
           createProperties.tor_proxy = getTorSocksProxy()
-          const userDataDir = app.getPath('userData')
-          try {
-            fs.mkdirSync(path.join(userDataDir, 'tor'), 0o0700)
-          } catch (e) {
-            // TODO(riastradh): Report errors other than EEXIST.
-          }
           if (process.platform === 'win32') {
             createProperties.tor_path = '"' + path.join(getExtensionsPath('bin'), 'tor.exe') + '"'
           } else {
--- a/app/browser/tabs.js
+++ b/app/browser/tabs.js
@@ -40,6 +40,8 @@ const ledgerState = require('../common/state/ledgerState')
 const {getWindow} = require('./windows')
 const activeTabHistory = require('./activeTabHistory')
 const path = require('path')
+const {channel} = require('../channel')
+const fs = require('fs')
 
 let adBlockRegions
 let currentPartitionNumber = 0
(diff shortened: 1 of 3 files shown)


app/filtering.js
```diff
@@ -692,9 +694,35 @@ const initPartition = (partition) => {
     options.parent_partition = ''
   }
   if (isTorPartition) {
+    // TODO(riastradh): Duplicate logic in app/browser/tabs.js.
```
darkdh: it is actually a bug @diracdeltas mentioned in https://github.com/brave/muon/pull/473#issuecomment-376620108

diracdeltas: @riastradh-brave can you split this out into a separate module so it's not in here twice? for instance you could add it as an exported method in `app/channel.js`

app/browser/tabs.js
```diff
@@ -1019,9 +1021,36 @@ const api = {
...
+              break
+            case 'developer':
+              portno = 9280
+              break
+          }
+          console.log(`channel ${channel()} tor port ${portno}`)
+          createProperties.tor_proxy = `socks5://127.0.0.1:${portno}`
+          const etcDir = getExtensionsPath('etc')
+          try {
+            fs.mkdirSync(path.join(etcDir, 'tor'), 0o0700)
```
diracdeltas: we shouldn't call synchronous fs methods for performance reasons. can you just make this directory and check it into git? i would prefer it be `app/extensions/brave/etc/tor` since Tor is part of the Brave "extension" mostly

app/browser/tabs.js
```diff
@@ -1019,9 +1021,34 @@ const api = {
...
+              portno = 9270
+              break
+            case 'developer':
+              portno = 9280
+              break
+          }
+          createProperties.tor_proxy = `socks5://127.0.0.1:${portno}`
+          const userDataDir = app.getPath('userData')
+          try {
+            fs.mkdirSync(path.join(userDataDir, 'tor'), 0o0700)
```
diracdeltas: you should make this directory on startup if it doesn't exist, instead of every time that a tor tab is created, for performance reasons. see app/sessionStore.js

app/browser/tabs.js
```diff
@@ -1019,9 +1021,34 @@ const api = {
...
+          } catch (e) {
+            // TODO(riastradh): Report errors other than EEXIST.
+          }
+          createProperties.tor_data_dir = path.join(userDataDir, 'tor', 'data')
```
darkdh: This can be done in muon so we can remove it here

app/filtering.js
```diff
@@ -692,9 +694,34 @@ const initPartition = (partition) => {
...
+        break
+    }
+    options.tor_proxy = `socks5://127.0.0.1:${portno}`
+    const userDataDir = app.getPath('userData')
+    try {
+      fs.mkdirSync(path.join(userDataDir, 'tor'), 0o0700)
+    } catch (e) {
+      // TODO(riastradh): Report errors other than EEXIST.
+    }
+    options.tor_data_dir = path.join(userDataDir, 'tor', 'data')
```
darkdh: ditto

riastradh-brave:
mkdir and user data lookup will be done in muon in https://github.com/brave/muon/pull/473 now, so those parts of the commit have been changed.  Now it's just setting the socks proxy url with a port number dependent on the channel.

app/browser/tabs.js
```diff
@@ -1019,9 +1020,10 @@ const api = {
           createProperties.parent_partition = ''
         }
         if (createProperties.isTor) {
+          // TODO(riastradh): Duplicate logic in app/filtering.js.
           createProperties.isolated_storage = true
           createProperties.parent_partition = ''
-          createProperties.tor_proxy = 'socks5://127.0.0.1:9050'
+          createProperties.tor_proxy = getTorSocksProxy()
```
diracdeltas: [minor] this function gets called every time a new tab is created. you could declare `const TOR_SOCKS_PROXY = getTorSocksProxy()` at the beginning of this file and just use the constant here since it won't change for a given build of Brave

PR #13673, opened 2018-03-30
Adding fireball search engine to Brave

jumde:
## Testing Instructions

1. Check `about:preferences#search`, verify entry for fireball search exists
2. Select fireball as the search engine, and enter a query in the URL bar to verify that search is completed successfully.


lukemulks:
@jumde do we have to include `t=brave` as a query string param for Fireball? 

IIRC, DDG required, but we don't have a similar requirement here and would be good to not pass if we can avoid it.

jumde:
Merged (master) - 273749d212f996e400c6b543a46200bd647c854e

bsclifton:
@jumde can you create an issue for this please? We track features via issues (you can tag with `release-notes\include` and then link to the test plan above

Milestone was also missing, so I set it as 0.24.x

jumde:
Updated

Issue #13668, opened 2018-03-30
[FEATURE REQUEST] Add an option to disable WebRTC in about:preferences


MkllTech:
### Description
Add an option to disable WebRTC in the settings menu (about:preferences).

**Why**: WebRTC can leak IP addresses even behind VPN.

BrendanEich:
This came up on reddit and our advice to use fingerprinting protection was not useful due to auth systems that do canvas readback (slider captchas etc). I think we must fix this issue so users do not face the bad choice of logging in over a vpn with such a captcha, leaking home ipaddr, and too late then enabling f.p. Cc: @diracdeltas

diracdeltas:
https://github.com/brave/browser-laptop/pull/13713

PR #13674, opened 2018-03-30
Issue 13668 - Add disable WebRTC to Shields

jumde:
Fix for: https://github.com/brave/browser-laptop/issues/13668

## Test Plan:

1. Visit https://www.expressvpn.com/webrtc-leak-test to verify IP address is not leaked.
2. Toggle Site Shields and WebRTC preferences in about:preferences to verify WebRTC works when WebRTC is enabled

### DIFF
--- a/app/browser/reducers/tabsReducer.js
+++ b/app/browser/reducers/tabsReducer.js
@@ -39,7 +39,7 @@ const getWebRTCPolicy = (state, tabId) => {
     siteSettings.getSiteSettingsForURL(allSiteSettings, tabValue.get('url'))
   const activeSiteSettings = siteSettings.activeSettings(tabSiteSettings, state, appConfig)
 
-  if (!activeSiteSettings || activeSiteSettings.fingerprintingProtection !== true) {
+  if (!activeSiteSettings || activeSiteSettings.disableWebRTC !== true) {
     return WEBRTC_DEFAULT
   } else {
     return WEBRTC_DISABLE_NON_PROXY
--- a/app/extensions/brave/locales/en-US/bravery.properties
+++ b/app/extensions/brave/locales/en-US/bravery.properties
@@ -13,6 +13,7 @@ blockAllCookies=Block all cookies
 blockAllFingerprinting=Block all fingerprinting
 blockPopups=Block Popups
 cookieControl=Cookie Control
+disableWebRTC=Disable WebRTC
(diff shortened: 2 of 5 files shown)


diracdeltas:
the original issue is for disabling webrtc in preferences: https://github.com/brave/browser-laptop/issues/13668. i think it makes more sense to do it there instead of in site settings to avoid adding more switches to site settings. cc @bradleyrichter

flamsmark:
I'm not sure which way I lean on this. I think it more likely that folks generally don't use WebRTC, and definitely don't want it usable as a background tracking vector. But if there are a couple of sites where I *do* use WebRTC, I probably don't want to leave it on in the background for everyone? Could we do it as a permission request rather than a shields toggle?

diracdeltas:
> But if there are a couple of sites where I do use WebRTC, I probably don't want to leave it on in the background for everyone? 

You wouldn't need to do that if we add webRTC as a global toggle but keep the current behavior where it's also tied to fingerprinting protection. If webrtc is globally disabled but you allow fingerprinting on a site, the latter should win so webrtc works on only that site

flamsmark:
Yeah, that works pretty well. Without any sort of prompt, it relies on the user behavior of "This site is broken somehow…" → try toggling fingerprinting protection. That pushes us in the direction of erring on the side of breaking more sites rather than risking unexpected tracking, but I'm okay with that.

diracdeltas:
@flamsmark actually my proposal doesn't work for Brendan's use case (user needs to disable FP on a site to use a captcha slider, but they don't want to allow webrtc) if the user changes FP to 'allow all' on a site. another option is we could add a 4th setting to the FP select menu that is  `Allow all fingerprinting except webrtc`

flamsmark:
@yan This UX that we're armchairing is getting pretty gnarly. Might benefit from some pro design work.

bradleyrichter:
Over time as we reach a broader audience, we need to be careful to decrease the amount of default site breakage. Providing shield switches for advanced users that are off by default may be OK on going. We are going to roll-up the advanced choices in a near-future release. This will guide most users to use the shields master switch to fix a broken site. 

we have more advanced toggles on deck as well:
- block all images
- block (and unblock) popups

That being said, we should still carefully consider adding per-site switches and assume people will not be able to understand any switch dependancies.

diracdeltas:
discussed in slack; we are just going to add a global option to never enable webrtc for now.

Issue #13625, opened 2018-03-27
Don't Generate Crash Reports when Reporting is Disabled


jonathansampson:
### Description
When Brave is _not configured_ to report crash data, the browser should not generate temporary crash files.

### Steps to Reproduce
1. Disable crash reporting in Brave
2. *Crash Brave*
3. See `%appdata%/brave` for crash files

**Actual result:**
Crash files are still generated.

**Expected result:**
No files should be generated when reporting is disabled.

**Reproduces how often:**
Always.

### Brave Version
0.21.24

**about:brave info:**
N/A

**Reproducible on current live release:**
Yes

### Additional Information
Reported by a user over Twitch stream.

akashnimare:
@jonathansampson this should fix it -
```javascript
exports.init = (enabled) => {
  if (enabled) {
    initCrashKeys()
    muon.crashReporter.setEnabled(enabled)
    console.log('Crash reporting enabled')
  } else {
    console.log('Crash reporting disabled')
  }
}

```

flamsmark:
Currently, we create reports but don't send them?

akashnimare:
> Currently, we create reports but don't send them?

Looking at the code, yes we don't send it. You need to dump the crash data to Sentry or somewhere else (ex - [here](https://github.com/zulip/zulip-electron/blob/master/app/main/crash-reporter.js)).

akashnimare:
@bsclifton let me know if you want me to send a PR with [this](https://github.com/brave/browser-laptop/issues/13625#issuecomment-376709536) fix.

bsclifton:
hi @akashnimare - please do PR this 😄

bsclifton:
Closing as `wontfix`. Let's revisit this in [brave-core](https://github.com/brave/brave-browser)

akashnimare:
What's going on here? Is this repo deprecated or something?

bsclifton:
@akashnimare kind of - we are in the process of moving to a new code-base. We'll definitely offer hotfix support (security fixes, etc) and critical bug fixes, but basically all new development is happening on the new code-base

You can download the developer channel release (which uses this new code-base) here:
https://brave.com/download-dev

And then there is information about the new repository here:
https://github.com/brave/brave-browser

PR #13677, opened 2018-04-02
preference: Fix crash reporter setting.

akashnimare:
This PR fixes an issue which was caused by crash-reporter.
We should only enable the crash reporter if the setting for
the same is turned on.

Fixes: #13625.


akashnimare:
@bsclifton FYI.

akashnimare:
I have no idea why the CI is failing.

bsclifton:
Thanks for the patch 😄 However, I'm going to close as the issue this fixes was closed

PR #13717, opened 2018-04-04
add WordCountTextArea component for ledger and sync

cezaraugusto:
screenshot from about:styles

screen shot 2018-04-04 at 1 49 02 am

Auditors: @NejcZdovc 
/cc @bsclifton as this can ease the task of bip39

Test Plan:

* Go to about:styles, search for "Word Count textarea"
* Ensure words are updated as you type
* Ensure words are updated when a word is removed
* Ensure you can copy content to clipboard without selecting any text
* Open console and check that at each new type there's a log called "changed!!" (`onChangeText` callback works)
* Ensure you can import the component to a preference page and keep the same styling

Default props:
* `clipboardDisabled` - disables copy to clipboard
* `autoFocus` - set immediate focus to the textarea
* `disabled` - disables the textarea (user can't type on it)
* `value` - the textarea's strings
* `onChangeText` - callback function that will be fired when text is changed

### DIFF
--- a/app/renderer/components/common/textbox.js
+++ b/app/renderer/components/common/textbox.js
@@ -6,6 +6,9 @@ const React = require('react')
 const ImmutableComponent = require('../immutableComponent')
 const {StyleSheet, css} = require('aphrodite/no-important')
 
+const ClipboardButton = require('../common/clipboardButton')
+const appActions = require('../../../../js/actions/appActions')
+
 const globalStyles = require('../styles/global')
 const commonStyles = require('../styles/commonStyles')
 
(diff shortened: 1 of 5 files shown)


app/renderer/components/common/textbox.js
```diff
@@ -77,6 +80,65 @@ class DefaultTextArea extends ImmutableComponent {
...
+
+    this.setState({wordCount})
+
+    if (this.props.onChangeText) {
+      this.props.onChangeText()
+    }
+  }
+
+  handleCopyToClipboard () {
+    if (!this.passphraseDisplay) {
```
bsclifton: we might want to call this something else, like `textAreaBox` (since it may not always be a passphrase)

bsclifton:
Should this also have a boolean property `readonly` for cases where we just want to display the code (but not allow editing)?

bsclifton:
master 4e6cd0055a81894d803ba86b5535627b1ac12c3e
0.23.x 615d26fd5b34f57cd2799233264a24c3911673f4

Issue #13780, opened 2018-04-09
No sanity checking in upload/publish scripts


RyanJarv:
### Description
The tools/upload.py and tools/publish.py scripts don't do any sanity checking and just add files to the existing draft it already exists. This means the person doing the build needs to make sure no one else starts it before a release.

### Steps to Reproduce
1. Run build
2. Run a second build with the same tag
3. Release the draft

**Actual result:**
Release will be a combination of files from the 1st and 2nd build

**Expected result:**
Upload fails

PR #13781, opened 2018-04-09
Clean up upload/publish tools, add initial tests

RyanJarv:
* upload.py now fails when an existing draft exists to avoid publishing the wrong version
* changed publish.py so it checks if tag is already released, asks for confirmation and doesn't upload draft when it doesn't currently exist
* moved duplicate functions to helper.py
* added initial python tests

Fixes: https://github.com/brave/browser-laptop/issues/13780

## Submitter Checklist:

## Test Plan:
Started adding tests in tools/test but not enabled in CI yet. Tested manually on fork.

## Reviewer Checklist:

Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "start-brk": "node ./tools/start.js --debug-brk=5858 -enable-logging --v=0 --enable-dcheck --user-data-dir-name=brave-development",
     "test": "cross-env NODE_ENV=test mocha \"test/**/*Test.js\"",
     "testsuite": "node ./tools/test.js",
-    "unittest": "cross-env NODE_ENV=test mocha \"test/unit/**/*Test.js\" --globals chrome,DOMParser,XMLSerializer",
+    "unittest": "python tools/test && cross-env NODE_ENV=test mocha \"test/unit/**/*Test.js\" --globals chrome,DOMParser,XMLSerializer",
     "unittest-cov": "node --harmony node_modules/istanbul/lib/cli.js cover ./node_modules/mocha/bin/_mocha --globals chrome,DOMParser,XMLSerializer --report lcovonly --report html --report text -x \"test/unit/**/*Test.js\" -- \"test/unit/**/*Test.js\"",
     "update-pdfjs": "rm -r app/extensions/pdfjs/; cp -r ../pdf.js/build/chromium/ app/extensions/pdfjs/",
     "update-psl": "./tools/updatepsl.sh",
--- a/tools/publish_release.py
+++ b/tools/publish_release.py
@@ -19,11 +19,6 @@ def main():
   release = get_draft(repo, get_tag())
   commit_tag = get_commit_tag(get_version())
 
-  print("Found draft {}, release this? [y/n]".format(release['tag_name']))
-  if raw_input() != 'y':
-    print("Ok, not releasing")
-    exit(1)
-
   print("[INFO] Releasing {}".format(release['tag_name']))
   publish_release(repo, release['id'], get_tag(), commit_tag)
 
--- a/tools/publish_release.py
+++ b/tools/publish_release.py
@@ -1,8 +1,12 @@
 #!/usr/bin/env python
-# for more info see https://developer.github.com/v3/repos/releases/
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import json
 import os
+import sys
+
 from lib.github import GitHub
 from lib.helpers import *
 import requests
--- a/tools/publish_release.py
+++ b/tools/publish_release.py
@@ -16,7 +16,7 @@ RELEASE_NAME = 'Dev Channel Beta'
 def main():
   repo = GitHub(get_env('GITHUB_TOKEN')).repos(BROWSER_LAPTOP_REPO)
 
-  release = get_draft(repo, get_tag())
+  release = get_draft(repo, get_tag_without_channel())
   commit_tag = get_commit_tag(get_version())
 
   print("[INFO] Releasing {}".format(release['tag_name']))
--- a/tools/publish_release.py
+++ b/tools/publish_release.py
@@ -4,73 +4,44 @@
 import json
 import os
 from lib.github import GitHub
+from lib.helpers import *
 import requests
 
-BROWSER_LAPTOP_REPO = 'brave/browser-laptop'
-TARGET_ARCH= os.environ['TARGET_ARCH'] if os.environ.has_key('TARGET_ARCH') else 'x64'
 RELEASE_NAME = 'Dev Channel Beta'
 
 def main():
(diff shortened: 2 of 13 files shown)


tools/publish_release.py
```diff
@@ -1,76 +1,49 @@
...
-    if release['draft']:
-      return release
+  release = get_draft(repo, get_tag())
+  commit_tag = get_commit_tag(get_version())
 
-  if tag_exists:
-    tag = 'do-not-publish-me'
-  return create_release_draft(github, tag)
+  print("Found draft {}, release this? [y/n]".format(release['tag_name']))
+  if raw_input() != 'y':
```
mbacchi: I haven't run this myself, but this assumes the user is manually executing this script. Is that the only way its done today?

RyanJarv: I thought just the upload was automated, not publishing?

RyanJarv: Didn't see anything running it in jenkins at least, is it being run somewhere else?

My thinking is we do automate the publish part it will at least be gated.

bsclifton: The whole process is automated actually- in jenkins.brave.com, you can view the following jobs:
- browser-laptop-build-linux
- browser-laptop-build-macos
- browser-laptop-build-win32
- browser-laptop-build-win64

Each of them runs the `./tools/cibuild.py` script which does npm install, builds binary, packages binary, and then uploads to GitHub

bsclifton: The existing process works fine as-is, except the fact that tags aren't considered when searching for the build on GitHub. Instead of blindly using any draft, it should also use the tag name

RyanJarv: Sorry for the delay, got that removed.

RyanJarv:
@mbacchi mind taking a look at this again?

package.json
```diff
@@ -34,7 +34,7 @@
     "start-brk": "node ./tools/start.js --debug-brk=5858 -enable-logging --v=0 --enable-dcheck --user-data-dir-name=brave-development",
     "test": "cross-env NODE_ENV=test mocha \"test/**/*Test.js\"",
     "testsuite": "node ./tools/test.js",
-    "unittest": "cross-env NODE_ENV=test mocha \"test/unit/**/*Test.js\" --globals chrome,DOMParser,XMLSerializer",
+    "unittest": "python tools/test && cross-env NODE_ENV=test mocha \"test/unit/**/*Test.js\" --globals chrome,DOMParser,XMLSerializer",
```
bsclifton: Could we make a new task called `tooltest`? It would be great to isolate `unittest` to purely the JavaScript unit tests 😄

bsclifton: or actually, since the output is just stdout (ex: `unittest-cov` wasn't modified), this should be great! Especially since this runs before push 👍

tools/test/test_upload.py
```diff
@@ -0,0 +1,35 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#!/usr/bin/env python
```
bsclifton: This should be the first line, right?

tools/upload.py
```diff
@@ -1,102 +1,64 @@
...
+  if releases:
+    raise(UserWarning("Draft with tag {} already exists".format(tag)))
 
 def upload_browser_laptop(github, release, file_path):
-  filename = os.path.basename(file_path)
-  print('upload_browser_laptop: ' + filename)
-  try:
-    for asset in release['assets']:
-      if asset['name'] == filename:
-        github.repos(BROWSER_LAPTOP_REPO).releases.assets(asset['id']).delete()
```
bsclifton: This logic is missing from the new script and is definitely important. There are times where a file gets partially uploaded and then network connection drops. When the build/upload job is re-triggered, it needs to wipe out the files on GitHub

RyanJarv: Thinking here is if the file exists something went wrong (network or otherwise) and we would want manual intervention.

Maybe makes sense to add retry to the upload?

tools/test/test_upload.py
RyanJarv: Yep, thanks. Fixed this

tools/upload.py
bsclifton: @RyanJarv would definitely be great to have a retry built into the upload script- right now the retry is done manually... Opening up Jenkins and re-queueing the failed build type (ex: win32). When that is manually re-queued, the delete is needed (otherwise GitHub will return an error since the file exists- which may only be a partial upload)

bsclifton: Here are two examples from this morning where the upload failed:
https://jenkins.brave.com/job/browser-laptop-build-win32/426/console
https://jenkins.brave.com/job/browser-laptop-build-win64/423/console

bsclifton:
@RyanJarv friendly reminder on the last comment I had left (needing to delete the existing files before uploading) for when you have time 😄 
https://github.com/brave/browser-laptop/pull/13781#discussion_r183622804

RyanJarv:
@bsclifton ty! apologize for the delay, will make sure to update this today or at least early next week

tools/upload.py
RyanJarv: @bsclifton added some retry logic if you have a sec to take a look at this again

RyanJarv:
(copying this from above since it's kinda hidden)
Added some retry logic and a few tests for that in the last commit

RyanJarv:
Ok thanks, will look into these.

On 3 this is intended, open to changing the behavior but the intention here was to fail hard to prevent the possibility of 2 scripts running at the same time (or some other bug) to do something unintended silently, hopefully this would't come up much with the retry logic but if we do get to this point I was thinking manual intervention should be required. Might make sense to have a force option that deletes the draft and starts the upload over so we can expose this option in jenkins though?

RyanJarv:
@bsclifton the last 2 commits take care of what you mentioned before.

Fixed the tag naming, added the '--force' option to overwrite the release if it already exists and now cleaning up the partial upload before retrying.

mbacchi:
I'm giving this a :+1:, I haven't run it to test but I'll be porting much of this to brave-core so will open issues if/when I find them.

Nice work @RyanJarv!

RyanJarv:
Looking at the CI job's again I realized that I had the behavior wrong here and the drafts should be merged instead of deleted/overwritten.

https://github.com/RyanJarv/browser-laptop/blob/fix/cibuild-draft-upload/tools/test/test_upload.py#L18

@bsclifton Can you confirm this is the behavior we want?

bsclifton:
I'm going to pull this into 0.23.x so that we can take advantage of the more robust script 😄

bsclifton:
master 875b9ef2c324a392fef2b3c568103f3e89c048e7
0.23.x 0d234c91120531e823bbcf2a5196b18b80f067c4

RyanJarv:
Got a quick fix here for avoiding depending on the python requests module here: https://github.com/brave/browser-laptop/pull/14429

PR #13783, opened 2018-04-09
Do not create a new <webview> element or <Frame /> component for each Tab (single-webview)

### LINKED ISSUE #13216: Enhance performance by attaching only active tabs to a webview (single-webview)
## Description
At the moment, a `<webview>` element is created for each open tab in a window. This creates the following issues:
- Additional DOM elements
- `<webview>` overhead in muon - it's a plugin that has rendering overhead - the guestview assumes that a guest (attached to a webview) is visible, even if we make the <webview> invisible through css
- Attach events firing for all tabs at the same time seems to create an event storm that creates even further performance impact by each event creating one or more redux actions, which then may also attempt to make render changes.

In order to fix this, we aim to only have a single webview per active / previewed tab, switching the guest contents that are attached to the webview when the active or previewed tab changes.

This will mainly fix startup with a profile containing many tabs, but should also improve general responsiveness and possible memory usage.

### Manual
Startup with a profile containing 4 windows with 150 tabs in each window. All tabs marked as unloaded: true (which is enforced in Brave on shutdown for all saved tabs)

#### Expected result
- App starts within a few seconds
- Only the active tab loads its content

#### Actual result
- App takes > 5 minutes to load
- Only the active tab loads its content

### Automated perf test from #11424 
Automated test (feature/perf-tests branch): https://github.com/brave/browser-laptop/pull/10517/commits/648bafdaaf1308d966d415b98beba5e22664d457#diff-e5c305031ff191e0f8d1fb021abeb435

  1. Start Brave
  2. Open 100 tabs each with unique URLs
  3. Close Brave
  4. Open Brave. It should remember the 100 tabs.
  5. Focus the URL bar, type a URL and hit enter to navigate.

**Actual result:**
Much slower with many tabs

**Expected result:**
Similar speed to a fresh profile

## Prerequisites

- [x] Ability to receive events for a tab in the render process without a webview.
This allows us to not move every bit of content functionality to the browser process from the renderer process which, although will be beneficial for performance too, will be too big of a refactor to do at once.
**Status:** Working implementation in [muon/single-webview2](https://github.com/brave/muon/compare/single-webview2) with [petemill/browser-laptop/single-webview2](https://github.com/brave/browser-laptop/compare/single-webview2...petemill:single-webview2?expand=1)

#### Content-related

- [x] Ability to load contents before tabs are attached to a GuestViewContainer.
Currently, if a tab does not have a `<webview>`, it will not proceed with loading in the background. This is the same reason that in the current version of browser-laptop, there is a delay in loading background tabs whilst the frame/<webview> is created and attached to. In the meantime 'Untitled' is displayed as the tab title. The WebContents SetOwnerWindow method is not called. Perhaps the contents are not activated until they are attached (`guest_view_base.cc::ActivateContents` checks if is attached)? It will also not fire its updated events when made active (although it will fire the `set-active` event, which we previously were not relying on). We need to allow tab's without `<webviews>` to load, whilst preserving lazy-load functionality via the `discarded: true` flag in a tab's `createProperties`, (e.g. background ctrl-click tabs)

- [x] Background tabs remain active but do not continue processing requests or finishing navigation.
        - e.g. a tab which console.logs and reloads on a setTimeout, javascript will continue to be processed whilst tab is not attached to a `<webview>` and `load-start` & `'did-get-response-details'` events will be received for tab, but not until the tab is re-attached will the tab send `navigation-entry-commited` & `did-navigate` & `dom-ready`, etc events
        - e.g. a tab which you switch to and whilst still loading switch away. The loading spinner will never stop because we don't receive those events.

 - [x] tab.discard() does not destroy/replace unattached tab contents
Tab content memory will not be released if the tab is discarded whilst it is detached. When it is eventually re-attached (e.g. when it is made active) the tab will immediately discard at that point and then load again, which kind of defeats the purpose.

- [x] many tabs startup crash: `[36162:775:0322/105702.875384:FATAL:navigator_impl.cc(156)] Check failed: 0.`

- [x] move never-displayed tabs to a new window

- [x] Background tab theme color is incorrect
  - workaround: wait for first window resize event

- [x] Lazy-load tabs are loaded initially then reloaded on first visit

- [ ] background loaded tabs have initial DOM window inner{Width,Height} size of 0x0 (webcompat) (minor)
    - `RenderWidgetHostViewGuest::GetViewBounds` returns `0 x 0` size
      `WebContentsImpl::GetSizeForNewRenderView` returns the default `300 x 300` size
- [x] detach then move tab to new window crash (check if fixed)

- [ ] clone tab (check if works)

#### Webview attaching / painting issues

- [x] Webview cannot attach anything else if `will-destroy` of contents has been emitted whilst attached to it via guest
    For all subsequent `<webview>.attachGuest()` calls, those contents do not get attach events, even though web view / guest view container thinks it is attached successfully.
    Perhaps something else is being destroyed and not re-created / reset, inside browser_plugin_guest or browser_plugin_embedder.
    A new `<webview>` can attach to the tabs fine.
    ~**Workaround:** Another reason to use a new `<webview>` each time a tab is made active~
- [x] Very often, when a guest attaches to a <webview> it has already been attached to previously, the webview is white and *will never paint, even when re-paint forced (hide and show in js)*
    ~**Workaround:** (see `webviewDisplay.js`) Create and use a new <webview> each time we want to attach and display a guest.~

- [x] harden race conditions for rapid attach / detach / destroy (rapid cmd-w, tab switching, detaching)

- [x] GuestView receives no events when not attached to webview, and duplicate events for guest when re-attached
   **Solution:** fixed in muon with https://github.com/brave/muon/commit/2fbabc61b64338c6b0549cf67c4a72030b977ab6 (new `set-window` event for assigning to first / subsequent window IDs, similar to `add-new-contents`), see https://github.com/brave/muon/issues/515 **requires review from @bridiver**

- [ ] Removing <webview> from DOM destroys most-recently-attached WebContents, even if detached
  ~**Solution:** fixed in muon with https://github.com/brave/muon/commit/32ebe0b1dbeb62e99fd00222c137afd1adf36025 (**requires review with @bridiver**)~ <-- reverted
  ~Modified `guest_view_container` not to call `guest_view.destroy()` on the last attached guest (`detach()` does not remove `this.guest`)
  We still want the guest to be destroyed when the web contents is, and that's ok because it's done in guest_view_base.cc WebContentsDestroyed()~
  **Status:** Not a big blocker if we're not re-creating webview every active tab change.

#### Solved by webview pool (2 webviews)

- [x] Very often, detaching a guest, attaching another guest and then re-attaching the original guest will result in a white <webview>, even if its to a different / new <webview> that has never been attached to before. Forcing a paint by resizing the <webview> or changing a css style attribute and then back again fixes this (although causing another flicker which we hide).
    **Workaround:** (see `webviewDisplay.js`) AttachGuest -> Hide <webview> -> Show <webview>. Unfortunate since this takes an additional > 50ms before we are sure the frame is visible.

- [x] <webview> has brief white flash between attaches
    **Workaround:** (see `webviewDisplay.js`) Use 2 single-webviews, only hide the currently attached webview when the newly attached webview is attached.

#### State-related

- [x] webview focus timing (especially re-open closed tab)
- [x] tab title / loading status if events happen before event handler can be attached
    - solution, handle all tab events immediately on web contents creation, don't wait for react render cycle

### Implementation
- [x] active frame display

- [x] implement tab preview

- [x] Ability to move tabs between windows, even if they have not had a guest attached to their assigned window yet. @bridiver agress we implement something similar to [WebExtension's `tab.move(tabId, { windowId, index })`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/move): https://github.com/brave/muon/issues/489.
    **Workaround:** Working workaround implemented via using new 'set-window-id' event to remove frame from departing window, and bypassing the `tab.detach()` call if `!tab.attached`. The tab 'gets' to the new window since we always make a moved frame active, so it gets attached to a `<webview>`. Not ideal since a move to a new window could happen internally to muon, but works for basic browser-laptop tab moving.

### Cleanup
- [x] Tidy up `frame.js` - consider non-React class, and do not use fake DOM element as a stand-in for eventemitter.
- [ ] remove logging / put behind flag
- [x] consider `tab-detached-at` should be emitted by embedder contents not tab

### Areas to focus
Focus areas for testing are contained on https://github.com/brave/browser-laptop/issues/13216 but essentially:
- Tab creation, switching, closing and restore on restart
- webcompat of various sites, especially for tabs loading in background through cmd/ctrl-click links as well as foreground loading)
- entry and exit of fullscreen via on-site html buttons / keyboard shortcut (<keyb>esc</keyb>), and switching away from tab
- origin settings (exceptions for flash / noscript when navigating to different origins in the same tab, as well as allow-once vs allow-always for all origin settings inc fullscreen https://github.com/brave/browser-laptop/pull/9475)
- widevine popup (background-loaded tabs especially)
- Autofill
- Brave Shield settings
- Ledger tab activity time logging
- SSL (+EV) certificate and lock icon status

petemill:
Fix #13216 of which the overall goal is to move towarrds making sure that performance of a renderer window is not affected by the number of tabs that are associated with that window.

### Removes `<Frame />` Component
This PR not only removes the `<webview>` element for each frame, the problems of which are addressed in the attached issue, but also removes its React wrapper: the `<Frame />` component. 

The main issue with this component for performance is that for each tab in a window, a new instance of that component was being created. Then, each time the store dispatched an action, the Frame's `mergeProps` function would run. Even though this is a very quick function, this could add up to blocking for ~10% of the window's JS thread time, when there are > 150 tabs. In addition, the component had a `componentDidUpdate` function which would then dispatch further store actions based on the `mergeProps result, which would cause every Frame (and every other React component in the window) to perform this cycle again.

Previously, this component listened to events on its associated Tab's WebContents via routing through the `<webview>`. Instead of keeping this model, creating a new React element for each tab, and providing direct access to the `WebContents` event handler, we create a single event handler function instance that is called for all events on _any_ tab in a window. Ideally these event handlers and dispatched actions should be performed by the browser process, but a change of that scope is too large to be done at once, especially with the already-large changes required by removing the multiple webviews.

Most event handler code is transported _as-is_ from `frame.js` to `rendererTabEvents.js` and `rendererShortcutHandler.js` with the exception of items that relied on internal component instance state, including `prevProps`. These should be especially tested:
- When navigating to a different origin, expire temporary site-settings for widevine, flash and noscript - https://github.com/brave/browser-laptop/commit/17c181420d24539096ad70a9d4488a316f73bf6a
- When changing full-screen via in-content control, keyboard shortcut (<keyb>esc</keyb>), or switching tab, ensure that content area and window go in and out of full screen
- Whenever frame state for a tab was changed after `componentDidUpdate`, the frame would dispatch an action to the browser each time, for each frame. This has been moved to a single-purpose function - `rendererFrameTracker.js`

### Converted `app-new-web-contents-added` action to IPC
- This action communicates to a window that it has a new frame, and adds it that window's frams state. Moved from store action to IPC because:
    - Events are received much more immediately by the renderer now due to:
        - Tabs load immediately, not just when webview is attached
        - Event handlers are not only setup when <Frame> component is created, there is a single event handler for all window tab events
    - These events must have the frame created in state
    - We can't simply listen for 'tab-inserted-at' as we need the frame data
    - We can't simply pass it in on window creation with shared memory as window creation is not at all the only way a frame can be added to a window (e.g. moving tabs to existing window, using buffer window)

### Creates `(Pooled)WebviewDisplay`
This is a vanilla JS class which controls the lifecycle of the `<webview>`, including attaching and detaching to different tabs. This is controlled by a React Component: `<GuestInstanceRenderer />` which is a direct child of `<Main />`.
This class contains all the workarounds for quirks of the `<webview />` including issues with not displaying content of an attached tab until the painting has been forced by hiding and showing, and handling fragility in timing when it comes to attaching to destroyed WebContents.
In order to prevent a 'white flash' between the detach of a previous tab and the attach of the next tab to display, we actually maintain a pool of 2 `<webview>` elements. We first attach the incoming Tab to an unused `<webview>`, make sure it is displaying, before we hide and detach the previously displayed tab from its `<webview>`. To this component, switching to a Tab, or viewing it as a preview is the same thing, it only cares about which tab should be currently displayed.
Should an incoing request to view a different Tab occur whilst the `WebviewDisplay` is still attaching to a new tab, it will either queue up the next Tab to be displayed after the current attachment is complete, or cancel the current attachment and begin attaching the newly-requested Tab instead, depending on the status of the attachment process.

### Handles tab window attachment differently
Relying on `did-attach` and `did-detach` had to be changed, since these events are about attachment to a webview. Previously, the only time a tab was detached from a webview was when moving window, and the tab would always get a new webview in another window, even if it was a background tab. Now, every time we switch active tab one tab is detached from the webview and another is attached. We made new events in muon that specifically inform about being removed from or added to windows: `tab-detached-at` and `tab-inserted-at`, as well as a new muon api to tell a tab to move to a different window.

Other major muon changes were on the topic of ensuring that the tab content was fetched and loaded even when never having been attached - previously the tab would just be created but not actually do anything until it was made active.

### Other optimizations
Plus a few more changes with the same goal of reducing performance impact of many tabs, windows, or items in state, such as:
- Not changing the state.tabs Immutable.List when deleting a tab (setting to null instead)
- Not forwarding frame objects for each tab back to every window in appState updates, since the window is where the frame state came from in the first place
- Setting the New Tab theme color immediately, rather than waiting for it to be computed from DOM content

There's definitely more to be done in this area, especially with the amount of events locking up the browser and window threads on app startup with even a medium-sized profile (#14082), but I had to draw the line somewhere.

### Logging
New flag `--debug-store-actions` now outputs each action at the time of dispatch in both the browser console, and each window console (with timing info and fancy colors).

Using the flag `--debug-tab-events` will now show specialized console output in the renderer windows (in addition to the previous log output in the main browser terminal / console) which will show
 - Each event from each Tab in the window
 - A console group for each attempt to attach a Tab to a `<webview>` in `WebviewDisplay`, including timings for each step.

A new Debug menu option _"Display Tab Identifiers"_ will now display: Tab Id, Index, Guest InstanceId, FrameKey and FrameIndex on each Tab, as well as the the Tab which is currently being displayed (or is in the process of attaching to a `<webview>`, or the reason why it is not):
image

### Misc
Addded support for `BRAVE_SHOW_FIRST_RUN_WELCOME=1` env variable in order to debug issues with the 'about:welcome' tab which is created on a blank profile (but disabled by default in dev environment).

## Submitter Checklist:

## Test Plan:
- Full pass
- Tests have been ongoing with the beta channel for a few weeks

## Reviewer Checklist:

Tests


app/renderer/components/main/main.js
```diff
@@ -736,31 +744,23 @@ class Main extends React.Component {
           : null
         }
       </div>
-      <div className='mainContainer'>
-        <TransitionGroup className={cx({
```
bsclifton: Since this is gone now, can we remove the package.json reference to `react-transition-group`?

app/renderer/rendererTabEvents.js
```diff
@@ -0,0 +1,567 @@
...
+    // Whilst any actions dispatched by this function will reach the store
+    // *after* the frame creation event, some of these events require the frame to
+    // already exist in order to make some processing decisions,
+    // though that should be refactored so that the state-processing is done in
+    // the reducers.
+    if (!frame || frame.isEmpty()) {
+      if (shouldDebug) {
+        console.debug('%cNo frame for event yet, queueing until later', 'color: red', tabId, eventType)
+      }
+      windowStore.on(`new-frame-${tabId}`, (newFrame) => {
```
bsclifton: should be `.once(`

app/browser/reducers/windowsReducer.js
```diff
@@ -24,6 +24,7 @@ const BrowserWindow = electron.BrowserWindow
 const firstDefinedValue = require('../../../js/lib/functional').firstDefinedValue
 const settings = require('../../../js/constants/settings')
 const getSetting = require('../../../js/settings').getSetting
+const { shouldDebugStoreActions, shouldDebugTabEvents } = require('../../cmdLine')
```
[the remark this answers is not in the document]
petemill: Thanks for the note @jumde! I haven't satisfied linter and some tests yet. Some of this code was written in an experimental mode with things added, removed, or changed very quickly. But right now we're pretty locked down on architecture and are in clean-up mode, so that's what I'm looking in to over the course of the next day or 2.

app/renderer/components/frame/guestInstanceRenderer.js
```diff
@@ -0,0 +1,205 @@
+const React = require('react')
+const {StyleSheet, css} = require('aphrodite/no-important')
+
+// Components
+const ReduxComponent = require('../reduxComponent')
+// comment out to not use pooled webview (2x webviews):
```
bsclifton: If time allows, it would be much better to check for an environment variable instead of having the notes about commenting in here. Something like:
```
let WebviewDisplay
if (process.env.USE_POOLED_WEBVIEW === false) {
  WebviewDisplay = require('../../webviewDisplay')
} else {
  WebviewDisplay = require('../../pooledWebviewDisplay')
}
```

app/browser/tabMessageBox.js
```diff
@@ -44,8 +57,15 @@ const tabMessageBox = {
...
-    process.on('window-prompt', (webContents, extraData, title, message, defaultPromptText,
-          shouldDisplaySuppressCheckbox, isBeforeUnloadDialog, isReload, muonCb) => {
+    process.on('window-prompt', (
+      webContents, extraData, title,
+      message, defaultPromptText, shouldDisplaySuppressCheckbox,
+      isBeforeUnloadDialog, isReload, muonCb
+    ) => {
+      const tabId = webContents.getId()
+      if (shouldDebugTabEvents) {
+        console.log(`Tab [${tabId}] window-confirm`)
```
bsclifton: this should be `window-prompt`

app/browser/reducers/tabsReducer.js
```diff
@@ -367,8 +436,10 @@ const tabsReducer = (state, action, immutableAction) => {
         }
         break
       }
-    case appConstants.APP_FRAME_CHANGED:
-      state = tabState.updateFrame(state, action, shouldDebugTabEvents)
+    case appConstants.APP_FRAMES_CHANGED:
+      for (const frameAction of action.get('frames').valueSeq()) {
```
bsclifton: might be good to check if `frames` is truthy (ex: `action.has('frames')`) before using

app/renderer/rendererFrameTracker.js
```diff
@@ -0,0 +1,64 @@
...
+      appActions.framesChanged(changedFrames)
+    }
+    if (shouldDebugStoreActions) {
+      console.log(`%cSpent ${window.performance.now() - t0}ms figuring out frame changes (${changedFrames.length} changed)`, 'color: #bbb')
+    }
+  }
+
+  // compare frames when state changes debounced, and only when thread is idle
+  const onWindowStoreChanged = debounce(() => {
+    idleCallbackId = idleCallbackId || window.requestIdleCallback(relayFrameChanges, { timeout: 1000 })
```
bsclifton: Could you put parens around this to make the logic obvious? Without them, this looks like the OR would only run if the assignment is falsey.  Should be like:
`idleCallbackId = (idleCallbackId || window.requestIdleCallback(relayFrameChanges, { timeout: 1000 }))`

app/renderer/rendererFrameTracker.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+    const lastFrames = frames
+    frames = mapFramesByKey(currentFrameState)
+    const changedFrames = []
+    for (const frameKey in frames) {
+      // does it exist in the last version of state?
+      const frame = frames[frameKey]
+      const lastFrame = lastFrames[frameKey]
+      if (!lastFrame || !lastFrame.equals(frame)) {
+        changedFrames.push(Immutable.Map().set('frame', frame))
```
bsclifton: You should be able to initialize this in one go; such as:
`changedFrames.push(Immutable.Map({frame}))`

app/renderer/rendererShortcutHandler.js
```diff
@@ -0,0 +1,237 @@
...
+const api = module.exports = {
+  handleActiveFrameShortcut(shortcut, e, args) {
+    const frameKey = frameStateUtil.getActiveFrame(windowStore.state).get('key')
+    handleShortcut(frameKey, shortcut, e, args)
+  },
+  handleFrameShortcut: handleShortcut
+}
+
+function handleShortcut (frameKey, shortcut, e, args) {
+  console.log('got shortcut', shortcut, frameKey)
```
bsclifton: does this console.log need to be removed / commented out?

app/renderer/rendererShortcutHandler.js
```diff
@@ -0,0 +1,237 @@
...
+  switch (shortcut) {
+    case 'toggle-dev-tools': {
+      appActions.toggleDevTools(frameStateUtil.getTabIdByFrameKey(windowStore.state, frameKey))
+      break
+    }
+    case 'stop': {
+      const tabId = frameStateUtil.getTabIdByFrameKey(windowStore.state, frameKey)
+      console.log('getting webcontets', tabId)
```
bsclifton: this (and below) - more console.logs possibly needing removal

app/browser/reducers/tabsReducer.js
```diff
@@ -291,6 +352,14 @@ const tabsReducer = (state, action, immutableAction) => {
         tabs.setTabIndex(action.get('tabId'), action.get('index'))
       })
       break
+    case appConstants.APP_TAB_SET_FULL_SCREEN: {
+      const isFullscreen = action.get('isFullScreen')
+      const tabId = action.get('tabId')
+      if (isFullscreen === true || isFullscreen === false) {
```
jumde: If the tab is already `FullScreen`, do we need to set it to full screen again?

petemill: This code was moved in https://github.com/brave/browser-laptop/pull/13783/commits/23f5b719fd8ee0a610e8d52685602f4fc37221f7. At this point, I believe that the tab content area is not fullscreen, but other UX may be (e.g. the window). Agreed that if that's the case, perhaps `isFullscreen` is not the best variable name here, although it's more of a state description - "This tab is in fullscreen mode or not, so tell the tab's content to go fullscreen too, or not)".

app/browser/reducers/tabsReducer.js
```diff
@@ -57,14 +57,51 @@ const getWebRTCPolicy = (state, tabId) => {
   }
 }
 
+function expireContentSettings (state, tabId, origin) {
+  // Expired Flash settings should be deleted when the webview is
+  // navigated or closed. Same for NoScript's allow-once option.
+  const tabValue = tabState.getByTabId(state, tabId)
```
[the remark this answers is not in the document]
diracdeltas: it's old functionality. you can enable flash in prefs, then go to https://www.onlinemictest.com/webcam-test-in-adobe-flash/, right click on the flash element, click 'allow once', then close the tab and re-visit it. flash should be re-blocked.

app/renderer/webviewDisplay.js
```diff
@@ -0,0 +1,51 @@
...
+module.exports = class SingleWebviewDisplay {
+  constructor ({containerElement, classNameWebview}) {
+    this.isAttached = false
+    this.webview = this.createWebview()
+    this.webview.classList.add(classNameWebview)
+    containerElement.appendChild(this.webview)
+  }
+
+  attachActiveTab (tabId) {
+    console.log('webviewDisplay: attaching tab id', tabId)
```
diracdeltas: extra console.log?

app/browser/windows.js
```diff
@@ -277,6 +285,22 @@ const api = {
...
+          if (shouldDebugWindowEvents) {
+            console.log(`window ${win.id} had ${!active ? 'in' : ''}active tab ${tabId} inserted at index ${index}`)
+          }
+        })
+        win.webContents.on('tab-strip-empty', () => {
+          // must wait for pending tabs to be attached to new window before closing
+          // TODO(petemill): race condition if multiple different tabs are moved at the same time
+          // ...tab-strip-empty may fire before all of those tabs are inserted to new window
+          win.webContents.once('detached-tab-new-window', () => {
+            console.log('departing tab made it to new window')
```
diracdeltas: extra console.log?

app/renderer/webviewDisplay.js
petemill: Thanks for the spot. I actually haven't done a pass of cleanup or lint fixing as we're still debugging a couple display issues.

app/browser/reducers/tabsReducer.js
petemill: This action does not get called without that property. Should we have a policy of trusting our actions in the way that they are currently being called, or optimize for possible future mistakes?

app/renderer/webviewDisplay.js
```diff
@@ -0,0 +1,51 @@
...
+      if (!webContents || webContents.isDestroyed()) {
+        return
+      }
+      this.webview.attachGuest(webContents.guestInstanceId, webContents)
+    })
+    this.isAttached = true
+  }
+
+  createWebview () {
+    console.log('creating a webview')
```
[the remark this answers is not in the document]
petemill: I've left these console.logs in as this is an experimental class at the moment, which we may switch to in an upcoming release.

app/renderer/components/main/main.js
petemill: I was just about to do this, but then realised that this may make the merge a bit more of a pain, especially package-lock.json, across branches. Perhaps a separate PR unless you think that it won't be a pain?

bsclifton: Separate PR is great 😄 👍 Let's leave it alone

petemill:
0.23.x b088714fb
0.22.x-release3 / 0.22.x 2371c55fec2ed9d53bd32ce87f3417d479a6c0b7

Issue #13791, opened 2018-04-10
use npm package for sync instead of github releases


diracdeltas:
I published the Sync staging branch (client only) https://github.com/brave/sync/commit/81d6b3342a9dfa3dcfab9be1a56cf0a45c260b93 as an npm package: https://www.npmjs.com/package/@brave/sync-client

so now we can install a specific version of sync instead of just the latest github version

PR #13792, opened 2018-04-10
install sync from NPM instead of github

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/13791

this doesn't need to block sync merge but it would probably be useful at some point


bsclifton:
Closing as we deferred the Sync v2 work until Brave Core and there hasn't been any action taken yet

This could still be useful though- let's reopen / talk about it in our weekly Triage meeting if you'd still like to do it

diracdeltas:
@bsclifton this is pretty straightforward and applies to brave core. in the future npm will be the definitive source for the sync client, so the browsers should switch to using the npm package (which is simpler anyway).

PR #13848, opened 2018-04-17
Cleanup of the eopkg commands on Solus for more simplicity

Jacalz:
Just some simple cleanup of the command for installing dependencys when building Brave under [Solus](https://solus-project.com). I later learned that there is no reason to use two separate commands because it can be compacted down to one command with "it" instead of "install".

This should be safe for merging pretty quickly 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,7 @@ dnf group install "Development Tools" "C Development Tools and Libraries"
 #### On Solus
 
 ````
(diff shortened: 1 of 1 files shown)


Jacalz:
This is ready to be merged @bsclifton :smiley:

Jacalz:
Thank you @bsclifton 🙂

PR #13856, opened 2018-04-18
Adding context menu item to history and bookmark entries for publisher list inclusion

ryanml:
Fixes: #6547

## Submitter Checklist:

## Test Plan:

1. Enable payments
2. Set minimum payment settings a bit high, (Ex: 1 minute, 5 visits)
3. Browse a few different sites to add history entries.
4. Bookmark a few sites naturally via the menu, create some custom bookmarks as well.
5. Create at least one bookmark with a bogus location like `asldjfoasidj`
6. Navigate to `about:bookmarks`
7.  For entries that have not been added to the ledger table, right click on them and confirm that "Add To Publisher List" is enabled and on the context menu. 
8. For the bogus entry, confirm that "Add To Publisher List" is disabled
9.  Use the context menu item to add a couple of the sites to the ledger.
10. Confirm that the items are present in the ledger table, and are set with the minimum visits and visit time set in payment settings.
11. Confirm that those items can not be double-added/stacked with the new context menu item.
12. Repeat steps 7 - 11 with the entries under `about:history`

## Reviewer Checklist:

- [x] Request a security/privacy review [as needed](https://github.com/brave/handbook/blob/master/development/security.md#how-to-request-a-security-review) if one was not already requested.

Tests


app/browser/api/ledger.js
```diff
@@ -706,14 +706,16 @@ const excludeP = (publisherKey, callback) => {
...
-const addSiteVisit = (state, timestamp, location, tabId) => {
+const addSiteVisit = (state, timestamp, location, tabId, manualAdd = false) => {
   if (!synopsis) {
     return state
   }
 
-  location = pageDataUtil.getInfoKey(location)
-  const locationData = ledgerState.getLocation(state, location)
-  const duration = new Date().getTime() - timestamp
+  location = manualAdd ? tldjs.getDomain(location) : pageDataUtil.getInfoKey(location)
```
NejcZdovc: Please add check if `tldjs.getDomain(location)` is null

app/browser/api/ledger.js
```diff
@@ -851,21 +853,28 @@ const shouldTrackTab = (state, tabId) => {
...
     if (shouldTrackTab(state, currentTabId)) {
       state = addSiteVisit(state, currentTimestamp, currentUrl, currentTabId)
     }
   }
 
+  if (manualAdd) {
+    const minimumVisits = getSetting(settings.PAYMENTS_MINIMUM_VISITS)
+    for (let v = 0; v < minimumVisits; v++) {
+      state = addSiteVisit(state, currentTimestamp, location, currentTabId, manualAdd)
+    }
```
NejcZdovc: I think that we need to return state here, because we don't want to reset `currentUrl` and `currentTimestamp`

js/contextMenus.js
```diff
@@ -376,6 +378,16 @@ const siteSingleDetailTemplate = (siteKey, type, activeFrame) => {
     )
   }
 
+  if (type === siteTags.HISTORY || type === siteTags.BOOKMARK) {
+    let publisherKey = tldjs.getDomain(siteDetail.get('location'))
```
NejcZdovc: This will not work for youtube and twitch url's

NejcZdovc:
We need to make sure that normal switching is not disrupted.

app/browser/api/ledger.js
diracdeltas: why can't this just use `pageDataUtil.getInfoKey`? I'm slightly concerned that pageDataUtil.getInfoKey includes the URL protocol whereas tldjs.getDomain doesn't

js/contextMenus.js
```diff
@@ -376,6 +378,16 @@ const siteSingleDetailTemplate = (siteKey, type, activeFrame) => {
     )
   }
 
+  if (type === siteTags.HISTORY || type === siteTags.BOOKMARK) {
```
diracdeltas: should this also check that payments is enabled?

ryanml: Very good catch, will update

app/extensions/brave/locales/en-US/menu.properties
```diff
@@ -4,6 +4,7 @@ actualSize=Actual Size
 addBookmark=Add Bookmark…
 addFolder=Add Folder…
 addToFavoritesBar=Add to Favorites Bar
+addToPublisherList=Add to Publisher List
```
diracdeltas: is it clear that people will know what this is referring to? cc @bradleyrichter @alexwykoff

app/browser/api/ledger.js
ryanml: adjusted

app/browser/api/ledger.js
ryanml: So here lays the problem I'm having and the way I chose to test it revealed this:

In the case we are adding something via the history item context menu, that site has already been visited and in some capacity captured in ledger state. So using getInfoKey() for ledgerState.getLocation(state, location) pulls in location data just fine.

The same will work for bookmarks added with the '+' (you would have had to visit the page already)

WHere tldjs.getDomain becomes helpful is if I make a bookmark for a site that I haven't visited yet and manually enter the location. In that case ledgerState.getLocation(state, location) returns nothing as it has no record of the site. This way a key for the domain can be generated without depending on getLocation()

With this in mind, I would be compelled to make the use of tldjs.getDomain a special case only for those bookmarks, not history entries since we can trust the ledgerState for the location data. Since the publisher key ends up being the url without the protocol, it doesn't seem to have any unintended side effects.

Let me know your thoughts :) @NejcZdovc @diracdeltas

app/browser/api/ledger.js
ryanml: adjusted

js/contextMenus.js
ryanml: updated

app/extensions/brave/locales/en-US/menu.properties
bradleyrichter: This needs a different string like "Include this site in Brave Payments contributions."

If that is too long for a menu, "Include site in Brave Payments"

ryanml: @bradleyrichter for comparison's sake:

image

image

ryanml: @bradleyrichter for comparison's sake:

screen shot 2018-04-19 at 12 44 52 pm

screen shot 2018-04-19 at 12 46 03 pm

app/browser/api/ledger.js
diracdeltas: @RyanML not sure i understand the problem since this isn't calling `ledgerState.getLocation(state, location)`. i'm just suggesting changing this line to `location = pageDataUtil.getInfoKey(location)` instead of `location = manualAdd ? tldjs.getDomain(location) : pageDataUtil.getInfoKey(location)`

diracdeltas: your comment would make sense for https://github.com/brave/browser-laptop/pull/13856/files#diff-3cdfdb124f718787d584e9e2b82fcedcR720

ryanml: @diracdeltas per our conversation, tldjs.getDomain() is no longer in use. Additionally, the context menu item is disabled for custom bookmark locations that are either invalid or without a protocol.

app/common/lib/ledgerUtil.js
```diff
@@ -132,6 +132,19 @@ const walletStatus = (ledgerData) => {
   return status
 }
 
+const shouldShowMenuOption = (state, location) => {
```
diracdeltas: ++

app/extensions/brave/locales/en-US/menu.properties
NejcZdovc: I would go with short version

ryanml:
squashed

app/browser/api/ledger.js
```diff
@@ -700,14 +700,21 @@ const excludeP = (publisherKey, callback) => {
...
 
-const addSiteVisit = (state, timestamp, location, tabId) => {
+const addSiteVisit = (state, timestamp, location, tabId, manualAdd = false) => {
   if (!synopsis) {
     return state
   }
 
   location = pageDataUtil.getInfoKey(location)
+
+  if (location == null) {
```
NejcZdovc: I think that we can move this check before `location = pageDataUtil.getInfoKey(location)`

js/contextMenus.js
```diff
@@ -376,6 +378,15 @@ const siteSingleDetailTemplate = (siteKey, type, activeFrame) => {
     )
   }
 
+  if (paymentsEnabled && (type === siteTags.HISTORY || type === siteTags.BOOKMARK)) {
+    let location = siteDetail.get('location')
+    let enabled = ledgerUtil.shouldShowMenuOption(state, location)
+    template.push(
```
NejcZdovc: I think that we should check if enabled and only if enabled we should add this template

app/browser/api/ledger.js
```diff
@@ -845,21 +852,29 @@ const shouldTrackTab = (state, tabId) => {
   return !isPrivate && !tabFromState.isEmpty() && ledgerUtil.shouldTrackView(tabFromState)
 }
 
-const addNewLocation = (state, location, tabId = tabState.TAB_ID_NONE, keepInfo = false) => {
+const addNewLocation = (state, location, tabId = tabState.TAB_ID_NONE, keepInfo = false, manualAdd = false) => {
```
NejcZdovc: Please add unit tests for this function

app/browser/api/ledger.js
```diff
@@ -700,14 +700,21 @@ const excludeP = (publisherKey, callback) => {
   })
 }
 
-const addSiteVisit = (state, timestamp, location, tabId) => {
+const addSiteVisit = (state, timestamp, location, tabId, manualAdd = false) => {
```
NejcZdovc: Please add unit tests for this function

NejcZdovc:
Test plan works great

app/browser/api/ledger.js
ryanml: Agreed, done

js/contextMenus.js
ryanml: Template only added if menu option should be enabled.

ryanml:
Tests updated

NejcZdovc:
master da59512
0.23 14193cf6e
0.22 624283ebb

PR #13900, opened 2018-04-23
Fixes publisher toggle (heart), button enables after time requirements satisfied.

### LINKED ISSUE #12675: URL Heart (include for payment) button disabled even after time requirements satisfied
image

image

site:
https://www.theguardian.com/cities/2018/jan/16/beginning-end-trump-real-estate-empire

ryanml:
Fixes #12675

## Submitter Checklist:

## Test Plan:
1. Enable payments
2. Select a minimum browsing time and visits in payment settings, (1 visit for ease)
3. Open web console for error notation.
4. Navigate to any web page. 
5. Browse the page for the minimum browsing time. 
6. Confirm that the publisher toggle button is enabled (Yellow if site was pinned when included in the ledger, deeper black if merely included in the ledger)
7. In the same tab, navigate to a different web page
8. Repeat steps 5 and 6.
9. Open a new tab, navigate to a new web page
10. Repeat steps 5 and 6

## Reviewer Checklist:

Tests

Given that there didn't seem to be an appropriate appConstant (or a relatively efficient/telling new one that could be made), it seems to make sense for the Publisher Toggle component to have an attribute of its internal state that can check its ledger status at an interval.

Currently, the interval avoids starting until after APP_TOP_SITE_DATA_AVAILABLE has dispatched. Doing this call to pageDataChanged before that action can break ledger activity recorder.

The interval will clear once the publisher toggle has reached an enabled state (there's no way for a site to become disabled merely by browsing, so this interval does not concern that path), any time the publisherToggle component dismounts, and reset if the tab loads to a new potential publisher.


ryanml:
WIP has been removed, I may have a way to eliminate the use of setTimeout

app/renderer/components/navigation/publisherToggle.js
```diff
@@ -54,6 +55,50 @@ class PublisherToggle extends React.Component {
...
 
+  setUpdateInterval () {
+    if (this.mounted === false) {
+      return
+    }
+    // The dispatch to update top sites occurs ~5 seconds after a
+    // page visit has started recording. It is important that a
+    // page data update is not triggered before this has happened.
+    const topSitesWait = ledgerUtil.milliseconds.second * 6
+    const updateWait = ledgerUtil.milliseconds.second * 3
```
NejcZdovc: could we instead of triggering this every 3s, trigger this only after `MIN_VISIT_TIME`?

NejcZdovc: also we have min count, so that means if you set this for 10, you need to visit this site 10 times. Maybe we should only add this timer if this visit would hit this minimum. This means that we would not add timer until you have 9 visits already logged and this is your 10 visit. Only in this case we would add timer, which would be min visit time

@ryanml wdyt?

app/renderer/components/navigation/publisherToggle.js
```diff
@@ -32,6 +32,7 @@ const fundUnverifiedPublisherImage = require('../../../extensions/brave/img/urlb
 class PublisherToggle extends React.Component {
   constructor (props) {
     super(props)
+    this.mounted = false
```
ryanml: Thanks for fixing 👍

app/renderer/components/navigation/publisherToggle.js
ryanml: I think this approach would be a lot more efficient yes, especially for minimums with multiple visits. It only would not work for one scenario:

Suppose your minimums are 1 visit for 1 minute, and you've made one visit but only for 41 seconds. Visiting the site again with this approach you'd have to wait 60 seconds for the full MIN_VISIT_TIME to pass.

To get around that, maybe we could calculate the time remaining, (In the above scenario, calculate 60000 - 41000 = 19000 and set the timeout to 19000). Does that sound agreeable?

NejcZdovc: yes let's do that

ryanml:
@NejcZdovc changes pushed, ready to review. The correct calculations should now be in place so an update is only triggered when it needs to.

app/common/lib/ledgerUtil.js
```diff
@@ -591,6 +592,45 @@ const getMediaProvider = (url, firstPartyUrl, referrer) => {
   return provider
 }
 
+const shouldSetTimeout = (publisherKey) => {
```
NejcZdovc: Please rename this to something more verbose like `hasRequiredVisits`

app/common/lib/ledgerUtil.js
```diff
@@ -591,6 +592,45 @@ const getMediaProvider = (url, firstPartyUrl, referrer) => {
...
+  if (publisherVisits === undefined) {
+    return minimumVisits === 1
+  }
+
+  const visitDifference = minimumVisits - publisherVisits
+
+  return (visitDifference === 1)
+}
+
+const getTimeoutWait = (publisherKey) => {
```
NejcZdovc: Please rename this to something more verbose like `hasRequiredTime`

app/common/lib/ledgerUtil.js
```diff
@@ -591,6 +592,45 @@ const getMediaProvider = (url, firstPartyUrl, referrer) => {
...
+  const minimumVisits = parseInt(getSetting(settings.PAYMENTS_MINIMUM_VISITS))
+
+  if (minimumVisits === 1) {
+    return true
+  }
+
+  const publisher = ledgerState.getPublisher(appStore.state, publisherKey)
+  const publisherVisits = publisher.get('visits')
+
+  if (publisherVisits === undefined) {
```
NejcZdovc: Can we just check here if `publisherVisits` is number?

app/common/lib/ledgerUtil.js
```diff
@@ -591,6 +592,45 @@ const getMediaProvider = (url, firstPartyUrl, referrer) => {
...
+  const publisher = ledgerState.getPublisher(appStore.state, publisherKey)
+
+  if (!publisher) {
+    return minimumVisitTime
+  }
+
+  const publisherDuration = publisher.get('duration')
+
+  if (
+    publisherDuration === undefined ||
```
NejcZdovc: Can we just check here if `publisherDuration` is number?

NejcZdovc:
@ryanml please add tests

ryanml:
@NejcZdovc tests added

NejcZdovc:
master 15a6784
0.23 b0e14b24a
0.22-release3 f5e4dc684

PR #13946, opened 2018-04-26
Merge Chromium 66 to 0.22.x

bsclifton:
TODOs: 
- [ ] rebase against 0.22.x (remove duplicate commits)
- [ ] merge to 0.22.x 
- [ ] cherry pick merge commit to master (if possible) OR PR the Chromium 66 changes against master


NejcZdovc:
@bsclifton I think that the only think that we need to do is version bump, all other commits not related to version bumps are already in 0.22 branch

PR #13986, opened 2018-05-02
Ensure all icons in bookmarks bar are vertically aligned

### LINKED ISSUE #13789: bookmark bar folder and file icons are misaligned when overflow indicator is visible
### Test plan
See https://github.com/brave/browser-laptop/pull/13986

### Steps to Reproduce

  1. Bookmark many items and at least one folder
  2. Switch to favicons only
  3. Resize window to show overflow indicator from not showing overflow indicator

**Actual result:**
Folder and default favicon (i.e. document icon) jump a couple of pixels vertically.

**Expected result:**
Same alignment as when overflow indicator is not present

**Reproduces how often:**
100%

### Brave Version

**about:brave info:**
0.22.x single webview
0.23.x source
master

No

### Additional Information
Caused by #13758

### LINKED ISSUE #13985: bookmarks toolbar does not reduce in height in text-only mode
### Test plan
See https://github.com/brave/browser-laptop/pull/13986

### Description
#13758 introduced fixed height on bookmarks bar to ensure it did not wrap to two lines. Looks like (possibly intentionaly) text-only mode in current release, the bar is a couple pixels shorter.

0.22.700

petemill:
Ensure all icons in bookmarks bar are vertically aligned
Fix #13789

Also restore different bookmarks bar height when in text-only mode vs with favicons.
Fix #13985

Reduce flicker of bookmark bar / overflow-indicator / last item by having aphrodite not remove the custom class-name on render.

Reduce complexity of bookmarks toolbar state by passing through ToolbarMode 'enum' to component.
Future-proof bookmark mode setting retrieval for state-selector-memoization by passing through appState

## Submitter Checklist:

## Test Plan:
1. Preferences > Bookmarks Bar > choose `Show favicons only`
2. Add 10 - 20 bookmarks to bookmarks bar (or pre-existing profile)
3. Bookmark the New Tab page (as it does not have a favicon)
4. Add a couple of folders to bookmarks bar
5. Extend window width so that there is no overflow indicator
6. Contract window width so that there is an overflow indicator
   - **Alignment of icons for folders, default favicon (New Tab page) and favicons should not differ whether there is the overflow indicator or not.**

7. Preferences > Bookmarks Bar > choose `Text only`
    - **Bookmarks bar should decrease in height by a couple of pixels**

## Reviewer Checklist:

Tests


petemill:
I've temporarily cherry-picked this to 0.22.x-single-webview until it is approved. If a change is required, that's ok since 0.22.x-single-webview is reset for each build until single-webview is merged to 0.22.x-release3

app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -196,23 +195,23 @@ class BookmarksToolbar extends React.Component {
   mergeProps (state, ownProps) {
     const currentWindow = state.get('currentWindow')
     const activeFrame = frameStateUtil.getActiveFrame(currentWindow) || Immutable.Map()
+    const bookmarkDisplayMode = bookmarkUtil.showOnlyFavicon()
+        ? 1
```
cezaraugusto: it's unclear to me what those numbers mean. IMO better to have them under https://github.com/brave/browser-laptop/blob/master/app/common/constants/settingsEnums.js and name them something like `bookmarkDisplayOptions`

app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -196,23 +195,23 @@ class BookmarksToolbar extends React.Component {
...
+        : bookmarkUtil.showFavicon()
+          ? 2
+          : 3
 
     const props = {}
     // used in renderer
     props.shouldAllowWindowDrag = !isWindows && windowState.shouldAllowWindowDrag(state, currentWindow, activeFrame, isFocused(state))
     props.toolbarBookmarks = bookmarksState.getBookmarksWithFolders(state, 0).take(110).map(item => item.get('key'))
-
+    props.textOnly = (bookmarkDisplayMode === 3)
```
cezaraugusto: if we follow the above comment here we could have

```
props.textOnly = (bookmarkDisplayMode === bookmarkDisplayOptions.textOnly)
```

or any other more meaningful name

app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -303,29 +316,29 @@ class BookmarksToolbar extends React.Component {
...
+        data-bookmarks-overflow-indicator
+        onClick={this.onMoreBookmarksMenu}
       >
-        <BrowserButton
-          bookmarksOverflowIndicator
-          iconOnly
-          size={globalStyles.spacing.bookmarksToolbarOverflowButtonWidth}
-          iconClass={globalStyles.appIcons.angleDoubleRight}
-          onClick={this.onMoreBookmarksMenu}
+        <svg className={css(styles.bookmarksToolbar__overflowIndicator__icon)} xmlns='http://www.w3.org/2000/svg'viewBox='0 0 320 512'>
```
cezaraugusto: is there any benefit of inlining SVGs instead of requiring the file here? if perf is the same I'd prefer the latter for readability.

app/renderer/components/bookmarks/bookmarksToolbar.js
```diff
@@ -283,11 +284,23 @@ class BookmarksToolbar extends React.Component {
...
+          !this.props.shouldAllowWindowDrag && styles.bookmarksToolbar_disallowDragging
+        ) +
+        // Ensure we do not remove overflow indicator on props change
+        // Aphrodite does not support data-attribute selectors :-(
+        // which would be nice to control this functionality.
+        // Instead, we must use a custom class, added by `calculateNonFirstRowItems`
+        // which gets overriden by this `className` attribute here.
+        (this.hasHiddenKeys
+          ? ' ' + css(styles.bookmarksToolbar_hasOverflow)
+          : '')
```
cezaraugusto: **nit** can't we put it inside `style` instead?

```
style={{ 
  '--bookmarks-toolbar-overflow-indicator-visibility': this.hasHiddenKeys && 'visible',
  '--bookmarks-toolbar-overflow-indicator-width': this.hasHiddenKeys && `${globalStyles.spacing.bookmarksToolbarOverflowButtonWidth} !important`
}}
```

it wouldn't parse if result is `false` and IMO it's easier to read. Again this is **nit**

app/renderer/components/bookmarks/bookmarksToolbar.js
petemill: I don't think that would be appropriate. Those numbers are here purely so that this component (only) will re-render when the mode changes. Putting that in a separate file only spreads logic that is otherwise fully self-contained in to another module. Plus, this is not a settings enum. This value is simply a tri-state value for a prop so that we can compare if the next update has a different display mode. It's irrelevant what the value is, apart from textOnly, which is why that is specified as a prop with a more understandable name.

app/renderer/components/bookmarks/bookmarksToolbar.js
petemill: The purpose of this prop is already to make a readable name: `textOnly` as that value *is* used in other places in the component, whereas the exact value of `bookmarkDisplayMode` is not considered anywhere else.

app/renderer/components/bookmarks/bookmarksToolbar.js
petemill: Benefit is:
- no extra network request
- ability to style through css

Sure, it can be imported as a separate component, e.g.  `<BookmarksToolbarOverflowIndicatorIcon />` and since this icon is changing in the icon project and I'm using this technique for most new icons, then I'll be sure to favor not having the `<svg>` paths make the other components look messy. I'd prefer to make that change for that release if that's alright with everyone else.

petemill:
image

app/renderer/components/bookmarks/bookmarksToolbar.js
bsclifton: if only a tri-state value, why not use a string value instead? 😄

app/renderer/components/bookmarks/bookmarksToolbar.js
bsclifton: being able to style through CSS is a huge win- I've fought angrily with SVGs when used as a value for `background-image` (and similar use-cases)

app/renderer/components/bookmarks/bookmarksToolbar.js
petemill: Scratch that... it was appropriate because there is already an enum that I didn't know about, and it's being called many many times for each action / render cycle, so I started using it and then went down a rabbit hole of optimizing it for readability and less state lookups... 🤷‍♀️  ✅

petemill:
Updated with changes and updated description to reflect that I ended up using and modifying the bookmarkUtil settings enum property. At this stage in the project I wouldn't normally consider changing that, but it sets that property up nicely for state-selection-memoization coming in #12105. Bookmark toolbar is a great candidate for that as if you have a full bookmarks bar, that's a lot of components and state selection repeating for every single action dispatch cycle (often many per second, when the data doesn't change anyway).

app/renderer/components/bookmarks/bookmarksToolbar.js
cezaraugusto: ya sorry missed the className prop here

cezaraugusto:
master (0.24.x) fdb0943
0.23.x 65a83fd37
0.22.x-release3 66d6cef96

PR #14083, opened 2018-05-10
Fixes for extension text and icons

Jacalz:
**This PR basically tries to do the following things:**

- Fixes the descriptions of the extensions to be the same as the text that comes with the extension. (Except for MetaMask because the description we had was way better than the one from the extension)

- It updates the extension icons of Bitwarden, Honey and Dashlane to use the latest logos. Also added Vimium icon in preparation for it becoming enabled.

- It also fixes the names of for Bitwarden and 1password to be the same as the one that comes with the extension and now both Dashlane and 1password say that they require an application to work. Dashlane still changes name when you enable it but that is because it should say that it needs the application.

- On the other side there are some cleanup and some changes for consistency in extensionUtils.js and also the preparation for adding Vimium support but it is commented out so it isn't enabled yet.

## Submitter Checklist:


app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -53,7 +53,7 @@ banSiteConfirmation=Are you sure you want to delete this site?
 beta=beta
 bitcoin=Bitcoin
 bitcoinBalance=Please transfer:&nbsp;
-bitwarden=bitwarden&reg;
+bitwarden=Bitwarden&reg;
```
jonathansampson: Should we go ahead and use the same value as extensions.properties? Otherwise the name changes in preferences before and after enabling the Bitwarden extension. (Same for other extension strings in this file)

Jacalz: I would suggest we do because then we can drop the &reg; part at the end...

Jacalz: This is now fixed :+1:

Jacalz:
This is ready for a re-review now @jonathansampson

bsclifton:
Thanks for the contribution @Jacalz - but I think with the move to brave-core coming, it makes sense to close this out

PR #14132, opened 2018-05-15
autoplay notification should follow spec

cezaraugusto:
Address #8956
Reported in https://twitter.com/BrendanEich/status/996258562346135552

Test Plan:

* When autoplay notification is visible, button order should be `deny`/`allow` and respective actions should be respected.


cezaraugusto:
tagged as `Completed work` cc @bsclifton for best prioritization

Issue #14043, opened 2018-05-07
When using Tor for private tabs, need UI changes that take Tor process startup time into account


davidtemkin:
### Description
While the Tor process is starting up, the browser should not permit a user to browse to a URL via Tor in a private tab. There should be (a) a visual indicator that shows something is happening, and (b) the UI should either make it impossible to enter a URL in this state -or- allow the URL to be entered, but only attempt loading once the Tor process is up and running and ready to receive requests. It should be impossible to get the proxy error.

### Steps to Reproduce
1. Create a new private tab and enable private browsing via Tor
2. While Tor is being initialized, attempt to browse to a URL

**Actual result:**
If you navigate to the URL before the Tor process is ready, an error message results: "Could not create a connection to the proxy server. An error occurred either in resolving its name, or in connecting a socket to it. Note that this does NOT include failures during the actual "CONNECT" method of an HTTP proxy."

**Expected result:**
No error message should occur. The user should not be able to navigate until the Tor process is ready; and a clear visual/UI state should block this navigation. 

**Reproduces how often:**
Anytime you get ahead of Tor process launch.

### Brave Version

**about:brave info:**
0.23.2

**Reproducible on current live release:**
no

diracdeltas:
there's at least 3 non-UI bugs here:

1. the tor process doesn't ever start on linux because the executable isn't marked as executable (fixed in https://github.com/brave/browser-laptop/pull/12987/commits/297afb24190fa72a7b245ed3b76179b590eced6d)
2. the tor process should start at startup instead of first tab launched (to-be-fixed in muon, partly done in browser-laptop https://github.com/brave/browser-laptop/commit/cb9be30b26932ed037555ebb490aaac4c85e3ce7)
3. the tor process seems more flakey than it should be on mac and windows. both @davidtemkin and @lukemulks reports that it seemed to never connect even after waiting a long time.

once tor init is moved to startup, we could do the following:
1. when user opens a private tab with Tor currently disabled, the Tor switch is shown greyed out if Tor is not done initializing (instead it shows the percent initialized) so the user can't enable it until it is done initializing
2. if Tor is currently enabled but Tor is not done initializing, the private tab shows the percent initialized and a warning sign saying that the user will not be able to browse unless they turn it to off. If they then try to navigate, the navigation is blocked.

diracdeltas:
[...]

PR #14146, opened 2018-05-16
Talk to the control socket using my old draft tor daemon manager.

riastradh-brave:
This doesn't solve #14043 yet, but it provides a candidate place to hook up a progress indicator.  See the `TODO(riastradh): Visually update a progress bar!` comment for the place where we should update a progress bar, currently https://github.com/brave/browser-laptop/pull/14146/commits/c2b3edab0024684f755be088715667edb7ce21b4#diff-8a9bac13ef6264feb2b6da7c18d86b3cR751.

- [x] Depends on https://github.com/brave/muon/pull/592 to launch tor with a control port.
- [x] Depends on https://github.com/brave/browser-laptop/pull/14145 to update the version of tor to one that has functional control ports, since apparently they were broken in the alpha version we had been using before.
- [ ] Depends on https://github.com/brave/browser-laptop/pull/14168, currently cherry-picked into the branch, to make the tests function.  Should be rebased once tor/0.23.x is rebased to have that commit.

## Submitter Checklist:

## Test Plan:
Launch Brave.  Confirm that the console output mentions tor bootstrapping.

## Reviewer Checklist:

Tests

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -747,58 +747,17 @@ function setupTor () {
     torDaemon.on('exit', () => console.log('tor: daemon exited'))
     torDaemon.on('launch', (socksAddr) => {
       console.log(`tor: daemon listens on ${socksAddr}`)
-      const control = torDaemon.getControl()
-      const bootstrapped = (event) => {
-        const args = event.split(' ')
-        if (args[0] === 'ERR' || args[0] === 'WARN') {
-          console.log(`tor: ${event}`)
-          return
-        }
-        if (args[1] !== 'BOOTSTRAP') {
-          return
-        }
-        for (let i = 0; i < args.length; i++) {
-          const [k, v] = args[i].split('=')
-          if (k === 'PROGRESS') {
-            // TODO(riastradh): Visually update a progress bar!
-            console.log(`tor: bootstrapped ${v}%`)
-            return
-          }
-        }
-        console.log(`tor: bootstrapped but not sure how much: ${event}`)
-      }
-      control.on('async-STATUS_CLIENT', (event, extra) => {
-        bootstrapped(event)
-      })
-      control.subscribe('STATUS_CLIENT', (err) => {
+      const bootstrapped = (err, progress) => {
         if (err) {
-          console.log(`subscribe STATUS_CLIENT: ${err}`)
+          console.log(`tor: bootstrap error: ${err}`)
           return
         }
-        const getinfoLine = (status, reply) => {
-          if (status !== '250') {
-            console.log(`tor: bootstrap error: ${reply}`)
-            return
-          }
-          const prefix = 'status/bootstrap-phase='
-          if (!reply.startsWith(prefix)) {
-            console.log(`tor: GETINFO status/bootstrap-phase: ${reply}`)
-            return
-          }
-          const event = reply.slice(prefix.length)
-          bootstrapped(event)
+        console.log(`tor: bootstrapped ${progress}%`)
+      }
+      torDaemon.onBootstrap(bootstrapped, (err) => {
+        if (err) {
+          console.log(`tor: error subscribing to bootstrap: ${err}`)
         }
-        control.cmd('GETINFO status/bootstrap-phase', getinfoLine,
-          (err, status, reply) => {
-            if (!err) {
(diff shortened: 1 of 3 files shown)


app/filtering.js
```diff
@@ -47,6 +48,8 @@ const headersReceivedFilteringFns = []
 let partitionsToInitialize = ['default', appConfig.tor.partition]
 let initializedPartitions = {}
 
+let torDaemon = new tor.TorDaemon()
```
diracdeltas: it looks like you don't reassign to this variable so better to use `const` instead of `let`

also this doesn't need to be in the module scope, right? so it can be declared in setupTor instead.

app/tor.js
```diff
@@ -0,0 +1,1253 @@
...
+      assert(i >= 0)
+      assert(i < directories.length)
+      fs.mkdir(directories[i], 0o700, (err) => {
+        if (err && err.code !== 'EEXIST') {
+          return callback(err)
+        }
+        loop(i + 1)
+      })
+    }
+    loop(0)
```
diracdeltas: it would be slightly clearer if you used `directories.forEach` instead of this recursive loop

app/tor.js
```diff
@@ -0,0 +1,1253 @@
...
+  }
+}
+
+module.exports.TorControl = TorControl
+module.exports.TorDaemon = TorDaemon
+module.exports.torControlParseKV = torControlParseKV
+module.exports.torControlParseQuoted = torControlParseQuoted
+module.exports.torrcEscapeBuffer = torrcEscapeBuffer
+module.exports.torrcEscapePath = torrcEscapePath
+module.exports.torrcEscapeString = torrcEscapeString
```
diracdeltas: please add unit tests for these exported methods (located in test/unit, `npm run unittest` to run

app/filtering.js
riastradh-brave: done in 453e327b099187651abd294ab409517431991378

app/tor.js
riastradh-brave: Callbacks make forEach a little tricky.  If this should be done with mkdirSync, I guess that's OK, but I figured it would be better to avoid that.  (I'm not _sure_ this is necessary, but the ordering of directory creation in muon vs browser-laptop is not obvious, so I figured it would be safer to keep this logic than delete it.)

app/tor.js
riastradh-brave: done in 428280759b7f9fe963d3bea7a9b7f169311ecf44

app/tor.js
diracdeltas: @riastradh-brave i see now that one of the directories is a subdir of the other. in that case you can just use fs-extra's mkdir instead of fs's mkdir, which will behave like `mkdir -p`: https://github.com/jprichardson/node-fs-extra

app/tor.js
diracdeltas: thanks for adding tests - could you also document the input types of any externally-callable methods/constructors using jsdoc so it's harder to get the types wrong? http://usejsdoc.org/ - see js/lib/request for an example

riastradh-brave: Will do.

riastradh-brave: done in 8f532a9a58155a82d65050fce4d647193d9dc894

app/tor.js
riastradh-brave: How does that help to do multiple ones in a loop?  I guess I could replace the whole thing by a single mkdirp for the deepest subdirectory, but I didn't want to embed the assumption that there is a single deepest subdirectory in that function.

riastradh-brave:
In https://github.com/brave/browser-laptop/pull/14146/commits/4b39b3a9386e6045a058cfab35b2b68b311f4166, I added some automatic tests for the logic to launch tor and wait for a notification that it has launched within a reasonable time, but they are broken because `fs.mkdir` doesn't work while running `npm run unittest`.

These tests may also be problematic because they deliberately put in a half-second delay to make sure we exercise both orders of (a) tor process startup, and (b) fs event watch notification logic setup.

What to do?

riastradh-brave:
Never mind, apparently this works now and I have no idea why it failed before.

PR #14217, opened 2018-05-22
remove unneccessary console warning

### LINKED ISSUE #14216: Unneccessary console warning "Frame did not have tab strip window id but tab did"
### Description
Often, when creating a new tab, the warning "Frame did not have tab strip window id but tab did..." will be displayed in the window console. This is not an important warning as it is expected that sometimes that tab will be added to the window strip before the frame is created in the window. That case is handled well.

0.22.714

petemill:
Fix #14216


petemill:
0.23.x 961cfdf2f
0.22.x 2242e6008

PR #14282, opened 2018-05-30
Handle null/undefined cases for frame entries gracefully so browser d…

bsclifton:
…oesn't crash

(there are still many methods which don't handle properly; this is a first go at it)

Fixes https://github.com/brave/browser-laptop/issues/14261
(however, there is still noise in the logs ("Cannot move frame to index..."))

Auditors: @petemill

## Submitter Checklist:

## Test Plan:
See https://github.com/brave/browser-laptop/issues/14261

## Reviewer Checklist:

Tests

### DIFF
--- a/app/renderer/reducers/frameReducer.js
+++ b/app/renderer/reducers/frameReducer.js
@@ -86,10 +86,12 @@ function setFrameChangedIndex (state, frame, currentIndex, newIndex) {
   let frames = frameStateUtil.getFrames(state)
   if (newIndex >= frames.size) {
     console.error(`Cannot move frame to index ${newIndex} from ${currentIndex} because it is invalid for a frame List of size ${frames.size}!`)
+    state.set('frames', frames)
     return state
   }
   // put the frame at the correct index
   frames = frames
+    .filter((frame) => !!frame)
     .splice(currentIndex, 1)
     .splice(newIndex, 0, frame)
   state = state.set('frames', frames)
@@ -242,7 +244,9 @@ const frameReducer = (state, action, immutableAction) => {
             !hasBeenActivated && isPotentialPhishingUrl(tab.get('url')))
         }
         if (!frame.get('hasBeenActivated')) {
-          state = state.setIn(['frames', index, 'hasBeenActivated'], true)
+          if (state.hasIn(['frames', index])) {
+            state = state.setIn(['frames', index, 'hasBeenActivated'], true)
+          }
         }
         state = frameStateUtil.updateTabPageIndex(state, tabId)
       }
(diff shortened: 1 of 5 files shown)


test/unit/state/frameStateUtilTest.js
```diff
@@ -32,87 +42,7 @@ describe('frameStateUtil', function () {
 
   after(function () {
     mockery.disable()
```
bsclifton: this file has a LOT of changes- mostly organizing the order of the methods being tested to actually match the order in frameStateUtil.js

app/renderer/reducers/frameReducer.js
```diff
@@ -86,10 +86,12 @@ function setFrameChangedIndex (state, frame, currentIndex, newIndex) {
   let frames = frameStateUtil.getFrames(state)
   if (newIndex >= frames.size) {
     console.error(`Cannot move frame to index ${newIndex} from ${currentIndex} because it is invalid for a frame List of size ${frames.size}!`)
+    state.set('frames', frames)
```
petemill: `state = state.set(...`

js/state/frameStateUtil.js
```diff
@@ -33,27 +33,31 @@ const matchFrame = (queryInfo, frame) => {
...
 
 function isFrameKeyActive (state, frameKey) {
   return getActiveFrameKey(state) === frameKey
 }
 
 function getFrames (state) {
-  return state.get('frames')
+  return state
+    ? state.get('frames')
+      ? state.get('frames').filter((frame) => !!frame)
```
petemill: Is this going to be expensive calling this iterative filter function many many times each time we render / check state for changes? That will be fixed with #12105, but in a different release, but just a little concerned this will affect performance as in a render-cycle, possibly hundreds of frameStateUtil functions will get called, and each one will then run this filter.

js/state/frameStateUtil.js
```diff
@@ -33,27 +33,31 @@ const matchFrame = (queryInfo, frame) => {
...
+      : Immutable.List()
+    : Immutable.Map()
```
petemill: Why would we ever want this function to return an `Immutable.Map`?

bsclifton:
Good catches! Had a few things there I didn't quite finish 😄 Will re-review and submit an update soon

bsclifton:
Closing as there is a much bigger issue happening here

In short, the page shown in steps to reproduce is causing ~92 web contents to be created... which then fails to get added to the tab strip. These are likely popups being blocked- but it's never being cleaned up properly

Will continue investigating

PR #14381, opened 2018-06-11
Torrent warning tor tabs

jumde:
## Fixes:
https://github.com/brave/browser-laptop/issues/14374

## Test Plan:

1. Open https://webtorrent.io/free-torrents in a tor private tab.
2. Open any `<torrent> | <magnet>` links.
3. Verify if the warning is displayed.

screen shot 2018-06-11 at 3 48 45 pm


app/browser/webtorrent.js
```diff
@@ -145,5 +145,7 @@ function getHeader (headers, headerName) {
 
 module.exports = {
   init,
+  isTorrentFile,
+  isMagnetURL,
```
diracdeltas: this approach looks good, but FWIW: @feross is it intentional that these methods check `details.url` instead of `details.firstPartyUrl`?

app/extensions/brave/locales/en-US/app.properties
```diff
@@ -237,6 +237,8 @@ streetAddress=Street Address
 submit=Submit
 tabsSuggestionTitle=Tabs
 topSiteSuggestionTitle=Top Site
+torrentBlockedInTor=Torrents are not allowed in tor private tabs.
```
diracdeltas: Can you change this to `For your privacy, torrents are blocked in private tabs when Tor is enabled.`

PR #14383, opened 2018-06-12
Fixes: #14373 Avoid favicon overlap in tor

jumde:
## Fixes
https://github.com/brave/browser-laptop/issues/14373

## Before

image

## After
screen shot 2018-06-11 at 5 49 53 pm


bradleyrichter:
I think @cezaraugusto may be able to assist on this one as he set up the rules. 

(looks like the new Tor icon broke the rule stack or something)


Regarding which icon to use, what if a user has 50 Tor tabs open? Is it better to see they are all Tor or better to distinguish them by different favicons?

rossmoody:
My vote would be to just establish the correct breakpoint here and keep the fav.ico. The edge case where you have 30 private tabs with tor protection open, I have to imagine you know they are a private collection and the user will get much more glanceability value out of site fav.ico's than a bundle of Tor icons.

diracdeltas:
if we get rid of the tor/private icons when there are a lot of tabs open, it would be great if they still appeared on the currently active tab somehow (for instance expanding the width of the active tab so that the icons are all visible)

rossmoody:
Oh that's a clever idea. Maybe swapping the site's fav.ico for the Tor icon on the tab that is active?

diracdeltas:
closing this since it's out of date

Issue #14459, opened 2018-06-19
Add the ability to open up non-Tor private tabs


bbondy:
### Description

Currently there are 2 modes: 
i) Use all private tabs as Tor private tabs
ii) Use all private tabs as normal private tabs

<img width="1077" alt="screen shot 2018-06-19 at 3 10 28 pm" src="https://user-images.githubusercontent.com/831718/41619069-edeff308-73d2-11e8-8da5-8cf36aecdc65.png">

Tor private tabs is an awesome feature, but they are slow.  So the choice between them seamlessly is important for us to deliver this feature correctly without angering our user base.

Having an all or nothing private tab option covers the use case that a privacy conscious user wants to be fully private.

But there are other use cases that people use private tabs for, the main one being a user just doesn't want cookies saved for a 1 time login.  For example logging into gmail with 1 time access when you don't want cookies saved. 

Others include you don't want history stored, but you are ok that it's not fully private, but still want fast browsing speed.  A lot of our user base are advanced users so we need to still keep them in mind.

Currently you have to switch the global option on or off. 

There was also talk about removing the Tor indicator. That needs to be there as a reminder of why a tab is slow.

### Steps to Reproduce

For each case there should be separate items for private tabs and for Tor private tabs. 

Case 1:

Add an option when right clicking on a link:
<img width="276" alt="screen shot 2018-06-19 at 2 56 31 pm" src="https://user-images.githubusercontent.com/831718/41619227-6e17a49a-73d3-11e8-8138-d159d4569528.png">

Case 2:

Add an option in the hamburger menu:

<img width="226" alt="screen shot 2018-06-19 at 2 56 43 pm" src="https://user-images.githubusercontent.com/831718/41619238-7a3e8176-73d3-11e8-82db-de8086982b76.png">
 
Case 3:

Add an option to the new tab drop down ~(sorry no screenshot, but just click and hold the new tab button)~
![image](https://user-images.githubusercontent.com/17010094/41619989-464c9de6-73d6-11e8-9c85-439ca800fb8f.png)

**Actual result:**

Only 1 private tab option is available.

**Expected result:**

Options for both private tabs and Tor private tabs should be available. 
Please also fail this testing if there is no Tor indicator for the Tor tabs.

**Reproduces how often:**

Always.

### Brave Version
0.22.810

**about:brave info:**

 
Brave: 0.22.810 
V8: 6.7.288.43 
rev: 8f30eeba0c2d6cccf1920a360b4774f820adbc58 
[...]

PR #14467, opened 2018-06-20
add tor menu items, make tor switch per tab

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/14459

test plan:
1. every menu that says 'open in new private tab' should also have an
   item that says 'open in new tor tab'
2. opening in a new tor tab should show the tor indicator on
3. opening in a new private tab should show the tor indicator off
4. ddg is enabled by default for now but can be turned off
5. cmd+click should open in the same type as the current tab
6. turning the tor switch on/off in a new tab should work
7. in a tor tab, right-click should show 'open in new tor tab' first

## Submitter Checklist:

## Reviewer Checklist:

Tests


js/constants/appConfig.js
```diff
@@ -151,7 +151,7 @@ module.exports = {
     'general.spellcheck-languages': Immutable.fromJS(['en-US']),
     'search.default-search-engine': 'Google',
     'search.offer-search-suggestions': false, // false by default for privacy reasons
-    'search.use-alternate-private-search-engine': false,
+    'search.use-alternate-private-search-engine': true,
```
diracdeltas: debatable; alternative (can be done in follow-up) is to make DDG a per-tab setting like Tor

app/common/commonMenu.js
```diff
@@ -77,20 +77,23 @@ module.exports.newPrivateTabMenuItem = () => {
...
+}
+
+module.exports.newTorTabMenuItem = () => {
+  return {
+    label: locale.translation('newTorTab'),
+    click: function (item, focusedWindow) {
+      ensureAtLeastOneWindow({
+        url: 'about:newtab',
+        isPrivate: true,
+        isTor: true
```
petemill: This will not result in a tor tab when there are 0 windows open, which is a common macOS scenario (as judging from the number of issues that were raised when I had a similar bug!).

In order to fix this we simply need to make the `openFramesInWindow` function inside _app/browser/windows.js_ aware of the `tor: ` property for the `createTabRequested` options argument.

app/extensions/brave/locales/en-US/newtab.properties
```diff
@@ -27,6 +27,8 @@ privateTabText1=Whether or not you use Tor, private tabs are not logged in page
 privateTabTitle=This is a Private Tab
 privateTabTorTitle=Make private tabs much more private with
```
petemill: With this change, perhaps we should not make it seem like a global setting a more like a setting just for the current tab: 'Make this tab much more private with Tor'?

app/common/commonMenu.js
diracdeltas: good catch thanks!

js/contextMenus.js
```diff
@@ -930,10 +926,12 @@ const addToPublisherListMenuItem = (location) => {
...
   if (!frame.get('isPrivate')) {
-    template.push(openInNewTabMenuItem(link, frame.get('isPrivate'), frame.get('partitionNumber'), frame.get('tabId')))
+    template.push(openInNewTabMenuItem(link, frame.get('partitionNumber'), frame.get('tabId')))
   }
+  const isTor = frameStateUtil.isTor(frame)
   template.push(
-    openInNewPrivateTabMenuItem(link, frame.get('tabId')),
+    openInNewPrivateTabMenuItem(link, frame.get('tabId'), isTor),
+    openInNewPrivateTabMenuItem(link, frame.get('tabId'), !isTor),
     openInNewWindowMenuItem(link, frame.get('isPrivate'), frame.get('partitionNumber')),
```
diracdeltas: bug: this should send isTor as the last argument

Issue #14641, opened 2018-07-03
[hackerone] favicon issue


diracdeltas:
https://hackerone.com/bugs?subject=brave&report_id=375965

LaurenWags:
Verified with macOS 10.12.6 using
* 0.23.35 72375c0
* Muon 7.1.6
* libchromiumcontent 67.0.3396.103

PR #14654, opened 2018-07-04
fix some favicon leaks

diracdeltas:
fix https://github.com/brave/browser-laptop/issues/14653
fix https://github.com/brave/browser-laptop/issues/14641

test plan: 
 1. open browser console and inspect the urlbar icon
 2. hit any search shortcut, ex: ':g' for google
 3. notice that the favicon for the search engine is loaded as a local resource instead of as a 'http(s)://' URL
4. open tor tab
5. browse to a few sites in the tor tab
6. long-click the back button. no favicons should appear.
7. go back a few sites, then repeat step 6 with the forward button.

### DIFF
--- a/app/renderer/components/common/contextMenu/contextMenuItem.js
+++ b/app/renderer/components/common/contextMenu/contextMenuItem.js
@@ -179,7 +179,7 @@ class ContextMenuItem extends ImmutableComponent {
       width: iconSize
     }
 
-    const icon = this.props.contextMenuItem.get('icon')
+    const icon = !this.props.isTor && this.props.contextMenuItem.get('icon')
     let faIcon
     if (icon) {
       iconStyle = Object.assign(iconStyle, {
--- a/js/data/searchProviders.js
+++ b/js/data/searchProviders.js
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const path = require('path')
 const {braveExtensionId} = require('../constants/config')
 
 /**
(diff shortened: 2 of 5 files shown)


js/data/searchProviders.js
```diff
@@ -13,12 +14,19 @@ const getFaviconUrl = (name, ext = 'ico') => {
   return `chrome-extension://${braveExtensionId}/img/favicons/${name}.${ext}`
 }
 
+/**
+ * Returns path of a favicon resource
+ */
+const getPath = (name, ext = 'ico') => {
+  return path.join(__dirname, '..', '..', 'img', 'favicons', `${name}.${ext}`)
```
diracdeltas: verified this works on packaged builds as well as in dev

app/renderer/components/common/contextMenu/contextMenuItem.js
```diff
@@ -179,7 +179,7 @@ class ContextMenuItem extends ImmutableComponent {
       width: iconSize
     }
 
-    const icon = this.props.contextMenuItem.get('icon')
+    const icon = !this.props.isTor && this.props.contextMenuItem.get('icon')
```
petemill: small nitpick here that IMO it would be preferrable to have the prop be more semantic, e.g. `shouldDisplayIcon`

js/data/searchProviders.js
petemill: @diracdeltas does it not work just to use the same `localImage` property which calls `getFaviconUrl`?

diracdeltas:
master: 3f8feb28e132ebd595ad93c0f08cbac46494fb16
0.24.x: a0e52edd7684a88128bb284f2e9e80711fa8c85e
0.23.x: c59de37b923f36ad123fccc271e52ea7fc875a3e

PR #14776, opened 2018-07-18
Add Vietnamese language packages

vituocgia:
Hello there
I've added Vietnamese package and translate from en-US strings to vi-VN
Has been tested, so Vietnamese people can use it easier
The translation is not finished, i'll take time and finish it when i have free time


vituocgia:
I just commit to track the translation progress

bsclifton:
Thanks for the PR @vituocgia! We actually have a translation process which you can sign up for and join. More info about that is here:
https://github.com/brave/browser-laptop/blob/master/docs/translations.md

We pull the translations periodically from Transifex, so these changes would get overwritten. For that reason (needs to happen in Transifex), I'll close this PR

PR #14790, opened 2018-07-20
Added support for the standard day period type.

phistuck:
Chrome has a bug - it returns the type as dayperiod instead of dayPeriod and this is probably going to change. Support both of the strings for now.

### DIFF
--- a/js/about/newTabComponents/clock.js
+++ b/js/about/newTabComponents/clock.js
@@ -20,7 +20,7 @@ class Clock extends React.Component {
           // hide blank strings
           return null
         }
-      } else if (component.type === 'dayperiod') {
+      } else if (component.type === 'dayperiod' || component.type === 'dayPeriod') {
         // hide day-period (AM / PM), it's rendered in a separate component
         return null
       }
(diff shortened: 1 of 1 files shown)


bsclifton:
@phistuck maybe instead of multiple checks, we could to a `toLower` and check once? What do you think?

1jj:
See
https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/7yWM-y_2j9k

phistuck:
@bsclifton - I am championing the Chrome (well, V8) fix for this and I looked for wrong cases and this was one of them. I am not using your project, so I do not know if anything is broken at the moment. I would imagine it is broken if you run it in Firefox or Safari, because they simply do not return `"dayperiod"`.

`toLowerCase` can be used, yes, though I imagine the performance is better when not transforming letter casing, if it matters. Also, having this small weirdness/ugliness in the code is a good incentive to remove it soon, when no browsers would return `"dayperiod"`.

phistuck:
Can you re-run the Travis CI build somehow? It seems to have failed unrelatedly.

NejcZdovc:
@phistuck re-run it again

phistuck:
Thank you. I am glad it passed. :)

PR #14804, opened 2018-07-22
random-lib@3.0.0

### LINKED ISSUE #6944: Misuse of Math.random() to generate random integers

fardog:
* fixes #6944

Updates to the latest version of random-lib, which fixes bias in the random integer methods of older versions.

**Note:** the newer algorithm is far slower than the old one; I typically don't recommend using it synchronously, as it is here. It will probably be fine, but I'll leave that to the maintainers to decide.

## Submitter Checklist:

## Reviewer Checklist:

Tests

### DIFF
--- a/app/browser/api/ledger.js
+++ b/app/browser/api/ledger.js
@@ -351,7 +351,7 @@ const onBootStateFile = (state) => {
   }
 
   if (client.sync(callback) === true) {
-    run(state, random.randomInt({min: ledgerUtil.milliseconds.minute, max: 5 * ledgerUtil.milliseconds.minute}))
+    run(state, random.intSync({min: ledgerUtil.milliseconds.minute, max: 5 * ledgerUtil.milliseconds.minute}))
   }
 
   module.exports.getBalance(state)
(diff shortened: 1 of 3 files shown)


riastradh-brave:
Thanks for the contribution!  We have a number of minor random sampling bugs that we should fix, as you have noted.  We discussed privately and decided to use our own samplers in our brave-crypto utility library.

diracdeltas:
closing as this has been superceded by #14843. thanks for the effort though!

PR #14851, opened 2018-07-26
Fix up random sampling with brave-crypto random samplers.

### LINKED ISSUE #6944: Misuse of Math.random() to generate random integers

riastradh-brave:
fix #6944

Some of these changes change the _distribution_ of probabilities to
the possible outcomes, where it seemed obvious that a uniform
distribution was intended but not previously attained.

Some of these changes also change the _support_, the set of possible
outcomes, where it made things simpler and was not obviously intended
one way or another.

Many of these are not _necessarily_ bugs, but if deviations from
uniform are actually intended they should be clearly justified.

Full details:

- Change Math.floor(Math.random() * n) to crypto.random.uniform(n).

  - app/extensions/brave/content/scripts/adInsertion.js
  - js/about/newtab.js (with care to preserve test spying in newTabPageTest.js)
  - js/about/preferences.js (which used |0 rather than Math.floor)
  - test/about/bookmarksManagerTest.js
  - test/lib/brave.js

  - This was obviously intended to be a uniform distribution on
    {0,1,2,...,n-1}.  It is nonuniform only because of IEEE 754
    binary64 floating-point approximation.

- Change Math.round(Math.random() * (n-1)) to crypto.random.uniform(n)
  when used as an index into an n-element array.

  - tools/lib/randomHostname.js, _chooseRandomLetter / ALPHABET
  - tools/lib/randomHostname.js, tld / TLDS
  - tools/lib/synopsisHelpers.js, publisher / PROTOCOL_PREFIXES
  - tools/lib/transactionHelpers.js, generateContribution: amount
    - (Adjusted to use the actual length of the array, not just the
      first four elements of an eight-element array.)

  - This gave half the probability to 0 and n - 1 that it gave to 1,
    2, 3, ..., n - 3, n - 2, a much bigger deviation from uniform than
    IEEE 754 binary64 floating-point approximations of the above.
    There was no obvious reason why this nonuniformity was desirable.

- Change Math.round(Math.random() * n) to crypto.random.uniform(n)
  when _not_ used as an index into an array.

  - tools/lib/synopsisHelpers.js, numHosts
  - tools/lib/synopsisHelpers.js, numVisits
  - tools/lib/transactionHelpers.js, generateTransaction: count
  - tools/lib/synopsisHelpers.js, duration

  - This actually changes the _support_ of the distribution to exclude
    n, but in the only places where this was used, it is not otherwise
    obvious that the caller intended to include n in the support.  It
    could have been a typo for Math.floor(Math.random() * n) without
    adverse consequences in these contexts.

- Change Math.round(Math.random()) to crypto.random.uniform(2).

  - tools/lib/randomHostname.js, numParts

  - This was obviously intended to be a uniform distribution on {0,1},
    and actually it may even be uniform under Math.random as typically
    implemented, but writing crypto.random.uniform(2) is clearer to
    flip a coin and this way we just avoid Math.random() altogether.

- Obscure corner cases:

  - app/renderer/rendererTabEvents.js, nonce
    - This used Math.random().toString() to generate a string `that is
      unlikely to collide'.  There are only 2^64 64-bit floating-point
      values, of which Math.random() returns a small subset.  A
      collision is expected after only a few billion uniform random
      samples, which is far from impossible.  I replaced it by a
      uniform random 256-bit quantity in hex, which will never collide
      even in the view of a conservative paranoid cryptographer unless
      there is a deeper bug breaking the PRNG.

  - tools/lib/randomHostname.js, partLen
    - The distribution originally used here puts higher probability on
      the minimum part length than everything else, and lower
      probability on the maximum part length than anything else.  I
      see no reason why this should not just be uniformly distributed
      on part lengths {minLength, minLength + 1, ..., maxLength - 1,
      maxLength}, so I replaced it by minLength + uniform(maxLength -
      minLength + 1).

  - tools/lib/transactionHelpers.js, generateBallots: votesToCast

    - The previous expression, Math.min(Math.round(Math.random() *
      votesRemaining) + 1, votesRemaining), assigned half weight to 1
      and half-again weight to votesRemaining - 1 (as compared to all
      other numbers) for no obvious reason.  I changed it to use 1, if
      only 1 vote is remaining, or 1 + uniform(votesRemaining - 1), if
      more than one vote is remaining.

P.S.  It is a bit confusing that we have a module named `brave-crypto`
which is usually used as `const crypto = require('brave-crypto')`
while there is also a standard nodejs module called `crypto` which is
usually used as `const crypto = require('crypto')`.  In one case these
are used in the same file, so I named ours `braveCrypto`.

## Submitter Checklist:

## Test Plan:
- npm run unittest
- npm run add-simulated-payment-history
- npm run add-simulated-synopsis-visits
- launch browser and browse around

## Reviewer Checklist:

Tests

### DIFF
--- a/app/common/lib/randomUtil.js
+++ b/app/common/lib/randomUtil.js
@@ -2,4 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-module.exports.random = () => Math.random()
+const crypto = require('brave-crypto')
+
+module.exports.uniform = (n) => crypto.random.uniform(n)
--- a/app/extensions/brave/content/scripts/adInsertion.js
+++ b/app/extensions/brave/content/scripts/adInsertion.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const crypto = require('brave-crypto')
+
 if (chrome.contentSettings.adInsertion == 'allow') {
   /**
    * Determines the ad size which should be shown
@@ -72,7 +74,7 @@ if (chrome.contentSettings.adInsertion == 'allow') {
     // generate a random segment
     // @todo - replace with renko targeting
     var segments = ['IAB2', 'IAB17', 'IAB14', 'IAB21', 'IAB20']
-    var segment = segments[Math.floor(Math.random() * 4)]
+    var segment = segments[crypto.random.uniform(segments.length)]
     var time_in_segment = new Date().getSeconds()
     var segment_expiration_time = 0 // no expiration
 
--- a/app/renderer/rendererTabEvents.js
+++ b/app/renderer/rendererTabEvents.js
@@ -3,6 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const Immutable = require('immutable')
+const crypto = require('brave-crypto')
 const electron = require('electron')
 
 const windowStore = require('../../js/stores/windowStore')
@@ -535,9 +536,8 @@ function showWidevineNotification (tabId, frame, noWidevineCallback, widevineCal
     return
   }
 
-  // Generate a random string that is unlikely to collide. Not
-  // cryptographically random.
-  const nonce = Math.random().toString()
+  // Generate a random string that is unlikely to collide.
+  const nonce = crypto.uint8ToHex(crypto.getSeed(32))
   const isWidevineEnabled = state.get('widevine') && state.getIn(['widevine', 'enabled'])
   if (isWidevineEnabled) {
     const message = locale.translation('allowWidevine').replace(/{{\s*origin\s*}}/, origin)
--- a/js/about/newtab.js
+++ b/js/about/newtab.js
@@ -29,7 +29,7 @@ const backgrounds = require('../data/backgrounds')
 
 // Utils
 const urlutils = require('../lib/urlutil')
-const {random} = require('../../app/common/lib/randomUtil')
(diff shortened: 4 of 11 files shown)


bsclifton:
master 7db1b6f37a6ad70d9a633ce945165bea3dd70dbf
0.24.x f8877ed9172632e62db4d0d5efd1fa85f6f34e4d
0.23.x 003a4b7f569770dc5a53e57c190d0cfd54923255

PR #14887, opened 2018-07-31
Fix/permission origin v2

riastradh-brave:
Stop-gap workaround for the most severe aspect of https://github.com/brave/browser-laptop/issues/14679.  Real problem is that we misinterpret the information that muon gives us and the results are bad.

## Submitter Checklist:

## Reviewer Checklist:

Tests

### DIFF
--- a/app/filtering.js
+++ b/app/filtering.js
@@ -549,14 +549,16 @@ function registerPermissionHandler (session, partition) {
         requestingOrigin = isPDFOrigin ? 'PDF Viewer' : 'Brave Browser'
         // display on all tabs
         mainFrameOrigin = null
-      } else if (isOpenExternal) {
+      } else {
+        requestingOrigin = getOrigin(requestingUrl) || requestingUrl
+      }
+
+      if (isOpenExternal) {
         // Open external is a special case since we want to apply the permission
         // for the entire scheme to avoid cluttering the saved permissions. See
         // https://github.com/brave/browser-laptop/issues/13642
         const protocol = urlParse(requestingUrl).protocol
         requestingOrigin = protocol ? `${protocol} URLs` : requestingUrl
-      } else {
-        requestingOrigin = getOrigin(requestingUrl) || requestingUrl
       }
 
       // Look up by host pattern since requestingOrigin is not necessarily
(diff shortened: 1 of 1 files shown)


bridiver:
we are determining the embedding origin in exactly the same way that Chrome is (`GetLastCommittedURL`). I think the problem is this overly permissive check that makes incorrect assumptions about the inputs.

bridiver:
we could potentially use `GetVisibleEntry` instead which shows the correct url in this particular case, but I feel like that's leaving the door open for some other as of yet undiscovered vulnerability that doesn't show up in `GetVisibleEntry` either. This just seems like a very dangerous bypass of all permission checks

bridiver:
actually I think the real problem is the use of the mainFrameOrigin for this check at all. We should be using the requestingUrl which is what chrome uses in all but one case

bridiver:
I believe you could do this same exploit with `GetVisibleEntry` if you used window.open to load `about:newtab` and then navigate to the same url. Since the new tab has been accessed by another tab the visible entry is no longer "safe" so it will be the last committed entry instead

diracdeltas:
v1:
screen shot 2018-07-31 at 6 27 25 pm

v2 (this branch):
screen shot 2018-07-31 at 6 26 56 pm

app/filtering.js
```diff
@@ -577,10 +577,6 @@ function registerPermissionHandler (session, partition) {
       } else if (isFullscreen && alwaysAllowFullscreen) {
         // Always allow fullscreen if setting is ON
         response.push(true)
-      } else if (isOpenExternal && (
-        // The Brave extension and PDFJS are always allowed to open files in an external app
-        isPDFOrigin || isBraveOrigin)) {
```
bridiver: I think this can actually be fixed by checking against the requestingUrl instead of the "mainFrameURL". I don't think the url of the webcontents is ever safe to use for any of these checks and Chrome also uses the requestingUrl in all but one case.

riastradh-brave:
@diracdeltas's original change was correct; I foolishly neglected to consider the _other_ branch of the `if` (which had scrolled off the top of the diff and out of my mind) when considering whether it would have any effect.  I have updated to remove my spurious removal of what is not, in fact, needless churn.

app/filtering.js
diracdeltas: @bridiver in this case the requesting URL is 'mailto:...' or whatever the external protocol is, so it would never be a PDF origin or Brave origin

bsclifton:
master 06905f4251e7c3336288656db3ee56905de8c2a6
0.24.x e951a03d29c7f3304c7cf2e0838f5dbf07f02d30
0.23.x 269b9741bbab30bae2441e8df2670ce64429c2b2
0.23.x-hotfix1 a6c310861ac17ad00fb590fdf038930c8f293dd1

PR #14944, opened 2018-08-03
Update muon 8.0.6

bsclifton:
Besides updating to Muon 8.0.6, this also fixes a dependency which was breaking Linux builds

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "file-loader": "0.11.2",
     "font-awesome": "^4.5.0",
     "fs-extra": "^2.1.2",
+    "home-path": "^1.0.6",
     "immutable": "^3.7.5",
     "immutablediff": "^0.4.2",
     "immutablepatch": "brave/immutable-js-patch",
--- a/.npmrc
+++ b/.npmrc
@@ -1,7 +1,7 @@
 runtime = electron
 target_arch = x64
-brave_electron_version = 8.0.4
+brave_electron_version = 8.0.6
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -107,6 +107,7 @@
     "file-loader": "0.11.2",
     "font-awesome": "^4.5.0",
     "fs-extra": "^2.1.2",
+    "home-path": "^1.0.6",
```
diracdeltas: is this suppoed to be here?

diracdeltas: oh i see, was it breaking linux builds?

seems like there should be a package-lock.json change for it

bsclifton: @diracdeltas yes - I believe it's a dependency of `electron-download`. Without this, the npm install fails on Linux. I'm not exactly sure where this broke, but you can see more details at https://jenkins.brave.com/view/laptop%20builds%20(child%20jobs)/job/browser-laptop-build-linux/536/console

bsclifton: Let me update the package-lock.json too...

bsclifton: @diracdeltas updated 😄 👍

DUDENICK:
lectron-download

bsclifton:
master 3a39fe728040b756f8bef441dc2f64d89e318a33
0.24.x a27f523ce91cb696ac56e5ac67fe75551165f497
0.23.x 41428f9fbe0e9ce41bae6e0625bc326c0e79b2fd

PR #15029, opened 2018-08-15
Harden use of geth

mrose17:
- catch “unlikely” but possible errors
- handle SIGHUP an dSIGINT

### DIFF
--- a/app/ethWallet-geth.js
+++ b/app/ethWallet-geth.js
@@ -28,6 +28,7 @@ const pidPath = isWindows ? '\\\\.\\pipe\\geth.pid' : path.join(gethDataDir, 'ge
 const gethProcessPath = path.join(getExtensionsPath('bin'), gethProcessKey)
 
 const configurePeers = async (dataDir) => {
+  const staticNodePath = path.join(dataDir, 'geth', 'static-nodes.json')
   try {
     const discoveryDomain = `_enode._tcp.${envNet}.${envSubDomain}.brave.com`
     let newNodes = await dns.resolveSrv(discoveryDomain)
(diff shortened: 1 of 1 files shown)


app/ethWallet-geth.js
```diff
@@ -104,29 +106,34 @@ const spawnGeth = async () => {
   }
 
   const gethOptions = {
-    stdio: process.env.GETH_LOG ? 'inherit' : 'ignore'
+    stdio: ((envNet === 'ropsten') || process.env.GETH_LOG) ? 'inherit' : 'ignore'
```
ryanml: is this included intentionally? i think that would should let `GETH_LOG` be the sole deciding indicator for logging, lest there is value in always having it debug on `ropsten`

app/ethWallet-geth.js
```diff
@@ -216,16 +243,19 @@ const restartGeth = async (tries = 3) => {
...
-  cleanupGeth(gethProcessId)
-})
-process.on('SIGINT', () => {
-  cleanupGeth(gethProcessId)
-  process.exit(2)
-})
+process.on('exit', cleanupGeth)
+process.on('SIGHUP', cleanupGethAndExit)
+process.on('SIGTERM', cleanupGethAndExit)
+process.on('SIGINT', cleanupGethAndExit)
```
ryanml: thanks for these ones ++

app/ethWallet-geth.js
mrose17: yes. it's intentional: if someone is on the test network, or someone turns on GETH_LOG, then we get logs, otherwise not. in other words, for production use, no logs by default; for testnet use, logs by default.

ryanml: ++

ryanml:
master aff1a96c4b130d7e878446820ded477b469f0609
0.24.x 84b416bfcb3fbfbfa642a09e67bc904381c2a41e
0.23.x 9949c3af3ced6781b16465de23af1922fb3cc944

PR #15309, opened 2018-10-26
Final browser-laptop update

bsclifton:
The final update for browser-laptop

Fixes https://github.com/brave/browser-laptop/issues/15315

DO NOT MERGE
- ~Needs more testing~
- ~Needs final binaries~
- Needs review with stakeholders

## Test plan
See https://github.com/brave/brave-browser/issues/1536


diracdeltas:
@bsclifton per discussion in slack, i would like to add some messaging/UI in this to help Linux users transition to the new repos (and also this should wait for https://github.com/brave/devops/issues/361).

app/updater.js
```diff
@@ -85,6 +86,35 @@ var scheduleUpdates = () => {
     }, appConfig.updates.appUpdateCheckFrequency)
   }
 
+  // Linux doesn't have an auto-update mechanism.
+  // Instead, deliver a periodic nag linking to instructions
+  if (isLinux() && appConfig.updates.deprecatedNagFrequency) {
+    electron.ipcMain.on(messages.NOTIFICATION_RESPONSE, (e, message, buttonIndex, persist) => {
+      appActions.hideNotification(message)
```
diracdeltas: i thought we decided to just make the notification unhideable on linux? fine with either way, just wondering if product spec has changed.

app/updater.js
```diff
@@ -85,6 +86,35 @@ var scheduleUpdates = () => {
...
+
+    const showDeprecationNotice = () => {
+      appActions.showNotification({
+        buttons: [
+          {text: 'OK'}
+        ],
+        options: {
+          persist: false,
+          advancedText: 'See instructions to upgrade to the latest version',
+          advancedLink: 'https://brave-browser.readthedocs.io/en/latest/installing-brave.html#linux'
```
diracdeltas: This is blocked on the doc text being updated with the new signing keys: https://github.com/brave/brave-browser/issues/1826

app/updater.js
bsclifton: Whoops- my bad on that. I'll fix that up 😄 So much for having a legit use for a `nagFrequency` variable 😂

diracdeltas:
looks fine when i pretended my mac was linux to trigger the notification, except kind of incongruous that the token grant notification also pops up:
screen shot 2018-11-06 at 12 56 32 pm

we probably want to either suppress it or prevent the old brave wallets from having any funds added to them?

diracdeltas:
On Mac, the message looks great but I don't think the auto-install of b-c actually worked. I see this in the console:

```
Extracting brave-core binaries from "/Users/yan/repos/browser-laptop/node_modules/electron-prebuilt/dist/Brave.app/Contents/Resources/Brave-Browser.pkg" into temp directory "/var/folders/hl/761qjwzn4qb7slh2yvlgwfxr0000gn/T/brave-upgrade"
Could not open product archive: /Users/yan/repos/browser-laptop/node_modules/electron-prebuilt/dist/Brave.app/Contents/Resources/Brave-Browser.pkg
Removing temp directory "/var/folders/hl/761qjwzn4qb7slh2yvlgwfxr0000gn/T/brave-upgrade"
```

diracdeltas:
`/Users/yan/repos/browser-laptop/node_modules/electron-prebuilt/dist/Brave.app/Contents/Resources/Brave-Browser.pkg` does not exist btw

bsclifton:
@diracdeltas for the auto-install to work, you have to run packaging; I'm about to queue a build for Nightly channel if you're interested in trying it out? 😄

bsclifton:
The only known issue that isn't fixed is captured with https://github.com/brave/brave-browser/issues/1914

Still looking through that... but I rebased this so that each commit is very meaningful. Latest is also pushed to Nightly channel for macOS + Windows. Ready for review!

app/darwinInit.js
```diff
@@ -99,10 +114,18 @@ if (process.platform === 'darwin') {
       } catch (e) {
       }
 
-      // relaunch and append argument expected in:
+      // if user has the Muon version of Brave as their default browser, let's
+      // ask the user to make Brave Core their new default. Passing this extra
+      // argument will prompt the user to ensure they agree with the choice
+      let defaultBrowserArg = ''
+      if (getDefaultBrowser() === 'com.electron.brave') {
```
darkdh: why can't we check default browser by https://github.com/brave/browser-laptop/blob/646212bafbce0b1bb8e93919435d27613b7af20c/app/index.js#L258?

app/darwinInit.js
```diff
@@ -99,10 +114,18 @@ if (process.platform === 'darwin') {
...
+        defaultBrowserArg = ' --make-default-browser'
+      }
+
+      // launch into freshly installed brave-core and append argument expected in:
       // https://github.com/brave/brave-browser/issues/1545
       console.log('Launching brave-core')
-      execSync(`open -a "${installedPath}/${appName}/" --args --upgrade-from-muon`)
+      execSync(`open -a "${installedPath}/${appName}/" --args --upgrade-from-muon${defaultBrowserArg}`)
```
darkdh: can we put space here instead of inside `defaultBrowserArg`?

app/darwinInit.js
darkdh: And by the method, you should be able to check default browser on Windows

bsclifton: No check on Windows needed... but `defaultProtocols.every()` looks like a much nicer way to do it, thanks! 😄

app/darwinInit.js
bsclifton: fixed

app/darwinInit.js
bsclifton: Fixed

app/darwinInit.js
```diff
@@ -0,0 +1,162 @@
...
+    if (!installerPath) {
+      console.log('brave-core installer not found')
+      return false
+    }
+
+    // brave-core is not installed; go ahead with silent install
+    const tempDir = path.join(os.tmpdir(), 'brave-upgrade')
+    try {
+      console.log(`Extracting brave-core binaries from "${installerPath}" into temp directory "${tempDir}"`)
+      execSync(`pkgutil --expand-full "${installerPath}" "${tempDir}"`)
```
jumde: Add a check to verify if `pkgutil` succeeded?

tools/buildPackage.js
```diff
@@ -147,11 +147,17 @@ if (isLinux) {
 } else if (isDarwin) {
   const macAppName = `${appName}.app`
   cmds.push('ncp ./app/extensions ' + path.join(buildDir, macAppName, 'Contents', 'Resources', 'extensions'))
+  cmds.push('cp ./res/Brave-Browser-0.56.9.pkg ' + path.join(buildDir, macAppName, 'Contents', 'Resources', 'Brave-Browser.pkg'))
```
jumde: Users update cadence is different, can we download the latest pkg/installer instead of embedding in the installer in resources? Users might end up installing an older version if the users update muon after a new version of brave-core is released.

app/updater.js
```diff
@@ -85,6 +86,21 @@ var scheduleUpdates = () => {
     }, appConfig.updates.appUpdateCheckFrequency)
   }
 
+  // Linux doesn't have an auto-update mechanism.
+  // Instead, show a persistent nag linking to instructions
+  if (isLinux()) {
+    appActions.showNotification({
+      buttons: [],
+      options: {
+        persist: false,
```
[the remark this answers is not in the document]
darkdh: `persist` means "remember this decision" and there is nothing needed to be memorized

tools/buildPackage.js
bsclifton: That would be nice- but since the brave-core has auto-update enabled, I don't think it's worth the extra complexity (ex: wget, have user wait, show a loading screen, then attempt).  Currently, install takes place when Muon is launched before any UI is shown. Windows has a stub installer which will fetch the latest, but that is very slow for folks outside the US (there is a known issue for that)

app/updater.js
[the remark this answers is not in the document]
bsclifton: That is correct- for example, if you wanted to remember that user chose to ALLOW auto-play for a given domain

js/about/newtab.js
```diff
@@ -257,6 +261,79 @@ class NewTabPage extends React.Component {
...
+      </div>
+    }
+
+    return <div className='deprecationNotice'>
+      <div>
+        <span className='note'>Hello!</span> This version of Brave {formattedMuonVersion} is no longer supported and will not be updated.
+        {
+          isLinux()
+            ? <span>&nbsp;To avoid potential security risks, please follow these <a href='https://brave-browser.readthedocs.io/en/latest/installing-brave.html#linux'>instructions</a> to upgrade to the latest version of the Brave Browser.</span>
+            : <span>&nbsp;To avoid potential security risks, please <a href='https://brave.com/download'>download the latest version</a> of the Brave Browser.</span>
```
diracdeltas: do we need to localize this?

bsclifton: I considered that- but since it's unlikely we'll get translations (we'd need to wait for them to be done in order to package them), I didn't translate it

cc: @bbondy @rebron

diracdeltas: seems reasonable

diracdeltas:
@bsclifton please ping us for a final security review when this is ready to merge, thanks

app/extensions/brave/locales/en-US/preferences.properties
```diff
@@ -379,7 +379,7 @@ syncShowQR=Show secret QR code. (Do not share!)
 syncSiteSettings=Saved site settings
 syncStart=I am new to Sync
 syncTitle=Brave Sync
-syncTitleMessage=Sync encrypted browser data between your devices securely and privately using Brave Sync.
+syncTitleMessage=This version of Sync is deprecated. You can clear any existing data you have
```
diracdeltas: needs a period at the end

app/renderer/components/preferences/syncTab.js
```diff
@@ -502,16 +496,11 @@ class SyncTab extends ImmutableComponent {
...
-              'fa-question-circle': true
-            })} />
-          </a>
           <div className={cx({
-            settingsListTitle: true,
-            [css(styles.subText)]: true
-          })} data-l10n-id='syncBetaMessage' />
+            settingsListTitle: true
+          })}>
+            Please download the <a className={css(commonStyles.linkText)} href='https://brave.com/download'>new version of Brave for desktop</a>
```
diracdeltas: needs period at the end

diracdeltas: also it is slightly better UX to have this open in a new tab instead of the current tab. `target=_blank` has some security pitfalls so you can use `aboutActions.createTabRequested`

bsclifton:
@diracdeltas feedback addressed in https://github.com/brave/browser-laptop/pull/15309/commits/3f06a9998d77947f12680c18b620d1eb3fe8b10c 😄

app/windowsInit.js
```diff
@@ -2,109 +2,246 @@
...
+      fs.writeFileSync(braveCoreUpgradeFile, `installed: ${new Date().getTime()}`)
+    } catch (e) {
+    }
+
+    // relaunch and append argument expected in:
+    // https://github.com/brave/brave-browser/issues/1545
+    try {
+      const installedPath = getBraveCoreInstallPath()
+      debugLog(`Launching brave-core at "${installedPath}/brave.exe" with "--upgrade-from-muon" flag`)
+      execSync(`"${installedPath}/brave.exe" --upgrade-from-muon`)
```
petemill: Shouldn't this be a `\` instead of a `/`. I'm sure Windows will accept it though, so just a nit

tools/buildPackage.js
```diff
@@ -112,10 +116,14 @@ cmds = cmds.concat([
 
 console.log('Building version ' + VersionInfo.braveVersion + ' in ' + buildDir + ' with Electron ' + VersionInfo.electronVersion)
 
+// Only way to give a different CFBundleName is to modify the 'app' name passed to electron-builder,
+// which also changes the output path.
+const tempModifiedAppName = (isDarwin && process.env.BUNDLE_NAME) ? process.env.BUNDLE_NAME : null
```
petemill: @bsclifton did we modify the build environment to contain this `BUNDLE_NAME` variable?

bsclifton: OMG - no, I didn't do that (yet). Thanks for the reminder 😄 (I'll mark this as resolved once it happens)

bsclifton: Looks like it was already there the whole time 😄 👍 (I think you added it- thanks!)
screen shot 2018-12-07 at 9 29 42 am

PR #15352, opened 2019-04-10
Force new tab page to show the "dashboard" (which has the deprecation…

bsclifton:
Addresses feedback by @LaurenWags:
https://bravesoftware.slack.com/archives/C0YL5KMA8/p1554929296100800

This change was already in 0.26.0; I just pulled it into this branch and then tested 😄


bsclifton:
@petemill yup- this will disable that from the get-go. As soon as new version is downloaded, the new tab page experience is streamlined 😄

